PR #2, opened 2019-09-19
[WIP][12.0] Add wms addon

guewen:
Bootstrap the base wms addon.

'wms' is the base with the mandatory dependencies.
'wms_demo' installs all the WMS addons and includes a demo configuration
for these addons.


wms/models/res_config_settings.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2019 Camptocamp (https://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    module_stock_vertical_lift_kardex = fields.Boolean(
```
rousseldenis, line 10: @guewen I don't really like this way of doing settings. IMHO each entry would be added in each module.

guewen: But the entry is for installing the module, so it wouldn't be there until you install it?

rousseldenis: Mmmmh yes you're right. That's because if the module is not present in the system, the click will lauch a stack trace (and that's not great for user)

guewen: Did you check in 12.0? I did this by mistake, forgot to `pip install -e` a module :) and I had no stack trace, nothing happened, the checkbox reverts to a blank box on save. A bit confusing, but no stack trace at least ;)

rousseldenis: Ok, good to know. It occured in past ones.

jgrandguillaume:
Hi @guewen 

Thanks you so much to provide this now. Little remark, the module are marked as "Enterprise" when logging on the runbot ! Any ideas why :) ?

A part from that, it looks good for a first push.

Regards,

Joël

jgrandguillaume (approved):

jgrandguillaume:
Propose to merge this to get something here for the OCA days. This is a very early draft

PR #5, opened 2019-11-08
[12.0][ADD] stock_putaway_storage_type_strategy

grindtildeath:


stock_putaway_storage_type_strategy/demo/location_storage_type.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
```
guewen:
```suggestion
<odoo noupdate="1">
```

stock_putaway_storage_type_strategy/demo/stock_putaway_rule.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
```
guewen:
```suggestion
<odoo noupdate="1">
```

stock_putaway_storage_type_strategy/models/location_storage_type.py
```diff
@@ -0,0 +1,74 @@
...
+    _inherit = 'stock.location.storage.type'
+
+    only_empty = fields.Boolean(
+        help="If checked, moves to the destination location will only be "
+             "allowed if there are not any existing quant nor planned move on "
+             "this location")
+    do_not_mix_lots = fields.Boolean(
+        help="If checked, moves to the destination location will only be "
+             "allowed if the location contains product of the same "
+             "stock.production.lot"
```
guewen:
```suggestion
             "lot."
```

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,160 @@
...
+        string='Packs storage strategy',
+        help='TODO',
+    )
+    display_pack_storage_strategy = fields.Boolean(
+        compute='_compute_display_pack_storage_strategy'
+    )
+
+    @api.depends(
+        'child_ids', 'location_id', 'location_id.pack_storage_strategy'
+    )
```
guewen, line 26: You should check `location_id.display_pack_storage_strategy` as with more than one level, the current depends would not detect the change.

Actually, I'm not sure we should hide the field on locations without children: when creating locations, you'll start by creating the zone/area, configure the strategy, then create the children locations. At worse, setting the strategy on a leaf will have no effect.

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,160 @@
...
+                dest_location.pack_storage_strategy == 'none' and
+                dest_location._package_storage_type_allowed(
+                    package_storage_type
+                )
+            ):
+                return dest_location
+            storage_locations = dest_location.get_storage_locations(
+                product=product
+            )
+            allowed_location = storage_locations.select_allowed_location(
```
guewen: I would rename it `select_first_allowed_location` for clarity.

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,160 @@
...
+        for location_storage_type in matching_location_storage_types:
+            if location_storage_type.only_empty:
+                if (
+                    not self._existing_quants() and
+                    not self._existing_planned_moves()
+                ):
+                    allowed_location_storage_types |= location_storage_type
+            elif location_storage_type.do_not_mix_products:
+                if location_storage_type.do_not_mix_lots:
+                    quant = self.env.context.get('putaway_quant')
```
guewen: Shouldn't we pass the quant around as a parameter in the `select_allowed_location` and `filter_restrictions` methods?

guewen (approved):
A couple of details but globally LGTM

stock_putaway_storage_type_strategy/models/location_storage_type.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+    do_not_mix_products = fields.Boolean(
+        help="If checked, moves to the destination location will only be "
+             "allowed if the location contains the same product."
+    )
+    # max_height = fields.Boolean(
```
rousseldenis, line 24: Remove if not necessary

stock_putaway_storage_type_strategy/models/location_storage_type.py
```diff
@@ -0,0 +1,74 @@
...
+    # max_volume = fields.Boolean(
+    #     help="If checked, moves to the destination location will only be "
+    #          "allowed if the packaging volume is lower than this maximum."
+    # )
+    # max_volume_value = fields.Float()
+
+    @api.constrains('only_empty', 'do_not_mix_lots', 'do_not_mix_products')
+    def _check_empty_mix(self):
+        for location_storage_type in self:
+            if (
```
rousseldenis, line 38: I like
```suggestion
            if any(location_storage_type.only_empty and (location_storage_type.do_not_mix_lots or 
                location_storage_type.do_not_mix_product) for location_storage_type in self):
```

 to avoid an indentation, but that's me.

stock_putaway_storage_type_strategy/models/location_storage_type.py
```diff
@@ -0,0 +1,74 @@
...
+                    "You cannot set 'Do not mix lots' without setting 'Do not"
+                    " mix products' constraint."
+                ))
+
+    @api.onchange('do_not_mix_products')
+    def _onchange_do_not_mix_products(self):
+        if not self.do_not_mix_products:
+            self.do_not_mix_lots = False
+
+    # @api.constrains('max_height', 'max_height_value', 'max_height_uom')
```
rousseldenis, line 66: Remove this if not necessary

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+
+    @api.depends(
+        'child_ids', 'location_id', 'location_id.pack_storage_strategy'
+    )
+    def _compute_display_pack_storage_strategy(self):
+        for location in self:
+            # Do not display on locations without children
+            if not location.child_ids:
+                location.display_pack_storage_strategy = False
```
rousseldenis, line 31: No need of affectation (for < v13), that's default

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,160 @@
...
+        return allowed_location_storage_types
+
+    def _existing_quants(self, product=None, lot=None):
+        domain = [('location_id', '=', self.id)]
+        if product is not None:
+            extra_domain = [('product_id', '!=', product.id)]
+            if lot is not None:
+                extra_domain = OR([extra_domain, [('lot_id', '!=', lot.id)]])
+            domain = AND([domain, extra_domain])
+        return self.env['stock.quant'].search(domain, limit=1)
```
rousseldenis, line 147: I would have built the domain in an other function, to make search(_prepare_domain_function()) and ease override

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,160 @@
...
+        domain = [
+            ('location_dest_id', '=', self.id),
+            ('move_id.state', 'not in', ('draft', 'cancel', 'done'))
+        ]
+        if product is not None:
+            extra_domain = [('product_id', '!=', product.id)]
+            if lot is not None:
+                extra_domain = OR([extra_domain, [('lot_id', '!=', lot.id)]])
+            domain = AND([domain, extra_domain])
+        return self.env['stock.move.line'].search(domain, limit=1)
```
rousseldenis: Same as above

stock_putaway_storage_type_strategy/models/stock_location.py
rousseldenis: Put this at the beginning

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+        allowed_location_storage_types = self.filter_restrictions(
+            matching_location_storage_types, product=product
+        )
+        return not self.allowed_stock_location_storage_type_ids or allowed_location_storage_types
+
+    def filter_restrictions(self, matching_location_storage_types, product=None):
+        allowed_location_storage_types = self.env['stock.location.storage.type']
+        for location_storage_type in matching_location_storage_types:
+            if location_storage_type.only_empty:
```
rousseldenis, line 115: Should these 'strategies' be called using built functions with getattr('xxx_only_empty') to ease override ?

rousseldenis (commented):
Some remarks.

stock_putaway_storage_type_strategy/models/stock_location.py
grindtildeath: > setting the strategy on a leaf will have no effect

This is actually why I wanted to hide the field as I don't think it make sense to display something meaningless... :man_shrugging:

guewen: It has meaning when you are creating the parent before the children (at this point, the parent is a leaf).

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+from odoo import api, models, fields
+from odoo.osv.expression import AND, OR
+
+
+class StockLocation(models.Model):
+
+    _inherit = 'stock.location'
+
+    pack_storage_strategy = fields.Selection(
+        selection=[('none', 'None'),
```
jbaudoux, line 12: We need to be consistent with the concept of storage kind here: https://github.com/OCA/stock-logistics-warehouse/pull/711/files#diff-bf18e95ba21ff1ac092947117a88571aR4

This should only be defined on location of kind "area".

There must also be a chaotic strategy that randomly select a location.

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+            )
+            if allowed_location:
+                return allowed_location
+        return self.browse()
+
+    def get_storage_locations(self, product=None):
+        self.ensure_one()
+        locations = self.browse()
+        if self.pack_storage_strategy == 'none':
+            locations = self
```
jbaudoux, line 82: We need here a random picker for chaotic storage on children locations

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+            product = quant.product_id
+        for rule in putaway_rules:
+            dest_location = rule._get_destination_location()
+            if (
+                dest_location.pack_storage_strategy == 'none' and
+                dest_location._package_storage_type_allowed(
+                    package_storage_type, quant, product=product
+                )
+            ):
+                return dest_location
```
jbaudoux, line 67: If strategy is none and location does not allow package, then jump to next rule otherwise you retest the same.
Nevertheless, we do not need this anymore following my comment just above in the code.

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+from odoo.osv.expression import AND, OR
+
+
+class StockLocation(models.Model):
+
+    _inherit = 'stock.location'
+
+    pack_storage_strategy = fields.Selection(
+        selection=[('none', 'None'),
+                   ('ordered_locations', 'Ordered Children Locations')],
```
jbaudoux, line 13: I cannot find any business use case where we would want an ordered_locations strategy

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+            return super()._get_putaway_rule_location(
+                product=product, category=category
+            )
+        putaway_rules = self.putaway_rule_ids.filter_rules(
+            product=product, category=category
+        )
+        if category and not product:
+            product = quant.product_id
+        for rule in putaway_rules:
+            dest_location = rule._get_destination_location()
```
jbaudoux, line 60: Either dest_location is a bin (fixed location strategy), or is an area (chaotic strategy).
In case of area, the storage type must be defined on the area location.

So we can check here "dest_location" is allowed for the storage type. Otherwise, jump to next rule.

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+                dest_location.pack_storage_strategy == 'none' and
+                dest_location._package_storage_type_allowed(
+                    package_storage_type, quant, product=product
+                )
+            ):
+                return dest_location
+            storage_locations = dest_location.get_storage_locations(
+                product=product
+            )
+            allowed_location = storage_locations.select_first_allowed_location(
```
jbaudoux, line 71: No need to check locations one by one. "Area" contains the allowed storage type.

jbaudoux (changes requested):

stock_putaway_storage_type_strategy/models/stock_location.py
guewen: It doesn't seem mandatory to me to bind this put-away concept to the location kind concept.
Could be a glue module to make the "UI-sugar" between them.

stock_putaway_storage_type_strategy/models/stock_location.py
guewen: Put in the first bin in the row in front of you everytime (put in the bin closest to you). As you have constraints, it will take the next available in the row.

stock_putaway_storage_type_strategy/models/stock_location.py
jbaudoux: Regarding my comment on chaotic, in fact we should have in this module the chaotic strategy and in https://github.com/OCA/wms/pull/6 only the ABC preference. So we should be able to implement chaotic without ABC.

stock_putaway_storage_type_strategy/models/stock_location.py
jbaudoux: I just cannot figure out who would like to work like that ;-) Not an issue of course. Just a side note.

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+        if not package_storage_type:
+            return super()._get_putaway_rule_location(
+                product=product, category=category
+            )
+        putaway_rules = self.putaway_rule_ids.filter_rules(
+            product=product, category=category
+        )
+        if category and not product:
+            product = quant.product_id
+        for rule in putaway_rules:
```
jbaudoux, line 59: There is a design issue. Put-away rules are fixed location rules, and storage strategy is dynamic/chaotic location rule. They must work independently.
When you set a chaotic storage strategy on an area, it must be applied even if you do not have a put-away rule.
You must first apply the put-away rules to find the dest location (standard odoo), then you must apply the storage strategy.

stock_putaway_storage_type_strategy/models/stock_location.py
```diff
@@ -0,0 +1,166 @@
...
+                location._package_storage_type_allowed(
+                    package_storage_type, quant, product=product
+                )
+            ):
+                return location
+        return self.browse()
+
+    def _get_ordered_children_locations(self):
+        return self.search(
+            [('id', 'child_of', self.ids), ('id', '!=', self.id)]
```
jbaudoux, line 99: We need to specify the wanted depth. If you have a storage strategy on an area having sub-areas and then bins, then I see 2 solutions:
* you only want to select bins. That's most likely where you will put the goods
* you only want to select a direct child (a sub-area). Then possibly you also have a storage strategy on the sub-area that selects a bin. That offers more flexibility but probably to specific
I would go for solution 1.

jbaudoux (changes requested):
Additionally, there is an issue when moving complete pack. As put-away is computed per quant, potentially you can end-up with quants sent to different locations, especially in case of chaotic storage. I can imagine a single product pack with multiple lots.

stock_putaway_storage_type_strategy/models/stock_location.py
guewen: In 2 you are speaking about the "recursive" strategies that we decided not to support with @jgrandguillaume , we only want bins.

stock_putaway_storage_type_strategy/models/stock_location.py
guewen: > it must be applied even if you do not have a put-away rule

why?

jbaudoux: Take an area filled in by internal replenishment (so no put-away rule from reception). You could still want to apply chaotic on the area.

guewen: As a temporary work-around, this commit allows this use case to work: https://github.com/OCA/wms/pull/5/commits/ffc6291a8cc3b326727ed0a93c0cdc1c611f67e8

Later, we'll probably want to take the final location returned by `StockLocation.get_putaway_strategy` and apply the storage type strategy on it.

One notable difference though, with the current behavior, you can have 2 rules as:

* Stock → Shelving
* Stock → Overflow

If it doesn't find any possible bin in Shelving, it will try in Overflow.
If we do the change we discuss, no place in Shelving means no put-away possible. Which in maybe not an issue in practice, I don't know.

grindtildeath:
Replaced by #8

PR #6, opened 2019-11-08
[12.0][ADD] stock_putaway_storage_type_abc

grindtildeath:


stock_putaway_storage_type_abc/models/stock_location.py
```diff
@@ -0,0 +1,62 @@
...
+
+    pack_storage_strategy = fields.Selection(
+        selection_add=[('abc', 'Chaotic ABC')]
+    )
+    display_abc_storage = fields.Boolean(
+        compute='_compute_display_abc_storage'
+    )
+    abc_storage = fields.Selection(ABC_SELECTION, required=True, default='b')
+
+    @api.depends('location_id', 'location_id.pack_storage_strategy')
```
guewen: I think this will invalidate only for one level (well, shouldn't be too much of an issue as long as it's not stored).
Couldn't you do the same check with `location_id.display_abc_storage`  beside `location_id.pack_storage_strategy`? If the parent displays ABC, so the children.

stock_putaway_storage_type_abc/models/stock_location.py
```diff
@@ -0,0 +1,62 @@
...
+        locations = self.search(
+            [('id', 'child_of', self.ids), ('id', '!=', self.id)],
+            order='abc_storage ASC'
+        )
+        product_abc = product.abc_storage
+        if product_abc == 'a':
+            # Already ordered, no need to reorder
+            pass
+        else:
+            # TODO improve ugly code
```
guewen: well, probably more efficient than doing 3 `filtered` on a large dataset?

guewen (approved):
Computed field to check, the rest LGTM

stock_putaway_storage_type_abc/models/stock_location.py
```diff
@@ -0,0 +1,67 @@
...
+                if loc.abc_storage == 'a':
+                    a_locations |= loc
+                elif loc.abc_storage == 'b':
+                    b_locations |= loc
+                elif loc.abc_storage == 'c':
+                    c_locations |= loc
+            if product_abc == 'b':
+                locations = b_locations | c_locations | a_locations
+            elif product_abc == 'c':
+                locations = c_locations | a_locations | b_locations
```
jbaudoux, line 66:
```suggestion
                locations = c_locations | b_locations | a_locations
```

stock_putaway_storage_type_abc/models/stock_location.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def get_storage_locations(self, product=None):
+        if not self.pack_storage_strategy == 'abc':
+            return super().get_storage_locations()
+        return self._get_abc_locations(product)
+
+    def _get_abc_locations(self, product):
+        locations = self.search(
+            [('id', 'child_of', self.ids), ('id', '!=', self.id)],
+            order='abc_storage ASC'
```
jbaudoux, line 45: That's probably not random. We should have a second order on random()

stock_putaway_storage_type_abc/models/stock_location.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+        return locations._sort_abc_locations(product.abc_storage)
```
jbaudoux, line 47: Instead of doing this, I would run twice the search method with:
- abc_storage >= product.abc_storage ; order='abc_storage ASC'
- abc_storage < product.abc_storage ; order='abc_storage DESC'

This will better allow to have more than just a/b/c as selection.

grindtildeath:
Replaced by #9

PR #7, opened 2019-11-25
[12.0] Move restrictions from stock_putaway_storage_type_strategy to base module

grindtildeath:
Original description:

The code added here was defined in `stock_putaway_storage_type_strategy` (#5).
However, we want these restrictions to be defined in the base module `stock_storage_type`.

Updates:

 - Dedicated tree view to edit package/location storage type relation
 - Rename fields to drop stock_ prefix
 - Small fixes (security group, demo data, view names, ...)

Depends on :
 - [ ] https://github.com/OCA/server-tools/pull/1738

PR #12, opened 2020-01-09
[13.0][MIG] stock_storage_type

grindtildeath:
Depends on:
 - [x] https://github.com/OCA/server-tools/pull/1751

Includes:
 - [x] https://github.com/OCA/wms/pull/7


simahawk (commented):
LG, please rebase and run pre-commit

grindtildeath:
Rebase done.

@simahawk @guewen @jgrandguillaume Here I'm wondering if we shouldn't merge stock_storage_type_putaway_strategy (#13) into this module, because stock_storage_type only defines fields that are unused and don't make any sense by themselves, i.e. without the other module to compute putaways and restrict stock moves according to storage types...

simahawk:
@grindtildeath you miss server-tools dependency 
```
odoo.exceptions.UserError: ("You try to install module 'stock_storage_type' that depends on module 'base_m2m_custom_field'.\nBut the latter module is not available in your system.", '')
```

stock_storage_type/models/stock_quant.py
```diff
@@ -0,0 +1,156 @@
...
+            )
+            if not lst_allowed_for_pst:
+                raise ValidationError(
+                    _(
+                        "Package storage type %s is not allowed into "
+                        "Location %s"
+                    ) % (pack_storage_type.name, location.name)
+                )
+            allowed = False
+            package_quants = quant.package_id.mapped('quant_ids')
```
jbaudoux: The constrain is evaluated at quant creation. At this stage, the package's quant is not up to date.

```
stock.quant(148,)
(Pdb) self.package_id
stock.quant.package(111,)
(Pdb) self.package_id.quant_ids
stock.quant()
```

stock_storage_type/models/stock_quant.py
```diff
@@ -0,0 +1,156 @@
...
+            package_products = package_quants.mapped('product_id')
+            package_lots = package_quants.mapped('lot_id')
+            other_quants_in_location = self.search(
+                [
+                    ('location_id', '=', location.id),
+                    ('id', 'not in', package_quants.ids)
```
jbaudoux: We must exclude all previous quants that are not there anymore, i.e. quantity = 0

```suggestion
                    ('id', 'not in', package_quants.ids),
                    ('quantity', '!=', 0)
```

stock_storage_type/models/stock_quant.py
```diff
@@ -0,0 +1,156 @@
...
+        Check if at least one location storage type allows the package storage
+        type into the quant's location
+        """
+        for quant in self:
+            location = quant.location_id
+            pack_storage_type = quant.package_id.package_storage_type_id
+            loc_storage_types = location.allowed_location_storage_type_ids
+            if (
+                not quant.package_id
+                or not pack_storage_type
```
jbaudoux: I would still evaluate the location storage type constrain like do not mix product even if the pack storage type is not defined and even if there is not package. This condition (having a package with a storage type) is only for the "lst_allowed_for_pst" check. For the other checks, we can still apply them based only the location storage type constrain

jbaudoux (changes requested):
What about extracting the location storage type in a separate module that allows to define restrictions on the location (empty, do not mix) without needing package.

This module would add the allowed storage type concept on the location storage type and manage package.

gurneyalex:
https://github.com/grindtildeath/wms/pull/5 fixes some performance issues

jbaudoux (approved):
Thanks for the fixes and perf improvements

jgrandguillaume (approved):
Tested & approved

grindtildeath:
I squashed all the commits with a group by authors. Hope this is fine for everyone involved

guewen:
> I squashed all the commits with a group by authors. Hope this is fine for everyone involved

Not such a big deal, but a commit message that contains 'do X AND Y', pretty often means you should have 2 separate changes.
For instance, the commit that replaces the XML domain by the field domain using web_domain_field was a good atomic commit that I intended to use as an example for a blog post. Now, if you want to understand, you have to read the whole commit message and reverse engineer what part of the commit's diff is related to what part of the commit's message.

Ref good to keep in mind: https://blog.mocoso.co.uk/talks/2015/01/12/telling-stories-through-your-commits/

Anyway, the test should be fixed by my last fixup, I'll squash it if the tests pass.

guewen (approved):

PR #13, opened 2020-01-13
[13.0][MIG] stock_storage_type_putaway_strategy

grindtildeath:
Depends on:
 - [ ] #12 
 - [ ] https://github.com/OCA/stock-logistics-workflow/pull/608

### DIFF
--- a/stock_storage_type_putaway_strategy/models/stock_location.py
+++ b/stock_storage_type_putaway_strategy/models/stock_location.py
@@ -177,14 +177,13 @@ class StockLocation(models.Model):
         return not (max_weight and pack_weight and pack_weight > max_weight)
 
     def _existing_quants(self, products=None, lot=None):
-
-        base_domain = [("location_id", "=", self.id)]
+        base_domain = [("location_id", "child_of", self.id)]
         domain = self._prepare_existing_domain(base_domain, products=products, lot=lot)
         return self.env["stock.quant"].search(domain, limit=1)
 
     def _existing_planned_moves(self, products=None, lot=None):
         base_domain = [
-            ("location_dest_id", "=", self.id),
+            ("location_dest_id", "child_of", self.id),
             ("move_id.state", "not in", ("draft", "cancel", "done")),
         ]
         domain = self._prepare_existing_domain(base_domain, products=products, lot=lot)
--- a/stock_storage_type_putaway_strategy/models/stock_location.py
+++ b/stock_storage_type_putaway_strategy/models/stock_location.py
@@ -77,7 +77,7 @@ class StockLocation(models.Model):
                 )
             ):
                 return pref_loc
-            storage_locations = pref_loc.get_storage_locations(product=product)
+            storage_locations = pref_loc.get_storage_locations(products=product)
             allowed_location = storage_locations.select_first_allowed_location(
                 package_storage_type, quant, product
             )
@@ -85,7 +85,8 @@ class StockLocation(models.Model):
                 return allowed_location
         return putaway_location
 
-    def get_storage_locations(self, product=None):
+    def get_storage_locations(self, products=None):
+        # TODO support multiple products? cf ABC
         self.ensure_one()
         locations = self.browse()
         if self.pack_putaway_strategy == "none":
@@ -94,10 +95,10 @@ class StockLocation(models.Model):
             locations = self._get_ordered_children_locations()
         return locations
 
-    def select_first_allowed_location(self, package_storage_type, quant, product):
+    def select_first_allowed_location(self, package_storage_type, quants, products):
         for location in self:
             if location._package_storage_type_allowed(
-                package_storage_type, quant, product
+                package_storage_type, quants, products
             ):
                 return location
         return self.browse()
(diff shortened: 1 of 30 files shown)


stock_storage_type_putaway_strategy/models/stock_location.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def _get_putaway_strategy(self, product):
+        putaway_location = super()._get_putaway_strategy(product)
+        quant = self.env.context.get("storage_quant")
+        return self._get_pack_putaway_strategy(
+            putaway_location, quant, product
+        )
+
+    def _get_pack_putaway_strategy(self, putaway_location, quant, product):
+        package_storage_type = False
```
Tonow-c2c, line 59: can you just add a docsting to show on the method description all return case possibly ?

stock_storage_type_putaway_strategy/models/stock_location.py
```diff
@@ -0,0 +1,200 @@
...
+        max_height = location_storage_type.max_height
+        return not (max_height and height and height > max_height)
+
+    def _max_weight_allowed(self, location_storage_type, quants):
+        pack_weight = quants.package_id.pack_weight
+        max_weight = location_storage_type.max_weight
+        return not (max_weight and pack_weight and pack_weight > max_weight)
+
+    def _existing_quants(self, products=None, lot=None):
+
```
Tonow-c2c:
```suggestion
```

stock_storage_type_putaway_strategy/models/stock_location.py
```diff
@@ -0,0 +1,200 @@
...
+        base_domain = [("location_id", "=", self.id)]
+        domain = self._prepare_existing_domain(base_domain, products=products, lot=lot)
+        return self.env["stock.quant"].search(domain, limit=1)
```
Tonow-c2c, line 182:
```suggestion
        return bool(self.env["stock.quant"].search(domain, limit=1))
```

but it's not very relevant is just to return on boolean like we expect with method `_existing_*` but don't change if you are not agree :wink:

stock_storage_type_putaway_strategy/models/stock_location_storage_type.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @api.depends(
+        "only_empty",
+        "do_not_mix_lots",
+        "do_not_mix_products",
+        "max_height",
+        "max_weight",
+    )
+    def _compute_has_restrictions(self):
+        for slst in self:
```
Tonow-c2c, line 20: are we sure they is default values for this compute ?

stock_storage_type_putaway_strategy/tests/test_storage_type_move.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from .common import TestStorageTypeCommon
+
+
+class TestStorageTypeMove(TestStorageTypeCommon):
+
+    def test_package_level_location_dest_domain(self):
+        pass
```
Tonow-c2c: you don't have time to do this tests or they are not commited ?

stock_storage_type_putaway_strategy/models/stock_location_storage_type.py
grindtildeath: What do you mean here? `has_restrictions` is always computed in the function since the result of `any` is used...

stock_storage_type_putaway_strategy/models/stock_location.py
grindtildeath: When I read `_existing_*`, I expect to get the existing records. IMO your suggestion would be valid if it red `_is_existing_*` but it's not case.

stock_storage_type_putaway_strategy/tests/test_storage_type_move.py
grindtildeath: Still WIP

stock_storage_type_putaway_strategy/models/stock_quant.py
```diff
@@ -0,0 +1,111 @@
...
+                if (
+                    loc_storage_type.max_weight and
+                    quant.package_id.pack_weight > loc_storage_type.max_weight
+                ):
+                    continue
+                # If we get here, it means there is a location storage type
+                #  allowing the package into the location
+                allowed = True
+                break
+            if not allowed:
```
jgrandguillaume, line 86: @grindtildeath I think here we should display different messages according to why it breaks. Having always the same error message is very confusing. In my case it happens trying to post an inventory with several quants in a bin with "only_empty" and I didn't understand what was wrong for a while.

stock_storage_type_putaway_strategy/models/stock_location.py
```diff
@@ -0,0 +1,199 @@
...
+        putaway_location = super()._get_putaway_strategy(product)
+        quant = self.env.context.get("storage_quant")
+        return self._get_pack_putaway_strategy(
+            putaway_location, quant, product
+        )
+
+    def _get_pack_putaway_strategy(self, putaway_location, quant, product):
+        package_storage_type = False
+        if quant:
+            package_storage_type = quant.package_id.package_storage_type_id
```
jgrandguillaume, line 61: if package_storage_type_id exists on package_id take it, otherwise look at the one stored on the quant

grindtildeath:
Merged into #12

PR #17, opened 2020-03-18
[13.0] Improve stock_available_to_promise_release

simahawk:
* use qty on hand
* add horizon to calculate qty
* compute qty by date (based on horizon)

### DIFF
--- a/stock_available_to_promise_release/models/res_company.py
+++ b/stock_available_to_promise_release/models/res_company.py
@@ -1,0 +1,13 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    stock_reservation_horizon = fields.Integer(
+        default=0,
+        help="Compute promised quantities for order planned to be shipped "
+        "until this number of days from today.",
+    )
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -50,7 +50,7 @@ class StockMove(models.Model):
             return 0.0
         available = self.product_id.with_context(
             location=self.warehouse_id.lot_stock_id.id
-        ).virtual_available
+        ).qty_available
         return max(
             min(available - self._previous_promised_qty(), self.product_qty), 0.0
         )
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -3,7 +3,7 @@
 
 from odoo import api, fields, models
 from odoo.osv import expression
-from odoo.tools import float_compare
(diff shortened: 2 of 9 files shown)


stock_available_to_promise_release/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    stock_reservation_horizon = fields.Integer(
+        default=15,
+        help="Compute reserved quantities for order planned to be shipped "
+        "up till a certain period in days.",
```
grindtildeath:
```suggestion
        "until this number of days.",
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -49,26 +49,47 @@ def _ordered_available_to_promise(self):
...
-        ).virtual_available
+            **self._order_available_to_promise_qty_ctx()
+        ).qty_available
         return max(
             min(available - self._previous_promised_qty(), self.product_qty), 0.0
         )
 
+    def _order_available_to_promise_qty_ctx(self):
+        return {
+            "location": self.warehouse_id.lot_stock_id.id,
```
grindtildeath, line 62: Maybe add a comment to explain this context is used in `_get_domain_locations`, took me some time to figure it out.

stock_available_to_promise_release/models/res_company.py
```diff
@@ -0,0 +1,13 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    stock_reservation_horizon = fields.Integer(
+        default=15,
```
grindtildeath: Maybe extract this to a `_default_stock_reservation_horizon` method to allow easier override. Also I feel like the default value should be 0 instead of a defined number...

grindtildeath (commented):
Small details, LG overall :+1:

stock_available_to_promise_release/models/res_company.py
simahawk: there's no need for a method: you can override it by overriding the field.
ok for `0` by default.

stock_available_to_promise_release/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class Company(models.Model):
+    _inherit = "res.company"
+
+    stock_reservation_horizon = fields.Integer(
+        default=0,
+        help="Compute reserved quantities for order planned to be shipped "
```
guewen:
```suggestion
        help="Compute promised quantities for order planned to be shipped "
```

guewen (approved):

PR #18, opened 2020-03-19
[13.0] Add stock_picking_type_shipping_policy - alpha

grindtildeath:
This module adds a Shipping Policy field on Operations Types in order to force
a specific Shipping Policy on Pickings according to their types.

### DIFF
--- a/stock_picking_type_shipping_policy/models/stock_picking_type.py
+++ b/stock_picking_type_shipping_policy/models/stock_picking_type.py
@@ -2,12 +2,6 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 from odoo import api, fields, models
 
-SHIPPING_POLICY_SELECTION = [
-    ("procurement_group", "Take from procurement group"),
-    ("force_as_soon_as_possible", "Force to as soon as possible"),
-    ("force_all_products_ready", "Force to when all products are ready"),
-]
-
 
 class StockPickingType(models.Model):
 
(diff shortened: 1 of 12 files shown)


stock_picking_type_shipping_policy/models/stock_picking_type.py
```diff
@@ -0,0 +1,25 @@
...
+
+    _inherit = "stock.picking.type"
+
+    @api.model
+    def _default_shipping_policy(self):
+        return "procurement_group"
+
+    shipping_policy = fields.Selection(
+        SHIPPING_POLICY_SELECTION,
+        default=lambda r: r._default_shipping_policy(),
```
simahawk, line 24: I think you don't need to split selection options to a var. You could delegate them to a method instead.
You don't need the default method but if you move options to a method then it's better if you keep it.

stock_picking_type_shipping_policy/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module adds a Shipping Policy field on Operations Types in order to force
+a specific Shipping Policy on Pickings according to their types.
```
simahawk, line 2: maybe worth to mention the use case?

simahawk (changes requested):
LG, just minor remarks

stock_picking_type_shipping_policy/readme/DESCRIPTION.rst
jgrandguillaume: Could be:

This is especially useful if you use a pick-pack-ship setup with the release of operation (stock_available_to_promise_release) module along side with the stock_routing_operation that may split operations by zone of the warehouse. In that case, you want to be sure the pack operations will wait all different picks to be processed before releasing the availability of the pack operation.

jgrandguillaume (approved):
LGTM a part from my comments

Tonow-c2c (approved):

simahawk (approved):

PR #19, opened 2020-03-20
[13.0] stock_available_to_promise_release: split picking on release

simahawk:
~~includes #17 that will be merged soon.
Please review last commit only.~~

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -144,7 +144,13 @@ class StockMove(models.Model):
                     # we don't want to deliver unless we can deliver all at
                     # once
                     continue
-                move.with_context(release_available_to_promise=True)._split(remaining)
+                move.with_context(release_available_to_promise=True)._release_split(
+                    remaining
+                )
+
+            if not move.picking_id.printed:
+                # Make sure the flag is set even if no split happens.
+                move.picking_id.printed = True
 
             values = move._prepare_procurement_values()
             procurement_requests.append(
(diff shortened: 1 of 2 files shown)


stock_available_to_promise_release/models/stock_move.py
```diff
@@ -145,3 +174,18 @@ def _run_stock_rule(self):
...
+        """Split move and create a new picking for it.
+
+        Instead of splitting the move and leave remaining qty into the same picking
+        we move it to a new one so that we can release it later as soon as
+        the qty is available.
+        """
+        # Rely on `printed` flag to make _assign_picking create a new picking.
+        # See `stock.move._assign_picking` and
+        # `stock.move._search_picking_for_assignation`.
+        self.picking_id.printed = True
```
guewen: Maybe worth to check and do the update only if `printed` is not already True, we can avoid several update queries in case of several moves split in the same picking

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -145,3 +174,18 @@ def _run_stock_rule(self):
...
+        new_move = self.browse(self._split(remaining_qty))
+        new_move._assign_picking()
```
guewen, line 194: is the call to `_assign_picking()` necessary? I would have thought `_split()` would call `_action_confirm()` which in turn calls `_assign_picking` (but maybe the conditions to call it are not fulfilled so you need to call it manually).

simahawk: ehh no, this is what we thought but I found that that flow happens only if you have moves to be assigned, which is not the case here. 
See
https://github.com/odoo/odoo/blob/f662c1582eb55a7f7fc1e8c0a3eea0e519f1aa64/addons/stock/models/stock_move.py#L1013
and 
https://github.com/odoo/odoo/blob/f662c1582eb55a7f7fc1e8c0a3eea0e519f1aa64/addons/stock/models/stock_move.py#L1034
Picking and picking type are already there because the move is copied.

guewen: :+1: alternatively, we could have emptied the `picking_id` with an override of `_prepare_move_split_vals` (https://github.com/odoo/odoo/blob/f662c1582eb55a7f7fc1e8c0a3eea0e519f1aa64/addons/stock/models/stock_move.py#L1442) when we have the `release_available_to_promise` context key, but calling `_assign_picking` should work as well.

simahawk: I added a comment

stock_available_to_promise_release/models/stock_move.py
simahawk: it's unlikely to happen AFAIS, I add the check anyhow.

guewen (approved):

PR #21, opened 2020-03-23
stock_available_to_promise_release: fix settings and qty computation

simahawk:


guewen (approved):

simahawk:
weird, `test_defer_creation_uom` is failing but I see no relation w/ this change.
Could be related to some change on 13.0 but I'm not able to reproduce w/ latest checkout locally :thinking:

simahawk:
@guewen can you pass by again to review qty fix?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -75,21 +75,46 @@ def _promise_reservation_horizon_date(self):
...
+        # either the move has to be released
+        # and priority is higher than the current one
+        domain_not_released = [
+            ("need_release", "=", True),
+            ("date_priority", "<", self.date_priority),
+        ]
+        # or it has been released already
+        # and the picking is printed
+        # and is not canceled or done
+        domain_released = [
```
guewen, line 100: oooh this is why we were using the `virtual_qty` and not the `qty_available`... now we have to search for much more moves :-(

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -75,21 +75,46 @@ def _promise_reservation_horizon_date(self):
...
+            ("need_release", "=", False),
+            ("picking_id.printed", "=", True),
```
guewen: Why do we need to search what is printed? I don't understand the relation, why a stock.picking that we did not print would not be considered in the quantity and why a printed one would be considered?
Besides this, this would be a performance issue as the ORM would search all the picking_ids where printed is True first, which can be a lot of records after some time.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -75,21 +75,46 @@ def _promise_reservation_horizon_date(self):
...
+            ),
+        ]
+        return expression.AND(
+            [base_domain, expression.OR([domain_not_released, domain_released])]
+        )
 
     def _previous_promised_qty(self):
         previous_moves = self.search(
             expression.AND(
+                # TODO: `!=` is suboptimal, consider filter out on recordset
```
guewen: I'm not sure it's suboptimal in this situation, maybe worth to check the query plan.

stock_available_to_promise_release/models/stock_move.py
simahawk: this is what we came up to w/ Joël. The point is that now we have the split of the move + split of the picking, thus we should look for moves that are attached to pickings that are not processed yet. OTOH we can find a way to do it better.

stock_available_to_promise_release/models/stock_move.py
jgrandguillaume: True, but unfortunately we can't use it anymore. When we release something, we must know its there, so having incoming shipment is not good.
We could use it by doing so though:
 Virtual qty [domain need_release=True&date_priority<self.available] - incoming qty 

Not sure it's better, that I cannot tell.

stock_available_to_promise_release/models/stock_move.py
jgrandguillaume: So on my side, I explain this with the way we account the moves to deduce from the current stock. So qty order to promise is:
 * Total in stock - Move to release (need_realease=True) - Move released not yet done (Printed=True as we set it on release now & state not done)

If @guewen you have any better domain for this I take it :) !

simahawk:
@guewen @jgrandguillaume updated

guewen (approved):

PR #27, opened 2020-05-01
[13.0] Add compatibility module for stock_picking_type_shipping_policy and...

guewen:
... stock_picking_group_by_partner_by_carrier (of OCA/stock-logistics-workflow).

In stock_picking_group_by_partner_by_carrier, the override of the method
_search_picking_for_assignation adds a domain on:

    ("move_type", "=", self.group_id.move_type)

(self is the move to assign in a picking).

When the move_type of the procurement group is "direct", the transfer is
normally created with the move_type "direct": when a first move is
assigned to a picking and none is found, it creates one with type
"direct", the second move should find the same and be assigned to it.

This is not what happens, with stock_picking_type_shipping_policy
configured to "force_all_product_ready", as it changes the way the first
stock.picking is created:

    def _get_new_picking_values(self):
        res = super()._get_new_picking_values()
        picking_type = self.mapped("picking_type_id")
        if picking_type.shipping_policy == "force_as_soon_as_possible":
            res["move_type"] = "direct"
        elif picking_type.shipping_policy == "force_all_products_ready":
            res["move_type"] = "one"
        return res

The first stock.picking is created with a move_type of "one", so when
the second move search a stock.picking, it can't find it as it looks for
a "direct" one.

This glue module changes the domain of the grouping to take into account the
picking type's shipping policy when both modules are installed.

- [ ] Depends on https://github.com/OCA/stock-logistics-workflow/pull/641


stock_picking_type_shipping_policy_group_by/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Picking Type Shipping Policy -" " Group By Partner and Carrier",
```
sebalix:
```suggestion
    "name": "Stock Picking Type Shipping Policy - Group By Partner and Carrier",
```

sebalix (approved):
Small comment, :+1:

jgrandguillaume (approved):
LGMT

PR #28, opened 2020-05-20
[13.0] Add stock_move_source_relocate

guewen:
Add relocation rules for moves.

Some use cases:

* Handle all the replenishments at the same place
* Trigger minimum stock rules or DDMRP buffers in one location

Behavior:

* When we try to assign a stock move and the move is not available, a
  rule matching the source location (sub-locations included), the picking
  type and an optional domain is searched
* If a relocation is found, the move source location is updated with
  the new one
* If the move was partially available, it is split in 2 parts:

 * one available part which keeps its source location
 * one confirmed part which is updated with the new source location


jgrandguillaume (approved):
Tested and works as expected

PR #31, opened 2020-06-12
[13.0] Add Dynamic Routing

guewen:
Moved from https://github.com/OCA/stock-logistics-warehouse/pull/788

# Description

Standard Stock Routes explain the steps you want to produce whereas the
“Dynamic Routing” defines how operations are grouped according to their final
source and destination location.

This allows for example:

* To parallelize transfers in two locations of a warehouse, splitting
  them in two different operation type
* To define pre-picking (wave) in some sub-locations, then roundtrip picking of
  the sub-location waves

Context for the use cases:

In the warehouse, you have a High-Bay which requires to place goods in a
handover when you move goods in or out of it. The High-Bay contains many
sub-locations.

A product can be stored either in the High-Bay, either in the Shelving zone.

When picking:

When there is enough stock in the Shelving, you expect the moves to have the
usual Pick(Highbay)-Pack-Ship steps. If the good is picked from the High-Bay, you will
need an extra operation: Pick(Highbay)-Handover-Pack-Ship.

This is what this feature is doing: on the High-Bay location, you define
a "routing rule". A routing rule selects a different operation type for the move.
The extra transfer will have the selected operation type, and be added
dynamically, on reservation, before the chain of moves.

When putting away:

A put-away rule targets the High-Bay location.
An operation Input-Highbay is created. You expect Input-Handover-Highbay.

You can configure a dynamic routing for the put-away on the High-Bay Location.
The operation type of the new Handover move will the one of the matching routing rule,
and its destination will be the destination of the operation type.

# Try on runbot

* In Inventory Settings, activate:

  * Storage Locations
  * Multi-Warehouses
  * Multi-Step Routes

The initial setup in the demo data contains locations:

* WH/Stock/Highbay
* WH/Stock/Highbay/Bin 1
* WH/Stock/Highbay/Bin 2
* WH/Stock/Handover

The "Highbay" location (and children) is configured to:

* create a pull routing transfer from Highbay to Handover when
  goods are taken from Highbay (using a new picking type Highbay → Handover)
* create a push routing transfer from Handover to Highbay when
  goods are put to Highbay (using a new picking type Handover → Highbay)

Steps to try the Pull Routing Transfer:

* In the main Warehouse, configure outgoing shipments to "Send goods in output and then deliver (2 steps)"
* Inventory a product, for instance "[FURN_8999] Three-Seat Sofa", add 50 items in "WH/Stock/Highbay/Bay A/Bin 1", and nowhere else
* Create a sales order with 5 "[FURN_8999] Three-Seat Sofa", confirm
* You'll have 3 transfers; a new one has been created dynamically for Highbay -> Handover.

Steps to try the Push Routing Transfer:

* In the "WH/Stock" location, create a Put-Away Strategy with:

  * "[DESK0004] Customizable Desk (Aluminium, Black)" to location "WH/Stock/Highbay/Bay A/Bin 1"
  * "[E-COM06] Corner Desk Right Sit" to location "WH/Stock/Shelf 1"

* Create a new purchase order of:

  * 5 "[DESK0004] Customizable Desk (Aluminium, Black)"
  * 5 "[E-COM06] Corner Desk Right Sit"

* Confirm the purchase
* You'll have 2 transfers:

  * one to move DESK0004 from Supplier → Handover and E-COM06 from Supplier → Shelf 1
  * one waiting on the other to move DESK0004 from Handover → WH/Stock/Highbay/Bay A/Bin 1 (the final location of the put-away)


stock_dynamic_routing/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Camptocamp (https://www.camptocamp.com)
+{
+    "name": "Stock Dynamic Routing",
+    "summary": "Dynamic routing for special locations",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-warehouse",
```
simahawk: `/wms` now ;)

stock_dynamic_routing/models/stock_routing.py
```diff
@@ -0,0 +1,262 @@
...
+
+        # free memory used for the cache
+        self.__cached_is_rule_valid_for_move.cache_clear()
+        return result
+
+    # Do not use ormcache, which would invalidate cache of other workers every
+    # time we clear it. We only need a local cache used for the duration of the
+    # execution of
+    @lru_cache()
+    def __cached_is_rule_valid_for_move(self, rule, move):
```
simahawk, line 244: why double underscore?

stock_dynamic_routing/tests/test_routing_pull.py
```diff
@@ -0,0 +1,911 @@
+# Copyright 2019 Camptocamp (https://www.camptocamp.com)
+
+from odoo.tests import common
+
+
+class TestRoutingPullCommon(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 9:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
on all classes possibly

simahawk (approved):
:brain: on :fire: :sweat_smile:

stock_dynamic_routing/models/stock_routing.py
guewen: I want to use a method for the `@lru_cache`, but in practice this method is really an internal thing of the algorithm, not meant to be overloaded in any way. Using double underscore makes it hard to overload (you would have to use `_StockRouting__cached_is_rule_valid_for_move` IIRC). I should write it in the docstring.

simahawk: yeah, that was the question: why making hard to override? :+1:

stock_dynamic_routing/models/stock_move.py
```diff
@@ -0,0 +1,443 @@
...
+            move._assign_picking()
+            # Note: we have to call _action_assign() here because if the move
+            # has been split because of partial availability, we want to ensure
+            # to reserve the move which has been "routed" first. Even if
+            # _action_assign() is called again, it should not be an issue
+            # because the move's state will be "assigned" and will be excluded
+            # by the method.
+            move.with_context(exclude_apply_dynamic_routing=True)._action_assign()
+
+        pickings_to_check_for_emptiness._dynamic_routing_handle_empty()
```
florian-dacosta, line 284: If I understand correctly, the routed moves are assigned 3 times.
The first, which is rollbacked, to know if yes or no a routing needs to be applied.
A second, right after the routing is applied, to ensure the reserved products are the same as during the 1st action_assign call.
And finally a third in the original `_action_assign` call.
To avoid this 3rd one, we could here return the routed moves.
Then in `_split_and_apply_routing` we could return something like `moves - routed_moves`.
Then, only the moves, which already had their routing applied will be returned to `_action_assign'.
So at the end, `_action_assign` will have been called : 
once for moves with no routing rules.
twice (one roll backed) on moves that already had been routed before
twice on the moves that just had been routed. (instead of 3 times).

What do you think?

Else it seems all good to me

florian-dacosta (approved):

stock_dynamic_routing/models/stock_move.py
guewen: The method `_action_assign` ignores moves with a state `assigned` and we know the routed moves have this state (they would not be routed if not assigned, and they can't be partially available because they are split to be available).

simahawk:
@guewen would you like to squash some commits?

PR #32, opened 2020-06-29
[13.0] sale_delivery_carrier_preference

grindtildeath:
This module allows to define preferred shipping methods for sales in order
to fine tune the selection of proper delivery carrier according to sale order
properties.

Go to Sales > Configuration > Sale orders > Preferred Shipping Methods to
define new Preferred Shipping Methods.

A Preferred Shipping Method can either define a specific Delivery Method or
use the partner's defined Delivery Method.

These Preferred Shipping Methods will then be filtered out according to the
order's shipping weight.


jgrandguillaume (approved):
LGTM

jbaudoux (approved):
LGTM

PR #33, opened 2020-07-01
[13.0] Allow to select locations with 'none' strategy in putaway sequence

guewen:
The putaway sequence for a package storage types defines where the goods
can be put and in which order.

For instance, if the "cardbox" storage type has this sequence:

1. Stock/Cardboxes Shelving: ordered locations
2. Stock/Cardboxes Reserve: ordered locations

It will first try to put the goods in Bin 1, then Bin 2 and finally Bin
3 if the others are full.

This commit aims to support the strategies with a 'none' strategy has
valid locations in the sequences, such as:

1. Stock: None
2. Stock/Cardboxes Shelving: ordered locations
3. Stock/Cardboxes Reserve: ordered locations

In this example, if a move line with a cardbox has a destination Stock,
no strategy will be applied, if the move line has a destination Cardbox
Shelving, the ordered location strategy is applied.

It allows to use a sort of buffer: the default destination is Stock,
then the destination is updated to Stock/Cardbox Reserve (upon the
choice of a user!) and a new computation of the putaway is asked.

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -203,7 +203,8 @@ class StockLocation(models.Model):
         if not package_locations:
             return dest_location
 
-        for pref_loc in package_locations.mapped("location_id"):
+        for package_sequence in package_locations:
+            pref_loc = package_sequence.location_id
             if (
                 pref_loc.pack_putaway_strategy == "none"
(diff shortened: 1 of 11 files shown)


grindtildeath (approved):
LGTM :+1:

jbaudoux (approved):
LGTM

PR #35, opened 2020-07-03
[13.0] Ignore unavailable carriers in preferences

guewen:
* in the 'partner' rule, ignore empty delivery method otherwise the
  wizard shows an empty selection
* filter the preferred carriers by the available carrier rules (country, ...)


jgrandguillaume (approved):
Thx for the fix

PR #36, opened 2020-07-07
Add shipping line on sale order confirmation

guewen:
The former implementation sets the preferred carrier, but it doesn't
completely reflects what is being done by the wizard.

This change adds the missing bits:

* When a carrier_id is set, it adds the shipping line with the computed
  price.
* If there is a warning message, it stores in on the sales order.
* If there is an error message, it prevents the confirmation of the
  sales order as no shipping could be added.
* Fix the check based on 'product_id.type != service', which would
  not add a shipping method as soon as we have a single service. There
  is a dedicated flag "is_delivery" to track the delivery line.

### DIFF
--- a/sale_delivery_carrier_preference/models/sale_order.py
+++ b/sale_delivery_carrier_preference/models/sale_order.py
@@ -1,7 +1,7 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
-from odoo import api, fields, models
+from odoo import _, api, exceptions, fields, models
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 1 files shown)


sale_delivery_carrier_preference/models/sale_order.py
```diff
@@ -1,26 +1,43 @@
...
-            ):
-                order.carrier_id = order.get_preferred_carrier()
-        return super().action_confirm()
+            if order.carrier_id or any(line.is_delivery for line in order.order_line):
+                continue
+            carrier = order.get_preferred_carrier()
+            if not carrier:
+                continue
+            vals = carrier.rate_shipment(order)
+            if vals.get("success"):
```
grindtildeath: :warning: `rate_shipment` will not necessarily return a dict if the method is not implemented for a specific carrier: https://github.com/odoo/odoo/blob/13.0/addons/delivery/models/delivery_carrier.py#L138-L159

sale_delivery_carrier_preference/models/sale_order.py
```diff
@@ -1,26 +1,43 @@
...
+                delivery_price = vals["price"]
+                order.set_delivery_line(carrier, delivery_price)
+                order.recompute_delivery_price = False
+                if vals.get("warning_message"):
+                    order.delivery_message = vals["warning_message"]
+            else:
+                raise exceptions.UserError(
```
grindtildeath, line 33: This will restrict usage of any `delivery.carrier` that doesn't have a rate integration. I'm not sure what must be done in such case but I'm not sure preventing the confirmation is right.

grindtildeath (changes requested):
Two things to pay attention to.

sale_delivery_carrier_preference/models/sale_order.py
guewen: Now it will raise an error only when success is explicitly set to False in a dict returned by `rate_shipment`.

grindtildeath (approved):

jgrandguillaume (approved):
LGTM thx !

PR #37, opened 2020-07-14
[13.0] Create module delivery_carrier_dangerous_good_warning

grindtildeath:
Depends on:
 - [ ] https://github.com/OCA/community-data-files/pull/80/

This module allows to display a warning on sale order and stock picking when
the selected delivery carrier is tagged to display the warning when it contains
a product being flagged as 'is_dangerous_good' according to the European
Agreement concerning the International Carriage of Dangerous Goods
by Road (ADR).

On any shipping method, one can activate warning by marking the checkbox
"Dangerous goods warning".

Screenshots:
image

image

image


l10n_eu_product_adr_delivery_carrier/models/delivery_carrier.py
```diff
@@ -0,0 +1,65 @@
...
+
+    adr_limited_amount_ids = fields.Many2many(
+        "limited.amount",
+        string="Warns according to ADR limited amount",
+        help="Raise warning when following limited amount are defined on "
+        "any product in the document.",
+    )
+
+    def _prepare_adr_warning(self, warnings, checked_object):
+        """Prepare warning dictionnary to be displayed according to warnings
```
simahawk:
```suggestion
        """Prepare warning dictionary to be displayed according to warnings
```

l10n_eu_product_adr_delivery_carrier/models/delivery_carrier.py
```diff
@@ -0,0 +1,65 @@
...
+        checked_object_str = ""
+        if checked_object not in ("move", "order_line"):
+            raise UserError(
+                _(
+                    "checked_object parameter %s used to call "
+                    "_prepare_dangerous_goods_warning is invalid."
+                )
+                % checked_object
+            )
+        elif checked_object == "move":
```
simahawk: I'd drop this part and pass the string as an argument. For plural issue, just use "move(s)".

l10n_eu_product_adr_delivery_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,28 @@
...
+        matching those defined on the carrier"""
+        warnings = {}
+        if self.carrier_id and self.carrier_id.adr_limited_amount_ids:
+            for move in self.move_lines:
+                if (
+                    move.product_id.limited_amount_id
+                    in self.carrier_id.adr_limited_amount_ids
+                ):
+                    warnings[move] = move.product_id.limited_amount_id
+                # Other "ADR" M2m fields to be defined on carrier
```
simahawk: you could add `_adr_validators` (or similar) which returns the callables for these checks. 
Then it's easy to add more validators w/out touching this code.

simahawk (changes requested):

simahawk:
non blocking comments, being alpha ;)

l10n_eu_product_adr_delivery_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def _check_adr(self):
+        """Display a warning if any move has a product with ADR settings
+        matching those defined on the carrier"""
+        warnings = {}
+        if self.carrier_id and self.carrier_id.adr_limited_amount_ids:
+            for move in self.move_lines:
+                if (
+                    move.product_id.limited_amount_id
+                    in self.carrier_id.adr_limited_amount_ids
```
jbaudoux: How to exclude a preferred carrier rule based on the fact that the stock available to promise contains a dangerous goods? I cannot give a picking_domain (see https://github.com/OCA/wms/pull/38) with this condition while it was the purpose of that domain

jbaudoux (changes requested):
This is lacking the glue between picking_domain in the carrier selection and this feature to check if the shipped goods are dangerous goods

delivery_carrier_dangerous_goods_warning/models/dangerous_good_warning_mixin.py
```diff
@@ -0,0 +1,81 @@
...
+        compute="_compute_dangerous_goods_carrier_warning"
+    )
+    dangerous_goods_carrier_warning = fields.Text(
+        compute="_compute_dangerous_goods_carrier_warning"
+    )
+
+    def _compute_contains_dangerous_goods(self):
+        for doc in self:
+            doc.contains_dangerous_goods = any(
+                [l.product_id.is_dangerous_good for l in doc[self._line_field_name]]
```
jbaudoux, line 29: Why did you dropped the feature regarding adr_limited_amount_ids ? We need a way to be able to configure the type of dangerous goods a carrier is not allowed to transport.
Like DHL cannot ship "DG" goods but well "LG" goods

jbaudoux (changes requested):
Why you don't use the sales exception?
It is required to block the confirmation of the SO if the carrier does not allow "DG" goods. If the SO is created through website or EDI, nobody will see the warning and it will automatically confirm. Sales exception will block the SO and somebody will have to manually review it.

delivery_carrier_dangerous_goods_warning/models/dangerous_good_warning_mixin.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+
+    def _search_contains_dangerous_goods(self, operator, value):
+        if operator not in ["=", "!="] or not isinstance(value, bool):
+            raise UserError(_("Operation not supported"))
+        line_model = self[self.line_field_name]._name
```
jbaudoux: When I tried, I got this error:
```
delivery_carrier_dangerous_goods_warning/models/dangerous_good_warning_mixin.py", line 35, in _search_contains_dangerous_goods
    line_model = self[self.line_field_name]._name
AttributeError: 'stock.picking' object has no attribute 'line_field_name'
```

jbaudoux (changes requested):

delivery_carrier_dangerous_goods_warning/models/dangerous_good_warning_mixin.py
grindtildeath:
```suggestion
        line_model = self[self._line_field_name]._name
```

delivery_carrier_dangerous_goods_warning/models/dangerous_good_warning_mixin.py
```diff
@@ -0,0 +1,81 @@
...
+        dangerous_goods_lines = self.env[line_model].search(
+            [("product_id.is_dangerous_good", "=", True)]
+        )
+        return [
+            (
+                "id",
+                "in" if value else "not in",
+                dangerous_goods_lines.mapped(self.line_doc_m2o_field_name).ids,
```
grindtildeath:
```suggestion
                dangerous_goods_lines.mapped(self._line_doc_m2o_field_name).ids,
```

delivery_carrier_dangerous_goods_warning/models/dangerous_good_warning_mixin.py
grindtildeath: Both sale exception and limited amounts are done as specific :wink:

grindtildeath:
Not used anymore, closing.

bosd:
@grindtildeath What is your current workflow for shipping ADR goods?
Can a different carrier be (automatically) selected when an (website) SO contains ADR goods?

PR #38, opened 2020-07-16
[13.0] Rename module to delivery_carrier_preference

grindtildeath:
Depends on:

 - [x] https://github.com/OCA/product-attribute/pull/651


delivery_carrier_preference_release_operations/__manifest__.py
```diff
@@ -1,15 +1,19 @@
...
     "category": "Operations/Inventory/Delivery",
     "website": "https://github.com/OCA/delivery-carrier",
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["delivery"],
-    "data": ["security/ir.model.access.csv", "views/delivery_carrier_preference.xml"],
+    "depends": [
+        "delivery_carrier_preference",
```
simahawk: autoinstall maybe?

simahawk (commented):
code review only. LG. I guess the build is red because of missing deps.

TDu (approved):

delivery_carrier_preference_release_operations/models/stock_move.py
```diff
@@ -0,0 +1,34 @@
...
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    def release_available_to_promise(self):
+        existing_backorders = {
+            pick.id: pick.mapped("backorder_ids").ids
+            for pick in self.mapped("picking_id")
+        }
+        res = super().release_available_to_promise()
```
jbaudoux: This step will apply the pull rules and the pull rules depend on the carrier.
The carrier must be computed based on what is available to promise and updated before the call to super

delivery_carrier_preference/models/delivery_carrier_preference.py
```diff
@@ -0,0 +1,157 @@
...
+                ("max_weight", "=", 0.0),
+                "|",
+                ("carrier_id", "in", company_carriers.ids),
+                ("carrier_id", "=", False),
+            ]
+        )
+        carriers_ids = list()
+        for cp in carrier_preferences:
+            if (
+                picking is not None
```
jbaudoux: As you implemented, we cannot define rules at sales which set the default carrier when not manually chosen, and rules at release of operation that remap if weight < max_weight.
Either we should have 2 set of rules, or being able to flag what is applied at sales and only at sales. E.g. weight rule should not be applied at sales. Customer preferred carrier should not be applied at picking.

jbaudoux (changes requested):
Some issues while testing.

Additionally, at sales order confirmation, I don't have the carrier set on the SO. I miss the delivery line on the SO.

jbaudoux (changes requested):
The field "picking_domain" is missing in the shipping method views

jgrandguillaume (approved):
LGTM

jgrandguillaume:
@grindtildeath Please add the oca_dependenciess

delivery_carrier_preference/models/stock_move.py
```diff
@@ -41,6 +41,7 @@ def release_available_to_promise(self):
             existing_carriers[picking.id] = picking.carrier_id.id
 
             picking.add_preferred_carrier()
+            picking.group_id.carrier_id = picking.carrier_id
```
guewen: I don't think that's correct, because if we grouped several SO in the same group, we'll update the method for all.
It seems we reach a shadowy part of the ties between:

* stock_picking_group_by_partner_by_carrier
* stock_available_to_promise_release
* delivery_carrier_preference

I'm not sure delivery_carrier_preference can really be generic (also because of the weird rules it uses) maybe it should only act as a glue for our project.

Some thoughts, situation at this line:

* The grouping by partner and by carrier has been done already, several sales order can the same procurement group
* Despite this, we are conditionally changing the group carrier, so it has an impact on the routing of the other pickings potentially for other sales orders (but the other pickings are not updated with the same carrier, in any case we have an issue)
* The routing is applied the line after in `super().release_available_to_promise()`, the routing depends on the carrier (at least in our case), so we should have it on the group at this point
* Which means we should split the procurement group *here* if the carrier is different as the previous one
* We do not expect to have chained transfers before the release, so we should not have any other transfer to move in this group (?)
* In the current implementation, the procurement group is indeed split, but only after the routing is applied. I have the feeling that moving the split of the group before only could resolve the issue, but there might be a reason I miss having it after, maybe we need the release to be done for a reason?
* What if we have 4 SO, 4 transfers all grouped in the same procurement group with carrier A, we release the 4 transfers, 2 of them are changed to carrier B, do we have to reapply the group by carrier again from stock_picking_group_by_partner_by_carrier? :exploding_head:

guewen: Rereading stock_picking_group_by_partner_by_carrier, I see I wasn't correct about what it does: it groups all the moves in the same picking, not different OUT pickings in the same group. We can have several sales order in the same group, but in the same OUT picking (+ backorders).

guewen: So **if** we assume that we have only one OUT at a time for a procurement group, setting the carrier on the line may work.
However, this method is doing this:

1. Setting new carrier on group
2. apply release/routing
3. Loop on backorders and assign a new group to each if they had a different carrier

Which means we'll have one new group per backorder (even if we are not really supposed to have several at a time?)

It seems more simple to do this:

1. Create a new group for the current transfer if its carrier is different and the group has other transfers, set the carrier on the group
2. apply release/routing

guewen: The problem is... before calling super, we don't know yet what will be released or not, and yet, we need to know and apply the carrier before releasing because that's what applies the routing.

Options: 
1. find a hook to place the carrier update
2. change the weight computation to take only in account the "available to promise" quantity


edit: 2. is already implemented

guewen (approved):

PR #39, opened 2020-07-20
[13.0] stock_storage_type: Add package storage type on product

grindtildeath:
Package storage type can be defined on products to compute automatically
a package storage type on packages if no packaging is defined and
the package only contains a single product.

### DIFF
--- a/stock_storage_type/models/stock_quant_package.py
+++ b/stock_storage_type/models/stock_quant_package.py
@@ -20,8 +20,9 @@ class StockQuantPackage(models.Model):
     @api.depends(
         "product_packaging_id",
         "product_packaging_id.package_storage_type_id",
-        "single_product_id",
-        "single_product_id.product_package_storage_type_id",
+        "quant_ids",
+        "quant_ids.product_id",
+        "quant_ids.product_id.product_package_storage_type_id",
     )
     def _compute_package_storage_type_id(self):
         for pack in self:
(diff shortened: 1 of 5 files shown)


stock_storage_type/models/stock_quant_package.py
```diff
@@ -7,31 +7,39 @@ class StockQuantPackage(models.Model):
...
+        readonly=False,
+        help="Package storage type for put-away computation. Computed "
+        "automatically from the packaging if set, or from the product if"
+        "the package contains only a single product.",
+    )
+
+    @api.depends(
+        "product_packaging_id",
+        "product_packaging_id.package_storage_type_id",
+        "single_product_id",
```
guewen: where is this field defined?

grindtildeath: https://github.com/OCA/stock-logistics-workflow/blob/13.0/stock_quant_package_product_packaging/models/stock_quant_package.py#L23 :wink:

guewen (approved):

jbaudoux (approved):
LGTM

stock_storage_type/models/stock_quant_package.py
```diff
@@ -7,31 +7,39 @@ class StockQuantPackage(models.Model):
...
+        "single_product_id.product_package_storage_type_id",
```
guewen: Logs show:

ERROR odoodb_test odoo.osv.expression: Non-stored field stock.quant.package.single_product_id cannot be searched. 

I propose to use the depends of single_product_id

```suggestion
        "quant_ids",
        "quant_ids.product_id",
        "quant_ids.product_id.product_package_storage_type_id",
```

PR #40, opened 2020-07-20
Make stock_storage_type depend on stock_putaway_hook

guewen:
The putaway based on storage type *must* be applied at the very end
of the putaway rules, because it depends on the putaway location found.

If we execute it by calling super() and we added different putaway
strategies in StockLocation, we have no way to be sure it is the last
called.

That's why now it depends on stock_putaway_hook, which allow to search
for alternative put-away hooks, and call _putaway_strategy_finalizer at
the very end, where we can plug the storage type rules.

Stock Putaway Hook is added by https://github.com/OCA/stock-logistics-workflow/pull/687

### DIFF
--- a/stock_storage_type/__manifest__.py
+++ b/stock_storage_type/__manifest__.py
@@ -12,6 +12,7 @@
     "application": False,
     "installable": True,
     "depends": [
+        "stock_putaway_hook",
         "base_m2m_custom_field",
         "stock_quant_package_dimension",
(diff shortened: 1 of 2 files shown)


sebalix (approved):

Tonow-c2c (approved):

grindtildeath (approved):
LGTM :+1:

PR #41, opened 2020-08-04
[13.0] Create module delivery_carrier_warehouse

grindtildeath:
This module allows to define a delivery on stock warehouse.

If a partner does not have a default carrier defined, the 'Add shipping' wizard
will use the carrier defined on the warehouse as default.


jgrandguillaume (commented):
LGTM

PR #42, opened 2020-08-05
[13.0] Create module delivery_carrier_preference_dangerous_goods - alpha

grindtildeath:
Depends on:
 - [ ] #38 
 - [ ] #37 

This modules adds consideration of 'Dangerous goods' in the selection of preferred
delivery carrier.

In other words, if the delivery carrier is set to display a warning when the
delivery order contains dangerous goods, this delivery carrier will not be
selected in the computation of preferred delivery carrier.


grindtildeath:
Not used anymore, closing.

PR #43, opened 2020-08-12
Add printed flag on all released transfers

guewen:
## Current behavior

When we release an OUT stock move, the OUT transfer is set to printed.
The origin transfers (e.g. Pick+Pack) are not.

## Expected behavior

When we release an OUT stock move, all the origin transfer (e.g.
Pick+Pack) are set to printed.

## Reason

The original reason for setting "printed" to True, is that when
"_assign_picking()" is called on a stock.move, the stock.move can be
added only in a stock.picking which is not printed yet. In other terms,
"printed" means the transfer has been started and shouldn't be modified.
When we "release" moves, we actually planify the work to be done, and if
we release again new moves, we don't expect transfers of the 2 waves of
releases to be merged together.

If we set the flag "printed" on the Out transfer only, we can end up
with one Packing transfer leading to 2 Out transfers, which generally
doesn't make sense.

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -174,10 +174,6 @@ class StockMove(models.Model):
                 )._release_split(remaining)
                 backorder_links[new_move.picking_id] = move.picking_id
 
-            if not move.picking_id.printed:
-                # Make sure the flag is set even if no split happens.
-                move.picking_id.printed = True
-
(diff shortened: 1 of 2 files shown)


simahawk (approved):
Seems legit. Thanks for the explanation. As usual :)

jgrandguillaume (approved):
LGTNM thx ! much better with this.

PR #46, opened 2020-08-27
[IMP] various improvements to stock_available_to_promise_release

i-vyshnevska:


stock_available_to_promise_release/__manifest__.py
```diff
@@ -3,11 +3,11 @@
...
 {
     "name": "Stock Available to Promise Release",
-    "version": "13.0.1.3.0",
+    "version": "13.0.1.4.0",
     "summary": "Release Operations based on available to promise",
     "author": "Camptocamp,Odoo Community Association (OCA)",
     "category": "Stock Management",
-    "depends": ["stock"],
+    "depends": ["stock", "delivery"],
     "data": [
```
guewen: This is a huge change, previously, the module needed only stock, having a dependency on `delivery` means we have to have `sale` as well.  BTW, there is `sale_stock_available_to_promise_release` on purpose to avoid a dependency on sale... Could you use such a glue module for `delivery`, I'd be ok with moving the delivery stuff in `sale_stock_available_to_promise_release` though.

guewen (changes requested):

guewen (approved):

guewen:
@i-vyshnevska could you squash your fixup?

PR #47, opened 2020-09-01
[13.0] Search ordered available to promise

santostelmo:
We want to give the possibility to search stock moves by `Ordered Available to Promise`

Also in Stock Moves To Release tree view, we just want to show moves with quantities available to release.


jgrandguillaume (approved):
Works well

sebalix:
@jgrandguillaume the build issue came from an Ubuntu APT repository not available during the build, should be OK now I suppose

PR #49, opened 2020-09-14
Ensure compatibility between release and dynamic routing

guewen:
Add a module with tests to check a release of moves with dynamic routing
correctly sets the "printed" flag on all the chain.

When we release moves, we set the "printed" flag on the released
transfers, so we do not include any new moves (like a "wave" of
release).

As the dynamic routing is applied when we assign the moves, new moves
and transfers may be created after the "assign". We therefore have
to set the printed flag after it.


jgrandguillaume (approved):
Works well !

jgrandguillaume:
@guewen Ok to merge this one ?

simahawk (approved):

PR #50, opened 2020-09-23
stock_storage_type: option to make 'height' required on packages

sebalix:


simahawk (approved):

sebalix:
It seems I broke the tests but I do not understand yet why...
EDIT: unable to reproduce it locally, but I'm based on an older version of the repository. Anyway I don't think the changes here are part of the issue.

sebalix:
Retriggered the build, now tests pass ¯\_(ツ)_/¯

guewen (approved):

PR #53, opened 2020-10-06
[13.0] Fix unpredictable package storage type on packages

guewen:
From the information I collected, the requirements for the
`package_storage_type_id` field on the packages are:

* if we have a package storage type on the related product packaging,
  set it on the package
* else if we have only one product in the package and the product has a
  package storage type, set it on the package
* users must be able to change the storage type, so we cannot keep it
  in sync as a related field would do

The solution using a stored computed field which is set only once is not
reliable: if the computation of the field is called before the product
packaging is set (by the auto assign method), and we have a single
product in the package, the product's storage type is set. We expect to
have the packaging's storage type in this case.

Instead, use a normal non-computed field and set it from:

* when creating a new package with a product packaging or writing a
  product packaging, overwrite the storage type with the new one
* the auto_assign_packaging method sets the "single product" storage
  type as fallback when there is no product packaging on the package

### DIFF
--- a/stock_storage_type/models/stock_quant_package.py
+++ b/stock_storage_type/models/stock_quant_package.py
@@ -10,41 +10,10 @@ class StockQuantPackage(models.Model):
 
     package_storage_type_id = fields.Many2one(
         "stock.package.storage.type",
-        compute="_compute_package_storage_type_id",
-        store=True,
-        readonly=False,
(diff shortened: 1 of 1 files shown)


TDu (approved):

jgrandguillaume (approved):
Functional tests ok

PR #55, opened 2020-10-14
[13.0] stock_available_to_promise_release: Fix release of operation containing canceled moves

jbaudoux:
If a delivery contains moves from multiple sales order and one of the SO
is canceled, the canceled moves should not be split not moved to a
separate operation.

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -155,7 +155,7 @@ class StockMove(models.Model):
         for move in self:
             if not move.need_release:
                 continue
-            if move.state not in ("confirmed", "waiting"):
+            if move.state not in ("confirmed", "waiting", "done", "cancel"):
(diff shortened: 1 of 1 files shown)


jgrandguillaume (approved):
LGTM

lmignon (commented):
@jbaudoux Can you add test for this case?

sebalix:
Test should be added to `sale_stock_available_to_promise_release` I suppose

guewen:
> Test should be added to `sale_stock_available_to_promise_release` I suppose

Maybe we don't need to test the whole flow but cancel a move manually?

simahawk:
@jbaudoux ping :)

sebalix:
@simahawk @lmignon @jgrandguillaume I open https://github.com/OCA/wms/pull/174 as I could not push on the current branch.
The current PR should be closed.

PR #57, opened 2020-10-21
Fix manager access on stock storage types

guewen:
All users have group_adv_location or group_stock_multi_locations when the options are activated.
Normal users should not be allowed to edit these models, only actual managers


guewen:
@grindtildeath @jgrandguillaume can you check?

jgrandguillaume (approved):
All good!

grindtildeath (approved):

PR #58, opened 2020-10-23
[13.0] Add stock_storage_type_buffer 

guewen:
This is an extensions for putaway with storage types (provided by
``stock_storage_type``).

In a warehouse, some areas where goods are put cannot be accessed
directly and need to be moved to a buffer first (for instance, because
they need a special machine to move from the buffer to the bins).

This module allows to define buffers in front of a set of locations.
When a buffer contains something or has move lines that reaches it, all
the locations are removed from the put-away computation.

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -339,6 +339,12 @@ class StockLocation(models.Model):
     def select_allowed_locations(
         self, package_storage_type, quants, products, limit=None
     ):
+        """Filter allowed locations for a storage type
+
+        ``self`` contains locations already ordered according to the
+        putaway strategy, so beware of the return that must keep the
+        same order
+        """
         # We have package who may be placed in a stock.location
         #
         # 1. On the stock.location there are location_storage_type and on the
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -78,6 +78,15 @@ class StockLocation(models.Model):
         help="technical field: the pending incoming "
         "stock.move.lines in the location",
     )
+    out_move_line_ids = fields.One2many(
+        "stock.move.line",
+        "location_id",
+        domain=[
+            ("state", "in", ("waiting", "confirmed", "partially_available", "assigned"))
+        ],
+        help="technical field: the pending outgoing "
+        "stock.move.lines in the location",
+    )
     location_will_contain_lot_ids = fields.Many2many(
         "stock.production.lot",
         store=True,
@@ -149,12 +158,17 @@ class StockLocation(models.Model):
             ) | rec.mapped("in_move_line_ids.lot_id")
 
     @api.depends(
-        "quant_ids.quantity", "in_move_ids", "in_move_line_ids",
+        "quant_ids.quantity",
+        "out_move_line_ids.qty_done",
+        "in_move_ids",
+        "in_move_line_ids",
     )
     def _compute_location_is_empty(self):
         for rec in self:
             if (
                 sum(rec.quant_ids.mapped("quantity"))
+                - sum(rec.out_move_line_ids.mapped("qty_done"))
+                > 0
                 or rec.in_move_ids
                 or rec.in_move_line_ids
             ):
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -386,8 +386,10 @@ class StockLocation(models.Model):
 
         # NOTE: self.ids is ordered as expected, so we want to filter the valid
         # locations while preserving the initial order
-        valid_location_ids = [id_ for id_ in self.ids if id_ in valid_locations.ids]
-        valid_locations = self.browse(valid_location_ids)
+        valid_location_ids = set(valid_locations.ids)
+        valid_locations = self.browse(
+            id_ for id_ in self.ids if id_ in valid_location_ids
+        )
         valid_locations = valid_locations._select_final_valid_putaway_locations(
             limit=limit
         )
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -312,6 +312,30 @@ class StockLocation(models.Model):
         )
         return pertinent_loc_storagetype_domain
 
+    def _allowed_locations_for_location_storage_types(
+        self, location_storage_types, quants, products
+    ):
+        valid_location_ids = set()
+        for loc_storage_type in location_storage_types:
+            location_domain = loc_storage_type._domain_location_storage_type(
+                self, quants, products
+            )
+            _logger.debug("pertinent location domain: %s", location_domain)
+            locations = self.search(location_domain)
+            valid_location_ids |= set(locations.ids)
+        return self.browse(valid_location_ids)
+
+    def _select_final_valid_putaway_locations(self, limit=None):
+        """Return the valid locations using the provided limit
+
+        ``self`` contains locations already ordered and contains
+        only valid locations.
+        This method can be used as a hook to add or remove valid
+        locations based on other properties. Pay attention to
+        keep the order.
+        """
+        return self[:limit]
+
     def select_allowed_locations(
         self, package_storage_type, quants, products, limit=None
     ):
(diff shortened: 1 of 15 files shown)


stock_storage_type_buffer/models/stock_location_storage_buffer.py
```diff
@@ -0,0 +1,113 @@
...
+        buffer_locations = self.buffer_location_ids.leaf_location_ids
+        return any(location.location_is_empty for location in buffer_locations)
+
+    def _help_message(self):
+        buffer_locations = self.buffer_location_ids.leaf_location_ids
+        buffer_location_count = len(buffer_locations)
+        leaf_locations = self.location_ids.leaf_location_ids
+        leaf_location_count = len(leaf_locations)
+
+        message = "<p>The buffer locations:<ul><br/>"
```
simahawk: would be better to use a qweb template

stock_storage_type_buffer/models/stock_location_storage_buffer.py
```diff
@@ -0,0 +1,113 @@
...
+                "<p>The buffers have no capacity because they all contain"
+                " goods or will contain goods due to move lines reaching them.</p>"
+            )
+        return message
+
+    def name_get(self):
+        result = []
+        for record in self:
+            name = ", ".join(
+                self.buffer_location_ids[: self.MAX_LOCATIONS_IN_NAME_GET].mapped(
```
simahawk:
```suggestion
                rec.buffer_location_ids[: self.MAX_LOCATIONS_IN_NAME_GET].mapped(
```

stock_storage_type_buffer/models/stock_location_storage_buffer.py
```diff
@@ -0,0 +1,113 @@
...
+                    "name"
+                )
+            )
+            if len(self.buffer_location_ids) > self.MAX_LOCATIONS_IN_NAME_GET:
```
simahawk:
```suggestion
            if len(rec.buffer_location_ids) > self.MAX_LOCATIONS_IN_NAME_GET:
```

stock_storage_type/models/stock_location.py
```diff
@@ -312,9 +312,39 @@ def _domain_location_storage_type_constraints(
...
+        for loc_storage_type in location_storage_types:
+            location_domain = loc_storage_type._domain_location_storage_type(
+                self, quants, products
+            )
+            _logger.debug("pertinent location domain: %s", location_domain)
+            locations = self.search(location_domain)
+            valid_location_ids |= set(locations.ids)
+        return self.browse(valid_location_ids)
+
+    def _select_valid_locations(self, limit=None):
```
simahawk: "select valid locations"... for what? I think is not clear here. Maybe `_select_valid_putaway_locations`?

simahawk (changes requested):

guewen:
@simahawk could you review my fixups, then I squash them?

stock_storage_type_buffer/models/stock_location_storage_buffer.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def buffers_have_capacity(self):
+        self.ensure_one()
+        buffer_locations = self.buffer_location_ids.leaf_location_ids
+        return any(location.location_is_empty for location in buffer_locations)
+
+    def _help_message_location_items(self, records):
+        items = records[: self.MAX_LOCATIONS_IN_HELP].mapped("display_name")
+        if len(records) > self.MAX_LOCATIONS_IN_HELP:
+            items.append("... and {} other locations.".format(
```
simahawk: make it translatable or move this to the template as well using a `remaining_[locations|buffer_locations]_count` key in rendering values.

guewen: I feel it would imply too much logic in the template with the qweb magic keys, so I ended up with providing the "remaining count" to the template, which seems simple and clean actually. Thanks for pointing this

guewen: @simahawk and I pushed a last fixup that use a sub-template to avoid repeating "... and {} other locations." in the 2 places

stock_storage_type_buffer/templates/stock_location_storage_buffer_templates.xml
```diff
@@ -0,0 +1,49 @@
...
+    <template id="storage_buffer_help_message">
+        <p>The buffer locations
+            <ul class="mt8">
+                <t t-foreach="buffer_locations" t-as="loc_text">
+                    <li>
+                        <span itemprop="name" t-esc="loc_text" />
+                    </li>
+                </t>
+                <t t-if="buffer_locations_remaining">
+                    <t t-set="remaining" t-value="buffer_locations_remaining" />
```
simahawk, line 14: or
```xml
<t t-call="stock_storage_type_buffer.storage_buffer_help_message_other_locations">
    <t t-set="remaining" t-value="buffer_locations_remaining" />
</t>
```
not blocking, of course

simahawk (approved):

stock_storage_type_buffer/templates/stock_location_storage_buffer_templates.xml
guewen: Oh nice! so much better indeed, fixed and squashed

jgrandguillaume (approved):
This is just so elegant! Thanks for the great job here.

stock_storage_type_buffer/models/stock_location_storage_buffer.py
```diff
@@ -0,0 +1,88 @@
...
+                    "reaching them.</p>"
+                )
+                continue
+
+            record.help_message = record._help_message()
+
+    def buffers_have_capacity(self):
+        self.ensure_one()
+        buffer_locations = self.buffer_location_ids.leaf_location_ids
+        return any(location.location_is_empty for location in buffer_locations)
```
jbaudoux, line 50: The buffer should also be considered empty as soon as
sum(location.out_move_lines_ids.qty_done) = sum(location.quant_ids.mapped("quantity"))
Probably to put in _compute_location_is_empty.
So when someone takes what is on the buffer by writing the done quantity on the move line, the buffer is physically empty and we should be able to put again something on it.
This detail is quite important for the selection of the right set of allowed locations.

stock_storage_type_buffer/models/stock_location.py
```diff
@@ -0,0 +1,54 @@
...
+        locations based on other properties. Pay attention to
+        keep the order.
+        """
+        if self:  # short-circuit computation when already 0 valid
+            self = self._filter_locations_blocked_by_buffer()
+        return super()._select_final_valid_putaway_locations(limit=limit)
+
+    def _filter_locations_blocked_by_buffer(self):
+        valid_location_ids = set(self.ids)
+        for storage_buffer in self.env["stock.location.storage.buffer"].search([]):
```
jbaudoux, line 29: You compute all the buffers availability even if the valid locations are not guarded by any buffer. I would have done it the other way around by checking each valid location if it is guarded by a buffer and filter it out if the buffer is not empty.

The pro:
- You compute only the required buffers. For each buffer location_is_empty should only be computed once if kept in cache.
- If you define the buffer on the location, you can read_group it at once to not read each location one by one which would be costly
- If there is a limit, you can stop the iteration on the locations when the limit is reached

guewen: We have to measure the cost of the buffer availability, we don't have a lot of them and it only queries indexed relations.

stock_storage_type_buffer/models/stock_location_storage_buffer.py
guewen: @jbaudoux can you check if this is what you expect: https://github.com/OCA/wms/pull/58/commits/aa7473a1f464a5a1e6078df79a8406510a1658c1 ?

jbaudoux: Perfect. Thanks

stock_storage_type_buffer/models/stock_location.py
guewen: The cost of `if storage_buffer.buffers_have_capacity():` is really low (between 0.01 and 0.06 per buffer on my test database after stress test), so for now I don't think it's worth changing this.

simahawk:
@jbaudoux ping :) @guewen any test to be added to make coverage happier?

simahawk:
@guewen ping

guewen:
The missing test coverage is on the help message and `name_get`. Is it worth to test these?

guewen:
Added tests and rebased

PR #60, opened 2020-11-04
[13.0][FIX] shopfloor, zone picking: ability to unload partial moves

sebalix:


shopfloor/models/stock_move.py
```diff
@@ -44,6 +44,13 @@ def extract_and_action_done(self):
         to first extract some move lines in a separate move, then validate it
         with this method.
         """
+        # Put remaining qty to process from partially available moves
+        # in their own move (which will be then 'confirmed')
+        partial_moves = self.filtered(lambda m: m.state == "partially_available")
+        for partial_move in partial_moves:
+            partial_move._split(partial_move.quantity_done)
+            partial_move._recompute_state()  # become 'assigned'
```
guewen: This is actually what we did in `split_other_move_lines`, do you have a case where `split_other_move_lines` is not called before? Or does it happen when `split_other_move_lines` is called with `intersection=True` (cause we didn't fix the issue in this case).

simahawk: good point :+1:

sebalix: On the `set_destination_all` endpoint we are not calling `split_other_move_line`, we are calling it on a previous endpoint when we scan the destination location with `/set_destination` (but if we scan a package, it is not called... maybe worth to add it there too).

sebalix: intersection not used in this flow.

guewen (approved):

simahawk (approved):
@sebalix can you squash?

sebalix:
@simahawk squashed

PR #63, opened 2020-11-06
[13.0] Add stock_dynamic_routing_reserve_rule

guewen:
Glue module between stock_dynamic_routing and stock_reserve_rule.
It adds a link to open reserve rules from dynamic routings, and
displays warning on dynamic routings when there are risks of
inconsistencies because of discrepancies between dynamic routing
and reservation rules.

For the record, these issues are:

How the application of the dynamic routing works:

1. StockMove._action_assign() is executed in a savepoint, to see what move
   lines are created and what are their source and destinations
2. If no dynamic routing matches the picking type and one of the lines:
   source / location (+ domain), the savepoint is released (we keep the
   initial _action_assign)
3. If we have a dynamic routing, the savepoint is rollbacked, we apply
   the dynamic routing (split moves if required, change picking type, add
   chained move), and call again StockMove._action_assign()

As in step 3, the picking type can be changed, and the reservation rules
are dependent of the picking type, if we have a reservation rule for the
initial picking type and no reservation rule for the new one, we may
get a different reservation which does no longer takes the good at the
place expected initially.

Depends on https://github.com/OCA/stock-logistics-warehouse/pull/1013

### DIFF
--- a/stock_dynamic_routing_reserve_rule/models/stock_routing.py
+++ b/stock_dynamic_routing_reserve_rule/models/stock_routing.py
@@ -1,0 +1,94 @@
+# Copyright 2020 Camptocamp (https://www.camptocamp.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from os.path import commonpath
+
+from odoo import _, api, fields, models
+
+
+class StockRouting(models.Model):
+    _inherit = "stock.routing"
+
+    reserve_rule_warning = fields.Html(compute="_compute_reserve_rule_warning")
+
+    def _removal_rules_common_locations(self, reserve_rules):
+        dest_removal_locations = reserve_rules.rule_removal_ids.location_id
+        location_names = dest_removal_locations.mapped("complete_name")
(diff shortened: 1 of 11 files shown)


guewen:
Rebased

stock_dynamic_routing_reserve_rule/models/stock_routing.py
```diff
@@ -0,0 +1,94 @@
...
+
+class StockRouting(models.Model):
+    _inherit = "stock.routing"
+
+    reserve_rule_warning = fields.Html(compute="_compute_reserve_rule_warning")
+
+    def _removal_rules_common_locations(self, reserve_rules):
+        dest_removal_locations = reserve_rules.rule_removal_ids.location_id
+        location_names = dest_removal_locations.mapped("complete_name")
+        locations_common_path = commonpath(location_names)
```
simahawk, line 17: TIL `commonpath` :)

simahawk (approved):

simahawk:
rebased and bumped manually

PR #64, opened 2020-11-09
[WIP][13.0][IMP] Allow reusable packs

hparfr:
Currently it's not possible to use the same pack twice in the shopfloor mobile.

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -590,6 +590,7 @@ class ClusterPicking(Component):
         if any(
             ml.picking_id != move_line.picking_id
             for ml in bin_package.planned_move_line_ids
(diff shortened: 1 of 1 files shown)


jgrandguillaume:
As said here https://github.com/OCA/wms/issues/78 this is not possible in cluster picking. You never mix eveal order per bin. If you do so, you're doing batch picking then which need another scenario.

- Batch picking: Pick several orders then sort during packing
- Cluster picking, pick and sort per order during picking time

hparfr:
>
    Batch picking: Pick several orders then sort during packing
    Cluster picking, pick and sort per order during picking time

Thanks for this clarification :) Cluster and batch picking were synonyms for me. It makes more sense now.

PR #66, opened 2020-11-10
 [13.0][FIX] shopfloor - api_key_demo not found

hparfr:
Can't call None.sudo()

### DIFF
--- a/shopfloor/services/service.py
+++ b/shopfloor/services/service.py
@@ -265,9 +265,10 @@ class BaseShopfloorService(AbstractComponent):
         defaults = super()._get_openapi_default_parameters()
         # Normal users can't read an API key, ignore it using sudo() only
         # because it's a demo key.
-        demo_api_key = self.env.ref(
-            "shopfloor.api_key_demo", raise_if_not_found=False
-        ).sudo()
+        demo_api_key = self.env.ref("shopfloor.api_key_demo", raise_if_not_found=False)
+        if demo_api_key:
+            demo_api_key = demo_api_key.sudo()
+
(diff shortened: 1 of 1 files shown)


shopfloor/services/service.py
```diff
@@ -267,7 +267,10 @@ def _get_openapi_default_parameters(self):
         # because it's a demo key.
         demo_api_key = self.env.ref(
```
simahawk: isn't this useless as we don't have the key at all?
Shall we do something better? Either add the key back for demo only or trash this bits and put "PUT_YOUR_KEY_HERE" in value?

hparfr: IMO it's more related to base_rest, you can expect the same behavior in shopinvader as well. If something is needed to be done; it should be on base_rest.

I'm in favor to remove this demo_api_key from shopfloor;  I don't use this feature btw.

guewen: It's useful when testing the module with demo data

guewen: It's useful when testing the module with demo data for faster development. When not using demo data, we don't care if the key is empty in swagger

guewen:
Travis fails because of pre-commit

shopfloor/services/service.py
simahawk: alright, fine then.

hparfr:
> Travis fails because of pre-commit

Fixed. It's now green

simahawk (approved):

simahawk:
@hparfr please, include module name in commits next time :wink:

PR #67, opened 2020-11-10
[13.0] Improve performance of available to promise computation

guewen:
When we have thousands of moves, the screen becomes hardly usable.
Make the comptutation faster.

* Compute the available quantity directly from quants with a read group
  (the qty_available computed field is too slow as it computes other
  quantities, and even much slower because of "mrp" overrides)
* The computation of the quantity promised to previous moves is very
  slow because it runs one query per move. Use a SQL query with LATERAL
  JOIN in order to run a single query for *all* the moves.

As I changed the computation, I went through it and improved / fixed
some things:

* Show a "previous_promised_qty" field which contains the result of
  the new query that returns the promised qty of previous moves. It
  is shown only with debug mode.
* Do no longer filter on the "previous moves" of the same warehouse:
  moves do not always have a "warehouse_id" set. Instead, all the
  quantities are computed on the same basis: in the view locations of
  all the warehouses, assuming that if we have the quantity anywhere
  in the stock, we can release (then make internal transfers).
* Compute both uom and real "ordered available to promise" quantity in
  the same compute method, the uom one being displayed to the user,
  while the real one is used during the release
* When both the priority and the date_priority of previous moves were
  equal, the previous quantity could be indeterministic, in this case,
  order by id in last resort
* Add tests for search methods

### DIFF
--- a/delivery_carrier_preference/models/stock_move.py
+++ b/delivery_carrier_preference/models/stock_move.py
@@ -22,13 +22,13 @@ class StockMove(models.Model):
         "product_id.packaging_ids",
         "product_id.packaging_ids.max_weight",
         "product_id.weight",
-        "ordered_available_to_promise",
+        "ordered_available_to_promise_uom_qty",
     )
     def _compute_estimated_shipping_weight(self):
         for move in self:
             prod = move.product_id
             move.estimated_shipping_weight = prod.get_total_weight_from_packaging(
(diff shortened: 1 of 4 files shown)


TDu (approved):

sebalix (approved):

simahawk (approved):
:fireworks:

PR #68, opened 2020-11-10
[FIX] shopfloor, zone picking: do not process line processed by someone else

sebalix:
First user scans the source location 'Zone sub-location 1' containing only one move line, then processes the next step 'set_line_destination'.

The second user scans the same source location, and should not find any line.

And since 9021c85c7085f883756ff64b797950a2192c3a5a has been merged, if no move line is found when scanning the source location, the message returned has to say that the location is empty.

### DIFF
--- a/shopfloor/tests/test_zone_picking_base.py
+++ b/shopfloor/tests/test_zone_picking_base.py
@@ -13,6 +13,20 @@ class ZonePickingCommonCase(CommonCase):
         cls.picking_type = cls.menu.picking_type_ids
 
     @classmethod
+    def setUpClassUsers(cls):
+        super().setUpClassUsers()
+        Users = cls.env["res.users"].sudo().with_context(no_reset_password=True)
+        cls.stock_user2 = Users.create(
+            {
+                "name": "Paul Posichon",
+                "login": "paulposichon",
+                "email": "paul.posichon@example.com",
+                "notification_type": "inbox",
+                "groups_id": [(6, 0, [cls.env.ref("stock.group_stock_user").id])],
+            }
+        )
+
+    @classmethod
     def setUpClassBaseData(cls, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


shopfloor/tests/test_zone_picking_base.py
```diff
@@ -12,6 +12,24 @@ def setUpClassVars(cls, *args, **kwargs):
...
+    def setUpClassUsers(cls):
+        super().setUpClassUsers()
+        Users = (
+            cls.env["res.users"]
+            .sudo()
+            .with_context({"no_reset_password": True, "mail_create_nosubscribe": True})
+        )
+        cls.stock_user2 = Users.create(
+            {
+                "name": "Paul Posichon",
```
guewen, line 21: :racing_car::racing_car::racing_car:
:joy:

guewen (approved):

shopfloor/tests/test_zone_picking_base.py
sebalix: For the reference: https://en.wikipedia.org/wiki/Joe_Bar_Team :p

shopfloor/tests/test_zone_picking_base.py
```diff
@@ -12,6 +12,24 @@ def setUpClassVars(cls, *args, **kwargs):
         cls.wh = cls.profile.warehouse_id
         cls.picking_type = cls.menu.picking_type_ids
 
+    @classmethod
+    def setUpClassUsers(cls):
+        super().setUpClassUsers()
+        Users = (
+            cls.env["res.users"]
+            .sudo()
+            .with_context({"no_reset_password": True, "mail_create_nosubscribe": True})
```
simahawk: use kw args otherwise you wipe the ctx and lose tracking_disable which is set in base test class :wink: 
FTR mail_create_nosubscribe is not needed if you don't wipe the ctx.

PR #73, opened 2020-11-13
[13.0] stock_available_to_promise: Rework release screens

guewen:
Releasing at move level is not really meaningful, even if we can do it, the main way should be from the Transfers.
The main changes (split in individual commits for easier review) are:

* On products, sales orders and transfers, a new smart button allows to open the Moves Allocations view
* The Moves Allocations view is now considered as reporting and as such moved in Inventory / Reporting
* Add a new view based on transfers: Transfers Allocations. On the model of Moves Allocations, it shows useful information to handle the releases waves
* This screen shows the count of moves ready to be released and a button to drilldown on these moves allocations
* Fix an issue with the "all at once" delivery type: previously, a transfer was considered "release ready" as long as at least 1 qty of each product was available, now the whole quantity of all moves has to be available
* New configuration field "Preparation Time", when transfers are released, their scheduled date is set at now + prep. time, so we can easily see if a transfer is late (stuck in the warehouse, ...)
* Adapt the count of transfers in the Delivery Types dashboard: transfers not yet released are only included in "Need Release" count. The "waiting" and "late" count include only the transfers already released, so we can track the transfers *currently in preparation*


stock_available_to_promise_release/models/res_company.py
```diff
@@ -11,3 +11,12 @@ class Company(models.Model):
         help="Compute promised quantities for order planned to be shipped "
         "until this number of days from today.",
     )
+    stock_release_max_prep_time = fields.Integer(
+        string="Transfer Releases Max Prep. Time",
```
sebalix, line 15:
```suggestion
        string="Transfer Releases Max Prep. Time (in minutes)",
```

stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -35,3 +133,30 @@ def _release_link_backorder(self, origin_picking):
...
+        """
+        self._after_release_set_printed()
+        self._after_release_set_expected_date()
+
+    def _after_release_set_printed(self):
+        self.filtered(lambda p: not p.printed).printed = True
+
+    def _after_release_set_expected_date(self):
+        prep_time = self.env.company.stock_release_max_prep_time
+        new_expected_date = fields.Datetime.now() + timedelta(minutes=prep_time)
```
sebalix: We could use the Odoo API to avoid importing `timedelta`?
```
new_expected_date = fields.Datetime.add(fields.Datetime.now(), minutes=prep_time)
```

stock_available_to_promise_release/views/res_config_settings.xml
```diff
@@ -28,6 +28,26 @@
                             </div>
                         </div>
                     </div>
+                    <div class="col-12 col-lg-6 o_setting_box">
+                        <div class="o_setting_right_pane">
+                            <label for="stock_release_max_prep_time" />
+                            <div class="text-muted">
+                                When your release transfers, their expected date
+                                is rescheduled to now + this preparation time.
```
sebalix:
```suggestion
                                is rescheduled to now + this preparation time (in minutes).
```

sebalix (approved):
Few comments, pre-approving

stock_available_to_promise_release/models/stock_picking.py
guewen: I wasn't aware of `Datetime.add`, I'll change

stock_available_to_promise_release/views/res_config_settings.xml
guewen: It's displayed "in minutes" next to the input field, so is it useful?

guewen: updated anyway

sale_stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    sale_date_expected = fields.Date(
+        string="Delivery Date", related="group_id.date_expected", store="True"
```
simahawk: `store=True`?

stock_available_to_promise_release/views/stock_picking_views.xml
```diff
@@ -7,15 +7,125 @@
...
+    </record>
+    <record id="stock_picking_release_action" model="ir.actions.act_window">
+        <field name="name">Transfers Allocations</field>
+        <field name="res_model">stock.picking</field>
+        <field name="type">ir.actions.act_window</field>
+        <field name="view_mode">tree,kanban,form</field>
+        <field name="search_view_id" ref="stock.view_picking_internal_search" />
+        <field
+            name="view_ids"
+            eval="[(5, 0, 0),
```
simahawk, line 108: why this 1st line?

simahawk (approved):
LG minor comments.

sale_stock_available_to_promise_release/models/stock_picking.py
guewen: I fixed this here https://github.com/OCA/wms/pull/73/files#diff-fca2e9865c2d75a9793dbe699bffdf976606d996b69d09cacd25660f2926e998R51 and copy-pasted the wrong version :facepalm:

stock_available_to_promise_release/views/stock_picking_views.xml
guewen: It resets the m2m, this is what's done in every core addon where `view_ids` is set. The alternative would be to use a `[(6, 0, ...)]` but I prefer consistency.

sale_stock_available_to_promise_release/models/stock_picking.py
guewen: Fixed (squashed the change in 4434c424a) and force-pushed with this change

simahawk:
@sbidoul seems the job for merge still does not trigger travis. Is there anything we can do or shall we stick to manual merge?

sbidoul:
@simahawk I've no idea... can you try again ? Maybe travis is slow to catch up ?

simahawk:
@sbidoul nada, stilll running Runbot only

sbidoul:
@simahawk travis seems to work normally there: https://github.com/OCA/queue/pull/287

simahawk:
@sbidoul WTH... I found why: https://github.com/OCA/wms/settings/hooks the travis hook is not there anymore :sob: 
I've just re-created it. Coverage is missing as well... How's that possible?

sbidoul:
@simahawk these webhooks should not be necessary: travis and codecov are configured at organization level, as well as all necessary webhooks (weblate, runbot, oca-github-bot).

simahawk:
@sbidoul ok, what else can we check?

sbidoul:
@simahawk ah, I see something. This repo, like some others, is still on travis-ci.org and disabled there. I activated there. What is strange is that I don't see it in the "migrate" tab...

/ocabot merge minor

PR #75, opened 2020-11-19
Shopfloor: fix sorting of lines by scheduled date

simahawk:
zone_picking and cluster_picking should order lines by scheduled date as well. Always.

ref: 1791

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -324,6 +324,7 @@ class ClusterPicking(Component):
             line.location_id.shopfloor_picking_sequence or "",
             line.location_id.name,
             -int(line.move_id.priority or 1),
+            line.move_id.date_expected,
             line.move_id.sequence,
             line.move_id.id,
             line.id,
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -317,22 +317,25 @@ class ZonePicking(Component):
                 locations, picking_type, package, product, lot
             )
         )
-        sort_keys_func, reverse = self._sort_key_move_lines(order)
-        move_lines = move_lines.sorted(sort_keys_func, reverse=reverse)
+        sort_keys_func = self._sort_key_move_lines(order)
+        move_lines = move_lines.sorted(sort_keys_func)
         return move_lines
 
     @staticmethod
     def _sort_key_move_lines(order):
-        """Return a `(sort_keys_func, reverse)` tuple for move lines."""
+        """Return a sorting function to order lines."""
+
         if order == "priority":
-            return lambda line: line.move_id.priority or "", True
+            # make prority negative to keep sorting ascending
+            return lambda line: (
+                -int(line.move_id.priority or "0"),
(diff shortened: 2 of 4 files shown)


shopfloor/services/cluster_picking.py
```diff
@@ -324,6 +324,7 @@ def _sort_key_lines(line):
             line.location_id.shopfloor_picking_sequence or "",
             line.location_id.name,
             -int(line.move_id.priority or 1),
+            line.move_id.date_expected or "",
```
guewen: `date_expected` being a datetime, we will have a comparison error if one is empty (can it?)

simahawk: AFAIS is always valued. I could add a safer default though.

simahawk:
```py
date_expected = fields.Datetime(
        'Expected Date', default=fields.Datetime.now, index=True, required=True,
        states={'done': [('readonly', True)]},
        help="Scheduled date for the processing of this move")
```
I can even drop the `or ""` IMO.

guewen: > I can even drop the or "" IMO.

:+1:

simahawk: done

shopfloor/tests/test_cluster_picking_select.py
```diff
@@ -263,6 +266,42 @@ def setUpClassBaseData(cls, *args, **kwargs):
...
+                [cls.BatchProduct(product=cls.product_b, quantity=1)],
+            ]
+        )
+        cls._simulate_batch_selected(cls.batch2, in_package=True)
+
+    def test_lines_order(self):
+        batch = self.batch2
+        picking1 = batch.picking_ids[0]
+        today = fields.Datetime.today()
+        future = date_utils.add(
```
sebalix: Note: `date_utils` functions are added as static methods on `fields.Date` and `fields.Datetime`, so you are not forced to import it if you want.

shopfloor/services/zone_picking.py
```diff
@@ -317,22 +317,25 @@ def _find_location_move_lines(
...
         )
-        sort_keys_func, reverse = self._sort_key_move_lines(order)
-        move_lines = move_lines.sorted(sort_keys_func, reverse=reverse)
+        sort_keys_func = self._sort_key_move_lines(order)
+        move_lines = move_lines.sorted(sort_keys_func)
         return move_lines
 
     @staticmethod
     def _sort_key_move_lines(order):
         """Return a `(sort_keys_func, reverse)` tuple for move lines."""
```
sebalix: Docstring needs to be updated as `reverse` parameter is not returned anymore.

shopfloor/tests/test_cluster_picking_select.py
simahawk: great. Good to know! Tnx!

simahawk:
@sebalix fixed

guewen (approved):

sebalix (approved):

Issue #78, opened 2020-11-20
[RFC] Cluster picking  location as destination


hparfr:
Hi,

For the cluster picking, I need to have multiple sale orders per "temporary destination" (= a bin, =a trolley...) and there is a picking unicity constraint on packs.

I think allowing to use locations instead of pack will solve my problem and allow to lower the entry barrier of `shopfloor` for smaller warehouses.
These locations will be sublocations of the picking type destination (ie: WH/PACK/BIN{1..n}).

I don't remember the good reasons to manage these "temporary destinations" only as pack and not as locations as well.

To start the implementation, I propose the following changes in cluster picking :
* rename `def scan_destination_pack` to `def scan_destination`
* allow `barcode` of `scan_destination` to be a location or a pack

[EDIT]
* there is also some changes in mobile_app, "package_dest" should be renamed in "destination" 
[/EDIT]

Do you agree with these changes ? @guewen @jgrandguillaume @simahawk

jbaudoux:
When you do the cluster picking, you collect and put the goods on a moving support (like a trolley) which is handled as a pack. You do that for multiple move lines. At the end, you place this support somewhere which is a location (which is physically located).
A location should be somewhere in the warehouse. You can go there and count the goods. It should not be moving.
A pack is something you build over time by putting items inside. Its location is known only once you have finished it and placed somewhere.
The pack can also be temporary. You can transfer its content to a customer pack.

hparfr:
>[a location] should not be moving

Says who ? Why ?

jbaudoux:
> > [a location] should not be moving
> 
> Says who ? Why ?

Because you move from location A to location B. You do not move the location otherwise where is it physically?
If your location is moving, how do you record where you place it? 

At least, that's the reason behind you where asking.

hparfr:
> If your location is moving, how do you record where you place it?

There is a difference between (A)moving a physical trolley and (B)move (=rename) a location.
Here we are talking about A.

So you track if the goods are in the trolley or if they are not. I don't kneed to know where is the trolley.
I don't need realtime geolocalisation.

jbaudoux:
The question is what happens after the picking? Someone will consume what is in that trolley to pack or ship. And it helps if you can guide that operator by indicating where are the goods, i.e. where you placed that trolley at the end of the picking.

hparfr:
In the warehouses I know, the following setup is usually in place:
1) the picker, picks goods from the alleys to a trolley,
[...]

PR #79, opened 2020-11-23
[WIP][13.0] shopfloor cluster_picking scan_destination

hparfr:
In cluster_picking allow to scan a location instead of a pack (destination).

See #78 

Some refactoring pending and tests to be included.


hparfr:
a new scenario will be proposed instead

PR #80, opened 2020-11-23
[13.0] shopfloor misc improvements

simahawk:
Several fixes/imp, see atomic commits for details.
Most changes namely:
* zone picking add counters per each zone
* zone picking improve zone list display
* zone picking make sure you get always fresh data when navigating back
* show move lines / positions counters per each menu item
* upgrade vuetify

ref: 1780


TDu (approved):
I only skim-read this...

shopfloor/services/zone_picking.py
```diff
@@ -249,9 +242,7 @@ def _data_for_location(self, zone_location, picking_type, location):
         }
 
     def _zone_lines(self, zones):
-        return self.env["stock.move.line"].search(
-            self._find_location_move_lines_domain(zones)
-        )
+        return self._find_location_move_lines(zones)
```
sebalix, line 245: Can't we do the following 
```suggestion
        return self.search_move_line.search_move_lines_by_location(zones)
```
and remove `_find_location_move_lines` method (which seems also used at different places but could be replaced by the component)? Even the `_zone_lines` method could be removed at this stage.

simahawk: I left it on purpose: if you want to completely customize the behavior at scenario level.

simahawk: @sebalix beside this the rest is good?

sebalix: gut!

sebalix (approved):

PR #83, opened 2020-11-25
Fix prefered delivery on backorder creation

TDu:
This is PR has only a unit test, trying to reproduce an issue encountered.
WIP


sebalix:
@TDu this PR is still WIP?

TDu:
@sebalix No I think the error I was investigating was related to a conflict between other modules. I wonder if I did not create https://github.com/OCA/wms/pull/97 to fix that problem.
Anyway closing.

PR #84, opened 2020-11-27
[FIX] shopfloor, checkout: ability to recover/restart

sebalix:
If the user starts to process a line, and for whatever reason he
stops there and restarts the scenario from the beginning, he should
still be able to find/recover the previous line.

For that purpose we are using the `shopfloor_user_id` field on move
lines (already used to manage currently processed lines in the Zone
picking and Location content transfer scenarios).

Ref. 1775

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -296,13 +296,16 @@ class Checkout(Component):
             if line.shopfloor_checkout_done:
                 continue
             line.qty_done = line.product_uom_qty
+            line.shopfloor_user_id = self.env.user
 
         picking = lines.mapped("picking_id")
         other_lines = picking.move_line_ids - lines
         self._deselect_lines(other_lines)
 
     def _deselect_lines(self, lines):
-        lines.filtered(lambda l: not l.shopfloor_checkout_done).qty_done = 0
+        lines.filtered(lambda l: not l.shopfloor_checkout_done).write(
+            {"qty_done": 0, "shopfloor_user_id": False}
+        )
 
     def scan_line(self, picking_id, barcode):
         """Scan move lines of the stock picking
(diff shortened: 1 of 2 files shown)


sebalix:
Right now we do not recover on the state `select_package`, I'm checking if it can be improved that way.
But at least the user is not blocked anymore with the current solution (he'll have to do it in two steps: `scan_document` then `select_line` to land on `select_package`, but we could get rid of `select_line` and go directly from `scan_document` to `select_package` if the package has already been processed in the past.

jgrandguillaume:
@sebalix  Pay attention, not sure this is ok because whatever the user I want to be able to recover the situation. 
e.g.

**Case 1)**

User 1:

Go to the scanner, profile Pack, menu Pack PAL
Scan CCOPAL-01 or choose the pack operation manually
Scan the PACK one time
-> Here you quit the barcode app

User 1:
Get back in it, same menu, find the operation again
You get nothing in it, nothing in the summary, no way to be able to finish the checkout


**Case 2)**

User 1:

Go to the scanner, profile Pack, menu Pack PAL
Scan CCOPAL-01 or choose the pack operation manually
Scan the PACK one time
-> Here you quit the barcode app

User 2:
Get back in it, same menu, find the operation again
You get nothing in it, nothing in the summary, no way to be able to finish the checkout

shopfloor/services/checkout.py
```diff
@@ -592,7 +595,9 @@ def _switch_line_qty_done(self, picking, selected_lines, switch_lines):
 
     @staticmethod
     def _filter_lines_unpacked(move_line):
-        return move_line.qty_done == 0 and not move_line.shopfloor_checkout_done
+        return (
+            move_line.qty_done == 0 or move_line.shopfloor_user_id == move_line.env.user
```
jgrandguillaume: May be with this ?

move_line.qty_done == 0 or move_line.shopfloor_user_id

simahawk: agree

simahawk (approved):
@sebalix squash pls :)

sebalix:
@simahawk done

sebalix:
@simahawk tests are broken... do not merge

TDu (approved):

sebalix:
@simahawk my DB was not up-to-date, tests are running fine

PR #90, opened 2020-12-01
[WIP][14.0][MIG] shopfloor

hparfr:
All the dependencies are not merged yet.
I installed modules from OCA/#14.0 branch when available, then pending PR then simply v13 of remaining modules.

Note : During the migration process (`git format-patch --keep-subject --stdout origin/14.0..origin/13.0 -- shopfloor | git am -3 --keep`) I had some conflicts, I don't know why, and there was so many commits, I had a lot of difficulties to find them back in the history. , so I applied blindly the "keep theirs" strategy. 

Note 2: Migrated with odoo-module-migrator (and this pending PR for pre-commit  https://github.com/OCA/odoo-module-migrator/pull/41)


simahawk:
@hparfr @guewen I think it's better to add the module to branch 14.0 and make it uninstallable/ignored and keep only migration relevant commits here. Is going to be way easier to review. What do you think?

hparfr:
Agree

sebastienbeau:
@guewen @simahawk @hparfr @bguillot I just push a V14 branch here https://github.com/OCA/wms/tree/14.0
I run pre-commit and I set module uninstallable

PR #91, opened 2020-12-02
Fix error when a picking has no move lines

guewen:
When computing date_priority, if a picking has no move lines or some
move lines have no date_priority, we can have errors because min()
is called without sequence or compares bool and datetime.


TDu (approved):

simahawk (approved):

simahawk:
@guewen please rebase on 13.0 and bump the version manually. I dunno why but merge jobs get stuck.

PR #97, opened 2020-12-11
[13.0][ADD] delivery_preference_glue_stock_picking_group

TDu:


delivery_preference_glue_stock_picking_group/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    def release_available_to_promise(self):
+        """After release reset to customer prefered carrier the backorder."""
+        pickings = (
+            self.filtered(self._filter_recompute_preferred_carrier)
```
simahawk: couldn't you put your condition in AND with `_filter_recompute_preferred_carrier`?

delivery_preference_glue_stock_picking_group/readme/ROADMAP.rst
```diff
@@ -0,0 +1,6 @@
+This is a glue module, so the question is.
+
+Is it necessary or there is a bug some where than cause this ?
+
+Maybe in the grouping done by the module `stock_picking_group_by_partner_by_carrier`
```
simahawk, line 5: eh, good question :smile:

simahawk (commented):
I'm not sure we need it :thinking: 
I'd discuss w/ Joël and JE as I have to solve a similar issue in another context.

delivery_preference_glue_stock_picking_group/models/stock_move.py
TDu: I changed that.

TDu:
@simahawk This is already used as a local addon. It is only a move to the OCA, plus the tests.

guewen (approved):

PR #99, opened 2020-12-11
[13.0] shopfloor: fix zone picking scan source

simahawk:
When a location is scanned and several items are found
ask to scan a package instead of saying that the location is empty.

The fix is in the 1st commit. The second commit is a small fix on messages declaration and 3rd one is a small refactoring to reduce the complexity of `_scan_source` and avoid searching w/ the same barcode all kind of records.

ref: 1895

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -382,43 +382,130 @@ class ZonePicking(Component):
         )
         return self._response_for_select_line(zone_location, picking_type, move_lines)
 
-    def _scan_source_location(self, zone_location, picking_type, location, **kw):
-        """Return the move line related to the scanned `location`.
-
-        The method tries to identify unambiguously a move line in the location
-        if possible, otherwise return `False`.
+    def _scan_source_location(
+        self, zone_location, picking_type, barcode, order="priority"
+    ):
+        """Search a location and find available lines into it.
         """
+        response = None
+        message = None
+        search = self.actions_for("search")
+        location = search.location_from_scan(barcode)
+        if not location:
+            return response, message
+
+        if not location.is_sublocation_of(zone_location):
+            response = self._response_for_start(
+                message=self.msg_store.location_not_allowed()
+            )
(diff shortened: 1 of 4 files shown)


shopfloor/services/zone_picking.py
```diff
@@ -383,48 +383,128 @@ def list_move_lines(self, zone_location_id, picking_type_id, order="priority"):
...
+        product, lot, package = self._find_product_in_location(location)
         if len(product) > 1 or len(lot) > 1 or len(package) > 1:
-            return False
+            response = self.list_move_lines(location.id, picking_type.id)
+            message = self.msg_store.several_products_in_location(location)
+            return response, message
+
         move_lines = self._find_location_move_lines(
-            location,
+            zone_location,
```
sebalix: I'm not sure to understand this change? If the user scan a location, it is to find the move lines within it, not the parent zone location?

simahawk: good catch! just a typo! :facepalm:

simahawk: fixed

simahawk:
@sebalix fixed, can you pass by again?

sebalix (approved):

simahawk:
FTR the issue on the build was due to a bug in odoo core fixed here https://github.com/odoo/odoo/pull/63587

shopfloor/services/zone_picking.py
```diff
@@ -383,48 +383,129 @@ def list_move_lines(self, zone_location_id, picking_type_id, order="priority"):
         return self._response_for_select_line(zone_location, picking_type, move_lines)
 
     def _scan_source_location(
-        self, zone_location, picking_type, location, order="priority"
+        self, zone_location, picking_type, barcode, order="prioriry"
```
guewen:
```suggestion
        self, zone_location, picking_type, barcode, order="priority"
```

guewen (approved):

PR #106, opened 2020-12-18
stock_available_to_promise_release: Fix quantity available to promise of released moves

jbaudoux:
When multiple moves of a same product are released, the quantity promised before the move was wrongly computed. It should take into account only the preceding moves and not all the released moves.

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -182,7 +182,7 @@ class StockMove(models.Model):
             }
         )
 
-        locations = self._ordered_available_to_promise_locations()
+        locations = moves._ordered_available_to_promise_locations()
 
         # Compute On-Hand quantity (equivalent of qty_available) for all "view
         # locations" of all the warehouses: we may release as soon as we have
@@ -197,7 +197,7 @@ class StockMove(models.Model):
                 ]
             )
         domain_quant = expression.AND(
-            [[("product_id", "in", self.product_id.ids)], location_domain]
+            [[("product_id", "in", moves.product_id.ids)], location_domain]
         )
         location_quants = self.env["stock.quant"].read_group(
             domain_quant, ["product_id", "quantity"], ["product_id"], orderby="id"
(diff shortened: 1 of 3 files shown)


TDu:
@jbaudoux Any ideas why the tests fail on the unrelated addon  `stock_storage_type_putaway_abc` ?
Also an explanation in the commit message how and why this fixes the quantity available to promise of released moves, would be helpful

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -82,7 +82,7 @@ def _previous_promised_qty_sql_lateral_where(self):
                 AND p_type.code = 'outgoing'
                 AND loc.parent_path LIKE ANY(%(location_paths)s)
                 AND (
-                    m.need_release = true
+                    COALESCE(m.need_release, False) = COALESCE(move.need_release, False)
```
guewen, line 85: This SQL query should never be called on moves if the flag `need_release` is not `True`, so this shouldn't be necessary: https://github.com/OCA/wms/pull/106/files#diff-fca2e9865c2d75a9793dbe699bffdf976606d996b69d09cacd25660f2926e998R248-R254

guewen: This line https://github.com/OCA/wms/pull/106/files#diff-fca2e9865c2d75a9793dbe699bffdf976606d996b69d09cacd25660f2926e998R208 should be changed to `for move in moves:`.

And in `_should_compute_ordered_available_to_promise`: https://github.com/OCA/wms/pull/106/files#diff-fca2e9865c2d75a9793dbe699bffdf976606d996b69d09cacd25660f2926e998R251, the filter on `need_release` be removed.

guewen:
I think the travis error should be fixed by a rebase

stock_available_to_promise_release/models/stock_move.py
jbaudoux: Thanks for the review. I fixed those

guewen (approved):
Thanks!

TDu (approved):

PR #112, opened 2021-01-07
[13.0] Ensure moves routed from different location are reserved before other

guewen:
Take this use case:

* When we pick goods in HB, we have to go through Handover
* When we pick goods and their priority is 3, we change the picking type
  to "PRIO"

We have a Pick transfer for 30 units. We have 12 units in Stock/HB, 18 in
Stock/Shelving, priority is 3.

When we reserve, the dynamic routing's code does:

* run _action_assign on the move in a savepoint, to peek where we can
  reserve, and rollback the savepoint
* according to the reservations, split the move in 12 and 18
* apply the routings for the units:

  * in shelving, the move picking type is changed to PRIO
  * the other move is updated to take goods from HB to Handover in the
    Handover picking type, and a new waiting move is created for Handover
    -> Output with the PRIO picking type (will be in the same transfer as
    the shelving move)

If case the reservation of the shelving move is done first, it could
reserve the goods in HB that were reserved for the other move. We end
up with: HB move without reservation, the shelving stock taking goods in
HB without going through handover.

To prevent moves of a less specific location (such as Stock) stealing
moves of a more specific location (such as Stock/HB), sort the moves
to assign by reserved parent path of their source locations, as we know
the most specific locations cannot reserve moves above them. The order
within one location is preserved.

### DIFF
--- a/stock_dynamic_routing/models/stock_move.py
+++ b/stock_dynamic_routing/models/stock_move.py
@@ -235,6 +235,7 @@ class StockMove(models.Model):
         the routing ones and creates a new chained move after it.
         """
         pickings_to_check_for_emptiness = self.env["stock.picking"]
+        move_ids_to_assign_per_location = defaultdict(list)
         for move in self:
             routing_rule = routing_details[move].rule
             if not routing_rule.method == "pull":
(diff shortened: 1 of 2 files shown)


stock_dynamic_routing/models/stock_move.py
```diff
@@ -273,13 +274,56 @@ def _apply_routing_rule_pull(self, routing_details):
...
+        # excluded by the method.
+        #
+        # Now among these moves, some may have their source location changed,
+        # for instance:
+        #
+        # * A move of 20 units has a source location: Stock
+        # * It is split to take
+        #   * 10 in Stock/Shelving
+        #   * 10 in Stock/Reserve
+        # * Stock/Reserve insert a new move to do Stock/Reserve -> Stock/Shelving
```
sebalix: I'm not sure to understand this sentence, the `Stock/Reserve -> Stock/Shelving` part especially?

stock_dynamic_routing/tests/test_routing_pull.py
```diff
@@ -910,3 +916,138 @@ def test_change_dest_move_source_chain(self):
...
+        # | 12x Product Stock → Output (waiting on the Handover moves)        |
+        # | 18x Product Stock → Output (available)                            |
+        # |   lines: 10 from Stock/Shelf1, 8 from Stock/Shelf2                |
+        # +-------------------------------------------------------------------+
+        #
+        # +-------------------------------------------------+
+        # | OUT/xxxx Waiting                                |
+        # | Output → Customer                               |
+        # | 18x Product Output → Customer (waiting)         |
+        # | 12x Product Output → Customer (waiting)         |
```
sebalix, line 1009: Question: is it possible to get only one move of 30 units waiting for the others?

sebalix (approved):

stock_dynamic_routing/tests/test_routing_pull.py
guewen: I don't remember why the next move is split, this is what I observed, so I documented it this way.

stock_dynamic_routing/models/stock_move.py
guewen: I'll improve the example

simahawk (approved):
thanks for the explanation :)

stock_dynamic_routing/models/stock_move.py
guewen: updated

stock_dynamic_routing/tests/test_routing_pull.py
sebalix: Probably linked to this PR https://github.com/OCA/wms/pull/232

PR #113, opened 2021-01-08
[IMP] shopfloor: create backorder when validating moves

sebalix:
**Before**: when moves are processed, we were managing two cases:

- if the current moves to process are all the moves remaining in a
  transfer, we validate the transfer

- if the current moves to process are a subset of available moves of a
  transfer, then we put these moves to process in a new transfer which
  is validated. All remaining moves stay in the current transfer.

**After**: a third case is added:

- if the current moves to process are equal to all available moves of
  the current picking - but there are still unavailable moves to process -
  then we validate the current transfer as usual, which will create a
  backorder with remaining/unavailable moves.

As this third case is very specific to Shopfloor, it has been implemented in a new `StockAction` component (used by all related scenarios).
Now all scenarios use this new component to validate stock moves, and the `extract_and_action_done` method on `stock.move` can stay generic.
The opportunity was taken to create a new method `split_unavailable_qty` on `stock.move` to put unavailable qty to a new move and update the current move to `assigned` (it was done directly in `extract_and_action_done` before).

### DIFF
--- a/shopfloor/actions/__init__.py
+++ b/shopfloor/actions/__init__.py
@@ -27,3 +27,4 @@ from . import search
 from . import inventory
 from . import savepoint
 from . import move_line_search
+from . import stock
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -1,0 +1,41 @@
+# Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.addons.component.core import Component
+
+
+class StockAction(Component):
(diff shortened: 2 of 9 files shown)


sebalix:
<s>I'm not sure about one thing: this change is maybe too specific, and should be outside of the `extract_and_action_done` method? This would mean create our own `extract_and_action_done`-like dedicated to Shopfloor (in a component?), and keep the current one on the move (which could be moved later in another generic module) that does what it says: extract the moves and validate them.

EDIT: I'll update the PR to put this new case outside of the `extract_and_action_done` method.</s>

i-vyshnevska (approved):

guewen (approved):

PR #104, opened 2020-12-15
[FIX] shopfloor: cluster picking, remove unavailable picking from validated batch

sebalix:
Unavailable transfers could be part of a batch (following a stock inventory happening at the same time for instance), as such we have to remove such transfers from the batch picking when this one is validated.
The transfers will be assigned to a new batch thereafter.

Ref. 1836

PR #114, opened 2021-01-11
[FIX] shopfloor_mobile, delivery, rename button 'Mark as done' to 'Make partial delivery'

sebalix:
Because as soon as all lines are processed (qty done filled), the related transfer is automatically validated. So the current button only serves one purpose: creating a partial delivery.

Ref. 1783.

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/delivery.js
+++ b/shopfloor_mobile/static/wms/src/scenario/delivery.js
@@ -33,7 +33,7 @@ const Delivery = {
             <div class="button-list button-vertical-list full">
                 <v-row align="center" v-if="state_is('deliver') && has_picking()">
                     <v-col class="text-center" cols="12">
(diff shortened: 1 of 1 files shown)


simahawk:
> Because as soon as all lines are processed (qty done filled), the related transfer is automatically validated. So the current button only serves one purpose: creating a partial delivery.
> 
> Ref. 1783.

hmm shouldn't we check if the picking is complete or not and display the label according to this?

simahawk:
hmm re-reading the code: we already mark the picking as done if all lines are processed. This button is to force it. Makes sense then.

simahawk (approved):
to be merged after https://github.com/OCA/wms/pull/105 is

i-vyshnevska (approved):

simahawk:
@sebalix https://github.com/OCA/wms/pull/104 must be rebased and merged 1st to fix the flappy test.

PR #115, opened 2021-01-11
[13.0] Add stock_checkout_sync

guewen:
Move https://github.com/OCA/stock-logistics-warehouse/pull/907

Depends on:

* https://github.com/OCA/stock-logistics-warehouse/pull/1051

Regroup moves at the same location when they reach an operation type.

Example of use case:

* We use several operation types for Picking Transfers (Pick Shelves, Pick Reserve)
* We use several operation types for Packing (Pack, Pack Post)
* The "Checkout Synchronization" flag is activated on the 2 Packing operation types
* The Packing can occur in several areas (Packing/PACK1, Packing/PACK2, Packing/PACK3)

Based on these assumptions, we have these pickings, all for the same order:

* SHELF001, pick 2 goods in the shelves
* RESER001, pick 2 goods in the reserve
* PACK001, pack 3 of the goods (2 from SHELF001, 1 from RESER001)
* PPOST001, pack the remaining good of the reserve

The operator uses the button "Checkout Sync" on a pick transfer,
for instance SHELF001, and select Packing/PACK2 as destination.
We want all the lines that reach PACK001 to be updated with:

* The move lines (except the one that goes to PPOST001) must now reach
  Packing/PACK2
* The same moves' destination must reach Packing/PACK2 as well:
  we constrain the user to move the goods there as we want to move
  all the goods at the same place for the packing.

Selection_848

Works best when used with ``stock_available_to_promise_release``:

* When using the mentioned module, we assume that we release moves (which
  creates the whole chain of moves) only when we know that we have the
  quantities in stock (otherwise the module splits the delivery). Any exception
  (which is now rare when you release this ways) that would lead to a backorder
  is handled by canceling the remaining line.
* When the module is not used, then the destination of the backorders may be
  changed at the same time

Compatible with ``stock_dynamic_routing``.


stock_checkout_sync/models/stock_move.py
```diff
@@ -0,0 +1,52 @@
...
+            if not dest_pickings:
+                continue
+
+            moves = (move | move.common_dest_move_ids).filtered(
+                lambda move: move.state not in ("done", "cancel")
+            )
+            for dest_picking in dest_pickings:
+                if dest_picking not in selected_pickings:
+                    selected_pickings[dest_picking] = self.env["stock.move"].browse()
+                selected_pickings[dest_picking] |= moves
```
simahawk: better to use a set of ids instead?

sebalix (approved):

stock_checkout_sync/tests/test_checkout_sync.py
```diff
@@ -0,0 +1,279 @@
...
+                    self.assertEqual(line.location_dest_id, location)
+                for dest in move.move_dest_ids:
+                    self.assertEqual(dest.location_id, location)
+
+
+class TestMoveCommonDestSyncLocation(CheckoutSyncCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
simahawk: this is done in common class

stock_dynamic_routing_checkout_sync/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2019 Camptocamp (https://www.camptocamp.com)
+{
+    "name": "Stock Routing Operations - Checkout Sync",
+    "summary": "Glue module",
```
simahawk: this summary should be improved

guewen:
@simahawk attended to your comments

simahawk (approved):

simahawk:
@guewen reminder: remove [this line](https://github.com/OCA/wms/blob/13.0/test-requirements.txt#L2) when merged.

simahawk:
> @guewen reminder: remove [this line](https://github.com/OCA/wms/blob/13.0/test-requirements.txt#L2) when merged.

@guewen  could you include the change here so that we don't need another PR later?

guewen:
>  could you include the change here so that we don't need another PR later?

@simahawk Done, also I fixed https://github.com/OCA/stock-logistics-warehouse/pull/855 tests

PR #124, opened 2021-01-14
[14.0] fix missing url key in manifests

hparfr:
because pre-commit complains otherwise


simahawk (commented):
make sense :) 
Shouldn't be opened on 13.0?

hparfr:
No need, there is not this precommit check  on v13.

hparfr:
@simahawk can you merge it? I need it in order to rebase my other pr and then have a green pre-commit

PR #127, opened 2021-01-14
[13.0] Shopfloor: use estimated package weight if weight is not set

sebalix:
For `shopfloor` and `stock_storage_type`.

Depends on https://github.com/OCA/stock-logistics-workflow/pull/777

Ref. 1900


shopfloor/actions/data.py
```diff
@@ -86,14 +90,28 @@ def package(self, record, picking=None, with_packaging=False, **kw):
...
-    def _package_parser(self):
-        return [
+    def _package_parser(self, picking=None):
+        parser = [
             "id",
             "name",
-            "pack_weight:weight",
             ("package_storage_type_id:storage_type", ["id", "name"]),
         ]
+        if picking:
```
guewen: to avoid duplicating the whole block, I would do this:

```python
    context_kwargs = {}
    if picking:
        context_kwargs["picking_id"] = picking.id
    parser.append(
        (
            "pack_weight:weight",
            lambda rec, fname, picking=picking: rec.pack_weight
            or rec.with_context(**context_kwargs).estimated_pack_weight,
         ),
    )
```

guewen (approved):

sebalix:
<s>Tests still need to be fixed in `test_checkout_list_package.py`</s>
Fixed

shopfloor/actions/data.py
sebalix: Done

TDu (approved):

shopfloor/actions/data.py
```diff
@@ -86,15 +90,25 @@ def package(self, record, picking=None, with_packaging=False, **kw):
...
-        return [
+    def _package_parser(self, picking=None):
+        parser = [
             "id",
             "name",
-            "pack_weight:weight",
             ("package_storage_type_id:storage_type", ["id", "name"]),
         ]
 
+        context_kwargs = {}
```
simahawk: I'd remove all these changes and add a new computed context dependent field instead (eg: [real|final|shopfloor]_weight) and replace the parser line `pack_weight:weight` w/ `real_weight:weight`. Then when you need to tweak the weight just pass proper ctx keys (eg: `package_real_weight=True`)

simahawk (changes requested):

simahawk:
base PR is getting merged... hopefully :crossed_fingers:

simahawk (approved):

PR #133, opened 2021-01-20
[13.0] shopfloor: adaptations for new base_rest API

guewen:
Following changes in base_rest: https://github.com/OCA/rest-framework/pull/71

The following changes are needed:

* Adapt "dispatch" method to new base_rest API with *args 
* Fix REST services tests
* Fix validators for new base_rest API

Details in commits.

I'll do a follow-up for the last commit, replacing the `__getattr__` override by a change in `base_rest`.
But merging this already would make the CI pass.

### DIFF
--- a/shopfloor/services/service.py
+++ b/shopfloor/services/service.py
@@ -14,6 +14,7 @@ from odoo.osv import expression
 
 from odoo.addons.base_rest.controllers.main import _PseudoCollection
 from odoo.addons.component.core import AbstractComponent, WorkContext
+from odoo.addons.component.exception import NoComponentError
 
 
 def to_float(val):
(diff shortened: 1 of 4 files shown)


simahawk (approved):

shopfloor/services/service.py
```diff
@@ -194,19 +197,40 @@ def _to_json(self, records):
...
+    def _get_validator_schema(self, method_name, usage_suffix):
         validator_component = self.component(
-            usage="%s.validator.response" % self._usage
+            usage="{}.{}".format(self._usage, usage_suffix)
         )
-        return validator_component._get_validator(method_name)
+        return getattr(validator_component, method_name)
+
+    def __getattr__(self, item):
+        # We have delegated the validator / return validators to dedicated
```
simahawk, line 209: can you add a `FIXME:` here?
Also, maybe is worth to involve @lmignon in the discussion on how to support this in core base_rest?

guewen: working on this here https://github.com/OCA/rest-framework/pull/99

guewen: and here https://github.com/OCA/wms/commit/e5905303d5a4ae719283b92a59fa84085e3f97be

guewen: fixme added

PR #135, opened 2021-01-20
[13.0] Add barcode on Package Storage Type, use it at reception screen

carlosserratoro:
While in the step 'Product Packaging' of the reception screen
of an incoming transfer, allow to scan a barcode for a package
storage type. If a package storage type is found, then fill it
automatically and set the product packaging also from it.

In the reception screen, the order in which the product packaging
and the package storage type has been reversed, because it is the
latter which is now used while scanning, so it comes first now
in the user interface.

### DIFF
--- a/stock_reception_screen/models/stock_reception_screen.py
+++ b/stock_reception_screen/models/stock_reception_screen.py
@@ -392,9 +392,13 @@ class StockReceptionScreen(models.Model):
         self.current_move_line_package = barcode
 
     def on_barcode_scanned_select_packaging(self, barcode):
-        """Auto-complete the package data."""
+        """Auto-complete the package data.
+
+        The package data is filled automatically depending on
+        the barcode scanned for the package storage type.
+        """
         packaging = self.current_move_product_packaging_ids.filtered(
-            lambda o: o.barcode == barcode
+            lambda o: o.package_storage_type_id.barcode == barcode
         )[:1]
         self._autocomplete_package_data(packaging)
 
--- a/stock_reception_screen/tests/test_reception_screen.py
+++ b/stock_reception_screen/tests/test_reception_screen.py
@@ -16,6 +16,7 @@ class TestReceptionScreen(SavepointCase):
         cls.storage_type_pallet = cls.env.ref(
             "stock_storage_type.package_storage_type_pallets"
         )
+        cls.storage_type_pallet.barcode = "123"
         cls.product = cls.env.ref("product.product_delivery_01")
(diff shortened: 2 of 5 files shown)


sebalix (approved):
LG, waiting for Travis

simahawk:
@carlosserratoro travis fails because of shopfloor. It has been patched and should be green w/ a rebase.

stock_reception_screen/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Stock - Reception screen",
     "summary": "Dedicated screen to receive/scan goods.",
-    "version": "13.0.1.1.3",
+    "version": "13.0.1.1.4",
```
simahawk: do not change the version unless you need a specific migration step, the OCA bot will do it on merge :wink:

stock_storage_type/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Stock Storage Type",
     "summary": "Manage packages and locations storage types",
-    "version": "13.0.1.6.2",
+    "version": "13.0.1.6.3",
```
simahawk: same

simahawk (changes requested):
can you split the commits by module?
Something like:

`stock_storage_type: add barcode`
`stock_reception_screen: select packaging via storage type`

carlosserratoro:
@simahawk thanks for your review. I attended all your changes. The PR is in green now.

guewen (approved):

simahawk (approved):

simahawk:
@carlosserratoro I won't block for commit messages, but as I reported in my comment above it's always better - when possible - to include the module name in the commit message. This is a good rule of thumb and it's even reported in Odoo's guidelines https://www.odoo.com/documentation/13.0/reference/guidelines.html#commit-message-full-description
Just keep it in mind for the next time :wink:

simahawk:
Thanks for your 1st contrib here! :tada: 

/ocabot merge minor

carlosserratoro:
> @carlosserratoro I won't block for commit messages, but as I reported in my comment above it's always better - when possible - to include the module name in the commit message. This is a good rule of thumb and it's even reported in Odoo's guidelines https://www.odoo.com/documentation/13.0/reference/guidelines.html#commit-message-full-description
> Just keep it in mind for the next time

OK, I'll do ─ I had understood that you preferred one commit per module, not that the name of the module had to be in the commit message, my fault (thanks for the link!)

PR #145, opened 2021-01-22
[13.0] Add shopfloor_workstation

TDu:
ref: 1943

### DIFF
--- a/shopfloor_mobile/static/wms/src/components/screen.js
+++ b/shopfloor_mobile/static/wms/src/components/screen.js
@@ -58,8 +58,11 @@ Vue.component("Screen", {
             return (
                 this.$root.authenticated &&
                 !this.$root.has_profile &&
+                // TODO: Routes should have a metadata flag `profile_required` ?
+                // to replace the following check.
                 this.$route.name != "profile" &&
-                this.$route.name != "settings"
+                this.$route.name != "settings" &&
+                this.$route.name != "workstation"
             );
         },
     },
@@ -174,6 +177,12 @@ Vue.component("Screen", {
                             <v-btn @click="$router.push({'name': 'profile'})">
                                 <v-icon>mdi-account-cog</v-icon>
                                 <span>{{ $t('app.profile_configure') }}</span>
+                            </v-btn>
+                        </v-col>
+                        <v-col class="text-center" cols="12">
+                            <v-btn @click="$router.push({'name': 'workstation'})">
+                                <v-icon>mdi-account-cog-outline</v-icon>
+                                <span>{{ $t('app.workstation_scan') }}</span>
                             </v-btn>
                         </v-col>
                     </v-row>
--- a/shopfloor_mobile/static/wms/src/i18n/i18n.fr.js
+++ b/shopfloor_mobile/static/wms/src/i18n/i18n.fr.js
@@ -28,6 +28,11 @@ const messages_fr = {
                 name: "Profil",
                 title: "Choisissez un profil",
             },
+            workstation: {
+                name: "Station de travail",
+                title: "Choisissez une station de travail",
+                workstation_updated: "Station de travail mise à jour",
+            },
             fullscreen: {
                 enter: "Go fullscreen",
                 exit: "Exit fullscreen",
--- a/shopfloor_mobile/static/wms/src/settings/workstation.js
+++ b/shopfloor_mobile/static/wms/src/settings/workstation.js
@@ -1,0 +1,66 @@
+/**
+ * Copyright 2021 Camptocamp SA (http://www.camptocamp.com)
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+ */
+
+export var Workstation = Vue.component("workstation", {
+    data: function() {
+        return {
+            usage: "workstation",
+            workstation_scanned: false,
+            scan_data: {},
+            scan_message: {},
+            search_input_placeholder: "Scan workstation code",
+        };
+    },
+    template: `
+        <Screen :screen_info="{title: $t('screen.settings.workstation.title'), klass: 'settings settings-workstation'}">
+            <template v-slot:header>
+                <user-information
+                    v-if="workstation_scanned"
+                    :message="scan_message"
+                    />
+            </template>
+            <searchbar
+                v-on:found="on_scan"
+                :input_placeholder="search_input_placeholder"
+                />
+
+            <div class="button-list button-vertical-list full">
+                <v-row align="center">
+                    <v-col class="text-center" cols="12">
+                        <btn-back/>
+                    </v-col>
+                </v-row>
+            </div>
+        </Screen>
+    `,
+    mounted() {
+        const odoo_params = {usage: this.usage};
+        this.odoo = this.$root.getOdoo(odoo_params);
+    },
+    methods: {
+        on_scan: function(scanned) {
+            this.odoo.call("setdefault", {barcode: scanned.text}).then(result => {
+                this.workstation_scanned = true;
+                // TODO : See how well a 404 when the shopfloor_workstation
+                // module is not installed will be handeled.
+                // Maybe there will be some this.$root.appconfig.features.xyz
+                // to test if installed.
+                this.scan_data = result.data;
+                this.scan_message = result.message;
+                if (this.scan_data) {
+                    if (this.scan_data.profile) {
+                        this.$root.trigger(
+                            "profile:selected",
+                            this.scan_data.profile,
+                            true
+                        );
+                    }
+                    // TODO: the success message will not be displayed, as we change screen !
+                    this.$root.$router.push({name: "settings"});
+                }
+            });
+        },
+    },
+});
--- a/shopfloor_workstation/readme/DESCRIPTION.rst
+++ b/shopfloor_workstation/readme/DESCRIPTION.rst
@@ -1,0 +1,5 @@
+Shopfloor workstation are useful for a quick configuration of the shopfloor
+mobile application and its connected user.
+
+To synchronize a connected user to a workstation setting, go to
+Settings - Workstation in the mobile application and scan a workstation barcode.
--- a/shopfloor_workstation/services/workstation.py
+++ b/shopfloor_workstation/services/workstation.py
@@ -1,0 +1,92 @@
+# Copyright 2021 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import _
+
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+
+
+class ShopfloorWorkstation(Component):
+    """
+    Workstation service for the mobile application.
+
+    This allows to scan a workstation bar code on the floor, which
+    will set a default profile on the mobile app and offer the option
+    to change some configuration of the connected user.
+
+    """
+
+    _inherit = "base.shopfloor.service"
+    _name = "shopfloor.workstation"
+    _usage = "workstation"
+    _description = __doc__
+
+    def setdefault(self, barcode):
+        """Endpoint that receives a scanned barcode."""
+        ws = self.env["shopfloor.workstation"].search([("barcode", "=", barcode)])
+        if ws:
+            ws.set_as_default_on_user(self.env.user)
+            message = {
+                "message_type": "info",
+                "body": _("Default workstation set to {}").format(ws.name),
+            }
+        else:
+            message = {
+                "message_type": "error",
+                "body": _("Workstation not found"),
+            }
+        return self._response(
+            message=message, data=self._convert_one_record(ws) if ws else {},
+        )
+
+    def _convert_one_record(self, record):
+        profile_data = None
+        if record.shopfloor_profile_id:
+            profile_data = {
+                "id": record.shopfloor_profile_id.id,
+                "name": record.shopfloor_profile_id.name,
+            }
+        return {
+            "id": record.id,
+            "name": record.name,
+            "barcode": record.barcode,
+            "profile": profile_data,
+        }
+
+
+class ShopfloorWorkstationValidator(Component):
+    """Validators for the Workstation endpoints"""
+
+    _inherit = "base.shopfloor.validator"
+    _name = "shopfloor.workstation.validator"
+    _usage = "workstation.validator"
+
+    def setdefault(self):
+        return {"barcode": {"type": "string", "nullable": True, "required": False}}
+
+
+class ShopfloorWorkstationValidatorResponse(Component):
+    """Validators for the Workstation endpoints responses"""
+
+    _inherit = "base.shopfloor.validator.response"
+    _name = "shopfloor.workstation.validator.response"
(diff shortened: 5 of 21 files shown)


shopfloor_mobile/static/wms/src/components/screen.js
```diff
@@ -59,7 +59,8 @@ Vue.component("Screen", {
                 this.$root.authenticated &&
                 !this.$root.has_profile &&
                 this.$route.name != "profile" &&
-                this.$route.name != "settings"
+                this.$route.name != "settings" &&
+                this.$route.name != "workstation"
```
simahawk, line 65: pls add a TODO that reminds to add a metadata flag to routes (eg: `profile_required`) then we can remove this hardcoded check.

shopfloor_mobile/static/wms/src/i18n/i18n.fr.js
```diff
@@ -28,6 +28,11 @@ const messages_fr = {
                 name: "Profil",
                 title: "Choisissez un profil",
             },
+            workstation: {
+                name: "Sation de travail",
```
simahawk:
```suggestion
                name: "Station de travail",
```

shopfloor_mobile/static/wms/src/settings/workstation.js
```diff
@@ -0,0 +1,72 @@
...
+        }
+    },
+    methods: {
+        on_scan: function(scanned) {
+            this.odoo.call("setdefault", {barcode: scanned.text}).then(result => {
+                this.workstation_scanned = true;
+                if (result.error) {
+                    console.error(result);
+                    this.scan_message = {
+                        message_type: "error",
```
simahawk: how do you know that this is the error? You should have it only if you get 404.

BTW We should have something like done here https://github.com/OCA/wms/blob/13.0/shopfloor_mobile/static/wms/src/scenario/mixins.js#L378 but for now is only for scenario.

shopfloor_mobile/static/wms/src/settings/workstation.js
```diff
@@ -0,0 +1,72 @@
...
+                        body:
+                            "Request could not be processed, maybe the module shopfloor_workstation is not installed.",
+                    };
+                    return;
+                }
+                this.scan_data = result.data;
+                this.scan_message = result.message;
+                if (this.scan_message.message_type === "info") {
```
simahawk: you should check if you have the profile or the workstation ID or a specific flag instead.

shopfloor_mobile/static/wms/src/settings/workstation.js
```diff
@@ -0,0 +1,72 @@
...
+                    this.$root.trigger(
+                        "profile:selected",
+                        this.scan_data.records[0].profile,
+                        true
+                    );
+                    // TODO: the success message will not be displayed, as we change screen !
```
simahawk, line 60: yeah, I did this workaround here https://github.com/OCA/wms/blob/13.0/shopfloor_mobile/static/wms/src/scenario/form.js#L58

shopfloor_workstation/services/workstation.py
```diff
@@ -0,0 +1,111 @@
...
+    This allows to scan a workstation bar code on the floor, which
+    will set a default profile on the mobile app and offer the option
+    to change some configuration of the connected user.
+
+    """
+
+    _inherit = "base.shopfloor.service"
+    _name = "shopfloor.workstation"
+    _usage = "workstation"
+    _expose_model = "shopfloor.workstation"
```
simahawk: expose_model is not needed if you don't use it

shopfloor_workstation/services/workstation.py
```diff
@@ -0,0 +1,111 @@
...
+    _description = __doc__
+
+    def setdefault(self, barcode):
+        """Endpoint that receives a scanned barcode."""
+        ws = self.env["shopfloor.workstation"].search([("barcode", "=", barcode)])
+        if self._set_workstation_as_default(ws):
+            message = {
+                "message_type": "info",
+                "body": "Default workstation set to {}".format(ws.name),
```
simahawk: make it translatable

shopfloor_workstation/tests/test_shopfloor_workstation.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.addons.shopfloor.tests.common import CommonCase
+
+
+class ShopfloorWorkstationCase(CommonCase):
+    def setUp(self):
```
simahawk: use setUpClass

simahawk (changes requested):

shopfloor_workstation/services/workstation.py
```diff
@@ -0,0 +1,111 @@
...
+            message=message, data={"size": len(ws), "records": self._to_json(ws)}
+        )
+
+    def _set_workstation_as_default(self, workstation):
+        """Apply changes to backend user configuration."""
+        if not workstation:
+            return False
+        if workstation.printing_printer_id:
+            # TODO : should the default action be checked ?
+            self.env.user.printing_printer_id = workstation.printing_printer_id
```
guewen: Does the user always has permission to write on this field?

shopfloor_mobile/static/wms/src/components/screen.js
TDu: Added just above

shopfloor_workstation/services/workstation.py
TDu: Yes he should, it is taken care of there https://github.com/OCA/report-print-send/blob/2160eedf36248d5b7660c5e41a4f03408271668c/base_report_to_printer/models/res_users.py#L28

shopfloor_workstation/tests/test_shopfloor_workstation.py
TDu: Done

shopfloor_workstation/services/workstation.py
TDu: Done

shopfloor_workstation/services/workstation.py
TDu: Removed

shopfloor_mobile/static/wms/src/settings/workstation.js
TDu: I have improved the check

TDu:
@simahawk I updated the PR with some fixes. I had to do a change on `odoo.js` which should be carefully reviewed, thanks.

shopfloor_mobile/static/wms/src/services/odoo.js
```diff
@@ -39,33 +39,29 @@ export class OdooMixin {
...
-                });
+                return {error: handler.call(this, response)};
             }
             return response.json();
         });
     }
-    _error_info(response, json) {
-        return _.extend({}, json, {
-            // strip html
-            description: $(json.description).text(),
```
simahawk: hmm this is required to properly display the error https://github.com/OCA/wms/blob/13.0/shopfloor_mobile/static/wms/src/scenario/mixins.js#L378

shopfloor_mobile/static/wms/src/settings/workstation.js
```diff
@@ -0,0 +1,76 @@
...
+        this.odoo = this.$root.getOdoo(odoo_params);
+        if (this.$route.params.identifier) {
+            this.getData(this.$route.params.identifier);
+        }
+    },
+    methods: {
+        on_scan: function(scanned) {
+            this.odoo.call("setdefault", {barcode: scanned.text}).then(result => {
+                this.workstation_scanned = true;
+                if (result.error) {
```
simahawk: what are you missing here that justifies the change above?

Also, I don't think this piece of code is required.
I would a TODO for something better: when app config is loaded return a mapping of features that are enabled.
This way you could simply check `this.$root.appconfig.features.xyz`

shopfloor_workstation/services/workstation.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+
+    _inherit = "base.shopfloor.service"
+    _name = "shopfloor.workstation"
+    _usage = "workstation"
+    _description = __doc__
+
+    def setdefault(self, barcode):
+        """Endpoint that receives a scanned barcode."""
+        ws = self.env["shopfloor.workstation"].search([("barcode", "=", barcode)])
```
simahawk, line 26:
```suggestion
        ws = self.env["shopfloor.workstation"].search([("barcode", "=", barcode)], limit=1)
```

shopfloor_workstation/services/workstation.py
```diff
@@ -0,0 +1,108 @@
...
+                "message_type": "info",
+                "body": _("Default workstation set to {}").format(ws.name),
+            }
+        else:
+            message = {
+                "message_type": "error",
+                "body": _("Workstation not found"),
+            }
+        return self._response(
+            message=message, data={"size": len(ws), "records": self._to_json(ws)}
```
simahawk: why do you return `records`? Here you should have a single record (the workstation) or nothing.
And you don't need `self._to_json`. Just run `_convert_one_record` or `ws.jsonify($parser, one=True)`

shopfloor_workstation/services/workstation.py
```diff
@@ -0,0 +1,108 @@
...
+            }
+        )
+
+    @property
+    def _record_schema(self):
+        return {
+            "id": {"coerce": to_int, "required": True, "type": "integer"},
+            "name": {"type": "string", "nullable": False, "required": True},
+            "barcode": {"type": "string", "nullable": False, "required": True},
+            "profile": {
```
simahawk, line 84: FTR, you could do `self.schemas._schema_dict_of(self.schemas._simple_record())`

shopfloor_workstation/views/shopfloor_workstation.xml
```diff
@@ -0,0 +1,67 @@
...
+        <field name="arch" type="xml">
+            <form>
+                <sheet>
+                    <widget
+                        name="web_ribbon"
+                        title="Archived"
+                        bg_color="bg-danger"
+                        attrs="{'invisible': [('active', '=', True)]}"
+                    />
+                    <group>
```
simahawk: name inner groups at least

shopfloor_mobile/static/wms/src/components/screen.js
```diff
@@ -176,6 +179,12 @@ Vue.component("Screen", {
                                 <span>{{ $t('app.profile_configure') }}</span>
                             </v-btn>
                         </v-col>
+                        <v-col class="text-center" cols="12">
+                            <v-btn @click="$router.push({'name': 'workstation'})">
+                                <v-icon>mdi-printer</v-icon>
```
simahawk: as the workstation is not just about printers, I'd pick a more generic icon like `cogs` or `account-cog-outline`

simahawk (changes requested):

shopfloor_mobile/static/wms/src/services/odoo.js
TDu: But where does the `description` comes from in the response, in the backend ?

TDu: The problem is https://github.com/OCA/wms/pull/145/files#diff-3b7d8d1af55526336a23050aa28b6605a0eb0625d5e129bde2cfd683d8ee42e5L42
crashes with 404 from Odoo (their body contains  html not json data.

shopfloor_workstation/services/workstation.py
TDu: In the latest changes I have added a constraint for the bar code to be unique so not  really necessary

shopfloor_workstation/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+Shopfloor workstation are useful for a quick configuration of the shopfloor
+mobile application and its connected user.
+
+To synchronize a connected user to a workstation setting, go to
+Settings - Workstation in the mobille application and scan a workstation barcode.
```
guewen:
```suggestion
Settings - Workstation in the mobile application and scan a workstation barcode.
```

shopfloor_workstation/services/workstation.py
```diff
@@ -0,0 +1,100 @@
...
+    _name = "shopfloor.workstation.validator"
+    _usage = "workstation.validator"
+
+    def search(self):
+        return {}
+
+    def get(self):
+        return {
+            "name_fragment": {"type": "string", "nullable": True, "required": False}
+        }
```
guewen: Are the search and get endpoints implemented by a super class of the component? Are they used?
I have the impression they do not exist/are not used and could be removed.

TDu: Thanks I removed the validator

shopfloor_workstation/readme/DESCRIPTION.rst
TDu: Fixed

TDu:
I would like to squash all the messy commits, Let me know when I can do it.

shopfloor_mobile/static/wms/src/settings/workstation.js
```diff
@@ -0,0 +1,69 @@
...
+                        <btn-back/>
+                    </v-col>
+                </v-row>
+            </div>
+        </Screen>
+    `,
+    mounted() {
+        const odoo_params = {usage: this.usage};
+        this.odoo = this.$root.getOdoo(odoo_params);
+        if (this.$route.params.identifier) {
```
simahawk: who's using this?

shopfloor_mobile/static/wms/src/settings/workstation.js
```diff
@@ -0,0 +1,69 @@
...
+                if (this.scan_data) {
+                    if (this.scan_data.profile) {
+                        this.$root.trigger(
+                            "profile:selected",
+                            this.scan_data.profile,
+                            true
+                        );
+                    }
+                    // TODO: the success message will not be displayed, as we change screen !
+                    this.$root.$router.push({name: "home"});
```
simahawk: why not getting back to settings page?

TDu: It was requested like that, but probably make more sense back to settings.
Also they may complain about an extra thumb click

shopfloor_mobile/static/wms/src/settings/workstation.js
TDu: Removed

simahawk (approved):
squash pls :)

sebalix (commented):
Need to add `report-print-send` repo in `oca_dependencies.txt` I think as the depend on `base_report_to_printer`

simahawk:
> Need to add `report-print-send` repo in `oca_dependencies.txt` I think as the depend on `base_report_to_printer`

this is not needed, we have MQT=PIP here.

If you have a pending PR add the pending PR to test-requirements.txt

simahawk:
hopefully fixed here https://github.com/OCA/wms/pull/173

PR #147, opened 2021-01-25
[13.0] Shopfloor Zone Picking: fix zero check

guewen:
## Fix "zero check" 

The client js application needs the move line's id to confirm or not the
zero quantity using "/is_zero". The backend did not send the move
line's data when going to the "zero_check" state.

## Improve wording on "zero check" screen

The zero check screen shows 2 buttons:

* Confirm stock = 0
* Confirm stock not empty

At this point, the normal situation would be that the stock is zero.
So "Confirm stock not empty" is confusing, as we do not confirm the
expected value of zero.

### DIFF
--- a/shopfloor/__manifest__.py
+++ b/shopfloor/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Shopfloor",
     "summary": "manage warehouse operations with barcode scanners",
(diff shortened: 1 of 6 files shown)


simahawk (approved):
Code review only. LG.

PR #148, opened 2021-01-25
[13.0][FIX] shopfloor: handle backorders

sebalix:
When moves are validated among all scenarios, we are validating related picking the standard way (with `action_done()`) when:

1) the moves to process are equal to all assigned moves of the current picking (so we can have waiting moves in the picking, in that case a normal backorder will be created by `action_done()`), in other terms it is triggered when we are validating the last moves related to one picking,

2) the moves to process are linked to unprocessed ancestor/source moves. This condition is required because we want to avoid some business logic bound to the `_create_backorder` standard method, like the backorder strategy configured through the `stock_picking_backorder_strategy` module.

In all other cases, we want to process the current moves in their own picking (use of `extract_and_action_done()` method), letting the current picking `confirmed/waiting` with the remaining qties.

Point 1) was already implemented, this commit implements the point 2).

Also, location content transfer scenario was not using the `validate_moves()` method (already used by all scenarios to validate
moves), this commit is fixing that.

Ref. 1966

### DIFF
--- a/shopfloor/__manifest__.py
+++ b/shopfloor/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Shopfloor",
     "summary": "manage warehouse operations with barcode scanners",
-    "version": "13.0.2.4.0",
+    "version": "13.0.2.4.1",
     "development_status": "Alpha",
     "category": "Inventory",
     "website": "https://github.com/OCA/wms",
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -32,10 +32,14 @@ class StockAction(Component):
     def _check_backorder(self, picking, moves):
         """Check if the `picking` has to be validated as usual to create a backorder.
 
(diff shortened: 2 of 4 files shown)


simahawk (approved):

sebalix:
The rule is a bit more complicated in fact, I'm putting the PR in WIP.

sebalix:
@simahawk implementation updated, the rule has to be applied for all scenarios (see the PR/commit description), not only location content transfer.

hparfr (approved):

hparfr:
@simahawk no /ocabot merge ?

simahawk:
@hparfr no. Most of the PRs get stuck because of the issue w/ travis .org/.com. The builds are messy and they need hours to get completed. I'm bumping version manually and merging manually to speed them up (and avoid useless builds BTW).

hparfr:
Ok no problem :evergreen_tree:

PR #137, opened 2021-01-20
[13.0] Add validator handlers delegating to component validators

guewen:
Following commit 81b9cfb3 which was a work-around, a new way to
customize the handlers for validators is added in base_rest:
https://github.com/OCA/rest-framework/pull/99

This commit uses the new component to dispatch the validators to
shopfloor's validator components.

PR #150, opened 2021-01-27
Add validator handlers delegating to component validators

bguillot:
Port #137 from 13.0 to 14.0.


simahawk:
already done in shopfloor_base

PR #144, opened 2021-01-22
[13.0] shopfloor_mobile: Add setting page improvements

TDu:
Move the log out button to the main setting page.
On the profile and language setting buttons, add the
active value.

PR #152, opened 2021-01-28
[13.0] Add workstation and setting improvements

TDu:
This is those two PR united
#145 
#144 
That are already reviewed.

And one commit that adds the name of the last scanned workstation on the button, like it is done for languages and profile
https://github.com/OCA/wms/commit/523be80ccf8175579015acbc3606a12fdad078d0

### DIFF
--- a/shopfloor_mobile/static/wms/src/main.js
+++ b/shopfloor_mobile/static/wms/src/main.js
@@ -87,6 +87,24 @@ const app = new Vue({
         app_info: function() {
             return shopfloor_app_info;
         },
+        available_languages: function() {
+            // FIXME: this should come from odoo and from app config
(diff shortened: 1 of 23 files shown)


shopfloor_mobile/static/wms/src/main.js
```diff
@@ -121,6 +122,21 @@ const app = new Vue({
                 this.$storage.set("profile", v);
             },
         },
+        has_set_workstation: function() {
```
simahawk: this is useless IMO. You can simply check `this.$root.workstation`

TDu: Thanks I removed it

simahawk:
https://github.com/OCA/wms/pull/173

PR #154, opened 2021-01-28
[13.0][IMP] stock_reception_screen: minor improvements

sebalix:
- invert the `set_location` and `select_packaging` steps
- filter the list of available destinations based on the package storage type
- product info: display the qty available, outgoing qty and the most recent expiry date

Ref. 1967

- Do not use anymore a compute+inverse field to handle the destination location of the processed move line to avoid sync issues. Instead a dedicated field is now used to handle the user's input (and its initial/default value)

Ref. 2039

### DIFF
--- a/stock_reception_screen/models/stock_reception_screen.py
+++ b/stock_reception_screen/models/stock_reception_screen.py
@@ -37,22 +37,22 @@ class StockReceptionScreen(models.Model):
         },
         "set_quantity": {
             "label": _("Set Quantity"),
-            "next_steps": [{"next": "set_location"}],
-            "focus_field": "current_move_line_qty_done",
-        },
-        "set_location": {
-            "label": _("Set Destination"),
             "next_steps": [
                 {
-                    "before": "_before_set_location_to_select_packaging",
+                    "before": "_before_set_quantity_to_select_packaging",
                     "next": "select_packaging",
                 }
             ],
-            "focus_field": "current_move_line_location_dest_id",
+            "focus_field": "current_move_line_qty_done",
         },
         "select_packaging": {
             "label": _("Select Packaging"),
+            "next_steps": [{"next": "set_location"}],
+        },
+        "set_location": {
+            "label": _("Set Destination"),
             "next_steps": [{"next": "set_package"}],
+            "focus_field": "current_move_line_location_dest_id",
         },
         "set_package": {
             "label": _("Set Package"),
@@ -570,7 +570,7 @@ class StockReceptionScreen(models.Model):
     def process_select_packaging(self):
         self.next_step()
 
-    def _before_set_location_to_select_packaging(self):
+    def _before_set_quantity_to_select_packaging(self):
         """Auto-complete the package data matching the qty (if there is one)."""
         qty_done = self.current_move_line_qty_done
         if qty_done:
@@ -698,9 +698,6 @@ class StockReceptionScreen(models.Model):
         assert self.current_step == "set_quantity", f"step = {self.current_step}"
         self.current_move_line_qty_done = qty_done
         self.process_set_quantity()
-        #   - set the destination
-        self.current_move_line_location_dest_id = location_dest
-        self.process_set_location()
         #   - set packaging data
         assert self.current_step == "select_packaging", f"step = {self.current_step}"
         self.product_packaging_id = product_packaging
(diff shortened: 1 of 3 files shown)


stock_reception_screen/models/stock_move_line.py
```diff
@@ -10,6 +10,11 @@ class StockMoveLine(models.Model):
     lot_life_date = fields.Datetime(
         string="End of Life Date", related="lot_id.life_date"
     )
+    original_location_dest_id = fields.Many2one(
+        comodel_name="stock.location",
+        ondelete="restrict",
+        string="Original default destination",
+    )
```
simahawk: Do you really need this field on move line? It seems used only for this screen.
Can't you have a mapping in for the reception screen only?
See below.

stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -124,8 +125,21 @@ class StockReceptionScreen(models.Model):
...
+    )
+    current_move_product_last_lot_life_date = fields.Datetime(
+        compute="_compute_current_move_product_last_lot_life_date",
+        string="Most recent exp. date",
+    )
     # current move line
     current_move_line_id = fields.Many2one(comodel_name="stock.move.line", copy=False)
+    current_move_line_original_location_dest_id = fields.Many2one(
+        related="current_move_line_id.original_location_dest_id"
+    )
```
simahawk:
```suggestion
    current_move_line_original_location_mapping = Serialized(default={})
    current_move_line_original_location_dest_id = fields.Many2one(
        compute="_compute_current_move_line_original_location_dest_id"
    )
```
In the compute lookup for the mapping that you build on create instead of storing the value on the move line itself.
What do you think?

stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -246,6 +284,28 @@ def _inverse_current_move_line_package(self):
         for wiz in self:
             wiz.current_move_line_package_stored = wiz.current_move_line_package
 
+    @api.depends("package_storage_type_id.location_storage_type_ids")
+    def _compute_allowed_location_dest_ids(self):
+        for wiz in self:
+            self.allowed_location_dest_ids = self.env["stock.location"].search(
```
simahawk:
```suggestion
            wiz.allowed_location_dest_ids = self.env["stock.location"].search(
```

simahawk (changes requested):

stock_reception_screen/models/stock_reception_screen.py
sebalix: I prefered to store the initial destination location to always propose it to the user (see the domain on the view) even if there are other places where to store the goods, because I suppose the user still don't know if the other places are free, so better to still propose the default value.
And it was store on create because when processing a line on the screen, we could change the destination, then cancel the screen. At this stage we have lost the original value computed by Odoo.

But by writing it maybe you're right: like the package, perhaps we should wait the validation of all the steps to make these data persistant on the move line... it'll require more work to do so.

stock_reception_screen/models/stock_reception_screen.py
sebalix: fixed

stock_reception_screen/models/stock_reception_screen.py
sebalix: I changed the implementation and dropped this field.

stock_reception_screen/models/stock_move_line.py
sebalix: After some checks, I rewrote the implementation of this, field removed.

jgrandguillaume (approved):
LGTM Thx !

simahawk (approved):

sebalix:
Issues have been found, do not merge this one please (ref. 2039)

guewen (approved):

simahawk:
> Issues have been found, do not merge this one please (ref. 2039)

@sebalix still the case?

guewen:
> still the case?

AFAIS it's fixed by the last commit

sebalix:
@simahawk sorry, yes fixed by the last commit

simahawk:
@sebalix squash pls :)

sebalix:
@simahawk done

PR #156, opened 2021-02-01
shopfloor: use db logging from rest_log

simahawk:
~Depends on https://github.com/OCA/rest-framework/pull/106~


shopfloor/models/shopfloor_log.py
```diff
@@ -1,130 +1,41 @@
...
-
-from odoo import api, fields, models, tools
-
-_logger = logging.getLogger(__name__)
+from odoo import models
+
+# TODO: drop this model on next release.
+# The feature has been moved to `rest_log`.
+#
+# It has been tried several times
```
lmignon, line 8: @simahawk Do you've tried to create a migration script using the methods [`rename_models`](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L747) and [`rename_tables`](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L683) from openupgradelib?

simahawk: yes, but the issue is not there. The issue is that when the ORM tries to delete the field options stored in ir.model.fields.selection it breaks because it does not find the model :man_shrugging: 
Maybe the problem stays in the ORM maybe in openupgradelib that misses this change.
I didn't want to spend time on digging into the issue tho.

lmignon (approved):

sebalix (approved):

PR #158, opened 2021-02-02
[13.0][IMP] shopfloor: Add standard messages for Checkout Packing Info

carlosserratoro:
This PR adds the option to choose predefined messages for the
parameter "Checkout Packing Info", that is set on the res.partner
and reused in other places. For this, the original text field
shopfloor_packing_info has been transformed into a Many2one that
points to the new model shopfloor.packing.info. Thus, the field
in the res.partner has been renamed to shopfloor_packing_info_id,
that has caused some renames in other files that used it. These
new messages are archived instead of deleted, to keep track of
old messages associated to existing (or archived) res.partners.

The new predefined messages can be created directly in the partner
or in the menu Inventory > Settings > Shopfloor.

To migrate the data, the existing data has been migrated by assuming
that the first line is the name while all the content is for the text
of the new model.

### DIFF
--- a/shopfloor/__manifest__.py
+++ b/shopfloor/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Shopfloor",
     "summary": "manage warehouse operations with barcode scanners",
-    "version": "13.0.2.5.0",
+    "version": "13.0.2.6.0",
     "development_status": "Alpha",
     "category": "Inventory",
     "website": "https://github.com/OCA/wms",
@@ -39,6 +39,10 @@
         # TODO: used for picking.carrier_id detail info.
         # This must be an optional dep
         "delivery",
+        # If "delivery" must be an optional dep, then sales_team must be also.
+        # Added to make normal sales users CRUD the shopfloor packing info.
+        # It is indirectly imported by "delivery".
+        "sales_team",
         #  OCA / product-attribute
         "product_packaging_type",
     ],
@@ -54,6 +58,7 @@
         "views/stock_picking_views.xml",
         "views/shopfloor_profile_views.xml",
         "views/shopfloor_log_views.xml",
+        "views/shopfloor_packing_info_views.xml",
         "views/menus.xml",
     ],
     "demo": [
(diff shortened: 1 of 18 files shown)


simahawk (changes requested):
please don't add another dependency here.
Move this feature to a new module and just hook to _data_for_packing_info

carlosserratoro:
> please don't add another dependency here.
> Move this feature to a new module and just hook to _data_for_packing_info

@simahawk I'm not sure if I understood you well.

* Regarding the dependency, I assume you mean ```sales_team```. I added it explicitly because it is needed for a permission. It's imported indirectly by the previous dependency that was already there: ```delivery```.

* Regarding the method ```_data_for_packing_info``` I assume you ask for it because the implementation changed. It changed because the field ```shopfloor_packing_info``` is now ```shopfloor_packing_info_id```, and that method uses that variable. The goal was not to add a new field ```shopfloor_packing_info_id```, but to substitute ```shopfloor_packing_info```.

simahawk:
> > please don't add another dependency here.
> > Move this feature to a new module and just hook to _data_for_packing_info
> 
> @simahawk I'm not sure if I understood you well.
> 
>     * Regarding the dependency, I assume you mean `sales_team`. I added it explicitly because it is needed for a permission. It's imported indirectly by the previous dependency that was already there: `delivery`.
> 
>     * Regarding the method `_data_for_packing_info` I assume you ask for it because the implementation changed. It changed because the field `shopfloor_packing_info` is now `shopfloor_packing_info_id`, and that method uses that variable. The goal was not to add a new field `shopfloor_packing_info_id`, but to substitute `shopfloor_packing_info`.

ok, but still: split it :wink:

carlosserratoro:
@simahawk Do you want to keep the field ```shopfloor_display_packing_info``` in the module ```shopfloor``` or do I extract it to the new module? If extracted, the method ```_data_for_packing_info``` will be just a placeholder returning the empty string in ```shopfloor``` and intended to be extended only.

simahawk:
> @simahawk Do you want to keep the field `shopfloor_display_packing_info` in the module `shopfloor` or do I extract it to the new module? If extracted, the method `_data_for_packing_info` will be just a placeholder returning the empty string in `shopfloor` and intended to be extended only.

Yes, leave it as hook to provide that info.

carlosserratoro:
Closed because was substituted by https://github.com/OCA/wms/pull/161.

PR #159, opened 2021-02-02
[13.0] stock_reception_screen exec picking action_done

TDu:
In the flow of the `stock_reception_screen` module, the `action_done` of
a stock.picking is never called. It is only done so on stock.move.

As the state of the stock.picking is computed it ends up with the
correct value, but any logic in its `action_done` function is never
executed.

This change fixes that.

### DIFF
--- a/stock_reception_screen/models/stock_reception_screen.py
+++ b/stock_reception_screen/models/stock_reception_screen.py
@@ -451,10 +451,18 @@ class StockReceptionScreen(models.Model):
         before checking the next move to process).
         """
         if self.current_move_line_id and self.current_move_id:
-            # We use the 'is_scrap' context key to avoid the generation of a
-            # backorder when validating the move (see _action_done() method in
-            # stock/models/stock_move.py).
-            self.current_move_id.with_context(is_scrap=True)._action_done()
+            moves_todo = self.picking_id.move_lines.filtered(
+                lambda m: m.state not in ["done", "cancel"]
+            )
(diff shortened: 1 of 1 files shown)


stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -451,10 +451,18 @@ def _validate_current_move(self):
         before checking the next move to process).
         """
         if self.current_move_line_id and self.current_move_id:
-            # We use the 'is_scrap' context key to avoid the generation of a
-            # backorder when validating the move (see _action_done() method in
-            # stock/models/stock_move.py).
-            self.current_move_id.with_context(is_scrap=True)._action_done()
+            moves_todo = self.picking_id.move_ids_without_package.filtered(
```
guewen: Is there a reason to look only moves without package? Looks dubious to me. If you have 2 remaining moves: one with a package level, one without, if you validate the latter, the picking is set to done and so the move with package level is moved to a backorder. IMO you want `move_lines`

TDu: Indeed fixed, thanks.

guewen (approved):

simahawk (approved):

sebalix:
There is an issue on this one: when it remains one move in the transfer, and we process only a partial qty of it, we should split the move in two so the picking is not validated directly. I'm working on a fix.

PR #161, opened 2021-02-03
[ADD] shopfloor_packing_info: Messages for Checkout Packing Info

carlosserratoro:
This PR adds the option to choose predefined messages for the
parameter "Checkout Packing Info", that is set on the res.partner
and reused in other places. For this, the original text field
shopfloor_packing_info has been transformed into a Many2one that
points to the new model shopfloor.packing.info. Thus, the field
in the res.partner has been renamed to shopfloor_packing_info_id,
that has caused some renames in other files that used it. These
new messages are archived instead of deleted, to keep track of
old messages associated to existing (or archived) res.partners.

This new model, shopfloor.packing.info has been moved to the new
module shopfloor_packing_info. Migration hooks have been provided
to migrate the data, that will happen if the old module is updated
at the same time in which the new module is installed.

The new predefined messages can be created directly in the res.partner
or in the menu Inventory > Settings > Shopfloor.

To migrate the data, the existing data has been migrated by assuming
that the first line is the name while all the content is for the text
of the new model.

ref: 1968


shopfloor_packing_info/hooks.py
```diff
@@ -0,0 +1,131 @@
...
+    if _column_exists(cr, "res_partner", "shopfloor_packing_info_tmp"):
+        copy_into_new_model(cr)
+        cr.execute(
+            """
+            ALTER TABLE res_partner
+            DROP COLUMN shopfloor_packing_info_tmp;"""
+        )
+
+
+def _column_exists(cr, table_name, column_name):
```
guewen: there is an implementation of this in `odoo.tools.sql.column_exists`

shopfloor_packing_info/models/shopfloor_packing_info.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import fields, models
+
+
+class ShopfloorPackingInfo(models.Model):
+    _name = "shopfloor.packing.info"
+    _description = "Shopfloor Packing Information"
+
+    name = fields.Char("Name", required=True)
+    text = fields.Text("Text", required=True)
+    active = fields.Boolean("Active", default=True)
```
guewen:
```suggestion
    name = fields.Char(required=True)
    text = fields.Text(required=True)
    active = fields.Boolean(default=True)
```

The string is generated from the field name already, better not to set it when it's the same

shopfloor_packing_info/services/checkout.py
```diff
@@ -0,0 +1,21 @@
...
+class CheckoutExt(Checkout):
+    _inherit = "base.shopfloor.process"
+
+    def _data_for_packing_info(self, picking):
+        res = super()._data_for_packing_info(picking)
+        if picking.picking_type_id.shopfloor_display_packing_info:
+            shopfloor_packing_info = (
+                picking.shopfloor_packing_info_id
+                and picking.shopfloor_packing_info_id.text
+                or ""
```
guewen: I recommend to never use the `... and ... or ...` ternary "operator", because it can be very misleading: see the comment on https://docs.python.org/3.3/faq/programming.html#is-there-an-equivalent-of-c-s-ternary-operator (can't be an issue here but a good habit IMO)

shopfloor_packing_info/views/shopfloor_packing_info_views.xml
```diff
@@ -0,0 +1,81 @@
...
+                        name="toggle_active"
+                        string="Archive"
+                        attrs="{'invisible': [('active', '=', False)]}"
+                    />
+                    <button
+                        type="object"
+                        name="toggle_active"
+                        string="Restore"
+                        attrs="{'invisible': [('active', '=', True)]}"
+                    />
```
guewen: Since 13.0 the archived button should use this:

```xml
                    <widget
                        name="web_ribbon"
                        title="Archived"
                        bg_color="bg-danger"
                        attrs="{'invisible': [('active', '=', True)]}"
                    />
```

shopfloor_packing_info/views/shopfloor_packing_info_views.xml
```diff
@@ -0,0 +1,81 @@
...
+            <search>
+                <field name="name" />
+                <field name="text" />
+                <field name="active" />
+                <separator />
+                <filter
+                    string="All"
+                    name="all"
+                    domain="[('active', 'in', [False, True])]"
+                />
```
guewen, line 50: did you find examples of this filter in core modules?

I find this:

```xml

                <filter
                    string="Archived"
                    name="inactive"
                    domain="[('active','=',False)]"
                />

```

and if I understand how the filter can be useful, but maybe we prefer UX consistency and use the `Archived` filter? (and so when you have this filter you don't need the active field in the tree view as you see either the archived either the unarchived records)

shopfloor_packing_info/views/shopfloor_packing_info_views.xml
```diff
@@ -0,0 +1,81 @@
...
+        <field name="res_model">shopfloor.packing.info</field>
+        <field name="view_mode">tree,form</field>
+        <field name="context">{}</field>
+        <field
+            name="view_ids"
+            eval="[(5, 0, 0),
+            (0, 0, {'view_mode': 'tree',
+                    'view_id': ref('shopfloor_packing_info.shopfloor_packing_info_tree_view')}),
+            (0, 0, {'view_mode': 'form',
+                    'view_id': ref('shopfloor_packing_info.shopfloor_packing_info_form_view')})]"
```
guewen, line 65: you shouldn't need to declare these if you have only one view of each and they'll be the default ones

shopfloor_packing_info/models/shopfloor_packing_info.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def unlink(self):  # pylint: disable=W8106
+        """Archive, not unlink"""
+        self.write({"active": False})
+
+    def toggle_active(self):
+        for record in self:
+            record.active = not record.active
```
guewen:
```suggestion
```

Already implemented in the base model

guewen (commented):
Added comments for details

shopfloor_packing_info/models/shopfloor_packing_info.py
```diff
@@ -0,0 +1,20 @@
...
+    _name = "shopfloor.packing.info"
+    _description = "Shopfloor Packing Information"
+
+    name = fields.Char("Name", required=True)
+    text = fields.Text("Text", required=True)
+    active = fields.Boolean("Active", default=True)
+
+    def unlink(self):  # pylint: disable=W8106
+        """Archive, not unlink"""
+        self.write({"active": False})
```
guewen: I'm not fund of this:

* if the record has been used, you'll have an error message as there is `ondelete='restrict'`, then you can archive it
* if the record has not been used, then I don't see the problem of deleting it (say you just created one by error)
* it goes against the conventions in other models, so is misleading for users

shopfloor_packing_info/tests/test_shopfloor_packing_info.py
```diff
@@ -0,0 +1,29 @@
...
+class TestShopfloorPackingInfo(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.packing_info = cls.env["shopfloor.packing.info"].create(
+            {"name": "Test Name", "text": "Test Text"}
+        )
+
+    def test_unlink(self):
+        partner = self.env["res.partner"].create(
```
simahawk: use an existing demo partner instead

shopfloor_packing_info/hooks.py
```diff
@@ -0,0 +1,131 @@
...
+            """
+            ALTER TABLE res_partner
+            ADD COLUMN shopfloor_packing_info_tmp TEXT;
+
+            UPDATE res_partner
+            SET shopfloor_packing_info_tmp=shopfloor_packing_info;"""
+        )
+        cr.execute(
+            """
+            DELETE FROM ir_ui_view WHERE id IN (
```
simahawk: is this needed? The view will be already deleted by the upgrade of shopfloor, no?

shopfloor_packing_info/hooks.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    for row in env.cr.dictfetchall():
+        partner_id = int(row["id"])
+        packing_info_full = row["packing_info_old"].strip()
+        if packing_info_full:
+            partner = res_partner.browse(partner_id)
+
+            # We assume the name is the first line, if there are several.
+            linesep_pos = packing_info_full.find(os.linesep)
```
simahawk, line 74: just `packing_info_full.splitlines()`?

shopfloor_packing_info/hooks.py
```diff
@@ -0,0 +1,131 @@
...
+            else:
+                packing_info_record = shopfloor_packing_info.create(
+                    {"name": packing_info_name, "text": packing_info_full}
+                )
+                partner.shopfloor_packing_info_id = packing_info_record
+                xml_id = hashlib.md5(packing_info_full.encode("utf-8")).hexdigest()
+                ir_model_data._update_xmlids(
+                    [
+                        {
+                            "xml_id": "shopfloor.packing_info_{}".format(xml_id),
```
simahawk: this xid is wrong. Also, I wouldn't add any xid because you did not have any record before, just strings inserted manually,

simahawk (changes requested):
I share all the remarks by Guewen and add few more :wink:

shopfloor_packing_info/models/shopfloor_packing_info.py
carlosserratoro: I agree. This was not a good idea...

shopfloor_packing_info/hooks.py
carlosserratoro: Shouldn't be needed. I'm going to remove it from here and place it in our own module, because in Travis I'm getting that error that says that a view still has a field that was removed (because of the copy of the view that is stored in the database). I thought that placing it here the problem would be fixed...

shopfloor_packing_info/hooks.py
carlosserratoro: I didn't know. Thanks!

shopfloor_packing_info/hooks.py
carlosserratoro: In this case it doesn't matter because it's a code that is executed once, but I didn't use splitlines nor split because it's much slower. See:

```python
>>> import timeit
>>>
>>> timeit.timeit("x.find('\\n')", setup="x = 'hi\\nhow\\nare\\nyou\\ndoing'")
0.10260158000164665
>>> timeit.timeit("x.find('\\n')", setup="x = 'hi\\nhow\\nare\\nyou\\ndoing'")
0.09811641899796086
>>> timeit.timeit("x.find('\\n')", setup="x = 'hi\\nhow\\nare\\nyou\\ndoing'")
0.09946276299888268
>>> 
>>> timeit.timeit('x.splitlines()', setup="x = 'hi\\nhow\\nare\\nyou\\ndoing'")
0.23253642599956947
>>> timeit.timeit('x.splitlines()', setup="x = 'hi\\nhow\\nare\\nyou\\ndoing'")
0.23402035599792725
>>> timeit.timeit('x.splitlines()', setup="x = 'hi\\nhow\\nare\\nyou\\ndoing'")
0.236937740999565
```

shopfloor_packing_info/hooks.py
carlosserratoro: You are right, the prefix should be ```shopfloor_packing_info```. The XML-id is placed here to ease exporting the existing records to clean them and then importing them in several databases without having to depend on the XML-id assigned by each database. Right now, this is actually not needed, and is thinking about the future when we do this field translatable. If you still don't want it, I can remove it for now...

simahawk: IMO it's useless. If you keep it I'd replace the module name w/ `__export__`

shopfloor_packing_info/views/shopfloor_packing_info_views.xml
carlosserratoro: I didn't search for any example, I just wanted the same funcionality that are in the scheduled actions by default, but here optional. I can change it by yours.

shopfloor_packing_info/views/shopfloor_packing_info_views.xml
carlosserratoro: That's the thing: right now there is just one view. By indicating the view we want, we protect against future views. This was done on purpose and I like to do it always this way (sometimes I use an explicit record of ```ir.actions.act_window.view``` for this... )

shopfloor_packing_info/hooks.py
carlosserratoro: Doesn't matter, I'll remove it.

carlosserratoro:
@simahawk @guewen https://github.com/OCA/wms/pull/161/commits/b3adbf214e16be433c82e03adbb8ed34d3679d2b has been added attending your comments.

shopfloor_packing_info/views/shopfloor_packing_info_views.xml
guewen: Generally you should never need this because the views you reference have a priority of 16, and when we add views that we don't want to be picked by default, we put a greater number such as 99. If you look in official or OCA addon, you'll never see this for default views, unless the views have a priority which would not make them the default.

guewen (approved):

shopfloor_packing_info/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+"id","name","model_id/id","group_id/id","perm_read","perm_write","perm_create","perm_unlink"
+"access_shopfloor_packing_info_user","shopfloor packing info user","model_shopfloor_packing_info","stock.group_stock_user",1,0,0,0
+"access_shopfloor_packing_info_manager","shopfloor packing info manager","model_shopfloor_packing_info","stock.group_stock_manager",1,1,1,0
```
guewen:
```suggestion
"access_shopfloor_packing_info_manager","shopfloor packing info manager","model_shopfloor_packing_info","stock.group_stock_manager",1,1,1,1
```

shopfloor_packing_info/demo/shopfloor_packing_info_demo.xml
```diff
@@ -0,0 +1,11 @@
...
+<odoo noupdate="1">
+    <record id="package_20kg" model="shopfloor.packing.info">
+        <field name="name">Package 20kg</field>
+        <field name="text">Max weight per package 20kg</field>
+    </record>
+    <record id="pallet_170cm" model="shopfloor.packing.info">
+        <field name="name">Pallet 1.7m</field>
+        <field name="text">Pallet height limit 1.7m</field>
+    </record>
+</odoo>
```
sebalix, line 11: Given these examples, is it not desirable to have several messages link to a partner, and concat them when needed? Like
```
Max weight per pallet 1000kg
Pallet height limit 1.7m
```
NOTE: but I don't know how this text field was used

PR #162, opened 2021-02-05
[13.0][FIX] shopfloor: handle better lines priority

sebalix:
**Before:**
when postponing lines to process, the priority was set to a
constant value 9999. But if all remaining lines to process have been
postponed, their priority was all set to 9999 (= no order) and it is
impossible to know what was the first lines which have been postponed
and that we should care about.
Also, the `shopfloor_postponed` boolean field was returning `True` if
the current priority was equal to 9999.

**After:**
Now to postpone a line we have to call the `shopfloor_postpone()`
method which will increment the priority of the line based on the max
priority used among all other lines of the current scope.
As a result, lines will always been ordered as expected, but as a
side-effect the `shopfloor_postponed` boolean field will always return
`True` as soon as the line has been postponed, but this is probably not
an issue in the lifecycle of the processed line.

Ref. 2018

### DIFF
--- a/shopfloor/models/priority_postpone_mixin.py
+++ b/shopfloor/models/priority_postpone_mixin.py
@@ -1,15 +1,12 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import api, fields, models
+from odoo import fields, models
 
 
 class PriorityPostponeMixin(models.AbstractModel):
     _name = "shopfloor.priority.postpone.mixin"
     _description = "Adds shopfloor priority/postpone fields"
 
-    # shopfloor_priority is set to this value when postponed
-    # consider it as the max value for priority
-    _SF_PRIORITY_POSTPONED = 9999
     _SF_PRIORITY_DEFAULT = 10
 
     shopfloor_priority = fields.Integer(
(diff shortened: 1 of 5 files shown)


shopfloor/models/priority_postpone_mixin.py
```diff
@@ -1,40 +1,33 @@
...
-    _SF_PRIORITY_POSTPONED = 9999
     _SF_PRIORITY_DEFAULT = 10
 
     shopfloor_priority = fields.Integer(
         default=lambda self: self._SF_PRIORITY_DEFAULT,
         copy=False,
         help="Technical field. Overrides operation priority in barcode scenario.",
     )
-
     shopfloor_postponed = fields.Boolean(
```
sebalix, line 17: @simahawk we are sending the value of this field to the client in the cluster picking scenario: https://github.com/OCA/wms/blob/13.0/shopfloor/services/cluster_picking.py#L395
With this PR this field will always return `True` as soon as it has been postponed at least one time, is it an issue? (I suppose no...).

simahawk: that's fine. It only sets a class on the line block to highlight it if required when the line is postponed.

simahawk (approved):

TDu (approved):

shopfloor/services/location_content_transfer.py
```diff
@@ -767,9 +767,13 @@ def postpone_package(self, location_id, package_level_id):
...
             return self._response_for_start(message=self.msg_store.record_not_found())
-        if package_level.exists():
-            package_level.shopfloor_postponed = True
         move_lines = self._find_transfer_move_lines(location)
+        if package_level.exists():
+            pickings = move_lines.mapped("picking_id")
+            sorter = self.actions_for("location_content_transfer.sorter")
+            sorter.feed_pickings(pickings)
+            package_levels = sorter.package_levels()
+            package_level.shopfloor_postpone(package_levels)
```
guewen: if we have move lines + package levels in a location, it should postpone at max(max(lines), max(package_levels)) I think

simahawk: @sebalix ?

sebalix: Indeed, if we are processing move lines and package levels from the same scope, I had to do it this way. Thanks!

sebalix: @simahawk check my last fixup!

simahawk: neat :)

simahawk (approved):
@sebalix can you squash?

sebalix:
@simahawk squashed

PR #163, opened 2021-02-05
[13.0] Optimize "location_is_empty" computed field on stock.location

lmignon:
This field is used only when applying a related location storage
type with "only_empty". As the field is computed each time
a stock.move, a stock.move.line or a quant references a location
and the computation reads all the quants and moves linked to
the location, we can prevent a lot of read and computation by
not computing this if it has no location storage type that will
need it. (ex 'Output' location).

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -153,6 +153,11 @@ class StockLocation(models.Model):
             for location in self.allowed_location_storage_type_ids
         )
 
+    def _should_compute_is_empty(self):
+        return self.usage == "internal" and any(
+            location.only_empty for location in self.allowed_location_storage_type_ids
+        )
+
     @api.depends(
         "quant_ids",
         "in_move_ids",
@@ -189,12 +194,12 @@ class StockLocation(models.Model):
         "out_move_line_ids.qty_done",
         "in_move_ids",
         "in_move_line_ids",
+        "allowed_location_storage_type_ids.only_empty",
     )
     def _compute_location_is_empty(self):
         for rec in self:
-            if rec.usage != "internal":
-                # No restriction should apply on customer/supplier/...
-                # locations.
+            if not rec._should_compute_is_empty():
+                # No restriction should apply
                 rec.location_is_empty = True
                 continue
             if (
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -143,19 +143,20 @@ class StockLocation(models.Model):
 
     def _should_compute_will_contain_product_ids(self):
         return self.usage == "internal" and any(
-            location.do_not_mix_products
-            for location in self.allowed_location_storage_type_ids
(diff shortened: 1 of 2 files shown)


stock_storage_type/models/stock_location.py
```diff
@@ -153,6 +153,11 @@ def _should_compute_will_contain_lot_ids(self):
             for location in self.allowed_location_storage_type_ids
         )
 
+    def _should_compute_is_empty(self):
+        return self.usage == "internal" and any(
+            location.only_empty for location in self.allowed_location_storage_type_ids
```
simahawk: does this flag exists on location? :thinking: 
```suggestion
            x.only_empty for x in self.allowed_location_storage_type_ids
```
maybe?

lmignon: @simahawk The same error exists in `_should_compute_will_contain_lot_ids`  and `_should_compute_will_contain_product_ids`. I'll modify the code for all these methods

lmignon: @simahawk Change done

simahawk (approved):

stock_storage_type/models/stock_location.py
```diff
@@ -143,14 +143,20 @@ def _compute_leaf_location_ids(self):
...
         return self.usage == "internal" and any(
-            location.do_not_mix_lots
-            for location in self.allowed_location_storage_type_ids
+            location_storage_type.do_not_mix_lots
+            for location_storage_type in self.allowed_location_storage_type_ids
+        )
+
+    def _should_compute_is_empty(self):
+        return self.usage == "internal" and any(
+            location_storage_type.only_empty
```
guewen, line 158: @lmignon we worked on the same thing  at the same time :sob: 
#164 

I reached the conclusion that we do want to write on the field even if the value change when we pick or put in a location, to force transactions to put or pick sequentially with a retry (
You can see my example in https://github.com/OCA/wms/pull/164/commits/9a3da83def369b9bc7f1975d50d3be2cb1ed9d6b)
But for other locations, we have to check if the value change and *not* write if it didn't change.

guewen: Also, `stock_storage_type_buffer` needed to handle `location_is_empty` invalidation: https://github.com/OCA/wms/pull/164/commits/402369071eceb08518c43fca5de9a94b6925e7bf#diff-29318822316101e435a8bb4690936de60502bf7b30abed00fac0ea9211f4d351

guewen: oops, I was not referring to the correct line, what I meant is:
https://github.com/OCA/wms/pull/164/commits/402369071eceb08518c43fca5de9a94b6925e7bf#diff-29318822316101e435a8bb4690936de60502bf7b30abed00fac0ea9211f4d351R214-R215

lmignon: @guewen I propose to close this one since yours is more complete.

PR #164, opened 2021-02-08
[13.0] stock_storage_type: optimize stored compute fields on stock.location

guewen:
Goal: avoid writes on stock.location when not necessary

3 stored computed fields on stock.location are used for storage type-based put-aways:

* Only Empty
* Will contain products
* Will contain lots

These fields are updated depending on the move/move lines/quants that goes in (or out for `location_is_empty`) of the location. These fields are actually used when:

* Only Empty: when we want to put something in a location, and it has a location storage type "Only Empty"
* Will contain products: when we want to put something in a location, and it has a location storage type "Do not mix products"
* Will contain lots: when we want to put something in a location, and it has a location storage type "Do not mix lots"

For the 2 last fields, the commits 461a26e changed them to be computed only if the location has a storage type with "do not mix...".  In this PR, the commit 40236907 does the same for `location_is_empty`. Details in the commit.

The previous commit e89b281 ensured that we would not compute these fields for non-internal locations, but even if not computed, the value was still written. The commit 9a3da83def3 updates the value for non-internal (or any that doesn't need the computation) locations only if it changes. We want to keep writing the values on locations that needs these constraints to be satisfied even if they don't change, to guarantee the end result is correct in case of concurrent transactions.

Another improvement that could be done is to try to acquire a `FOR UPDATE NOWAIT` lock when a location has been selected for a put-away and skip to the next in case it can't, which can prevent a retry of the transaction in case of concurrent errors.

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -161,13 +161,16 @@ class StockLocation(models.Model):
     )
     def _compute_location_will_contain_product_ids(self):
         for rec in self:
-            products = self.env["product.product"].browse()
-            if rec._should_compute_will_contain_product_ids():
-                products = (
-                    rec.mapped("quant_ids.product_id")
-                    | rec.mapped("in_move_ids.product_id")
-                    | rec.mapped("in_move_line_ids.product_id")
-                )
+            if not rec._should_compute_will_contain_product_ids():
+                if rec.location_will_contain_product_ids:
+                    no_product = self.env["product.product"].browse()
+                    rec.location_will_contain_product_ids = no_product
+                continue
+            products = (
+                rec.mapped("quant_ids.product_id")
+                | rec.mapped("in_move_ids.product_id")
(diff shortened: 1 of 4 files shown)


simahawk:
Have you seen https://github.com/OCA/wms/pull/163 ?

guewen:
Nope, I started working on this on Friday afternoon :sob:

stock_storage_type/models/stock_location.py
```diff
@@ -153,6 +155,11 @@ def _should_compute_will_contain_lot_ids(self):
             for location in self.allowed_location_storage_type_ids
         )
 
+    def _should_compute_location_is_empty(self):
+        return self.usage == "internal" and any(
+            location.only_empty for location in self.allowed_location_storage_type_ids
```
lmignon: /s/location/location_storage_type

lmignon (approved):
LGTM  (Code review)

sebalix (approved):

jbaudoux (approved):

PR #169, opened 2021-02-11
[13.0] Split shopfloor core -> shopfloor_base

simahawk:
See atomic commits for details. Key changes:

* `shopfloor` base stuff moved to `shopfloor_base`
* `shopfloor` keeps all scenario and stock related features
* `schema` is an action component, not a service component anymore (in line w/ `data`)
* `scan_anything` made pluggable with components (eg: you can search what you want)
* new model `shopfloor.scenario` add to hold per-scenario configuration via YAML (special flags on s.menu are still there in `shopfloor`)
* shopfloor menu moved to root menu as an app (a specific menu entry allows to see logs from shopfloor app only)
* no change yet in `shopfloor_mobile`

**TODO:**

- [x] Update README
- [x] Change license to LGPL on shopfloor_base
- [x] Add missing TODO to ROADMAP

Next: split shopfloor_mobile /  shopfloor_mobile_base.


simahawk:
@lmignon @sebalix @hparfr here's the split. If possible, if no blocking change is detected, I'd like to speed up this merge to avoid painful rebase/cherry-pick/merge as we are still pushing fix/imp to the app. Also, this would allow to battle test this change in our customer project in real life and to speed up the split of the mobile module.

I've tested some scenario and everything seems fine (even if nothing has changed that much from the functional POV) but it's always better to see it live in action.

If you'll have improvement proposals I'll track them down and consider them for the next round.

shopfloor_base/README.rst
```diff
@@ -0,0 +1,193 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/wms-13-0/wms-13-0-shopfloor
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/285/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Shopfloor is a barcode scanner application for internal warehouse operations.
```
lmignon: This description should be updated :smirk:

shopfloor_base/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,38 @@
+Profiles
+~~~~~~~~
+
+In Inventory / Configuration / Shopfloor / Profiles.
```
lmignon: The base shopfloor_base no more depends on `stock`.  `Inventory` as root into the path still relevant?

shopfloor_base/models/shopfloor_scenario.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class ShopfloorScenario(models.Model):
+    _name = "shopfloor.scenario"
+    _description = "Shopfloor Scenario"
+
+    name = fields.Char(required=True, translate=True)
+    key = fields.Char(required=True)
+    options = Serialized(compute="_compute_options", default={})
+    options_edit = fields.Char(help="Configure options via YAML")
```
hparfr: yaml ? 
Why not a sparse field like in storage ?

shopfloor_base/models/shopfloor_scenario.py
```diff
@@ -0,0 +1,70 @@
...
+except ImportError:
+    _logger.debug("`yaml` lib is missing")
+
+
+class ShopfloorScenario(models.Model):
+    _name = "shopfloor.scenario"
+    _description = "Shopfloor Scenario"
+
+    name = fields.Char(required=True, translate=True)
+    key = fields.Char(required=True)
```
hparfr: what is the purpose of key ? 
Not blocking but, can't we work without it ?

shopfloor_base/__manifest__.py
```diff
@@ -0,0 +1,41 @@
...
+    "category": "Inventory",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp, BCIM, Akretion, Odoo Community Association (OCA)",
+    "maintainers": ["guewen", "simahawk", "sebalix"],
+    "license": "AGPL-3",
+    "application": True,
+    "depends": [
+        "base_jsonify",
+        "base_rest",
+        "rest_log",
```
lmignon: Why do we have a dependency on `rest_log`? The only motivation seems to put some entries into the odoo menu :thinking:

lmignon (commented):
At first read LGTM, some comments into the code.

shopfloor_base/services/service.py
```diff
@@ -0,0 +1,267 @@
...
+
+        return response
+
+    _requires_header_menu = False
+    _requires_header_profile = False
+
+    def _get_openapi_default_parameters(self):
+        defaults = super()._get_openapi_default_parameters()
+        # Normal users can't read an API key, ignore it using sudo() only
+        # because it's a demo key.
```
hparfr, line 112: IMO This code should be put somewhere else but not in the base module.

shopfloor_base/services/service.py
```diff
@@ -0,0 +1,267 @@
...
+
+    @property
+    def data_detail(self):
+        return self._actions_for("data_detail")
+
+    @property
+    def msg_store(self):
+        return self._actions_for("message")
+
+    # TODO: maybe to be proposed to base_rest
```
hparfr, line 176: May be it's time ?

shopfloor_base/readme/CONFIGURE.rst
simahawk: yeah sorry, I forgot to mention that I still have to cleanup READMEs :sweat_smile:

hparfr (commented):
Just a code read for the moment.
Some minor remarks. Thanks for this great work.

shopfloor_base/__manifest__.py
simahawk: well, is not required indeed. Maybe for now I can move it to `shopfloor` as we expect it to be there. Other long term option is having a dumb module `shopfloor_rest_log` but it seems too much for just having a menu item.

shopfloor_base/models/shopfloor_scenario.py
simahawk: The sparse field is computed on it BUT this gives you much more flexibility in adding tech configurations w/out relying or real fields.
I find it very handy.

shopfloor_base/models/shopfloor_scenario.py
simahawk: it is mandatory: you really want 1 and only one scenario to match the usage of a component.

shopfloor_base/services/service.py
simahawk: These params are mandatory for the whole app ATM. I won't change in this PR. I can add a TODO to discuss about it later.

shopfloor_base/models/shopfloor_scenario.py
hparfr: Why not a db id then ?

shopfloor_base/services/service.py
simahawk: well, IDK, Current solution is a bit hack-ish and w/ the new api we might be able to add this validation via decorators.
I'd keep it as it is now and get back to it on a second round.

shopfloor_base/models/shopfloor_scenario.py
hparfr: Introducing a whole lib just for that, it's a bit overkill no ?

It's less hype but JSON does job.

shopfloor_base/services/service.py
hparfr: Ok

shopfloor_base/services/service.py
hparfr: ok

shopfloor_base/models/shopfloor_scenario.py
simahawk: IDG your point. How could you do it via ID? This is something the integrator/dev should provide and it should match the usage on both back and front. It's exactly what was used before in `scenario` as string. 
I could change it to `usage` if you prefer the name but we need something like this anyway.

shopfloor_base/models/shopfloor_scenario.py
simahawk: JSON sucks for manual editing. Adding pyyaml is not overkilling IMO.
It's an easy way to provide an easy way to edit options w/out getting insane w/ your own parser or hacks to display the values in a nice way (eg: body/body_str in se.index.config).

shopfloor_base/models/shopfloor_scenario.py
hparfr: I suggested id because I misunderstood the meaning of key.

Does it replace that : https://github.com/OCA/wms/pull/169/files/ebce44f5d58c96bc8cc208c3f3e87d2f32d0abcd#diff-6b10fc808919519f099c8197a6ee5aadca6444fb042593847fe1066d23228657L84 ?

shopfloor_base/models/shopfloor_scenario.py
hparfr: From an IT perspective : less dependencies = less conflicts + easier install

YAML has it's drawbacks too. It's still an hacky way to set things up, it's fine for the moment and I don't block it's addition even if I think it's not the good place to put it.

One more question: what are the extra settings you want to set?

simahawk: Extra settings can be whatever you want. They are translated into `self.options` in [services](https://github.com/OCA/wms/blob/13.0/shopfloor/services/service.py#L284) and they can come from the work ctx as well.
You see an example here on how it can be used https://github.com/OCA/wms/blob/13.0/shopfloor/services/checkout.py#L79
Here you see the config for existing scenario https://github.com/simahawk/wms/blob/shopfloor-split/shopfloor/data/shopfloor_scenario_data.xml#L5
The point is that these options should be "dynamic" and ppl should be able to use them as they prefer, even with nested values or list of values, etc.

That been said, I'll remove YAML and rely on good writing by configurators :)

shopfloor_base/models/shopfloor_scenario.py
simahawk: Yes, it replaces the string `scenario` that was on the menu model.

simahawk:
@lmignon @hparfr done:

* update readme
* add shopfloor_rest_log (LGPL) and dropped dep on rest_log
* use json instead of yaml for options edit
* RE-LICENSE shopfloor_base TO LGPL
* rebase on 13.0 (all latest changes included)

simahawk:
rest_log license fix is here https://github.com/OCA/rest-framework/pull/114

simahawk:
@lmignon @hparfr are we happy here? :)

shopfloor/models/shopfloor_menu.py
```diff
@@ -2,51 +2,13 @@
...
-        "location_content_transfer",
-    )
+    _inherit = "shopfloor.menu"
 
-    name = fields.Char(translate=True)
-    sequence = fields.Integer()
-    profile_id = fields.Many2one(
-        "shopfloor.profile", string="Profile", help="Visible on this profile only"
-    )
     picking_type_ids = fields.Many2many(
```
hparfr, line 9: `picking_type_ids` in not in the menu form view

shopfloor_base/models/shopfloor_scenario.py
hparfr: Personally, I prefer when this kind of fields are field Selection. From an UX perspective, I don't see why a user has to enter it itself. 

Because it's really a minor point, it's ok to leave it as is for the moment.

hparfr (approved):
Code and functional review done.

Thank @simahawk for this big refactoring.

shopfloor_base/models/shopfloor_scenario.py
simahawk: well, in theory the user is not the one that should edit this. I don't expect anyone to create scenario via UI as it would be useless. So, on the next round we should probably:

* make it not editable if not w/ very high permission
* make it a select

shopfloor/models/shopfloor_menu.py
simahawk: ahh, good catch thanks!

simahawk: fixed all missing fields / view override

simahawk:
> Code and functional review done.
> 
> Thank @simahawk for this big refactoring.

thanks for your review! :wink:

simahawk:
I've already changed major version.

/ocabot merge minor

simahawk:
test failing and one possible permission issue: I'll check in a while

simahawk:
bah, there's a test failing here https://travis-ci.com/github/OCA/wms/jobs/484254095#L1972 but I cannot reproduce it.
Likely something has changed somewhere in another PR or module or repo because the build on master is green but other PRs that do not contain the same changes as this one are failing for the same error. (eg: https://github.com/OCA/wms/pull/172)

simahawk:
Test fail because of this change https://github.com/odoo/odoo/commit/b33e72d0bf027fb2c789b1b9476f7edf1a40b0a6

simahawk:
last 3 commits fixes compatibilty w/ odoo core change https://github.com/odoo/odoo/commit/b33e72d0bf027fb2c789b1b9476f7edf1a40b0a6

- shopfloor: fix stock.move.extract_and_action_done
- shopfloor: improve _fill_stock_for_moves in tests
- shopfloor: fix package level unlink

simahawk:
Similar PR for mobile https://github.com/OCA/wms/pull/175

PR #170, opened 2021-02-11
[13.0] Shopfloor checkout: display carrier and validate packaging against it

simahawk:


shopfloor/actions/message.py
```diff
@@ -506,3 +506,11 @@ def package_open(self):
             "message_type": "info",
             "body": _("Package has been opened. You can move partial quantities."),
         }
+
+    def packaing_invalid_for_carrier(self, packaging, carrier):
```
```suggestion
    def packaging_invalid_for_carrier(self, packaging, carrier):
```

shopfloor/services/checkout.py
```diff
@@ -734,8 +734,19 @@ def scan_package_action(self, picking_id, selected_line_ids, barcode):
...
+        # Scan delivery packaging
         packaging = search.generic_packaging_from_scan(barcode)
         if packaging:
+            carrier = picking.carrier_id
+            # Validate against carrier
+            if carrier and not self._packaging_good_for_carrier(packaging, carrier):
+                return self._response_for_select_package(
+                    picking,
+                    selected_lines,
+                    message=self.msg_store.packaing_invalid_for_carrier(
```
sebalix:
```suggestion
                    message=self.msg_store.packaging_invalid_for_carrier(
```

sebalix (commented):
LG, small typo + should we update the comment in the manifest file there? https://github.com/OCA/wms/blob/13.0/shopfloor/__manifest__.py#L40-L42 Or it is still relevant?

simahawk:
> LG, small typo + should we update the comment in the manifest file there? https://github.com/OCA/wms/blob/13.0/shopfloor/__manifest__.py#L40-L42 Or it is still relevant?

Good catch!

simahawk:
done

TDu (approved):

PR #171, opened 2021-02-11
stock_storage_type: putaway strategy "do not mix" should give priorit…

### LINKED ISSUE #168: stock_storage_type: putaway strategy "do not mix" should give priority to location that already contains the product (the one with the less qty)
As stated [here](https://github.com/OCA/wms/pull/132#pullrequestreview-571363775) it would be nice to improve the put away strategy for 'do not mix product" to give priority to location that already contains the product.

A solution could be to reorder the list of valid stock locations into the the method [`_select_final_valid_putaway_locations`](https://github.com/OCA/wms/blob/13.0/stock_storage_type/models/stock_location.py#L377). This would require to modify the method signature to pass at least the package_storage_type..

something like this

```python

def _select_final_valid_putaway_locations(
    self, package_storage_type, quants, products, limit=None
):
    locations = self
    if package_storage_type.do_not_mix_products:
        # give priority to the location that already contains the product, the one with the less qty first
        locations = locations.sorted(lambda l:sum(l.mapped('quant_ids.qty')) or 99999)
    return locations[:limit]
``` 

My main concern here is that we break the original order but do we agree that's the expected result in this specific case?

lmignon:
…y to location that already contains the product (the one with the less qty)

fixes #168

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -522,10 +522,10 @@ class StockLocation(models.Model):
                 if (float_compare(item["quantity"], 0, precision_digits=2) > 0)
             ]
         valid_location_ids = set(valid_locations.ids) - set(loc_ordered_by_qty)
-        valid_locations = self.browse(loc_ordered_by_qty)
-        valid_locations |= self.browse(
+        ordered_valid_location_ids = loc_ordered_by_qty + [
             id_ for id_ in self.ids if id_ in valid_location_ids
-        )
+        ]
+        valid_locations = self.browse(ordered_valid_location_ids)
         return valid_locations
 
     def _get_ordered_leaf_locations(self):
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -3,6 +3,7 @@
 import logging
 
 from odoo import api, fields, models
+from odoo.tools import float_compare
 
 _logger = logging.getLogger(__name__)
 
@@ -506,24 +507,11 @@ class StockLocation(models.Model):
         (the ones with less qty first)
         """
         valid_no_mix = valid_locations.filtered("do_not_mix_products")
-        qty_by_loc = {}
+        loc_ordered_by_qty = []
         if valid_no_mix:
             StockQuant = self.env["stock.quant"]
             domain_quant = [("location_id", "in", valid_no_mix.ids)]
-            qty_by_loc = {
-                item["location_id"][0]: item["quantity"]
-                for item in StockQuant.read_group(
-                    domain_quant,
-                    ["location_id", "quantity"],
-                    ["location_id"],
-                    orderby="id",
-                )
-            }
-        loc_by_qty = []
-        if valid_no_mix:
-            StockQuant = self.env["stock.quant"]
-            domain_quant = [("location_id", "in", valid_no_mix.ids)]
-            loc_by_qty = [
+            loc_ordered_by_qty = [
                 item["location_id"][0]
                 for item in StockQuant.read_group(
                     domain_quant,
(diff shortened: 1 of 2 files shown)


stock_storage_type/models/stock_location.py
```diff
@@ -475,6 +497,55 @@ def select_allowed_locations(
...
+        if valid_no_mix:
+            StockQuant = self.env["stock.quant"]
+            domain_quant = [("location_id", "in", valid_no_mix.ids)]
+            qty_by_loc = {
+                item["location_id"][0]: item["quantity"]
+                for item in StockQuant.read_group(
+                    domain_quant,
+                    ["location_id", "quantity"],
+                    ["location_id"],
+                    orderby="id",
```
```suggestion
                    orderby="quantity",
```

stock_storage_type/models/stock_location.py
```diff
@@ -475,6 +497,55 @@ def select_allowed_locations(
...
+                )
+            }
+        return valid_locations.sorted(self._get_allowed_location_sorter(qty_by_loc))
```
```suggestion
        loc_by_qty = []
        if valid_no_mix:
            StockQuant = self.env["stock.quant"]
            domain_quant = [("location_id", "in", valid_no_mix.ids)]
            loc_by_qty = [
                item["location_id"][0]
                for item in StockQuant.read_group(
                    domain_quant,
                    ["location_id", "quantity"],
                    ["location_id"],
                    orderby="quantity",
                )
                if (float_compare(item["quantity"], 0) > 1)
             ]
        valid_location_ids = set(valid_locations.ids) - set(loc_by_qty)
        valid_locations = self.browse(loc_by_qty)
        valid_locations |= self.browse(
            id_ for id_ in self.ids if id_ in valid_location_ids
        )
        return valid_locations
```

jbaudoux (changes requested):
Thanks for the PR.

Just a performance issue. Sorting a big set takes to much time. The set is already sorted, we just need to put the few locations having quantity at the beginning.

stock_storage_type/models/stock_location.py
jbaudoux: and do not convert to dict to not lose the sorting

stock_storage_type/models/stock_location.py
lmignon: @jbaudoux A 'set' is not ordered. Moreover, you lost the new order into the call to self.browse since you iterate on the original order `self.ids`. An other point, the read_group could return 0 if we've negative quants. Finally you must also take into account that the qty into a location could be the same as another one. In such case, the initial order must be used to sort the locations with the same qty. All these cases are covered by the `sorter` method.

jbaudoux: @lmignon I updated the code. It wasn't right indeed.
I tried your PR on real customer DB and here is a result of the 2 methods:
len(self.ids): 8183
Without sorting as suggested: 0.001138 
Your version with sorting: 0.402326

lmignon: With real data to stress the logic it's always better... :smirk:

lmignon:
@jbaudoux I've to check why the tests are red now even with the error into typo are fixed... the sorting expected into the tests is no more the result one.

lmignon:
@jbaudoux I fixed your proposed code. Tests are now :green_circle: on my computer. Can you make another try to see if the performances are still better with your proposed improvements?

jbaudoux (approved):
Works perfectly now. Thanks

stock_storage_type/models/stock_location.py
```diff
@@ -475,6 +498,36 @@ def select_allowed_locations(
...
+                for item in StockQuant.read_group(
+                    domain_quant,
+                    ["location_id", "quantity"],
+                    ["location_id"],
+                    orderby="quantity",
+                )
+                if (float_compare(item["quantity"], 0, precision_digits=2) > 0)
+            ]
+        valid_location_ids = set(valid_locations.ids) - set(loc_ordered_by_qty)
+        valid_locations = self.browse(loc_ordered_by_qty)
```
simahawk: instead of intersection can't you browse once like this?
```py
ids = loc_ordered_by_qty + set([id_ for id_ in self.ids if id_ in valid_location_ids])
self.browse(ids)
```
```

simahawk (commented):
LG, one question

simahawk:
NOTE: I've restarted a [build](https://travis-ci.com/github/OCA/wms/builds/216921185) because I want to see if it fails as #169 and #172

simahawk:
fails here as well. I'm cheking odoo core as it seems something has changed in stock

simahawk:
> fails here as well. I'm cheking odoo core as it seems something has changed in stock

I confirm: I've checked out latest odoo 13.0 and it fails locally as well now :tada: :sob:

simahawk:
This https://github.com/odoo/odoo/commit/b33e72d0bf027fb2c789b1b9476f7edf1a40b0a6

lmignon:
> This [odoo/odoo@b33e72d](https://github.com/odoo/odoo/commit/b33e72d0bf027fb2c789b1b9476f7edf1a40b0a6)

:sob:

simahawk:
fixed here https://github.com/odoo/odoo/pull/66517

stock_storage_type/models/stock_location.py
lmignon: `set`will loose the initial order....

simahawk:
hopefully I've fixed shopfloor conflict w/ odoo core change in the last 3 commits in #169

simahawk:
@lmignon you can rebase finally :tada:

lmignon:
@simahawk done!

lmignon:
@simahawk Can we merge this one?

simahawk:
I thought it was already done :sweat_smile: 

/ocabot merge minor

lmignon:
Now it's :smirk: Thank you @simahawk

PR #172, opened 2021-02-11
[13.0][IMP] Reception screen warning messages

santostelmo:
### Context: 
Currently we use `web_notify` to show warning messages in the reception screen. But we noticed some bug:
If several tabs are open in the browser loading reception screen, sometimes the message is raised in the tab which is not currently visible. So far I coudn't find the solution for this bug.

### As an alternative solution I propose:
- Warning messages are stored in a HTML field and shown in the view if needed
- Dependency of `web_notify` can be removed.

### DIFF
--- a/stock_reception_screen/models/stock_reception_screen.py
+++ b/stock_reception_screen/models/stock_reception_screen.py
@@ -190,6 +190,23 @@ class StockReceptionScreen(models.Model):
     package_height = fields.Integer()
     # == / Packaging fields ==
 
+    warn_notification = fields.Char(default=False)
+    warn_notification_html = fields.Html(compute="_compute_warn_notification")
+
+    @api.depends("warn_notification")
+    def _compute_warn_notification(self):
+        warn_massage_html = """
+        <div class="alert alert-warning" role="alert">
+            {}
+        </div>
+        """
+        if self.warn_notification:
+            self.warn_notification_html = warn_massage_html.format(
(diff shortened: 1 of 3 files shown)


stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -350,6 +367,7 @@ def on_barcode_scanned(self, barcode):
 
     def on_barcode_scanned_select_product(self, barcode):
         """Try to find the corresponding product based on the barcode."""
+        self.warn_notification = False
```
sebalix: Instead of resetting `warn_notification` value on each defined step, is it possible to do this in `button_save_step()` method (and also in `button_next_pack()` I suppose), this is the method called each time a user validate a step to go to the next one.

sebalix (commented):
Nice! Small comment.

simahawk (approved):
pre approving

simahawk:
Tests fail because of https://github.com/odoo/odoo/commit/b33e72d0bf027fb2c789b1b9476f7edf1a40b0a6

sebalix (approved):

simahawk:
@santostelmo pls rebase

simahawk:
I could push to the branch, I did it :stuck_out_tongue_winking_eye:

santostelmo:
Thank you @simahawk

PR #180, opened 2021-02-26
[13.0][FIX] stock_reception_screen: split the move before validating

sebalix:
When we process a partial qty on the last move of the operation, the
operation should not be validated otherwise it'll create a backorder for
the remaining qty to receive.
Instead we have to split the current move with the remaining qty to
receive, so we get two moves and we only validate the current one.
Before commit 707013a8105ae5cf86008b0a3014dbeb1fba0cb6 it was not necessary
because the call to `_action_done` were already taking care of splitting
the partial processed qty to a new move.

Ref. 2139

### DIFF
--- a/stock_reception_screen/models/stock_reception_screen.py
+++ b/stock_reception_screen/models/stock_reception_screen.py
@@ -1,7 +1,12 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import _, api, exceptions, fields, models
+from odoo import _, api, exceptions, fields, models, tools
+
+
+def gt(value1, value2, digits):
+    """Return True if value1 is greater than value2"""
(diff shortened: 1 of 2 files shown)


TDu (approved):

stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -490,12 +490,16 @@ def process_select_product(self):
...
-        """
+        """Split the current move with the move line qty done and validate it."""
         if self.current_move_line_id and self.current_move_id:
+            remaining_qty = (
+                self.current_move_id.product_uom_qty
+                - self.current_move_line_id.qty_done
+            )
+            # We don't want to create a new move with a negative qty if we
+            # receive more than expected.
+            if remaining_qty > 0:
```
simahawk: `float_compare` not necessary?

sebalix: fixed

simahawk (approved):

PR #187, opened 2021-03-02
[FIX] stock_reception_screen: last line received with button "Next Pack"

carlosserratoro:
Fixes a crash that appeared when there were several lines to
pick and the last one was packed with the button "Next Pack".
For the case in which no more moves were waiting, there was
an assertion that was activated.

A test has been added to check this scenario. Because of the
test added, some refactoring in the file for tests has been done.

### DIFF
--- a/stock_reception_screen/models/stock_reception_screen.py
+++ b/stock_reception_screen/models/stock_reception_screen.py
@@ -755,8 +755,9 @@ class StockReceptionScreen(models.Model):
         package_storage_type = self.package_storage_type_id
         package_height = self.package_height
         # Validate the current package
-        if not self.process_set_package():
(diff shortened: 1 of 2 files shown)


sebalix:
Seems legit, but I was wondering how they reproduce it knowing that the "Next pack" button is invisible if the current step is != than `set_package`, do you know? (with two screens open from different browser tabs/computers probably?)

simahawk (approved):
seems legit but good question by @sebalix

carlosserratoro:
@sebalix the button is visible in the step in which you select a package. Is here when the crash happens, not before.

sebalix (approved):

PR #189, opened 2021-03-03
[FIX] stock_dynamic_routing_reserve_rule: fix Reservation Rules

carlosserratoro:
Button to access the reservation rules from a dynamic routing
crashed because of a bad name in a field, probably because of
an old refactoring.


ajaniszewska-dev (approved):

sebalix (approved):
Indeed, it was done here: https://github.com/OCA/stock-logistics-warehouse/pull/1013

PR #190, opened 2021-03-03
[13.0] shopfloor_mobile: split core -> shopfloor_mobile_base

simahawk:
In this episode:

* shopfloor_mobile: moved core features to shopfloor_mobile_base (only WMS related stuff remains in shopfloor_mobile)
* add page_registry -> register your own page routes and automatically collect setting pages
* add utils registry -> register your own utils to be used everywhere in components
* add config registry -> easily plug computed / reactive root config properties that are stored in local storage too
* add shopfloor_workstation_mobile (offers a good example on how to add new config options)
* add shopfloor_example: originated from shopfloo_mobile_custom_example, offers a complete example on how to define your own process
* login required by route param
* profile required by route param
* misc fix/imp

TODO: adjust READMEs and any other small glitch.
Replaces #175

PR #191, opened 2021-03-05
[14.0][MIG] shopfloor_base

hparfr:


rvalyi (approved):
shopfloor_base doesn't depend on anything that changed between v13 and v14, hence the migration is trivial. So I approve.

simahawk (approved):
skimmed through last commits: all good.

This is the next one: https://github.com/OCA/wms/pull/190
It contains some small fixes for backend as well :wink:

PR #192, opened 2021-03-08
[13.0] shopfloor_mobile: checkout display qty using qty widget 

simahawk:


shopfloor_mobile/static/wms/src/components/scenario_picking_detail/picking_select.js
```diff
@@ -76,8 +76,19 @@ Vue.component("picking-select-package-content", {
             <div class="lot" v-if="record.lot">
                 <span class="label">Lot:</span> <span>{{ record.lot.name }}</span>
             </div>
-            <div class="qty">
-                <span class="label">Qty:</span> <span>{{ record.qty_done }} / {{ record.quantity }}</span>
+            <div class="qty done">
```
sebalix, line 79: `done` is not used at all in the code, right? Same question for `requested`, I can't find them.

sebalix (approved):

shopfloor_mobile/static/wms/src/components/scenario_picking_detail/picking_select.js
simahawk: nope, it's just to identify the two lines if different styles are needed

PR #194, opened 2021-03-11
[13.0][IMP] shopfloor: checkout, use of delivery packages

sebalix:
<s>PR based on https://github.com/OCA/wms/pull/193, only the two last commits are relevant.</s>

A delivery package is a package with a delivery packaging defined ('packaging_id' field). It is a package different than internal ones (which are stored in the warehouse with their own constraints like height...).
Here we enforce the use of such delivery packages when we perform a checkout, this means that we can not use internal packages to perform a delivery.

* `list_dest_package` method ('/select_dest_package') is now listing only delivery packages
* `scan_dest_package` and 'set_dest_package' methods accept only a delivery package
* `scan_package_action` method accept only a delivery package
* Add a new state/screen when the user click on the "_New package_" button to list available delivery packaging (`select_delivery_packaging` state) for the current transfer and select one to create the package (existing `new_package` method which now accepts a packaging as parameter).

Ref. 2168

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -92,15 +92,17 @@ class Checkout(Component):
         return ""
 
     def _response_for_select_dest_package(self, picking, move_lines, message=None):
-        packages = picking.mapped("move_line_ids.result_package_id")
+        packages = picking.mapped("move_line_ids.result_package_id").filtered("packaging_id")
         if not packages:
+            # FIXME: do we want to move from 'select_dest_package' to
+            # 'select_package' state? Until now (before enforcing the use of
+            # delivery package) this part of code was never reached as we
+            # always had a package on the picking (source or result)
+            # Also the response validator did not support this state...
             return self._response_for_select_package(
                 picking,
                 move_lines,
-                message={
-                    "message_type": "warning",
-                    "body": _("No valid package to select."),
-                },
+                message=self.msg_store.no_valid_package_to_select(),
             )
         picking_data = self.data.picking(picking)
         packages_data = self.data.packages(
@@ -617,26 +619,30 @@ class Checkout(Component):
         return move_line.qty_done > 0 and move_line.shopfloor_checkout_done
 
     def _is_package_allowed(self, picking, package):
-        existing_packages = picking.mapped("move_line_ids.result_package_id")
+        """Check if a package is allowed as a destination/delivery package.
+
+        A package is allowed as a destination one if it is present among
+        `picking` lines and qualified as a "delivery package" (having a
+        delivery packaging set on it).
+        """
+        existing_packages = picking.mapped("move_line_ids.result_package_id").filtered(
+            "packaging_id"
+        )
         return package in existing_packages
 
     def _put_lines_in_package(self, picking, selected_lines, package):
         """Put the current selected lines with a qty_done in a package
 
-        Note: only packages which are already a destination package for another
+        Note: only packages which are already a delivery package for another
         line of the stock picking can be selected. Packages which are the
-        source packages are allowed too (we keep the current package), but
-        since Odoo set the value of the result package to the source package by
-        default, it works by default.
+        source packages are allowed too only if it is a delivery package (we
+        keep the current package).
         """
         if not self._is_package_allowed(picking, package):
             return self._response_for_select_package(
                 picking,
                 selected_lines,
-                message={
-                    "message_type": "error",
-                    "body": _("Not a valid destination package").format(package.name),
-                },
+                message=self.msg_store.dest_package_not_valid(package),
             )
         return self._put_lines_in_allowed_package(picking, selected_lines, package)
 
@@ -676,10 +682,10 @@ class Checkout(Component):
     def scan_package_action(self, picking_id, selected_line_ids, barcode):
         """Scan a package, a lot, a product or a package to handle a line
 
-        When a package is scanned, if the package is known as the destination
-        package of one of the lines or is the source package of a selected
-        line, the package is set to be the destination package of all then
-        lines to pack.
+        When a package is scanned (only delivery ones), if the package is known
+        as the destination package of one of the lines or is the source package
+        of a selected line, the package is set to be the destination package of
+        all the lines to pack.
 
         When a product is scanned, it selects (set qty_done = reserved qty) or
         deselects (set qty_done = 0) the move lines for this product. Only
@@ -730,6 +736,12 @@ class Checkout(Component):
 
         package = search.package_from_scan(barcode)
         if package:
+            if not package.packaging_id:
+                return self._response_for_select_package(
+                    picking,
+                    selected_lines,
+                    message=self.msg_store.dest_package_not_valid(package),
+                )
             return self._put_lines_in_package(picking, selected_lines, package)
 
         # Scan delivery packaging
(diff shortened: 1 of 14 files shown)


shopfloor/services/checkout.py
```diff
@@ -120,6 +122,19 @@ def _response_for_select_dest_package(self, picking, move_lines, message=None):
             message=message,
         )
 
+    def _response_for_select_delivery_packaging(self, picking, packaging, message=None):
+        return self._response(
+            next_state="select_delivery_packaging",
+            data={
+                # We don't need to send the 'picking' as the mobile frontend
+                # already this info
```
```suggestion
                # already has this info after `select_document` state
```

shopfloor/services/checkout.py
```diff
@@ -619,26 +637,28 @@ def _filter_lines_checkout_done(move_line):
         return move_line.qty_done > 0 and move_line.shopfloor_checkout_done
 
     def _is_package_allowed(self, picking, package):
-        existing_packages = picking.mapped("move_line_ids.result_package_id")
+        """A package is allowed as a destination one if it is present among
```
simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

shopfloor/services/checkout.py
```diff
@@ -678,10 +698,10 @@ def _create_and_assign_new_packaging(self, picking, selected_lines, packaging=No
...
         """Scan a package, a lot, a product or a package to handle a line
 
-        When a package is scanned, if the package is known as the destination
-        package of one of the lines or is the source package of a selected
-        line, the package is set to be the destination package of all then
-        lines to pack.
+        When a package is scanned (only delivery ones), if the package is known
+        as the destination package of one of the lines or is the source package
+        of a selected line, the package is set to be the destination package of
+        all then lines to pack.
```
```suggestion
        all the lines to pack.
```

shopfloor/services/checkout.py
```diff
@@ -758,7 +784,42 @@ def scan_package_action(self, picking_id, selected_line_ids, barcode):
...
+                    "package_carrier_type",
+                    "=",
+                    picking.carrier_id.delivery_type or "none",
+                ),
+            ]
+        )
+
+    def list_delivery_packaging(self, picking_id, selected_line_ids):
+        """List available delivery packaging based on the carrier of the
+        picking to create the new package.
```
```suggestion
        """List available delivery packaging for given picking.

```

shopfloor/services/checkout.py
```diff
@@ -823,17 +887,20 @@ def list_dest_package(self, picking_id, selected_line_ids):
...
-    def _set_dest_package_from_selection(self, picking, selected_lines, package):
+    def _set_dest_package_from_selection(
+        self, picking, selected_lines, package, barcode=None
+    ):
         if not package:
-            return self._response_for_select_dest_package(picking, selected_lines)
+            return self._response_for_select_dest_package(
+                picking,
+                selected_lines,
+                message=self.msg_store.package_not_found_for_barcode(barcode),
```
simahawk: what if barcode is none?

simahawk: probably you should not call `_set_dest_package_from_selection` if you don't have a package

shopfloor/services/checkout.py
```diff
@@ -1277,6 +1356,15 @@ def _schema_select_package(self):
             "picking": {"type": "dict", "schema": self.schemas.picking()},
         }
 
+    @property
+    def _schema_select_delivery_packaging(self):
+        return {
+            "picking": {"type": "dict", "schema": self.schemas.picking()},
```
```suggestion
            "picking": self.schemas._schema_dict_of(self.schemas.picking()),
```

shopfloor/tests/test_checkout_list_delivery_packaging.py
```diff
@@ -0,0 +1,122 @@
+# Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
```
simahawk: 2021

shopfloor/tests/test_checkout_list_package.py
```diff
@@ -115,29 +117,42 @@ def setUpClassBaseData(cls):
...
+        cls.move_line1.result_package_id = cls.delivery_package
         # We'll put only product A and B in the destination package
         cls.move_line1.qty_done = cls.move_line1.product_uom_qty
         cls.move_line2.qty_done = cls.move_line2.product_uom_qty
         cls.move_line3.qty_done = 0
 
         cls.picking = picking
-        cls.package = cls.move_line1.result_package_id
+
+    def _get_allowed_packages(self):
```
simahawk: pass the picking instead of using self.picking

simahawk (changes requested):

simahawk:
#193 merged

shopfloor/services/checkout.py
sebalix: I'll check this point.

sebalix: Yes it's a bit hackish, but this happens only when `_set_dest_package_from_selection` is called by `set_dest_package` (which has a `package_id` as parameter), so the only way to have `barcode=None` here is we send a wrong `package_id` to `set_dest_package`, and then yes the error message is not really appropriate (I talked about this in the related test).
Maybe the best way here is to return the generic "record not found" error message before calling `_set_dest_package_from_selection` yes.

shopfloor_mobile/static/wms/src/scenario/checkout.js
```diff
@@ -257,6 +271,14 @@ const Checkout = {
                 },
             };
         },
+        select_delivery_packaging_manual_select_options: function() {
+            return {
+                list_item_options: {
+                    loud_title: true,
+                    fields: [{path: "packaging_type"}, {path: "shipper_package_code"}],
```
simahawk: for `shipper_package_code` maybe better to show a label?

simahawk (commented):
my PR has been merged, rebase pls :)

shopfloor_mobile/static/wms/src/scenario/checkout.js
sebalix: done, same label than the backend

shopfloor/services/checkout.py
sebalix: Fixed, we now test that the package exists in `set_dest_package` and `scan_dest_package` before calling `_set_dest_package_from_selection`. The error message differs between the two methods if the package can't be found.

simahawk (approved):

TDu (approved):

sebalix:
@simahawk don't merge now, following my test with @jbaudoux I have to change small things regarding this (and some other fixes not related to this PR particularly)

sebalix:
Changes applied since the last push:

- list of delivery packaging are now sorted by name
- on the list of delivery packaging, no need to click on the _"Confirm"_ button once the packaging has been selected: it is chosen directly when selected
- on the list of delivery packaging, the carrier is displayed like the other screens to help the user to choose the right packaging if needed
- remove of `shipper_package_code` data on listed delivery packaging (not useful)
- change the error message _"Packaging {} does not match carrier {}."_ en _"Packaging {} is not allowed for carrier {}."_ when we scan a delivery packaging not related to the current carrier
- when we cancel a line or a package, get back on the `select_line` screen instead of the `summary` one

Ready for reviews/merge.

PR #199, opened 2021-03-15
sf_mobile_base: prevent scan multiple times

simahawk:
The scanning interface uses a simple input text
which is valued and submitted by the scanning device automatically.
While sending the request the focus stays on the input
hence if the user scans multiple times the same barcode
multiple requests are sent to the backend.

This change prevents this by checking if the app
is still waiting for data from the backend.

By doing so, several conflict errors are avoided
because the same action cannot be applied twice or more
for the same endpoint till the UI is unlocked.


TDu (approved):
LG, should the user be inform of the situation ?

simahawk:
> LG, should the user be inform of the situation ?

it's already ok: the UI shows a loader, it's normal that if you scan something nothing happens.

PR #203, opened 2021-03-18
[13.0] shopfloor_example: Fix validators

lmignon:


simahawk:
@lmignon wrong `shop*` :rofl:

shopfloor_example/services/partner_service.py
```diff
@@ -65,17 +65,17 @@ class ShopfloorCheckoutValidator(Component):
     _name = "shopfloor.partner_example.validator"
     _usage = "partner_example.validator"
 
-    def _validator_scan(self):
+    def scan(self):
```
simahawk: hmm I  think I forgot to rename them after moving to their own component :sweat_smile:

shopfloor_example/services/partner_service.py
```diff
@@ -15,7 +15,9 @@ class PartnerExampleService(Component):
     _description = __doc__
 
     @restapi.method(
-        [(["/scan/<string:identifier>"], "GET")], auth="api_key",
+        [(["/scan/<string:identifier>"], "GET")],
+        output_param=restapi.CerberusValidator("scan"),
```
simahawk, line 19: that's why they were working: by default lookup to _validate_*.
I  guess `auth` is not required anymore tnx to @acsonefho fix, right?

lmignon:
@simahawk The api is now properly declared and the documentation is working... (except a bug into base_rest when a method doesn't define input schema or only path parameters, the default parameters are not taken into account)
When you use the `@restapi` You must explicitly declare your input/ouput params.
```python
    @restapi.method(
        [(["/partner_list"], "GET")],
        input_param=restapi.CerberusValidator("partner_list"),
        output_param=restapi.CerberusValidator("partner_list"),
        auth="api_key",
    )
    def partner_list(self, **params):
        """Return list of available partners.
        """
```
image

shopfloor_example/services/partner_service.py
lmignon: by default, not validated at all....
We have to test if auth is no more required...

simahawk:
> @simahawk The api is now properly declared and the documentation is working... (except a bug into base_rest when a method doesn't define input schema or only path parameters, the default parameters are not taken into account)
> When you use the `@restapi` You must explicitly declare your input/ouput params.

I see! Thanks!
Would you mind to check if it is still working for shopinvader? I think you can see docs but you cannot test them.

lmignon:
@simahawk `base_rest` fix is on its way...
image

lmignon:
@simahawk With this fix  https://github.com/OCA/rest-framework/pull/130/commits/06553358a6e7af927da9bd2522115ed26176aa19 on rest_framework the auth is no more required.

simahawk (approved):
@lmignon ouch I lost track of this.... can we merge it as it is?

lmignon:
@simahawk yes you can.

simahawk:
merci :)

/ocabot merge patch-

PR #207, opened 2021-03-26
[13.0][IMP] shopfloor add cluster_batch_picking

hparfr:
Add new scenario: cluster_batch_picking


hparfr:
Replaced by a v14 branch on akretion/wms.

PR #209, opened 2021-03-26
[13.0][IMP] shopfloor changes in checkout

hparfr:
Add a new mode: "scan & pack" to quickly scan products.

When activated on the stock.picking.type = checkout it will change the view in the mobile app.

Capture d’écran 2021-04-15 à 15 59 42

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -15,8 +15,8 @@ class DataAction(Component):
             record.with_context(location=record.id), self._location_parser, **kw
         )
 
-    def locations(self, record, **kw):
-        return self.location(record, multi=True)
+    def locations(self, records, **kw):
+        return [self.location(rec, **kw) for rec in records]
 
     @property
     def _location_parser(self):
@@ -165,6 +165,7 @@ class DataAction(Component):
             "id",
             "qty_done",
             "product_uom_qty:quantity",
+            "shopfloor_checkout_done:done",
             ("product_id:product", self._product_parser),
             ("lot_id:lot", self._lot_parser),
             ("location_id:location_src", self._location_parser),
@@ -222,10 +223,17 @@ class DataAction(Component):
             "display_name",
             "default_code",
             "barcode",
+            (
+                "barcode_ids:barcodes",
+                lambda record, fname: self._product_barcode_list(record[fname]),
+            ),
             ("packaging_ids:packaging", self._product_packaging),
             ("uom_id:uom", self._simple_record_parser() + ["factor", "rounding"]),
             ("seller_ids:supplier_code", self._product_supplier_code),
         ]
+
+    def _product_barcode_list(self, rec):
+        return self._jsonify(rec, self._simple_record_parser(), multi=True,)
 
     def _product_packaging(self, rec, field):
         return self._jsonify(
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -251,8 +251,10 @@ class DataAction(Component):
     @ensure_model("stock.picking.batch")
     def picking_batch(self, record, with_pickings=False, **kw):
         parser = self._picking_batch_parser
+
         if with_pickings:
             parser.append(("picking_ids:pickings", self._picking_parser))
+
         return self._jsonify(record, parser, **kw)
 
     def picking_batches(self, record, with_pickings=False, **kw):
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -264,6 +264,15 @@ class MessageAction(Component):
             ),
         }
 
+    def x_units_put_in_location(self, qty, product, location):
+        # TODO refactor with x_units_put_in_package
+        return {
+            "message_type": "success",
+            "body": _("{} {} put in {}").format(
+                qty, product.display_name, location.name
+            ),
+        }
+
     def cannot_move_something_in_picking_type(self):
         return {
             "message_type": "error",
@@ -550,4 +559,10 @@ class MessageAction(Component):
                 "Pick + Pack mode ON: the picking {0.name} has no carrier set. "
                 "The system couldn't pack goods automatically."
             ).format(picking),
+        }
+
+    def stock_issue_for_line(self, name):
+        return {
+            "message_type": "info",
+            "body": _("Stock issue declared for {}").format(name),
         }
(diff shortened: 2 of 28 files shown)


hparfr:
@simahawk @sebalix   can you please review this

.gitignore
```diff
@@ -73,3 +73,7 @@ docs/_build/
 
 # OCA rules
 !static/lib/
+
```
simahawk: normally I keep these in my global .gitignore

shopfloor/actions/data.py
```diff
@@ -9,6 +9,13 @@
 class DataAction(Component):
     _inherit = "shopfloor.data.action"
 
+    @property
+    def _full_partner_parser(self):
```
simahawk: this is the simpler record parser. Also, "full" should go to the `data_detail` action.

shopfloor/actions/data.py
```diff
@@ -48,6 +55,20 @@ def _picking_parser(self):
             "scheduled_date",
         ]
 
+    @property
+    def _full_picking_parser(self):
```
simahawk: `data_detail` again?

shopfloor/actions/data.py
```diff
@@ -165,10 +186,17 @@ def _move_line_parser(self):
             "id",
             "qty_done",
             "product_uom_qty:quantity",
+            "shopfloor_checkout_done:done",
             ("product_id:product", self._product_parser),
             ("lot_id:lot", self._lot_parser),
             ("location_id:location_src", self._location_parser),
             ("location_dest_id:location_dest", self._location_parser),
+            (
```
simahawk: this will be more expensive than expected. Can't we make it conditional?

shopfloor/actions/data.py
```diff
@@ -243,9 +271,27 @@ def _product_supplier_code(self, rec, field):
...
-        return self._jsonify(record, parser, **kw)
+        if with_pickings == "full":
+            parser.append(("picking_ids:pickings", self._full_picking_parser))
+
+        data = self._jsonify(record, parser, **kw)
+
+        if with_pickings == "full":
+            pickings = record.picking_ids
+
+            for i, picking in enumerate(pickings):
```
simahawk: IMO this should be done in the service where you need it.
Anyway, these move lines computation shall be done in the full_picking_parser.

shopfloor/actions/message.py
```diff
@@ -249,6 +249,15 @@ def x_units_put_in_package(self, qty, product, package):
             ),
         }
 
+    def x_units_put_in_location(self, qty, product, location):
+        # TODO refactor with x_units_put_in_package
```
simahawk, line 268: maybe just `x_units_put_in(self, qty, product, destination)` and just pick the name of destination.

shopfloor/actions/message.py
```diff
@@ -504,6 +513,12 @@ def packaging_invalid_for_carrier(self, packaging, carrier):
             ),
         }
 
+    def stock_issue_for_line(self, name):
```
simahawk: this is for?

shopfloor/actions/schema.py
```diff
@@ -38,6 +60,12 @@ def move_line(self, with_packaging=False, with_picking=False):
             "package_dest": self._schema_dict_of(
                 self.package(with_packaging=with_packaging), required=False
             ),
+            "suggested_package_dest": self._schema_list_of(
```
simahawk: why do you need these?

shopfloor/services/exception.py
```diff
@@ -0,0 +1,117 @@
+from functools import wraps
```
simahawk, line 1: and this is for?

shopfloor_mobile/static/wms/src/scenario/checkout.js
```diff
@@ -320,15 +338,102 @@ const Checkout = {
         return {
             usage: "checkout",
             initial_state_key: "select_document",
+            lastScanned: null,
             states: {
+                scan_products: {
+                    on_scan: scanned => {
```
simahawk: way  too much logic here in the definition of the state IMO

shopfloor_mobile_base/static/wms/src/components/detail/detail_card.js
```diff
@@ -58,4 +65,7 @@ Vue.component("item-detail-card", {
         </p>
     </div>
   `,
+    renderTriggered() {
```
simahawk: ?

shopfloor_mobile_base/static/wms/src/components/detail/detail_mixin.js
```diff
@@ -85,14 +85,18 @@ export var ItemDetailMixin = {
             // Defining defaults for an Object property
             // works only if you don't pass the property at all.
             // If you pass only one key, you'll lose all defaults.
-            const opts = _.defaults({}, this.$props.options, {
```
simahawk, line 88: why this change?

simahawk (changes requested):
@hparfr please clean up this PR. Is a bit messy.

1. provide and extended description of the changes and document methods and classes w/ docstrings
2. split mobile changes to their own atomic commit or split them to another PR
3. also, the checkout behavior you need seems more a new checkout scenario: why don't you create a new one? Is it because of the mobile template? If so, we could find a better way to handle it.

shopfloor/actions/schema.py
FranzPoize: We need these when we process order with multiple products. We need to suggest the destination for the following product that go in the same package.

shopfloor/services/exception.py
FranzPoize: This was an attempt to find a better way to handle common error in the services. wraps is used to wraps the decorated function and allows us to keep the signature and name of the decorated function.

shopfloor_mobile_base/static/wms/src/components/detail/detail_mixin.js
FranzPoize: There's no good reason to use defaults here. And standard javascript has nearly all utility function that underscore/lodash provides. In the end the plan was to remove underscore from the codebase if that's possible and stick to vanilla js for object literal and array manipulation.

hparfr:
Replaced by a v14 branch on akretion/wms.

PR #212, opened 2021-04-02
[13.0] fix mobile demo mode

lmignon:


simahawk (approved):
my bad, this find/replace was not finished :sweat_smile: 
tnx!

shopfloor_mobile_base/templates/assets.xml
```diff
@@ -217,10 +217,5 @@
             t-attf-src="/shopfloor_mobile_base/static/wms/src/demo/demo.core.js?v=#{app_version}"
             type="module"
         />
-        <script
```
simahawk, line 220: hmm I'm not sure about this... I think this should be left and add demo data for non wms stuff. Can you add a todo?

lmignon: This one is to avoid an error into the console but is not required to have the demo working. I'll comment and add a TODO

simahawk: merci

PR #218, opened 2021-04-19
[14.0][MIG] shopfloor_mobile_base 

hparfr:


hparfr:
There is issues with eslint.

In eslint.yml , I did the following changes
`
parserOptions:
  ecmaVersion: 2017
  sourceType: "module"
`
To get rid of eslint error  ` 7:1  error  Parsing error: 'import' and 'export' may appear only with 'sourceType: module'`


There is still an issue with spread operator.

And with no-unused-vars https://eslint.org/docs/rules/no-unused-vars

I'm not very fan of removing all unused variables from declarations because when we have some inheritance, I think it helps.

hparfr:
I switched to ecma 2018 in order to have the spread operator and allowed unused variables in arguments.

I will port #202  and #198  after the merge of this PR because it's quite hard do re-do the migration work / conflict resolution.

@simahawk can you please review this PR ?

simahawk (approved):
thanks @hparfr !

Question: shall we port eslint changes to v13?

hparfr:
For sure having the same rules across v13 and v14 will ease the (back/forward) ports.  :man_dancing:

PR #219, opened 2021-04-19
[14.0][MIG] shopfloor_mobile

hparfr:
depends of #218 

After split of shopfloor_mobile and shopfloor_mobile_base


hparfr:
Replaced by 251

PR #214, opened 2021-04-02
[ADD] stock_reception_screen_measuring_device

gurneyalex:
this is a glue module between stock_reception_screen and stock_measuring_device

Depends on:
* https://github.com/OCA/stock-logistics-warehouse/pull/1121

PR #224, opened 2021-04-23
s_reception_screen_measuring_device: fix device release

mmequignon:
related to https://github.com/OCA/wms/pull/214

### DIFF
--- a/stock_reception_screen_measuring_device/models/stock_reception_screen.py
+++ b/stock_reception_screen_measuring_device/models/stock_reception_screen.py
@@ -65,12 +65,12 @@ class StockReceptionScreen(models.Model):
             self._notify(error_msg)
             return UserError(error_msg)
 
-        self.product_packaging_id._assign_measuring_device(device)
+        self.product_packaging_id._measuring_device_assign(device)
         return True
 
     def cancel_measure_current_packaging(self):
         self.ensure_one()
-        self.product_packaging_id._release_measuring_device()
+        self.product_packaging_id._measuring_device_release()
         return True
 
     def _notify(self, message):
--- a/stock_reception_screen_measuring_device/models/stock_reception_screen.py
+++ b/stock_reception_screen_measuring_device/models/stock_reception_screen.py
@@ -18,6 +18,11 @@ class StockReceptionScreen(models.Model):
         store=True,
         help="Indicates if the package have any measurement missing.",
     )
+    display_package_dimensions = fields.Char(
+        string="Dimensions (lxhxw)",
+        compute="_compute_package_dimensions",
+        help="Dimensions of the package in mm (length x height x width)",
+    )
     scan_requested = fields.Boolean(
         help="A scan from the measuring device was requested",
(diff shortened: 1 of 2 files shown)


simahawk (approved):

stock_reception_screen_measuring_device/models/stock_reception_screen.py
```diff
@@ -18,6 +18,11 @@ class StockReceptionScreen(models.Model):
         store=True,
         help="Indicates if the package have any measurement missing.",
     )
+    display_package_dimensions = fields.Char(
```
sebalix, line 21:
```suggestion
    package_dimensions = fields.Char(
```
No need to add a verb here.

stock_reception_screen_measuring_device/models/stock_reception_screen.py
```diff
@@ -33,6 +38,22 @@ def _compute_scan_requested(self):
...
+        "product_packaging_id.width",
+        "product_packaging_id.height",
+    )
+    def _compute_package_dimensions(self):
+        for record in self:
+            pack = record.product_packaging_id
+            dimension_values = [pack.lngth, pack.height, pack.width]
+            if all(dimension_values):
+                record.display_package_dimensions = " x ".join(
+                    [str(val) for val in dimension_values]
```
sebalix: Add mm unit would be nice
```suggestion
                    [f"{str(val)}mm" for val in dimension_values]
```

stock_reception_screen_measuring_device/models/stock_reception_screen.py
```diff
@@ -18,6 +18,11 @@ class StockReceptionScreen(models.Model):
         store=True,
         help="Indicates if the package have any measurement missing.",
     )
+    display_package_dimensions = fields.Char(
+        string="Dimensions (lxhxw)",
+        compute="_compute_package_dimensions",
+        help="Dimensions of the package (length x height x width)",
```
sebalix:
```suggestion
        help="Dimensions of the package in mm (length x height x width)",
```

stock_reception_screen_measuring_device/models/stock_reception_screen.py
mmequignon: "display" in "display_name" is not a verb.
Here it's just to say that's dimensions meant to be displayed.
But I can remove it if you still want it to be removed.

simahawk:
@mmequignon pls address @sebalix comments but the "display" one ;)

stock_reception_screen_measuring_device/models/stock_reception_screen.py
sebalix: If it "means to be displayed" then it is a boolean (to display or not something), but if the field contains qualitative information regarding the current record (we are stocking a string here), then `package_dimensions` is enough, no need to add more.
It's my POV, non blocking.

sebalix (approved):

PR #225, opened 2021-04-27
[13.0][FIX] Shopfloor, loc. content transfer: put lines in separate transfer

sebalix:
As soon as we scan a location with the location content transfer, put all move lines to process in its own move+transfer to ensure they won't be updated by new reservations (because of another user moving goods to this location in the middle of a location content transfer process).

Test the following scenario:

1) Operator-1 processes the first pallet with the "zone picking" scenario
   to move the goods to PACK-1:

    move1 PICK -> PACK-1 'done'

2) Operator-2 with the "location content transfer" scenario scan
  the location where the first pallet is (PACK-1):
    - the app should found one move line
    - this move line will be put in its own transfer in any case
    - as such the app should ask the destination location (as there is
      only one line)

    move1 PACK-1 -> SHIP (still handled by the operator so not 'done')

3) Operator-1 processes the second pallet with the "zone picking" scenario
   to move the goods to PACK-2:

    move1 PICK -> PACK-2 'done'

    - this will automatically update the reservation (new move line) in
      the transfer previously processed by Operator-2.

4) Operator-2 then finishes its operation regarding the first pallet
  without any trouble.

5) Operator-2 with the "location content transfer" scenario scan
  the location where the second pallet is (PACK-2), etc

Ref. 2456

### DIFF
--- a/shopfloor/models/stock_picking.py
+++ b/shopfloor/models/stock_picking.py
@@ -1,6 +1,6 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 
 
 class StockPicking(models.Model):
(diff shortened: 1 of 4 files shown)


shopfloor/models/stock_picking.py
```diff
@@ -51,3 +51,53 @@ def _check_move_lines_map_quant_package(self, package):
         ):
             return False
         return super()._check_move_lines_map_quant_package(package)
+
+    def split_assigned_move_lines(self, move_lines=None):
```
sebalix, line 55: This method could be moved to `stock_helper` probably

mmequignon (approved):
LG : technical review

jbaudoux (approved):

PR #226, opened 2021-04-28
[13.0] Shopfloor: Improve check valid destination + align checks in all scenario

jbaudoux:
We ensure the destination is either valid regarding the picking default destination or the move destination. With the push rules in the module stock_dynamic_routing in OCA/wms, it is possible that the move destination is not anymore a child of the picking default destination (as it is the last pushed move that now respects this condition and not anymore this one that has a destination to an intermediate location)

Extract the location checks in a component to ensure the same checks are applied in all scenario

cc @sebalix @simahawk

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -1,4 +1,5 @@
-# Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
+# Copyright 2020-2021 Camptocamp SA (http://www.camptocamp.com)
+# Copyright 2020-2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import _, fields
 from odoo.osv import expression
@@ -933,24 +934,20 @@ class ClusterPicking(Component):
             return self._unload_end(batch)
 
         first_line = fields.first(lines)
-        picking_type = fields.first(batch.picking_ids).picking_type_id
         scanned_location = self._actions_for("search").location_from_scan(barcode)
         if not scanned_location:
             return self._response_for_unload_all(
                 batch, message=self.msg_store.no_location_found()
             )
-        if not scanned_location.is_sublocation_of(
-            picking_type.default_location_dest_id
-        ) or not scanned_location.is_sublocation_of(
-            lines.mapped("move_id.location_dest_id"), func=all
-        ):
+        if not self.is_dest_location_valid(lines.move_id, scanned_location):
             return self._response_for_unload_all(
                 batch, message=self.msg_store.dest_location_not_allowed()
             )
 
-        if not scanned_location.is_sublocation_of(first_line.location_dest_id):
-            if not confirmation:
-                return self._response_for_confirm_unload_all(batch)
+        if not confirmation and self.is_dest_location_to_confirm(
+            first_line.location_dest_id, scanned_location
+        ):
+            return self._response_for_confirm_unload_all(batch)
 
         self._unload_write_destination_on_lines(lines, scanned_location)
         completion_info = self._actions_for("completion.info")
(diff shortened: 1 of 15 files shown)


shopfloor/services/single_pack_transfer.py
```diff
@@ -191,19 +192,25 @@ def _create_package_level(self, package):
...
-    def _is_move_state_valid(self, move):
-        return move.state != "cancel"
-
-    def _is_dest_location_valid(self, move, scanned_location):
-        """Forbid a dest location to be used"""
+    def _is_move_state_valid(self, moves):
+        return all(move.state != "cancel" for move in moves)
+
+    def _is_dest_location_valid(self, moves, scanned_location):
+        """We ensure the destination is either valid regarding the picking
```
sebalix: Better to have a summary sentence and the details separated by a blank line to be inline with PEP8:
```suggestion
        """Ensure the destination location is valid for given moves.

        We ensure the destination is either valid regarding the picking [...]
        """
```

sebalix (approved):
LG, minor comment about docstring

shopfloor/services/single_pack_transfer.py
jbaudoux: fixed

shopfloor/actions/stock.py
```diff
@@ -10,6 +10,36 @@ class StockAction(Component):
     _inherit = "shopfloor.process.action"
     _usage = "stock"
 
+    def is_src_location_valid(self, process, location):
```
simahawk: I would pass `picking_types` instead of process.
Alternatively you could move these to the process mixin but I think here is better because we can use them everywhere.

shopfloor/actions/stock.py
```diff
@@ -10,6 +10,36 @@ class StockAction(Component):
...
+        destination location or the move destination location. With the push
+        rules in the module stock_dynamic_routing in OCA/wms, it is possible
+        that the move destination is not anymore a child of the picking default
+        destination (as it is the last pushed move that now respects this
+        condition and not anymore this one that has a destination to an
+        intermediate location)
+        """
+        return location.is_sublocation_of(
+            moves.picking_id.location_dest_id, all
+        ) or location.is_sublocation_of(moves.location_dest_id, all)
```
simahawk:
```suggestion
            moves.picking_id.location_dest_id, func=all
        ) or location.is_sublocation_of(moves.location_dest_id, func=all)
```

shopfloor/services/cluster_picking.py
```diff
@@ -25,27 +26,31 @@ class ClusterPicking(Component):
...
     First phase, picking:
 
-    * Pick a good (move line) from a source location, scan it to confirm it's
-      the expected one
-    * Scan the label of a Bin (package) in a roller-cage, put the good inside
-      (physically). Once the first move line of a picking has been scanned, the
-      screen will show the same destination package for all the other lines of
-      the picking to help the user grouping goods together, and will prevent
-      lines from other pickings to be put in the same destination package.
+    * Pick a goods (move line) from a source location, scan to confirm it's
```
simahawk: thanks for updating this!

shopfloor/services/single_pack_transfer.py
```diff
@@ -215,11 +204,11 @@ def validate(self, package_level_id, location_barcode, confirmation=False):
...
             )
 
-        # if we have more than one move, we should assume they go to the same
-        # place
-        move_line = package_level.move_line_ids[0]
-        move = move_line.move_id
-        if not self._is_move_state_valid(move):
+        # Do not use package_level.move_ids, this is only filled in when the
+        # moves have been created from a manually encoded package level, not
+        # when a package has been reserved for existing moves
```
simahawk, line 208: :exploding_head:

simahawk (commented):
LG I'd only change `is_src_location` signature to get picking types explicitly.

shopfloor/actions/stock.py
jbaudoux: @simahawk I did it on purpose to allow to add any other condition in the process to validate source location

shopfloor/actions/stock.py
jbaudoux: done

jbaudoux:
rebased

shopfloor/actions/stock.py
simahawk: to do so, you should add this method to the service instead, and use this one by default, eg:
```py
[...service class...]
def is_src_location_valid(self, location):
    return self._action_for("stock").is_src_location_valid(self.picking_types, location)
```
then if you need customization for the specific service, you override it.

jbaudoux: @simahawk I moved the checks to the service. So by default they share all the same checks but it could be service specific. I find it better like that. Can you update your review? Thanks

shopfloor/services/zone_picking.py
```diff
@@ -602,19 +601,15 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
         # if `confirmation is True
         # Ask confirmation to the user if the scanned location is not in the
         # expected ones but is valid (in picking type's default destination)
-        if not location.is_sublocation_of(
-            self.picking_type.default_location_dest_id
-        ) or not location.is_sublocation_of(
-            move_line.move_id.location_dest_id, func=all
-        ):
+        stock = self._actions_for("stock")
```
simahawk: seems you don't need to call `stock` anymore

simahawk (approved):
@jbaudoux please cleanup left overs and squash. Pre-approving :wink:

shopfloor/services/zone_picking.py
jbaudoux: removed

jbaudoux:
squashed

PR #227, opened 2021-05-07
[13.0] shopfloor: fix wrong message for zone_picking.set_destination

mmequignon:
When trying to set a destination, it can happen that message is empty when next step is zero check.
This PR normalizes the message in such case.

ref: 2526

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -693,6 +693,7 @@ class ZonePicking(Component):
         # the field ``shopfloor_user_id`` is updated with the current user
         move_line.shopfloor_user_id = user or self.env.user
 
+    # flake8: noqa: C901
     def set_destination(
         self, move_line_id, barcode, quantity, confirmation=False,
     ):
@@ -771,7 +772,10 @@ class ZonePicking(Component):
                 )
                 if response:
                     if extra_message:
-                        response["message"]["body"] += "\n" + extra_message["body"]
+                        if response.get("message"):
+                            response["message"]["body"] += "\n" + extra_message["body"]
+                        else:
+                            response["message"] = extra_message
                     return response
 
         # When the barcode is a package


shopfloor/services/zone_picking.py
```diff
@@ -769,9 +769,15 @@ def set_destination(
...
-                    if extra_message:
+                if response and extra_message:
+                    if response.get("message"):
                         response["message"]["body"] += "\n" + extra_message["body"]
+                    else:
+                        # _handle_pick_pack_same_time_for_location successful
+                        response["message"] = {
+                            "body": extra_message["body"],
+                            "message_type": "success",
+                        }
```
sebalix: if there no `response['message']` just do `response['message'] = extra_message`, you should not enforce the message type here.

shopfloor/services/zone_picking.py
```diff
@@ -693,6 +693,7 @@ def _set_move_line_as_done(self, move_line, quantity, package, user=None):
         # the field ``shopfloor_user_id`` is updated with the current user
         move_line.shopfloor_user_id = user or self.env.user
 
+    # flake8: noqa: C901
```
simahawk, line 696: ?

simahawk (commented):
pls rewrite the commit msg. You can use what I've set in the PR title w/out the version. The version is irrelevant and is actually ugly when you port the commit to other version ;)

TDu (approved):
Improvement looks good,

shopfloor/services/zone_picking.py
mmequignon: too much "if" in this method.
And it looks a bit complicated to split.

simahawk: I see, can you add a TODO pls?

sebalix (approved):

PR #228, opened 2021-05-11
[13.0][ADD] s_r_screen_mrp_subcontracting: compatibility layer

sebalix:
New module to make compatible the reception screen with the `mrp_subcontracting` module.

With the reception screen we allowed to receive goods move by move (each time validated), and this new module makes this flow compatible with subcontracted goods.

Ref. 2472

### DIFF
--- a/stock_reception_screen/models/stock_reception_screen.py
+++ b/stock_reception_screen/models/stock_reception_screen.py
@@ -494,32 +494,46 @@ class StockReceptionScreen(models.Model):
             # Go to the next step automatically if only one move has been found
             self.process_select_move()
 
+    def _split_move(self, move, qty):
+        # Hook intended to be overridden
+        return move._split(qty)
+
+    def _action_done_picking(self):
+        """Called by '_validate_current_move' when processing the last move
+        to perform the validation at the picking level.
+        """
+        return self.picking_id.action_done()
+
+    def _action_done_move(self, move):
+        """Called by '_validate_current_move' when processing a partial qty to
+        perform the validation at the move level.
+        """
+        # We use the 'is_scrap' context key to avoid the generation of a
+        # backorder when validating the move (see _action_done() method in
+        # stock/models/stock_move.py).
+        return move.with_context(is_scrap=True)._action_done()
+
     def _validate_current_move(self):
         """Split the current move with the move line qty done and validate it."""
         if self.current_move_line_id and self.current_move_id:
-            remaining_qty = (
-                self.current_move_id.product_uom_qty
-                - self.current_move_line_id.qty_done
-            )
+            qty_done = self.current_move_line_id.qty_done
+            qty_remaining = self.current_move_id.product_uom_qty - qty_done
             # We don't want to create a new move with a negative qty if we
             # receive more than expected.
             digits = self.env["decimal.precision"].precision_get(
                 "Product Unit of Measure"
             )
-            if gt(remaining_qty, 0, digits=digits):
-                self.current_move_id._split(remaining_qty)
+            if gt(qty_remaining, 0, digits=digits):
+                self._split_move(self.current_move_id, qty_remaining)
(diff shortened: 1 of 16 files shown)


stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -494,32 +494,45 @@ def process_select_product(self):
             # Go to the next step automatically if only one move has been found
             self.process_select_move()
 
+    def _split_move(self, move, qty):
+        # Hook intended to be overridden
+        return move._split(qty)
+
+    def _action_done_picking(self):
+        # Called by '_validate_current_move' when processing the last move
```
simahawk: these could be moved to docstrings IMO

stock_reception_screen_mrp_subcontracting/models/stock_reception_screen.py
```diff
@@ -0,0 +1,54 @@
...
+from odoo import models
+
+
+class StockReceptionScreen(models.Model):
+    _inherit = "stock.reception.screen"
+
+    def _split_move(self, move, qty):
+        # Overridden to remove the link between the move of finished product
+        # and the move to receive to ease the split
+        move_is_subcontract = move.is_subcontract
```
simahawk, line 13: shouldn't the rest run only if is subcontract as done for the other methods?

stock_reception_screen_mrp_subcontracting/tests/test_reception_screen.py
```diff
@@ -0,0 +1,86 @@
...
+
+from odoo import fields
+
+from odoo.addons.stock_reception_screen.tests.common import Common
+
+
+class TestReceptionScreen(Common):
+    @classmethod
+    def setUpClass(cls):
+        super(TestReceptionScreen, cls).setUpClass()
```
simahawk:
```suggestion
        super().setUpClass()
```

stock_reception_screen_mrp_subcontracting/wizards/mrp_product_produce.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class MrpProductProduce(models.TransientModel):
+    _inherit = "mrp.product.produce"
+
+    def _record_production(self):
+        # Overridden to unlink existing move lines when the components
+        # consumption is recorded. As a result we will get only new move lines
+        # related to the production, avoiding issues when processing them with
+        # the reception screen.
```
simahawk, line 14: shall we run this only when ctx key from reception screen is there?

stock_reception_screen_mrp_subcontracting/tests/test_reception_screen.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields
+
+from odoo.addons.stock_reception_screen.tests.common import Common
+
+
+class TestReceptionScreen(Common):
```
TDu: Could be rename as well as the file to better reflect the name of the module

stock_reception_screen_mrp_subcontracting/models/stock_reception_screen.py
sebalix: We could do that yes to make things clearer (it doesn't change anything because we're only updating the cached value with the original one here).

stock_reception_screen_mrp_subcontracting/wizards/mrp_product_produce.py
sebalix: The context key is not enough (and doesn't make sense too) but instead run this only when the destination move is related to a reception transfer, because we usually call this wizard before using the reception screen (and this add move lines on the reception transfer difficult to handle with the reception screen...).

TDu (approved):

simahawk (approved):

PR #230, opened 2021-05-21
[130][IMP] shopfloor mobile carrier display

TDu:
In the backend add the carrier computed from the module
`stock_picking_delivery_link` into the picking data send.

In the frontend: checkout, set the color of the carrier card in green

* Add the carrier on the cluster picking last screen.
* On the checkout and the cluster picking improve the carrier display by using the one added in the backend if needed.
* Set the carrier card in green.

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -43,6 +43,7 @@ class DataAction(Component):
             "note",
             ("partner_id:partner", self._partner_parser),
             ("carrier_id:carrier", self._simple_record_parser()),
+            ("ship_carrier_id:ship_carrier", self._simple_record_parser()),
             "move_line_count",
             "package_level_count",
             "bulk_line_count",
--- a/shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
+++ b/shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
@@ -79,6 +79,13 @@ const ClusterPicking = {
             </div>
 
             <div class="unload-all" v-if="state_is('unload_all')">
+                <item-detail-card
+                    v-if="current_carrier()"
+                    :card_color="utils.colors.color_for('success')"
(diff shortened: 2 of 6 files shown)


shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
```diff
@@ -79,6 +79,13 @@ const ClusterPicking = {
             </div>
 
             <div class="unload-all" v-if="state_is('unload_all')">
+                <item-detail-card
+                    v-if="current_carrier()"
+                    :card_color="utils.colors.color_for('detail_carrier_card')"
```
simahawk: do not another color for this pls. Use "success" or "done"

shopfloor/actions/data.py
```diff
@@ -43,6 +43,7 @@ def _picking_parser(self):
             "note",
             ("partner_id:partner", self._partner_parser),
             ("carrier_id:carrier", self._simple_record_parser()),
+            ("ship_carrier_id:ship_carrier", self._simple_record_parser()),
```
simahawk, line 46: IDT we need another carrier here. The carrier is always for shipping. I would change the `carrier` parser to something like:
```py
[...]
("carrier_id:carrier", self._carrier_parser()),
[...]
def _carrier_parser(self, rec, fname):
    return (rec.carrier_id or rec.ship_carrier_id).jsonify(self._simple_record_parser(), one=True)
```
Maybe @jgrandguillaume could shed some light on this need.

simahawk (changes requested):

shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
TDu: Done

simahawk (approved):

simahawk:
/ocabot merge nobump

@TDu would you mind fwd porting to v14?

PR #231, opened 2021-05-27
Add easy pack change in zone picking (shopfloor)

TDu:
On the zone picking scenario it is possible to select a move line
by scanning the pack assigned to the move line.

But for some situation (block storage) where a location contains
multiple pack of the same type it may be difficult to scan the exact
pack, and much easier to scan an identical one in front of the user.

This change allows for this so when scanning a pack that can replace a
pack in the existing move line, the user is offered the option to do so,
by scanning the same pack a 2nd time.


shopfloor/actions/message.py
```diff
@@ -72,6 +78,15 @@ def package_not_found(self):
             "body": _("This package does not exist anymore."),
         }
 
+    def package_different_change(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "You scanned a different package with the same product, "
+                "do you want to change pack ? Scan it again to confirm"
```
```suggestion
                "do you want to change pack? Scan it again to confirm"
```
space is only for French I guess :smile:

shopfloor/services/zone_picking.py
```diff
@@ -549,7 +584,10 @@ def scan_source(self, barcode):
             self._scan_source_lot,
         )
         for handler in handlers:
-            response, message = handler(barcode)
+            if handler == self._scan_source_package:
```
simahawk: I would allow handlers to receive kw args, then you can drop this IF block

shopfloor/services/zone_picking.py
```diff
@@ -478,9 +488,33 @@ def _scan_source_package(self, barcode):
...
         if move_lines:
             response = self._response_for_set_line_destination(first(move_lines))
+            return response, message
+        if package.location_id:
+            # Check if the package selected can be a substitute on a move line
+            move_lines = self._find_location_move_lines(
+                locations=package.location_id,
+                product=package.product_packaging_id.product_id,
+            )
+        if move_lines:
```
simahawk, line 499: shouldn't this block be indented under `if package.location_id`?

shopfloor/services/zone_picking.py
```diff
@@ -478,9 +488,33 @@ def _scan_source_package(self, barcode):
...
+            if not confirmation:
+                message = self.msg_store.package_different_change()
+                response = self._response_for_select_line(
+                    move_lines, message, confirmation_required=True
+                )
+            else:
+                change_package_lot = self._actions_for("change.package.lot")
```
simahawk, line 506: I'd move this to its own method

simahawk (changes requested):
LG overall, some minor questions/requests

shopfloor/services/zone_picking.py
TDu: No, it does not have too, but the return case in it can be removed.
Which I did now

shopfloor/services/zone_picking.py
TDu: Actually the next line could be removed and the one after moved to the function call.
Not sure there is much to move, then...

shopfloor/services/zone_picking.py
TDu: My first thought and then changed to the additional if clause.
Changed

shopfloor/actions/message.py
TDu: Space after the question mark in French, yes.
Fixed

mmequignon (approved):
Code review only

shopfloor/services/zone_picking.py
```diff
@@ -478,12 +488,33 @@ def _scan_source_package(self, barcode):
         move_lines = self._find_location_move_lines(package=package)
         if move_lines:
             response = self._response_for_set_line_destination(first(move_lines))
+            return response, message
+        if package.location_id:
```
sebalix: You should check that the location of the scanned package is a sublocation of `self.zone_location` as now you are calling `_find_location_move_lines` with this location (by default the location used was `self.zone_location`).

TDu: Thanks for your review, I added the check.

simahawk:
@TDu pls squash

TDu:
@simahawk Squash done

simahawk (approved):

PR #232, opened 2021-06-02
[13.0] stock_dynamic_routing: fix compatibility with stock_move_source_relocate module

jbaudoux:
After applying the dynamic routing, call action_assign on all moves (also those non re-routed) to ensure that other hooks on that method get the complete list of moves.
Once all moves have been re-routed, propagate the changes to the destination moves.

Before this change, if you have product with 10 in stock and create a pick+ship for 15, with a dynamic routing on the pick to pull from location A to location Output/Area1, it will split the pick stock move in 10 and 5, update the destination location on the move of 10, propagate the change of source location on the ship move that gets split also. Then if you have a relocate rule for non-assigned pick from location A to Output/Area1, the pick move of 5 is re-routed, then the source location of the ship move of 5 is updated. When printing the delivery slip, you have 2 lines for the same product as the ship move was split.

After this change, it will split the pick stock move in 10 and 5, update the locations on the move of 10, then the move 5 gets relocated. Finally it propagates the change on the ship move that sees that both orig moves
are going to the same destination location and then does not split the ship move. When printing the delivery slip, you have 1 line for the product as the ship move was not split.

cc @jgrandguillaume @guewen

### DIFF
--- a/stock_dynamic_routing/models/stock_move.py
+++ b/stock_dynamic_routing/models/stock_move.py
@@ -1,7 +1,8 @@
 # Copyright 2019-2020 Camptocamp (https://www.camptocamp.com)
+# Copyright 2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 import uuid
-from collections import defaultdict, namedtuple
+from collections import OrderedDict, defaultdict, namedtuple
 
 from psycopg2 import sql
 
@@ -236,13 +237,27 @@ class StockMove(models.Model):
         """
         pickings_to_check_for_emptiness = self.env["stock.picking"]
         move_ids_to_assign_per_location = defaultdict(list)
-        for move in self:
-            routing_rule = routing_details[move].rule
-            if not routing_rule.method == "pull":
+        move_ids_to_assign_nonrelocated = []
+        next_moves_to_update = self.browse()
+        routing_to_apply = [
+            (move, detail.rule) for move, detail in routing_details.items()
+        ]
+        for move, routing_rule in routing_to_apply:
+            if routing_rule.method == "push":
+                # In this case, we should not assign the move inside the
+                # pull dynamic routing. The push move must first be added before the
+                # initial move is assigned. Otherwise, when the destination
+                # location of the initial move is changed by the push rule, the
+                # putaway won't be recomputed as it is already assigned.
+                continue
+
+            if not routing_rule:
+                move_ids_to_assign_nonrelocated.append(move.id)
                 continue
 
             if move.picking_id.picking_type_id == routing_rule.picking_type_id:
                 # already correct
+                move_ids_to_assign_nonrelocated.append(move.id)
                 continue
 
             # we expect all the lines to go to the same destination for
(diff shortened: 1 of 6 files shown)


jbaudoux:
TODO add test for described case

jgrandguillaume:
Hi @jbaudoux ,

That's a nice one, well done. Look, for what I'm able to judge (and after some headache in the understanding) I think this change is ok.

I would really appreciate you write a test case as you mentioned before we accept to merge it.

jbaudoux:
> TODO add test for described case

done

jbaudoux:
TODO refactor & add test for insert move case

jbaudoux:
> TODO refactor & add test for insert move case

refactored code for insert move to prevent recurring calls and properly propagate re-routing. Test added for that case

simahawk (commented):
LG and seems to make sense but as I don't have a deep knowledge on routing I wait for others to approve.

sebalix (approved):
Seems OK, I suppose :)

mmequignon (approved):
Code looks good

jbaudoux:
I fixed the push routing to ensure putaway is applied on the initial move

sebalix:
@jbaudoux one test is failing on `odoo/odoo`
```python
FAIL: LocationContentTransferSetDestinationAllCase.test_set_destination_all_with_partially_available_move_with_ancestor
Traceback (most recent call last):
  File "/home/travis/build/OCA/wms/setup/shopfloor/odoo/addons/shopfloor/tests/test_location_content_transfer_set_destination_all.py", line 161, in test_set_destination_all_with_partially_available_move_with_ancestor
    params={"location_id": self.content_loc.id, "barcode": sub_shelf1.barcode},
  File "/home/travis/build/OCA/wms/setup/shopfloor_base/odoo/addons/shopfloor_base/services/service.py", line 27, in dispatch
    return super().dispatch(method_name, *args, params=params)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/odoo/addons/rest_log/components/service.py", line 36, in dispatch
    return super().dispatch(method_name, *args, params=params)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/odoo/addons/base_rest/components/service.py", line 158, in dispatch
    res = method(*args, **secure_params)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/odoo/addons/base_rest/restapi.py", line 61, in response_wrap
    response = f(*args, **kw)
  File "/home/travis/build/OCA/wms/setup/shopfloor/odoo/addons/shopfloor/services/location_content_transfer.py", line 477, in set_destination_all
    self._set_all_destination_lines_and_done(pickings, move_lines, scanned_location)
  File "/home/travis/build/OCA/wms/setup/shopfloor/odoo/addons/shopfloor/services/location_content_transfer.py", line 434, in _set_all_destination_lines_and_done
    stock.validate_moves(move_lines.move_id)
  File "/home/travis/build/OCA/wms/setup/shopfloor/odoo/addons/shopfloor/actions/stock.py", line 30, in validate_moves
    moves_todo.extract_and_action_done()
  File "/home/travis/build/OCA/wms/setup/shopfloor/odoo/addons/shopfloor/models/stock_move.py", line 104, in extract_and_action_done
    assert new_picking.state == "assigned"
AssertionError
```
I restart the Travis job to check as all was green on `OCA/OCB`...

stock_dynamic_routing/models/stock_move.py
```diff
@@ -236,13 +237,27 @@ def _apply_routing_rule_pull(self, routing_details):
...
-        for move in self:
-            routing_rule = routing_details[move].rule
-            if not routing_rule.method == "pull":
+        move_ids_to_assign_nonrelocated = []
+        next_moves_to_update = self.browse()
+        routing_to_apply = [
+            (move, detail.rule) for move, detail in routing_details.items()
+        ]
+        for move, routing_rule in routing_to_apply:
+            if routing_rule.method == "push":
```
sebalix, line 246: The `method` selection field has no default value nor is required, so we could have no value here
```suggestion
            if routing_rule.method != "pull":
```

sebalix: Or having no value means it should be applied on both push & pull?

sebalix:
Test fails on `odoo/odoo` because of this new commit https://github.com/odoo/odoo/commit/591a726f4901b863f7a13581942e7e6b352c1728 still not synced on `OCA/OCB`, checking why.

sebalix:
A fix has been pushed here: https://github.com/OCA/wms/pull/275

sebalix:
@jbaudoux I restarted the build, should be OK now, Odoo made a fix regarding the mentioned issue here: https://github.com/odoo/odoo/pull/75019

sebalix:
@jbaudoux tests on `stock_move_source_relocate_dynamic_routing` are broken

jbaudoux:
@sebalix Thanks for the info. I reverted your last suggestion (== "push" to != "pull") as it was the only change I did since the tests were green

jgrandguillaume (approved):
LGTM

PR #234, opened 2021-06-04
[13.0]  stock_storage_type_putaway_abc: improve location sorting

TDu:
Before this change, the list of location returned by the module would be
sorted by abc type first and then by max height. This sets a higher
priority for the abc type than the max height paramter on selecting a
suitable location.

Resulting in a less than optimal usage of height storage space.

Changing the sorting to be done first by max height and then by abc type
sould improve on that.

Also improve performance in the sorting of max height.

### DIFF
--- a/stock_storage_type_putaway_abc/models/stock_location.py
+++ b/stock_storage_type_putaway_abc/models/stock_location.py
@@ -9,16 +9,13 @@ ABC_SELECTION = [("a", "A"), ("b", "B"), ("c", "C")]
 
 
 def gather_location_ids(abc_sorted, max_heights_sorted, locations_grouped):
-    """Return a list of location IDs sorted on `abc_sorted` then
-    on `max_heights_sorted`.
+    """Return a list of location IDs sorted on `max_heights` then `abc_sorted`
     """
     location_ids = []
-    for abc_key in abc_sorted:
-        group_ids_per_height = locations_grouped.get(abc_key)
-        if not group_ids_per_height:
-            continue
-        for max_height in max_heights_sorted:
-            location_ids.extend(group_ids_per_height.get(max_height, []))
+    for max_height in max_heights_sorted:
+        for abc_key in abc_sorted:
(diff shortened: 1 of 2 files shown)


simahawk (approved):
LG

stock_storage_type_putaway_abc/models/stock_location.py
```diff
@@ -75,12 +72,9 @@ def _sort_abc_locations(self, product_abc):
...
             # keep a list of available max heights
             max_heights.add(line["max_height"])
-        # sort max heights and take care to put any 0 value at the end
+        # sort max heights keeping zero value at the end
         max_heights = list(max_heights)
-        max_heights.sort()
-        if 0 in max_heights:
-            max_heights.pop(max_heights.index(0))
-            max_heights.append(0)
+        max_heights.sort(key=lambda x: x if x else float("inf"))
```
sebalix, line 77: Didn't know about `float('inf')`, thanks!

sebalix (approved):

jbaudoux:
Merge sort algo is quick but as it makes one sql query for each height, it is at the end slower than python bubble sort.
I did some tests with an additional sequence sorting (which allow to sort after abc) and performance drops to much.
For mixed pallets with many put-away to compute on large locations table, this is too slow. While testing the speed, I have end-up with a new PR https://github.com/OCA/wms/pull/237 which is about 5-10x faster depending on the amount of different heights you have.

Nevertheless, your PR functionally speaking works and performance should not be affected by this change

lmignon:
@jbaudoux @jgrandguillaume @sebalix Shouldn't this RP be closed in favour of #237 ?

simahawk:
> @jbaudoux @jgrandguillaume @sebalix Shouldn't this RP be closed in favour of #237 ?

@jbaudoux ping :)

jbaudoux:
> > @jbaudoux @jgrandguillaume @sebalix Shouldn't this RP be closed in favour of #237 ?
> 
> @jbaudoux ping :)

@TDu will do it ;)

PR #237, opened 2021-06-04
[13.0] Change sorting of locations in the pack putaway computation

jbaudoux:
Invert the sorting for abc packs storage strategy to first apply the max_height which is a physical constraint, then apply the abc preference.

Add a pack putaway sequence to be able to favor some locations (e.g. level). For ordered children strategy, finalize by sorting by location name. For abc, random selection is applied at last as it was already.

Replaced multi-steps python sorting by an unique SQL quicksort query to quickly compute the result on large stock_location table. This is especially important for operations containing many moves requiring a putaway computation.

Improved READMEs

Removed alpha status

cc @lmignon

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -1,6 +1,9 @@
-# Copyright 2019 Camptocamp SA
+# Copyright 2019-2021 Camptocamp SA
+# Copyright 2019-2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 import logging
+
+from psycopg2 import sql
 
 from odoo import api, fields, models
 from odoo.tools import float_compare
@@ -49,6 +52,7 @@ class StockLocation(models.Model):
         "locations according to the restrictions defined on their "
         "respective location storage types.",
     )
+    pack_putaway_sequence = fields.Integer()
     storage_location_sequence_ids = fields.One2many(
         "stock.storage.location.sequence",
         "location_id",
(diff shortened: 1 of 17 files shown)


stock_storage_type/models/stock_location.py
```diff
@@ -361,10 +365,37 @@ def get_storage_locations(self, products=None):
...
-            locations = self._get_ordered_leaf_locations()
+        else:
+            locations = self._get_sorted_leaf_locations(products)
         return locations
 
+    def _get_sorted_leaf_locations_orderby(self, products):
+        orderby = []
+        if self.pack_putaway_strategy == "ordered_locations":
+            orderby = [
+                "CASE WHEN max_height > 0 THEN max_height ELSE 'Infinity' END",
```
lmignon, line 385: We could create a computed field to replace the `CASE THEN ELSE` syntax.

jbaudoux: It is already a stored computed field but the value is 0 when not defined and should be put after positive values

jbaudoux:
FIXME some tests are non deterministic

jgrandguillaume (approved):
This is really a very nice addition and change my friend ! Well done :=)

stock_storage_type/models/stock_location.py
lmignon: If we encounter perf issue into the sorting by postgres, we could have a `max_height_sort_indice` computed from max_height and create an index `(max_height_sort_indice, pack_putaway_sequence, name)`

lmignon (approved):

stock_storage_type/models/stock_location.py
jbaudoux: @lmignon You could also create your index with
CREATE INDEX .... ON stock_location((CASE WHEN max_height > 0 THEN max_height ELSE 'Infinity' END));

simahawk: @jbaudoux I'd add a TODO comment

stock_storage_type_putaway_abc/readme/DESCRIPTION.rst
```diff
@@ -1,2 +1,17 @@
 This module implements chaotic storage 'ABC' according to Package Storage Type
 and Location Storage Type.
+
+The locations and products get an 'a', 'b' or 'c' storage (by default 'b').
+
+For the computation of the putaway, the locations are sorted first by max
+height which is a physical constain to respect, then by 'ABC' following that:
```
simahawk:
```suggestion
height which is a physical constraint to respect, then by 'ABC' as following:
```

simahawk (approved):
just minor remarks

stock_storage_type/models/stock_location.py
jbaudoux: @simahawk A TODO for what? Even with an INDEX and large location table, in my case postgresql favors seq scan probably due to subsequent sorting

stock_storage_type_putaway_abc/readme/DESCRIPTION.rst
jbaudoux: Thanks. I fixed also the generated README.rst

simahawk:
@jbaudoux pls squash ;)

jbaudoux:
@simahawk squashed

PR #238, opened 2021-06-08
[FIX] shopfloor: single pack transfer, handle backorders

sebalix:
When processing a package with the single pack transfer scenario which
belongs to a transfer having several other packages, the remaining ones
are not put in a backorder and as such can not be processed anymore.

This commit ensure to put the processed package/package level in its own
move which will be validated in its own transfer, keeping the remaining
packages to process in the current transfer.

Ref. 2400

### DIFF
--- a/shopfloor/services/single_pack_transfer.py
+++ b/shopfloor/services/single_pack_transfer.py
@@ -256,7 +256,14 @@ class SinglePackTransfer(Component):
         # on the move lines
         package_level.location_dest_id = scanned_location
         stock = self._actions_for("stock")
-        stock.validate_moves(package_level.move_line_ids.move_id)
+        package_moves = package_level.move_line_ids.move_id
+        package_move_lines = package_level.move_line_ids
+        for package_move in package_moves:
+            # Check if there is no other lines linked to the move others than
+            # the lines related to the package itself. In such case we have to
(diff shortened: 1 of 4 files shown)


shopfloor/services/single_pack_transfer.py
```diff
@@ -256,7 +256,14 @@ def _set_destination_and_done(self, package_level, scanned_location):
         # on the move lines
         package_level.location_dest_id = scanned_location
         stock = self._actions_for("stock")
-        stock.validate_moves(package_level.move_line_ids.move_id)
+        package_moves = package_level.move_line_ids.move_id
```
simahawk: @sebalix am I wrong or we have done this N times in many places for the same reason?
Can't we move this to `stock` action?

sebalix: Hum yes, it has been done like this at least in the location content transfer. I'll try to move it to the stock action.

sebalix: I added a commit adding a new method `put_package_level_in_move` in the `stock` action, used by the single pack transfer and location content transfer scenarios.

simahawk (approved):

mmequignon (approved):

TDu (approved):

PR #240, opened 2021-06-16
13.0 Fix failing branch in stock-logistics-warehouse for stock_vertical_lift_storage_type

jbaudoux:
Ensuring order when tray cells have all the same name

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -385,6 +385,7 @@ class StockLocation(models.Model):
                 "CASE WHEN max_height > 0 THEN max_height ELSE 'Infinity' END",
                 "pack_putaway_sequence",
                 "name",
+                "id",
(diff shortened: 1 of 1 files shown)


stock_storage_type/tests/test_storage_type_move.py
```diff
@@ -371,5 +371,3 @@ def _levels_for(packages):
                 move_line.qty_done = move_line.product_uom_qty
 
         second_level.location_dest_id = third_level.location_dest_id
-        with self.assertRaises(ValidationError):
```
jbaudoux: FIXME This error is not raised anymore but the contraint _check_storage_type is still there

jbaudoux: I put back this check. However, I don't get why the constrains is not loaded by the ORM when I run the test locally

PR #241, opened 2021-07-02
13.0: s_s_available_to_promise_release: Sale report product line availability

mmequignon:
Displays the line availability on sale report.
- "On order" if product's route is MTO
- "Yes" if product is fully available
- A qty if product is partially available
- A date if no stock and there's an active replenishment
- "No" otherwise

### DIFF
--- a/sale_stock_available_to_promise_release/models/product_product.py
+++ b/sale_stock_available_to_promise_release/models/product_product.py
@@ -1,0 +1,21 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_next_replenishment_date(self):
+        self.ensure_one()
+        move_line = self.env["stock.move"].search(
+            [
+                ("product_id", "=", self.id),
+                ("state", "not in", ["done", "cancel"]),
+                ("picking_type_id.code", "=", "incoming"),
+            ],
+            limit=1,
+            order="date_expected asc",
+        )
+        return move_line and move_line.date_expected or False
--- a/sale_stock_available_to_promise_release/models/sale_order_line.py
+++ b/sale_stock_available_to_promise_release/models/sale_order_line.py
@@ -1,13 +1,67 @@
 # Copyright 2019 Camptocamp (https://www.camptocamp.com)
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
-from odoo import models
+from odoo import api, fields, models
+from odoo.tools import float_compare, float_is_zero
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    def _prepare_procurement_values(self, group_id=False):
-        values = super()._prepare_procurement_values(group_id)
-        values["date_priority"] = self.order_id.date_order
-        return values
+    availability_status = fields.Selection(
+        selection=[
+            ("mto", "On order"),
+            ("full", "Fully Available"),
+            ("partial", "Partially Available"),
+            ("restock", "Restock ordered"),
+            ("no", "Not available"),
+        ],
+        compute="_compute_availability_status",
+    )
+    expected_availability_date = fields.Datetime(compute="_compute_availability_status")
+    available_qty = fields.Float(
+        digits="Product Unit of Measure", compute="_compute_availability_status"
+    )
+
+    @api.depends(
+        "move_ids.ordered_available_to_promise_uom_qty",
+        "product_id.route_ids",
+        "product_uom_qty",
+    )
+    def _compute_availability_status(self):
+        for record in self:
+            # Fallback values
+            availability_status = "no"
+            expected_availability_date = False
+            available_qty = sum(
+                record.mapped("move_ids.ordered_available_to_promise_uom_qty")
+            )
+            # required values
+            product = record.product_id
+            rounding = product.uom_id.rounding
+            mto_route = self.env.ref("stock.route_warehouse0_mto")
+            # MTO product
+            if mto_route in product.route_ids:
+                availability_status = "mto"
+            # Fully available
+            elif (
+                float_compare(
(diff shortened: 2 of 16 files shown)


sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -1,13 +1,73 @@
...
+                )
+                >= 0
+            )
+            is_partially_available = (
+                float_compare(available_qty, 0, precision_rounding=rounding) == 1
+            )
+            is_not_available = float_is_zero(available_qty, precision_rounding=rounding)
+            mto_route = self.env.ref("stock.route_warehouse0_mto")
+            is_mto = mto_route in product.route_ids
+            # If mto, status = mto
```
TDu: Non useful comment in my opinion

sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -1,13 +1,73 @@
...
+            # required values
+            product = record.product_id
+            rounding = product.uom_id.rounding
+            is_fully_available = (
+                float_compare(
+                    available_qty, record.product_uom_qty, precision_rounding=rounding
+                )
+                >= 0
+            )
+            is_partially_available = (
```
TDu: If it is fully available it would be better to not do this computation as well no ? Improved  performance

sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -1,13 +1,73 @@
...
+                float_compare(available_qty, 0, precision_rounding=rounding) == 1
+            )
+            is_not_available = float_is_zero(available_qty, precision_rounding=rounding)
+            mto_route = self.env.ref("stock.route_warehouse0_mto")
+            is_mto = mto_route in product.route_ids
+            # If mto, status = mto
+            if is_mto:
```
TDu: Same remark than above, if `is_mto` is True both previous computation where not needed.
I think code could be improved.
Should not be difficult because it has tests :smile:

sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -1,13 +1,73 @@
...
+        ],
+        compute="_compute_availability_status",
+    )
+    expected_availability_date = fields.Datetime(compute="_compute_availability_status")
+    available_qty = fields.Float(
+        digits="Product Unit of Measure", compute="_compute_availability_status"
+    )
+
+    @api.depends(
+        "is_mto",
```
TDu: Not sure about this one

sale_stock_available_to_promise_release/models/product_product.py
```diff
@@ -0,0 +1,23 @@
...
+
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_next_replenishment_date(self):
+        self.ensure_one()
+        picking_type_in = self.env.ref("stock.picking_type_in")
```
jbaudoux: You should search for picking type with code incoming. And ideally take incoming moves where the destination location is a child of the warehouse view location if you compute that in the scope of a SO where a warehouse given.

sale_stock_available_to_promise_release/models/product_product.py
```diff
@@ -0,0 +1,23 @@
...
+        move_line = self.env["stock.move"].search(
+            [
+                ("product_id", "=", self.id),
+                ("date_expected", ">=", fields.Date.today()),
+                ("state", "not in", ["done"]),
```
jbaudoux: done, cancel

sale_stock_available_to_promise_release/models/product_product.py
```diff
@@ -0,0 +1,23 @@
...
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_next_replenishment_date(self):
+        self.ensure_one()
+        picking_type_in = self.env.ref("stock.picking_type_in")
+        move_line = self.env["stock.move"].search(
+            [
+                ("product_id", "=", self.id),
+                ("date_expected", ">=", fields.Date.today()),
```
jbaudoux: could be expected in the past and not yet received. Then expected date is today

jbaudoux (changes requested):

sale_stock_available_to_promise_release/models/sale_order_line.py
jbaudoux: :+1:

sale_stock_available_to_promise_release/models/product_product.py
jgrandguillaume: +1 for picking type = incoming

sale_stock_available_to_promise_release/models/product_product.py
jgrandguillaume: +1

sale_stock_available_to_promise_release/models/product_product.py
jgrandguillaume: So the filter should find everything that has not been received yet -> product = product, state not in cancel or done

mmequignon: `>=` will select moves where date is today as well, then we're fine, right?

sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -1,13 +1,73 @@
...
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    def _prepare_procurement_values(self, group_id=False):
-        values = super()._prepare_procurement_values(group_id)
-        values["date_priority"] = self.order_id.date_order
-        return values
+    availability_status = fields.Selection(
```
jgrandguillaume, line 11: Please add this in the order line before unit_price (with the three dot option that allow to hide it)

jbaudoux, line 11: Show on SO line before unit price. Make visibility configurable.

TDu (approved):

jbaudoux (approved):

PR #247, opened 2021-07-16
[14.0][MIG] shopfloor

hparfr:
Another try to migrate shopfloor


simahawk:
@hparfr to ease reviews it would be better to:

1. create a new branch based on v14
2. pick the module from v13
3. make it uninstallable
4. open a new PR
5. merge it in 14.0
6. rebase the PR here

This way it will be much easier.

hparfr:
@simahawk ok I will do that.

I had so many conflicts in the merging and wasted a lot time across all the dependencies and changes. Having a green CI is a already a big achievement \o/

oca_dependencies.txt
```diff
@@ -1 +1,9 @@
 # See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
+
```
simahawk, line 2: we don't need this

shopfloor/actions/move_line_search.py
```diff
@@ -92,21 +92,19 @@ def _sort_key_move_lines(order):
...
             )
         return lambda line: line
 
-    def counters_for_lines(self, lines, priority_selection=("2", "3")):
+    def counters_for_lines(self, lines):
         # Not using mapped/filtered to support simple lists and generators
-        priority_lines = [
-            x for x in lines if x.picking_id.priority in priority_selection
-        ]
+        priority_lines = [x for x in lines if x.picking_id.priority == 1]
```
sebalix: Not sure it is working? Should be a string IMO
```suggestion
        priority_lines = [x for x in lines if x.picking_id.priority == "1"]
```
https://github.com/odoo/odoo/blob/8e11a0b9683c9593f50ab0c290a1bdf857e5141e/addons/stock/models/stock_move.py#L20

shopfloor/services/delivery.py
```diff
@@ -253,7 +253,7 @@ def _deliver_product(self, picking, product):
         # but also if we have one product as a package and the same product as
         # a unit in another line. In both cases, we want the user to scan the
         # package.
-        if packages and len({l.package_id for l in lines}) > 1:
+        if packages and len({m.package_id for m in lines}) > 1:
```
sebalix, line 256: Why this change?

shopfloor/services/delivery.py
```diff
@@ -303,7 +303,7 @@ def _deliver_lot(self, picking, lot):
         # package, but also if we have one lot as a package and the same lot as
         # a unit in another line. In both cases, we want the user to scan the
         # package.
-        if packages and len({l.package_id for l in lines}) > 1:
+        if packages and len({m.package_id for m in lines}) > 1:
```
sebalix, line 306: Same question

shopfloor/services/location_content_transfer.py
```diff
@@ -817,6 +818,9 @@ def stock_out_package(self, location_id, package_level_id):
             # split the move to process only the lines related to the package.
             package_move.split_other_move_lines(package_move_lines)
             lot = package_move.move_line_ids.lot_id
+            # We need to set qty_done at 0 because otherwise
+            # the move_line will not be deleted
+            package_move.move_line_ids.write({"qty_done": 0})
```
sebalix, line 823: Ok, now Odoo 14.0 doesn't unreserve lines having a `qty_done > 0`: https://github.com/odoo/odoo/blob/8e11a0b9683c9593f50ab0c290a1bdf857e5141e/addons/stock/models/stock_move.py#L704-L713
I'm wondering if this is the right thing to do here, if the `qty_done` has been set this means an operator processed it, resetting it to 0 is probably not what we want :thinking:

shopfloor/services/location_content_transfer.py
```diff
@@ -872,6 +876,9 @@ def stock_out_line(self, location_id, move_line_id):
         move = move_line.move_id
         package = move_line.package_id
         lot = move_line.lot_id
+        # We need to set qty_done at 0 because otherwise
+        # the move_line will not be deleted
+        move_line.qty_done = 0
```
sebalix, line 881: Same remark, to check later

shopfloor/tests/test_menu_counters.py
```diff
@@ -9,8 +9,8 @@ def test_menu_search(self):
             self.menu1.id: {
                 "lines_count": 2,
                 "picking_count": 2,
-                "priority_lines_count": 2,
-                "priority_picking_count": 2,
+                "priority_lines_count": 0,
+                "priority_picking_count": 0,
```
sebalix, line 13: I suppose we should have 2 priority lines and 1 priority picking related to the `menu1` here based on the configuration of pickings in `test_menu_base.py` above, link to my previous comment https://github.com/OCA/wms/pull/247/files#r675964910

sebalix (commented):
Well done, thanks! Few minor comments, once fixed/checked I think we should merge this PR ASAP to ease the port of others PRs we did over time.

shopfloor/services/delivery.py
hparfr: Because of linters :disappointed:

shopfloor/services/location_content_transfer.py
hparfr: So how you want to achieve it ?

sebalix: I don't know yet, previously with the 13.0 implementation we didn't care about it, so this should not block the merge for now.

shopfloor/tests/test_menu_counters.py
sebalix: So the result is OK here: `picking1` and `picking3` are assigned but has a low priority (set to `"0"`), while the picking2 which is as urgent is still `confirmed` so it has no line counted here.

sebalix (approved):

sebalix:
@simahawk if it's OK for you we could merge this one

simahawk (approved):

PR #248, opened 2021-07-19
[14.0][MIG] shopfloor : re-do commits from v13

hparfr:
Keep v14 in sync with v13.

shopfloor is not migrated in this PR.


hparfr:
ping @sebastienbeau

PR #249, opened 2021-07-20
[13.0][FIX] s_r_screen: subcontracting fix with components tracked by lot

sebalix:
When receiving goods with the reception screen we are able to process
a partial quantity, splitting the current move and validating it on its
own. By convenience, the stock.move.line related to the finished moves
(in the production order) are removed and recreated (by duplicating move
lines received) when the reception is validated.

This last part has been rewritten in 'stock_reception_screen_mrp_subcontracting'
module to satisfy a technical need for the reception screen, but this was
causing an issue when receiving subcontracted goods with components
tracked by lots.
The fix here is to disable the blocking check which compares the produced
qty of the production order (which is 0 because all produced move lines have
been removed) to the qty received on the move when the reception is
validated by the reception screen.

Ref. 2670

### DIFF
--- a/stock_reception_screen_mrp_subcontracting/tests/test_reception_screen.py
+++ b/stock_reception_screen_mrp_subcontracting/tests/test_reception_screen.py
@@ -10,6 +10,10 @@ class TestReceptionScreenMrpSubcontracting(Common):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls._prepare_reception_screen()
+
+    @classmethod
+    def _prepare_reception_screen(cls):
         # product.product_delivery_02 has a subcontracted BoM defined with
         # 'mrp_subcontracting' installed
         cls.picking = cls._create_picking_in(partner=cls.env.ref("base.res_partner_12"))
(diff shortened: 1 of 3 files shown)


TDu (approved):

stock_reception_screen_mrp_subcontracting/tests/test_reception_screen.py
```diff
@@ -18,7 +22,106 @@ def setUpClass(cls):
...
+                default_subcontract_move_id=move.id,
+            )
+            wiz_model = self.env["mrp.product.produce"].with_context(context)
+            wiz = wiz_model.create({})
+            for line in wiz.move_raw_ids.move_line_ids:
+                line.qty_done = line.product_uom_qty
+            wiz.do_produce()
+
     def test_reception_screen_with_subcontracted_products(self):
+        """Receive subcontracted goods."""
```
mmequignon, line 40: Is it possible to split this? You already added the comments where it would make sense to do so.

mmequignon (approved):

stock_reception_screen_mrp_subcontracting/tests/test_reception_screen.py
simahawk: @sebalix are you going to do it here?

mmequignon:
@simahawk @sebalix let's merge this

stock_reception_screen_mrp_subcontracting/tests/test_reception_screen.py
simahawk: TBD in the future IF we keep it for v14

PR #140, opened 2021-01-22
[14.0][PORT] 113 from 13.0

hparfr:
Port #113 from 13.0 to 14.0.

PR #252, opened 2021-07-29
[13.0->14.0][oca-port-pr-#113] shopfloor: create backorder when validating moves

sebalix:
Port of #113 from 13.0 to 14.0.


sebalix:
I open this one to get the "same" commit than the one existing in 13.0 to ease the comparison between branches.
Commit https://github.com/OCA/wms/commit/6c133e450806a2e2031aa819a3d6ed8096aba4d0 (13.0) as been ported to https://github.com/OCA/wms/commit/8f5075e42dd4e07d930a7157c65d2a3ed5e1a106 (14.0) but the title has changed (`[IMP]` removed), that's why the following PR is empty.

Having this commit will allow the tool [`oca-port-pr`](https://github.com/OCA/maintainer-tools/pull/504) to not detect the original commit as missing from 14.0 branch.

hparfr (approved):
Quite strange. The #113 was ported by oca-port-pr in #140

sebalix:
@hparfr yes, but the commit message has changed from `[IMP] shopfloor: ...` to `shopfloor: ...`, that's why they are considered different. Anyway, even this new empty commit, `oca-port-pr` will still consider the 13.0 commit as not ported to 14.0 because of the difference in updated files (none here).
So I'll close this for now, and re-open it if I found a better solution in `oca-port-pr` :)

sebalix:
@hparfr also, feel free to test https://github.com/OCA/maintainer-tools/pull/504 , your feedback will be welcome :)

hparfr:
erratum, I was not using oca-port-pr but this one https://gist.githubusercontent.com/sbidoul/a865d293adad10fd7e4dd32d71fa43b8/raw/a90cf0bb108ad7ede012385c821d524ce008b835/port-pr.py

sebalix:
@hparfr I didn't know it, thanks for sharing!

sebalix:
@hparfr I found why the commit message was truncated, we have to use the `--keep` option with the `git am`, and the script you posted earlier doesn't do it, so it truncates patterns between `[]` brackets (see https://stackoverflow.com/questions/13339027/git-am-should-ignore-something-in-commit-message-startswith)
The migration guide of OCA mentions it however: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0#technical-method-to-migrate-a-module-from-130-to-140-branch

hparfr:
Thanks @sebalix

PR #254, opened 2021-07-29
[14.0][FW] [ADD] shopfloor_packing_info: Messages for Checkout Packing Info

sebalix:
Port of #161 from 13.0 to 14.0.

### DIFF
--- a/shopfloor_packing_info/hooks.py
+++ b/shopfloor_packing_info/hooks.py
@@ -16,26 +16,21 @@
 # process:
 #     1. pre-init will copy, in a temporary column, the contents
 #        of the column that is going to be removed when the update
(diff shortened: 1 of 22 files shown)


shopfloor_packing_info/hooks.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2021 Camptocamp SA (https://www.camptocamp.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+# The field shopfloor_packing_info on res.partner contained a text
```
simahawk, line 4: we should probably get rid of this on v14

sebalix:
Closing in favor of https://github.com/OCA/wms/pull/279 which contains translations etc.

PR #259, opened 2021-08-03
[13.0->14.0][oca-port-pr-#135] Add barcode on Package Storage Type, use it at reception screen

sebalix:
Port of #135 from 13.0 to 14.0.


sebalix:
Regarding `stock_reception_screen`, the commit related to it has been ported because the addon was existing in the 14.0 branch (but not installable), but for now we don't know if we will keep it for 14.0 or if it'll be replaced by a Shopfloor scenario.

Anyway, we could merge the related commit, even if one day we choose to completely remove it from 14.0.

mmequignon (approved):

simahawk (approved):

PR #265, opened 2021-08-05
[13.0->14.0][oca-port-pr-#112] Ensure moves routed from different location are reserved before other

sebalix:
Port of #112 from 13.0 to 14.0.

This module is not yet migrated (`"installable": False`) but the migration will be easier if missing commits are ported first.


sebalix:
Closing in favor of https://github.com/OCA/wms/pull/267

PR #271, opened 2021-08-06
[14.0][FW] shopfloor_workstation: multiple ports from 13.0

sebalix:
Port of the following PRs from 13.0 to 14.0:
- #173
- #169

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,6 +82,10 @@ repos:
         args:
           - --color
           - --fix
+        # SPECIAL CASE: barcode scanner app use Vue.js +
+        # advanced JS (modules, classes, etc)
+        # which are not supported by OCA rules yet.
+        exclude: shopfloor_*/*
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ addons:
   apt:
     packages:
       - expect-dev # provides unbuffer utility
+      - cups
+      - libcups2-dev
 
 stages:
   - test
--- a/setup/shopfloor_workstation/odoo/addons/shopfloor_workstation
+++ b/setup/shopfloor_workstation/odoo/addons/shopfloor_workstation
@@ -1,0 +1,1 @@
+../../../../shopfloor_workstation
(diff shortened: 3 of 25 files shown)


sale_stock_available_to_promise_release/__manifest__.py
```diff
@@ -6,6 +6,7 @@
     "version": "13.0.1.3.0",
     "summary": "Integration between Sales and Available to Promise Release",
     "author": "Camptocamp,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/wms",
```
sebalix: Note to myself: to remove

sebalix: done

mmequignon (approved):

simahawk:
Probably the commit [Disable ESLint for all shopfloor modules](https://github.com/OCA/wms/pull/271/commits/4dd751587161527cee7dbd89d0982d8c1f0df5dc) is not wanted in v14 

@hparfr could you review here?

sebalix:
@simahawk I skip it at first and Travis was not happy, so I finally included it here

simahawk:
travis was not happy  because it implies to fix linting. Let's wait for @hparfr feedback because he worked on this subject AFAIS.
Anyway, what about shopfloor_workstation_mobile? Will it come in a new PR?

sebalix:
@simahawk yes, but I could include it here if you prefer

hparfr:
I already disabled (err -> warn) some stupid rules. If there some more, we can disable them as well.

Personally, I dislike most of the rules, in particular the cosmetic ones, like upper case in comments :rage:

simahawk:
> @simahawk yes, but I could include it here if you prefer

no that's ok.



> I already disabled (err -> warn) some stupid rules. If there some more, we can disable them as well.
> 
> Personally, I dislike most of the rules, in particular the cosmetic ones, like upper case in comments rage

@hparfr ok, then, let's go like this. Would be nice to have some linting tho somehow in the future ;)

simahawk (approved):

hparfr:
I mean, is better to disable problematic rules than exclude everything.

simahawk:
ops

sebalix:
@hparfr oh sorry, I merge too fast then, so let's fix the JS lint in another PR

simahawk:
@sebalix no it's ok. Go ahead. I blocked my merge build... we did at the same time :smile:

sebalix:
Work to migrate `shopfloor_workstation_mobile` is done here:
- https://github.com/OCA/wms/pull/273

PR #243, opened 2021-07-07
13.0: s_s_available_to_promise_release: fixes

mmequignon:
Fix availability status which was computed on sections and delivery lines.

PR #184, opened 2021-02-28
[MIG] stock_available_to_promise_release: Migration to 14.0

dzungtran89:

PR #94, opened 2020-12-10
[13.0][FIX] 'Release ready' based on the shipping policy of the order

sebalix:
Ref. 1671

PR #274, opened 2021-08-10
[14.0][MIG] sale_stock_available_to_promise_release (+multiple ports from 13.0)

sebalix:
Port of the following PRs from 13.0 to 14.0:
- <s>#94</s> Already ported here: https://github.com/OCA/wms/pull/270
- #241
- #243

Depends on:
- https://github.com/OCA/wms/pull/184

### DIFF
--- a/sale_stock_available_to_promise_release/models/sale_order_line.py
+++ b/sale_stock_available_to_promise_release/models/sale_order_line.py
@@ -1,13 +1,67 @@
 # Copyright 2019 Camptocamp (https://www.camptocamp.com)
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
-from odoo import models
+from odoo import api, fields, models
+from odoo.tools import float_compare, float_is_zero
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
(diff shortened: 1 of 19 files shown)


mmequignon (approved):

sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -1,13 +1,87 @@
...
+        "move_ids.ordered_available_to_promise_uom_qty",
+        "product_id.route_ids",
+        "route_id",
+        "product_uom_qty",
+        "display_type",
+        "is_delivery",
+    )
+    def _compute_availability_status(self):
+        for record in self:
+            if record.display_type or not record.product_id:
```
simahawk: not a blocker but I would split this in smaller methods

simahawk (approved):

sale_stock_available_to_promise_release/models/sale_order_line.py
sebalix: it's not really smaller, but I put the computation of the data in its own method to ease the override in https://github.com/OCA/wms/pull/274/commits/2d5fdb4f2324355bf84a60853757fbf47214b234

PR #283, opened 2021-08-26
[13.0] stock_storage_type: Fix putaway compution when a strategy is defined

jbaudoux:
When the strategy has to find a leaf sub-location (strategy = ordered children location), do not return the locations defined on a storage type putaway sequence (this is not a child location).

Example: sequence is defined to first look in WH/Stock/area1 and then in WH/Stock/area2. On area1 the strategy is to find a leaf sub-location under area1. If area1 has temporarily no sub-location, area1 should not be returned as a valid leaf location.

Also, when searching for a leaf location, exclude non-active locations

cc @jgrandguillaume @TDu @lmignon

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -112,6 +112,11 @@ class StockLocation(models.Model):
     leaf_location_ids = fields.Many2many(
         "stock.location",
         compute="_compute_leaf_location_ids",
+        help="technical field: all the leaves locations",
+    )
+    leaf_child_location_ids = fields.Many2many(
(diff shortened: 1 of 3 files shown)


lmignon (approved):
Makes sens. LGTM (code review)

jbaudoux:
+ added active filter on locations in SQL query

jbaudoux:
Rewrote commit message with better description

TDu (approved):
LG, but I do not get all the subtlety of the change.

jgrandguillaume (approved):
LGTM

PR #284, opened 2021-09-01
[14.0][IMP] shopfloor: added support for multiple barcodes

hparfr:
Add support of multi-barcode. It can be implemented (or not) in scenarios.

### DIFF
--- a/shopfloor/__manifest__.py
+++ b/shopfloor/__manifest__.py
@@ -18,6 +18,7 @@
         "shopfloor_base",
         "stock",
         "stock_picking_batch",
+        "product_multi_barcode",
         "base_jsonify",
         "base_rest",
         "base_sparse_field",
--- a/shopfloor/actions/search.py
+++ b/shopfloor/actions/search.py
@@ -38,7 +38,10 @@ class SearchAction(Component):
         model = self.env["product.product"]
         if not barcode:
             return model.browse()
-        product = model.search([("barcode", "=", barcode)], limit=1)
+        product = model.search(
+            ["|", ("barcode", "=", barcode), ("barcode_ids.name", "=", barcode)],
+            limit=1,
+        )
         if not product:
             packaging = self.env["product.packaging"].search(
                 [("product_id", "!=", False), ("barcode", "=", barcode)], limit=1
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -225,10 +225,21 @@ class DataAction(Component):
             "display_name",
             "default_code",
             "barcode",
+            (
+                "barcode_ids:barcodes",
+                lambda record, fname: self._product_barcode_list(record[fname]),
+            ),
             ("packaging_ids:packaging", self._product_packaging),
             ("uom_id:uom", self._simple_record_parser() + ["factor", "rounding"]),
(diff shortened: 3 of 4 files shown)


shopfloor/__manifest__.py
```diff
@@ -18,6 +18,7 @@
         "shopfloor_base",
         "stock",
         "stock_picking_batch",
+        "product_multi_barcode",
```
simahawk, line 21: shall we move this feature to a glue module `shopfloor_product_multi_barcode`?

shopfloor/actions/search.py
```diff
@@ -38,7 +38,10 @@ def product_from_scan(self, barcode):
         model = self.env["product.product"]
         if not barcode:
             return model.browse()
-        product = model.search([("barcode", "=", barcode)], limit=1)
+        product = model.search(
+            ["|", ("barcode", "=", barcode), ("barcode_ids.name", "=", barcode)],
```
simahawk, line 42: we can split the domain generation to a separated method to be able to override it from the other module

simahawk (changes requested):

hparfr:
Yep, that the kind of feature you see as a must have, but not every implementation need / want it.
I want some of the scenarios (like reception) be part of shopfloor and use multibarcode.

So instead of making these changes, what do you think @simahawk of change all scenarios and always use `barcodes` or `barcode_ids` in all shopfloor(_.+)* modules? In that way, the support of multibarcode is everywhere and can be opted-in with the installation of glue shopfloor_barcode"
with a default implementation in shopfloor like : 
```py
barcodes = [ product.barcode ]
# or 
barcode_ids = [ {"id": False, "name": product.barcode } ]
```

hparfr:
ping @sebastienbeau

simahawk:
> Yep, that the kind of feature you see as a must have, but not every implementation need / want it. I want some of the scenarios (like reception) be part of shopfloor and use multibarcode.
> 
> So instead of making these changes, what do you think @simahawk of change all scenarios and always use `barcodes` or `barcode_ids` in all shopfloor(_.+)* modules? In that way, the support of multibarcode is everywhere and can be opted-in with the installation of glue shopfloor_barcode" with a default implementation in shopfloor like :
> 
> ```python
> barcodes = [ product.barcode ]
> # or 
> barcode_ids = [ {"id": False, "name": product.barcode } ]
> ```

You mean that the UI should always expect a list?
BTW is it useful to expose the IDs?

hparfr:
> You mean that the UI should always expect a list?


No, you only scan one barcode.
The use case is when you have to consider two products with different barcode to be identical.

>BTW is it useful to expose the IDs?

Yes, I need it when scanning two products in a row to detect if we are scanning the same product or not.

simahawk:
> > You mean that the UI should always expect a list?
> 
> No, you only scan one barcode. The use case is when you have to consider two products with different barcode to be identical.

You got me wrong: I mean that if you want to support multiple barcodes seamlessly on the front, you have to expect a list of barcodes back when you return the product info and display them accordingly where needed.

> > BTW is it useful to expose the IDs?
> 
> Yes, I need it when scanning two products in a row to detect if we are scanning the same product or not.

but here you search only by name -> the ID seems useless

PR #287, opened 2021-09-10
[13.0] shopfloor: support barcode that do not send end of line + trim value

jbaudoux:
When using multiple apps on the barcode, sometimes you cannot reconfigure the barcode to append an end of line to the scanned value.
The searchbar will auto detects if the code is scanned or manually typed. By default, the search will automatically execute when scanned and no more input is fed. The end of line can still be send. By default, when typing, you have to press enter to submit the input.

Also trim spaces from the input value

cc @simahawk @lmignon @sebalix @jgrandguillaume

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
+++ b/shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
@@ -13,6 +13,7 @@ Vue.component("searchbar", {
     data: function() {
         return {
             entered: "",
+            debounceWait: this.autosearch,
         };
     },
     props: {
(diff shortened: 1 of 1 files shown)


hparfr:
Why not adding a submit button on the search bar ?

jbaudoux:
> Why not adding a submit button on the search bar ?

Because when you scan you don't want to click each time on a button

hparfr:
I think adding a debounce is not a good strategy for a barcode app. It's not an autocomplete search where the user is prompted with suggestions.
This kind of code leads to side effects and are painful for users. 

Another solution, is to print a "\n" barcode and stick somewhere.

hparfr (changes requested):
For the reasons stated before I think this code shouldn't be merged in shopfloor_mobile_base

jbaudoux:
> I think adding a debounce is not a good strategy for a barcode app. It's not an autocomplete search where the user is prompted with suggestions. This kind of code leads to side effects and are painful for users.
> 
> Another solution, is to print a "\n" barcode and stick somewhere.

@hparfr I changed the code to only "auto submit" when the input is scanned and not when it is typed.When you type, you can press enter when you're done and want to trigger the search.

I also added a fix to emptying the field when enter is pressed and the search is launched. Otherwise, the field was still filled on the next step.

shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
```diff
@@ -33,18 +36,76 @@ Vue.component("searchbar", {
...
+    },
     methods: {
         search: function(e) {
-            e.preventDefault();
             // Talk to parent
+            if (!this.entered) return;
             this.$emit("found", {
                 text: this.entered,
-                type: e.target.dataset.type,
+                type: String,
```
simahawk: Please restore this line. The type here is the type of event, for internal usage.
When you bind to `found` you can pass a specific type through the data attribute to mark the event w/ a specific "type".
Call it `kind` if you want ;)

simahawk (changes requested):
Beside the event type thing, LG

shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
simahawk: check where/how this is used https://github.com/OCA/wms/search?q=input_data_type

jbaudoux: fixed, thanks

jbaudoux: Hmm, now it is broken. I need to investigate

jbaudoux: solved

simahawk (approved):

simahawk:
@hparfr can you update your review?

simahawk:
@hparfr ping

hparfr:
I still dislike this feature. It's too fragile to put time to detect changes.

Our hardware is not rtos. We may have latency with scanners, js runtime may be paused by the OS or the browser
Doing a quick test with my computer, I can trigger the search while typing !

I will not be surprised if we encounter two barcode in a packaging (sku + batch number). Fixing supplier packaging is hard.

I have a feeling you are putting some complexity in a base oca module because you can't tell your customer to simply configure his scanner the right way. I know it's not always possible, but I always prefer generic oca modules.

Anyway, if it's seems a lot burden to externalize this feature in his own module. I suggest then: (1) make the feature-switch to opt-in instead of the currently opt-out.

From a code perspective, I just have one remark: (2) always prefer a pure JS implementation of debounce instead of relying on \_.

(1) is blocking IMO. (2) is optional.

jbaudoux:
@hparfr I don't share the same vision as you. Having to configure the scanner firmware in a specific way to have the app running properly can also hinder adhesion and cause frustration.
When typing the search should not be triggered automatically as it is opt our by default here https://github.com/OCA/wms/pull/287/files#diff-a4869cf6c39b3a560f8a947c384f07161b8476c8e348b935da970f90a2fbfd63R60

Any other opinions? @lmignon @simahawk @jgrandguillaume

lmignon:
This fix works like a charm in production for several weeks now. In our case, it's not possible to change the scanner configuration since the configuration will impact all the installed applications. The switch between the user input mode and the scanner input one is smooth and transparent for the User. It provide a very good UX and is visually explicit. I'm not aware of any trouble introduced by this change. On a UX perspective I am in favour of merging it.

hparfr (approved):
>When typing the search should not be triggered automatically as it is opt our by default here

Ok then.

PR #290, opened 2021-09-22
[MIG] stock_move_source_relocate: Migration to 14.0

dzungtran89:


simahawk (approved):

stock_move_source_relocate/models/stock_move.py
```diff
@@ -60,9 +60,9 @@ def _apply_source_relocate_rule(self, relocation, reserved_availability, roundin
             # A part of the quantity could be reserved in the original
             # location, so keep this part in the move and split the rest
             # in a new move, where will take the goods in the relocation
-            new_move_id = self._split(need)
-            # recheck first move which should now be available
-            new_move = self.browse(new_move_id)
+            new_move_vals = self._split(need)
+            new_move = self.create(new_move_vals)
+            new_move._action_confirm(merge=False)
```
sebalix: In v13 `_split` method was returning the current move if no split occurred, now in v14 you have to check that the returned dictionary is not empty before calling `create`:
```suggestion
            new_move = self    # fallback if no split occurs
            new_move_vals = self._split(need)
            if new_move_vals:
                new_move = self.create(new_move_vals)
                new_move._action_confirm(merge=False)
```

sebalix (changes requested):

mmequignon (approved):

sebalix (approved):

PR #295, opened 2021-09-23
[MIG] stock_reception_screen: Migration to 14.0

dzungtran89:


stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -500,13 +503,16 @@ def process_select_product(self):
 
     def _split_move(self, move, qty):
         # Hook intended to be overridden
-        return move._split(qty)
+        new_move_vals = move._split(qty)
+        if new_move_vals:
+            new_move = self.env["stock.move"].create(new_move_vals)
+            new_move._action_confirm()
```
sebalix, line 509: `_split_move` has to return a recordset:
```python
            [...]
            new_move._action_confirm()
            return new_move
        return move
```

stock_reception_screen/models/stock_reception_screen.py
```diff
@@ -609,21 +615,23 @@ def process_set_expiry_date(self):
...
             )
             self.warn_notification = _(
                 "You cannot set a date prior to previous one ({})".format(
                     previous_life_date_str
                 )
             )
             return
-        self.current_move_line_lot_id.life_date = self.current_move_line_lot_life_date
+        self.current_move_line_lot_id.expiration_date = (
+            self.current_move_line_lot_life_date
```
sebalix: I would rename all `life_date` terms to `expiration_date` to be inline with the standard (this implies a pre-migration script to rename the column).

sebalix (commented):
@dzungtran89 the module is not installable, can you check?

stock_reception_screen/models/stock_reception_screen.py
dzungtran89: Hi @sebalix , I'll update the PR

stock_reception_screen/__manifest__.py
```diff
@@ -27,6 +27,6 @@
         "views/stock_reception_screen.xml",
         "views/manual_barcode.xml",
     ],
-    "installable": True,
+    "installable": False,
```
sebalix: @dzungtran89 sorry my mistake... the module was already flagged `installable` in the `pre-commit` commit, I was reviewing only your migration commit and didn't see the change.

stock_reception_screen/models/stock_reception_screen.py
sebalix: You forgot to return `new_move` in the `if` block (or does GH rendering is tricking me?)

dzungtran89: it's mybad actually

stock_reception_screen/__manifest__.py
dzungtran89: @sebalix okay, I missed this comment too :(. Just pushed the PR

sebalix (approved):
LG! Thank you for all the forward ports.

hparfr (approved):

PR #307, opened 2021-10-01
[14.0][FIX] shopfloor_batch_automatic_creation: fix typo

nguyenhk99:


shopfloor_batch_automatic_creation/tests/test_batch_create.py
```diff
@@ -178,7 +178,7 @@ def test_volume(self):
 
     def test_cluster_picking_select(self):
         self.menu.sudo().batch_create = True
-        self.menu.sudo().batch_create_max_pickings = 2
+        self.menu.sudo().batch_create_max_picking = 2
```
simahawk, line 181: seems legit, but: does it have any effect on tests? Seems not if it was passing before :thinking:

hailangvn: Thanks @simahawk , we are still scratching our head as we don't understand why the test was passing before AND the test was failed recently.

simahawk:
@sebalix fine for you?

sebalix (approved):

PR #313, opened 2021-10-21
[14.0] dotfiles update needs manual intervention

sbidoul:
Dear maintainer,

After updating the dotfiles, `pre-commit run -a`
fails in a manner that cannot be resolved automatically.

Can you please have a look, fix and merge?

Thanks,


simahawk:
Fixed by rolling back some changes in `.eslintrc.yml`. 

Here's the breaking change:

image

@Yajo @sbidoul what is the good practice - if any - to handle such exceptions by repo? (beside disabling such check on every JS file where it happens)

sbidoul:
> what is the good practice - if any - to handle such exceptions by repo

Ideally, variations should be handled in the template, with copier questions.
For the case here it would not work I suppose.

yajo:
Yes, although copier will respect your manual overrides as long as they don't interfere with the update (while updating).

PR #315, opened 2021-10-26
[14.0] Shopfloor mobile base - refactor auth

simahawk:
Auth is now pluggable. See atomic commits for details.

In short:
- auth is now pluggable on frontend thanks to the new auth registry
- 2 new modules for handling auth: `shopfloor_mobile_base_auth_user` + `shopfloor_mobile_base_auth_api_key`
- the base auth for shopfloor_base is still `api_key` which gets overriden by the 2 modules above (which are not compatible together as you can have only one type of auth at time)
- shopfloor_mobile_base installs `shopfloor_mobile_base_auth_api_key` to make sure current installations work as before

Support for multiple auth at the same time per different app will come with https://github.com/OCA/wms/pull/316.

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/main.js
+++ b/shopfloor_mobile_base/static/wms/src/main.js
@@ -219,6 +219,9 @@ new Vue({
             this.$router.push({name: "login"});
             this.trigger("logout:after");
         },
+        is_authenticated: function () {
+            return this.authenticated ? true : false;
+        },
         // Likely not needed anymore
         loadJS: function (url, script_id) {
             if (script_id && !document.getElementById(script_id)) {
--- a/shopfloor_mobile_base/static/wms/src/main.js
+++ b/shopfloor_mobile_base/static/wms/src/main.js
@@ -135,11 +135,6 @@ new Vue({
             let params = _.defaults({}, odoo_params, {
                 debug: this.demo_mode,
                 base_url: this.app_info.base_url,
-                // TODO: move out to its own handler
-                // when full aut decoupling happens
-                headers: {
-                    "API-KEY": this.apikey,
-                },
             });
             let OdooClass = null;
             if (this.demo_mode) {
@@ -181,9 +176,9 @@ new Vue({
                     self.appconfig = result.data;
                     self.authenticated = true;
                     self.$storage.set("appconfig", self.appconfig);
+                    return result;
                 } else {
-                    // TODO: any better thing to do here?
-                    console.log(result);
+                    return result;
                 }
             });
         },
@@ -215,6 +210,7 @@ new Vue({
             });
         },
         login: function (evt, data) {
+            evt.preventDefault();
             const self = this;
             this.trigger("login:before");
             const auth_handler = this._get_auth_handler();
@@ -225,13 +221,20 @@ new Vue({
                     .catch((error) => {
                         self.trigger("login:failure", error);
                     });
+            } else {
+                // TODO: we might want to enforce every authentication handler
+                // to provide the on_login handler to avoid such case
+                this._on_login();
             }
         },
         _on_login: function () {
             const self = this;
-            this._loadConfig().then(function () {
-                self.authenticated = true;
-                self.trigger("login:success");
+            this._loadConfig().then(function (result) {
+                if (!result.error) {
+                    self.trigger("login:success");
+                } else {
+                    self.trigger("login:failure");
+                }
             });
         },
         logout: function () {
@@ -245,6 +248,10 @@ new Vue({
                     .catch(function () {
                         self.trigger("logout:failure");
                     });
+            } else {
+                // TODO: we might want to enforce every authentication handler
+                // to provide the on_logout handler to avoid such case
+                this._on_logout();
             }
         },
         _on_logout: function () {
--- a/shopfloor_mobile_base/static/wms/src/main.js
+++ b/shopfloor_mobile_base/static/wms/src/main.js
@@ -147,14 +147,18 @@ new Vue({
             } else {
                 OdooClass = Odoo;
             }
+            const auth_handler = this._get_auth_handler();
+            params = _.merge({}, params, auth_handler.get_params(this));
+            // TODO: allow auth_handler to return OdooClass?
+            return new OdooClass(params);
+        },
+        _get_auth_handler: function (force) {
             const auth_type = this.app_info.auth_type;
             const auth_handler = auth_handler_registry.get(auth_type);
             if (_.isUndefined(auth_handler)) {
                 throw "Auth type '" + auth_type + " not supported";
             }
-            params = _.merge({}, params, auth_handler.get_params(this));
-            // TODO: allow auth_handler to return OdooClass?
-            return new OdooClass(params);
+            return auth_handler;
         },
         loadConfig: function (force) {
             if (this.appconfig && !force) {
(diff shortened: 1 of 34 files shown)


simahawk:
@lmignon @hparfr you might be interested in this.
Also, we'll soon add test coverage via Cypress for core frontend.

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -178,9 +176,9 @@ new Vue({
                     self.appconfig = result.data;
                     self.authenticated = true;
                     self.$storage.set("appconfig", self.appconfig);
+                    return result;
                 } else {
-                    // TODO: any better thing to do here?
-                    console.log(result);
+                    return result;
```
TDu: That `else` can probably be dropped, and only one return statement be kept.

TDu (approved):

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -136,26 +135,25 @@ new Vue({
...
+            const auth_handler = this._get_auth_handler();
+            params = _.merge({}, params, auth_handler.get_params(this));
+            // TODO: allow auth_handler to return OdooClass?
+            return new OdooClass(params);
+        },
+        _get_auth_handler: function (force) {
             const auth_type = this.app_info.auth_type;
             const auth_handler = auth_handler_registry.get(auth_type);
             if (_.isUndefined(auth_handler)) {
                 throw "Auth type '" + auth_type + " not supported";
```
petrus-v: I know this is not part to this current work but I suppose we could fix that in the mean time? 
```suggestion
                throw "Auth type '" + auth_type + "' not supported";
```

shopfloor_mobile_base/static/wms/src/router.js
```diff
@@ -57,7 +61,12 @@ const router = new VueRouter({
 });
 router.beforeEach(async (to, from, next) => {
     await Vue.nextTick();
-    if (!router.app.authenticated && to.meta.requiresAuth && !router.app.demo_mode) {
+    // debugger;
```
petrus-v:
```suggestion
```

petrus-v (approved):
I've leave 2 suggestions but won't block for that

I've tested both modules on my local env and passed the authentication process.

Also I wonder if we should let some documentation for fresh odoo install. As far I understand it's let to administrator to choose/install the right authentication module but at least one of those modules is required to make shopfloor_mobile_base auth process passed. Also considering functional runbot users that gives a try !

simahawk:
> I've leave 2 suggestions but won't block for that
> 
> I've tested both modules on my local env and passed the authentication process.

thanks for your review!

> Also I wonder if we should let some documentation for fresh odoo install. As far I understand it's let to administrator to choose/install the right authentication module but at least one of those modules is required to make shopfloor_mobile_base auth process passed. Also considering functional runbot users that gives a try !

The migration script solves this by installing api_key by default (because is the default today)

https://github.com/OCA/wms/pull/315/files#diff-82957e5675b1afd643a33048884d9b2ff90ba2edd6d8bcf2de3253bb0d5f79c3

I won't spend more time on this because the refactoring with `shopfloor.app` backend on #316 will solve the issue by allowing users to set the auth system via app config.

petrus-v:
> > I've leave 2 suggestions but won't block for that
> > I've tested both modules on my local env and passed the authentication process.
> 
> thanks for your review!
> 
> > Also I wonder if we should let some documentation for fresh odoo install. As far I understand it's let to administrator to choose/install the right authentication module but at least one of those modules is required to make shopfloor_mobile_base auth process passed. Also considering functional runbot users that gives a try !
> 
> The migration script solves this by installing api_key by default (because is the default today)
> 
> https://github.com/OCA/wms/pull/315/files#diff-82957e5675b1afd643a33048884d9b2ff90ba2edd6d8bcf2de3253bb0d5f79c3
> 
> I won't spend more time on this because the refactoring with `shopfloor.app` backend on #316 will solve the issue by allowing users to set the auth system via app config.

Reading migration script I don't think any of those module will be installed on a fresh new install which is good to let administrator/integrator to choose the right way ! Any way thanks a lot for this refactor, as mentioned I won't block on this PR, documentation could be improved afterwards.

shopfloor_mobile_base/static/wms/src/main.js
petrus-v: For record I were getting this error before installing  shopfloor_mobile_base_auth_api_key, to avoid loosing time we could suggested developer to ensure one of shopfloor mobile auth module is properly installed !

```suggestion
                throw "Auth type '" + auth_type + "' not supported, did you install the proper shopfloor auth module ?";
```

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -211,14 +207,59 @@ new Vue({
...
+                    });
+            } else {
+                // TODO: we might want to enforce every authentication handler
+                // to provide the on_login handler to avoid such case
+                this._on_login();
+            }
+        },
+        _on_login: function () {
+            const self = this;
+            this._loadConfig().then(function (result) {
```
hparfr, line 230:
```suggestion
            return this._loadConfig().then(function (result) {
```

be consistent with auth_handler.on_login

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -211,14 +207,59 @@ new Vue({
...
+                // to provide the on_logout handler to avoid such case
+                this._on_logout();
+            }
+        },
+        _on_logout: function () {
             this.authenticated = false;
             this._clearAppData();
             this.$router.push({name: "login"});
-            this.trigger("logout:after");
+            this.trigger("logout:success");
```
hparfr, line 259:
```suggestion
            this.trigger("logout:success");
            return Promise.resolve()
```
be consistent

shopfloor_mobile_base_auth_user/static/wms/src/login.js
```diff
@@ -0,0 +1,115 @@
...
+            .post("auth/login", data, true)
+            .then(function (response) {
+                if (response.error) {
+                    return def.reject();
+                }
+                return def.resolve();
+            })
+            .catch(function (error) {
+                return def.reject();
+            });
```
hparfr, line 45:
```suggestion
        return odoo
            .post("auth/login", data, true)
            .then(function (response) {
                if (response.error) {
                    return Promise.reject();
                }
                return response;
            })
 ```

File: shopfloor_mobile_base_auth_user/static/wms/src/login.js [hunk 7]
```diff
@@ -0,0 +1,115 @@
...
+        const odoo = $root.getOdoo({base_url: "/session/"});
+        return odoo
+            .post("auth/logout", null, true)
+            .then(function () {
+                return def.resolve();
+            })
+            .catch(function () {
+                return def.reject();
+            });
+    }
```
hparfr, line 59:
```suggestion
        const odoo = $root.getOdoo({base_url: "/session/"});
        return odoo
            .post("auth/logout", null, true);
    }
```

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -136,26 +135,25 @@ new Vue({
...
                 OdooClass = OdooMocked;
             } else {
                 OdooClass = Odoo;
             }
+            const auth_handler = this._get_auth_handler();
+            params = _.merge({}, params, auth_handler.get_params(this));
+            // TODO: allow auth_handler to return OdooClass?
+            return new OdooClass(params);
+        },
+        _get_auth_handler: function (force) {
```
hparfr, line 150: useless force param ?

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -211,14 +207,59 @@ new Vue({
...
+                    .catch((error) => {
+                        self.trigger("login:failure", error);
+                    });
+            } else {
+                // TODO: we might want to enforce every authentication handler
+                // to provide the on_login handler to avoid such case
+                this._on_login();
+            }
+        },
+        _on_login: function () {
```
hparfr, line 228:
```suggestion
        _default_on_login: function () {
```
or use the same arguments than _on_login

shopfloor_mobile_base_auth_user/static/wms/src/login.js
```diff
@@ -0,0 +1,115 @@
...
+             * we'll have to provide `credentials` same-origin|include
+             * to the fetch method.
+             * */
+
+            headers: {},
+        };
+    }
+
+    on_login($root, evt, data) {
+        evt.preventDefault();
```
hparfr, line 31:
```suggestion
    on_login($root, data) {
```
`evt.preventDefault();` already done in parent

hparfr (changes requested):
Thanks for this work.
Some remarks about the code, didn't test it yet

simahawk:
@hparfr thanks for you review! We'll propose your changes into https://github.com/OCA/wms/pull/324 so that we have test coverage. They are not blocking as far as I see as they don't break the new API added here.

simahawk:
I've just moved the version bump commit to the end. More fixes and improvements will be done on #324  to ensure we don't break anything.

/ocabot merge nobump

simahawk:
> @hparfr thanks for you review! We'll propose your changes into #324 so that we have test coverage. They are not blocking as far as I see as they don't break the new API added here.

Here's the commit https://github.com/OCA/wms/pull/324/commits/aff5c0488e2a2b264570c8dbff185bd4c5797695

PR #316, opened 2021-10-26
[14.0] Shopfloor: add app backend with automatic routing registration

simahawk:
Depends on
- [x] https://github.com/OCA/wms/pull/315 (Included here)
- [x] https://github.com/OCA/rest-framework/pull/214
- [x] https://github.com/OCA/rest-framework/pull/208
- [x] https://github.com/OCA/edi/pull/469
- [x] https://github.com/OCA/rest-framework/pull/219

The main change that this PR brings is: you can now have multiple shopfloor apps and configure them differently (eg: different url, different menus, different auth, etc).

NOTE: if you want to support the old URL of the app for your own project, you'll have to add your own controller, like this:

```
from odoo import http
from odoo.addons.shopfloor_mobile_base.controllers.main import (
    ShopfloorMobileAppMixin,
)


class ShopfloorMobileAppController(http.Controller, ShopfloorMobileAppMixin):
    @http.route(
        ["/shopfloor_mobile/app", "/shopfloor_mobile/app/<string:demo>"],
        auth="public",
    )
    def load_app(self, demo=False, **kw):
        # keep the old URL for backward compat
        # after v14 refactoring w/ shopfloor.app
        shopfloor_app = http.request.env.ref("my.new.shopfloor.app")
        return self._load_app(
            shopfloor_app, demo=True if demo else False, **kw
        )

```
I've chosen not to support this by default because the app can be created in several ways and there's no default app anymore.


shopfloor_base/models/shopfloor_app.py
```diff
@@ -0,0 +1,262 @@
...
+        compute_sudo=True,
+        help="""
+        Base route for endpoints attached to this app,
+        internal controller-ready version.
+        """,
+    )
+    url = fields.Char(compute="_compute_url", help="Public URL to use the app.")
+    api_docs_url = fields.Char(compute="_compute_url", help="Public URL for api docs.")
+    auth_type = fields.Selection(
+        selection="_selection_auth_type", default="user_endpoint"
```
jbaudoux, line 62: The default value needs the module shopfloor_mobile_base_auth_user that is not a dependency

shopfloor_base/models/shopfloor_app.py
```diff
@@ -0,0 +1,262 @@
...
+            routes = rec._registered_routes()
+            rec.registered_routes = "\n".join(sorted([x[1].rule for x in routes]))
+
+    @api.depends("profile_ids")
+    def _compute_profile_required(self):
+        for rec in self:
+            rec.profile_required = bool(rec.profile_ids)
+
+    def _selection_auth_type(self):
+        return self.env["endpoint.route.handler"]._selection_auth_type()
```
jbaudoux, line 115: user_endoint needs the module shopfloor_mobile_base_auth_user
Either it should then be in the dependencies or that selection option should be added by that module.

jbaudoux (changes requested):

shopfloor_base/models/shopfloor_app.py
jbaudoux: Even with the module installed, I still get:
[Vue warn]: Error in created hook: "Error: Auth type 'user_endpoint' not supported"

shopfloor_base/models/shopfloor_app.py
simahawk: no, this comes from `endpoint_route_handler` the way you implement it in the front is up to you.

shopfloor_base/models/shopfloor_app.py
simahawk: I think this is solved now.

sebalix:
@simahawk I pushed your patch in https://github.com/OCA/wms/pull/316/commits/aa64e673e4318324a9b709a8faac5ecb38d0ad79 (with a fix in the query parameter to make it a tuple)

sebalix (approved):
Tested some Shopfloor scenario locally :+1:

shopfloor_base/models/shopfloor_app.py
```diff
@@ -0,0 +1,265 @@
...
+    def action_open_app_docs(self):
+        return {
+            "type": "ir.actions.act_url",
+            "name": self.name,
+            "url": self.api_docs_url,
+            "target": "new",
+        }
+
+    def action_view_menu_items(self):
+        action = self.env.ref("shopfloor_base.action_shopfloor_menu").read()[0]
```
sebalix:
```suggestion
        action = self.env["ir.actions.act_window"]._for_xml_id("shopfloor_base.action_shopfloor_menu")
```

shopfloor_mobile_base_auth_api_key/models/shopfloor_app.py
```diff
@@ -0,0 +1,30 @@
...
+        string="Allowed API key groups",
+    )
+
+    def _allowed_api_key_ids(self):
+        return self.auth_api_key_group_ids.auth_api_key_ids.ids
+
+    def action_manage_api_key_groups(self):
+        action = self.env.ref(
+            "auth_api_key_group.auth_api_key_group_act_window"
+        ).read()[0]
```
sebalix:
```suggestion
        action = self.env["ir.actions.act_window"]._for_xml_id("auth_api_key_group.auth_api_key_group_act_window")
```

sebalix (dismissed):

sebalix:
Rebased to resolve conflicts on `shopfloor` and `shopfloor_base` manifest files

sebalix (approved):
But Travis is still unhappy regarding tests...

shopfloor_base/models/shopfloor_app.py
```diff
@@ -0,0 +1,268 @@
...
+        compute_sudo=True,
+        help="""
+        Base route for endpoints attached to this app,
+        internal controller-ready version.
+        """,
+    )
+    url = fields.Char(compute="_compute_url", help="Public URL to use the app.")
+    api_docs_url = fields.Char(compute="_compute_url", help="Public URL for api docs.")
+    auth_type = fields.Selection(
+        selection="_selection_auth_type", default="user_endpoint"
```
jbaudoux, line 62: Sad you cannot enable multiple auth types

jbaudoux (approved):

simahawk:
I've restored the commits that got lost in the last few force pushes. That's why it was failing.

PR #318, opened 2021-11-03
[14.0] shopfloor: add module icon

petrus-v:


simahawk (changes requested):
thanks for your contrib.
Sorry, but this doesn't belong to `shopfloor_base`.
The WMS app stays in `shopfloor`. Can you move it there?

Note that in the future we'll be able to have different apps based on shopfloor_base (WIP on https://github.com/OCA/wms/pull/316) and I'll probably move icons at app config level.

Which means: we should find a generic icon for the shopfloor_base menu -> If you'd like to submit proposals, that would be great!

petrus-v:
thanks for your feedback, (i m discovering those set of shopfloor modules) for you, what would means generic as shopfloor_base as menu icon ?

A scanner 
image

a warehouse
image

a config device
image

something else ?

simahawk:
`shopfloor_base` allows you to create mobile apps w/ support scanner feature. IMO it recall the concept of "mobile" (phone and or tablet) and "scanner".

`shopfloor` applies this to the WMS context -> should have something to deal w/ WH, packaging or stuff like that.

petrus-v:
@simahawk thanks for details, so I've moved this icon app to shopfloor module.

Can I propose an icon for shopfloor_base based on the device and the gear proposed above in this thread? in an other PR

petrus-v:
I've open #319 with an icon proposal with a "generic" icon, I think it would be proper to carry on the discuss there and keep this thread regarding the shopfloor icon proposal.

In fact i'm not sure if it's relevant to change the shopfloor_base icon in this shopfloor module :thinking: probably better to wait #316 to add properly set the app icon on the declared shopfloor.app if needs. At least remove the menu icon overwrite !

simahawk:
> I've open #319 with an icon proposal with a "generic" icon, I think it would be proper to carry on the discuss there and keep this thread regarding the shopfloor icon proposal.
> 
> In fact i'm not sure if it's relevant to change the shopfloor_base icon in this shopfloor module thinking probably better to wait #316 to add properly set the app icon on the declared shopfloor.app if needs. At least remove the menu icon overwrite !

We'll need an "Odoo app" icon for the main menu anyway. Then we'll have shopfloor.app specific icons to be used for end users' devices (eg: what you'll see after installing the app or in the favicon).

Hence, we should probably have a specific main menu entry for the Shopfloor WMS app. The WMS menu could then show items from apps of a given type (which tells me I should add a "category" to the app).

Let's drop the menu item in shopfloor_base and just add the icon which can be used later (we can use it already in the manifest of the mobile module).

Then in shopfloor we add the menu item but we use the actions from the base module.
All in all, today is the only real app implemented. Then if you implement another kind of app, you'll have to add your own main menu entry.

Sounds good?

mileo:
Can you add an icon to the main menu app to?

hparfr (approved):

PR #246, opened 2021-07-12
Add shopfloor_delivery_shipment_mobile

TDu:
- [x] Depends on https://github.com/OCA/wms/pull/245

This module adds the front end part for a new scenario
delivery shipment

PR #321, opened 2021-11-03
[14.0][FW] shopfloor_mobile_base: detail card improvement 

sebalix:
Port of #246 from 13.0 to 14.0.

This commit is a port of a chunk from 14e674b with a different commit message.


simahawk:
@sebalix can we change the commit msg? It has nothing to do w/ `shopfloor_delivery_shipment_mobile`

sebalix:
@simahawk it was part of the commit in 13.0 branch when the delivery shipment has been developed. If we change the commit message, then it'll pop-up with `oca-port`, but we could still blacklist it afterwards... Ok to rename it then.

simahawk (approved):

TDu (approved):

PR #335, opened 2021-12-06
[14.0][IMP] delivery_carrier_preference: Extract test setup in common class

mmequignon:


delivery_carrier_preference/tests/__init__.py
```diff
@@ -1,3 +1,4 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from . import common
```
simahawk: :fire: this is not a test

mmequignon: So it shouldn't be referenced at all, even if imported elsewhere?

sebalix: There is no need to import `common` in `__init__.py` as it is a Python module without test. It will be imported by tests modules if needed.

mmequignon: done

simahawk (approved):

sebalix (approved):

PR #236, opened 2021-06-04
[WIP][13.0][IMP] shopfloor: new scenario 'manual product transfer'

sebalix:
New scenario allowing to move a product or a lot from a source location to another one (similar to the "location content transfer" but limited to a specific product/lot).

Requires:
- https://github.com/OCA/stock-logistics-workflow/pull/851
- https://github.com/odoo/odoo/pull/71726 (required by `stock_restrict_lot`)

TODO:
- [x] Backend
- [x] Frontend
- [ ] Put in a separate module

Ref. 2152

PR #337, opened 2021-12-09
[14.0][MIG] shopfloor_manual_product_transfer(_mobile)

sebalix:
Original PR which was blocked in 13.0 because of missing features (features that exist in 14.0):
- https://github.com/OCA/wms/pull/236

The scenario `Manual product transfer` has been extracted in its own addons:
- `shopfloor_manual_product_transfer`
- `shopfloor_manual_product_transfer_mobile`


shopfloor/models/stock_location.py
```diff
@@ -69,3 +69,10 @@ def planned_qty_in_location_is_empty(self, move_lines=None):
         planned = remaining - sum(move_lines.mapped(move_line_qty_field))
         compare = float_compare(planned, 0, precision_rounding=0.01)
         return compare <= 0
+
+    def should_bypass_reservation(self):
+        self.ensure_one()
+        # TODO: This could be set as an option on shopfloor.menu
```
simahawk: it should IMO. Not a blocker.

shopfloor_manual_product_transfer/services/manual_product_transfer.py
```diff
@@ -0,0 +1,668 @@
...
+        quants = self.env["stock.quant"].search(
+            [("location_id", "=", location.id), ("quantity", ">", 0)]
+        )
+        if not quants:
+            return self._response_for_start(
+                message=self.msg_store.location_empty(location)
+            )
+        return self._response_for_scan_product(location)
+
+    def _find_user_move_lines_domain(self, location, product, lot=None):
```
simahawk, line 125: can't we reuse the search move line action?

shopfloor_manual_product_transfer/services/manual_product_transfer.py
```diff
@@ -0,0 +1,668 @@
...
+    def confirm_quantity(
+        self,
+        location_id,
+        product_id,
+        quantity,
+        lot_id=None,
+        barcode=None,
+        confirm=False,
+    ):
+        """Confirm the quantity to move by scanning the product/lot a second
```
simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

simahawk (approved):

TDu (approved):
Thanks

shopfloor_manual_product_transfer/services/manual_product_transfer.py
sebalix: I checked the `_search_move_lines_by_location_domain` method, while some parts could be reused, we will have to delete/replace the following criteria from the returned domain:

- `("location_id", "child_of", locations.ids)`
- `("qty_done", "=", 0)`
- `("state", "in", ("assigned", "partially_available"))`
- `"|",
                ("shopfloor_user_id", "=", False),
                ("shopfloor_user_id", "=", self.env.uid),`

The only thing reusable at this stage is the domain on `product_id` and `lot_id`... so better to not reuse it IMO.
But I'll change the criteria `("state", "not in", ("cancel", "done"))`, I remember it makes things slower over time in the execution plan of PostgreSQL, and it's better to filter on `confirmed`, `assigned` and `partially_available` states as there are less records compared to `cancel`+`done`.
EDIT: I found a lot of `state not in cancel/done` domains in other scenarios, and we didn't see any performance issue until now, so I'll keep it like this.

shopfloor/models/stock_location.py
sebalix: Done

shopfloor_manual_product_transfer/services/manual_product_transfer.py
sebalix: fixed

PR #342, opened 2021-12-21
[14.0][FIX] shopfloor_mobile: Display location as TODO as long as the artic…

lmignon:
…le is not scanned

Before this change, the location into the screen displayed at the start of a line processing was always marked as done. With this change, the location is marked as done only when the location \ product \ lot has been scanned.


simahawk (approved):
Seems legit

jbaudoux (approved):

PR #345, opened 2021-12-28
[13.0] Shopfloor: Control the display of the virtual keyboard (Gboard) + SECURITY FIX

jbaudoux:
## Manage virtual keyboard
On an android device having no physical keyboard, the virtual keyboard (Gboard) is always displayed by default when an input field has focus which is what we have on each screen when something must be scanned. This is quite annoying as it hides most of the screen info on small mobile devices.

This change hides by default the virtual keyboard while keeping focus to collect the scanned input. The field shows a glowing light red to indicate that it is ready to collect a scanned input.
When clicking on the input field already having the focus, this will toggle the virtual keyboard and a cursor in the input field. You can still scan the input. If the input field lost the focus and you click on it, it will only acquire the focus for scanning and show back the glowing light red to indicate is it back ready to collect data.

## Login Security fix
Set input type="password" for the login input field to prevent Gboard to automatically add the api key in the location dictionary suggestions

@simahawk @lmignon @sebalix @hparfr

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
+++ b/shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
@@ -4,6 +4,8 @@
  * @author Raphaël Reverdy <raphael.reverdy@akretion.com>
  * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
  * @author Simone Orsi <simahawk@gmail.com>
+ * Copyright 2021 BCIM (http://www.bcim.be)
+ * @author Jacques-Etienne Baudoux <je@bcim.be>
  * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
  */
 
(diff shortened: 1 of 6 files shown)


jbaudoux:
Here is a preview. The real virtual keyboard is displayed as a white square in the video as it is not part of the browser but on top of it. 
VirtualKeyboard

jbaudoux:
ShopfloorLogin

I removed the "(API key)" from the placeholder and put it back above as label as it was before for a better display on xs screen.

sebalix (approved):

shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
```diff
@@ -35,9 +45,18 @@ Vue.component("searchbar", {
         },
     },
     mounted: function() {
-        this.$root.event_hub.$on("screen:reload", this.on_screen_reload);
+        // As the inputMode is set to none when inserted in the DOM, we need to force the focus
+        if (this.autofocus) this.$refs.searchbar.focus();
```
simahawk, line 49: genuine question: why not `this.autofocus()`?

simahawk (approved):

PR #346, opened 2021-12-29
[13.0] shopfloor: fix computation of contained packagings

jbaudoux:
If a packaging is not a multiple of the next packaging, the displayed info was wrong.
In this example, a vendor pallet has 54 TU and a stored pallet has 42 TU.
The vendor pallet was computed as 1 Pallet which is wrong.
The new implementation looks to further packagings until it finds a multiple.

## Before
QtyByPkg-before

## After
QtyByPkg-after

cc @simahawk @sebalix @lmignon @jgrandguillaume

### DIFF
--- a/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
+++ b/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
@@ -1,6 +1,7 @@
 /**
  * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
  * @author Simone Orsi <simahawk@gmail.com>
+ * Copyright 2021 Jacques-Etienne Baudoux (BCIM)
  * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
  */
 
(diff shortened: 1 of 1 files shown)


shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -194,14 +195,23 @@ export var PackagingQtyPickerMixin = {
...
+                qty_per_pkg,
+                remaining,
+                elected_next_pkg;
             const packaging = this.sorted_packaging;
             _.forEach(packaging, function(pkg, i) {
-                if (packaging[i + 1]) {
-                    const next_pkg = packaging[i + 1];
+                const next_pkgs = packaging.slice(i + 1);
+                remaining = undefined;
+                _.every(next_pkgs, function(next_pkg) {
```
simahawk, line 206: I don't get the usage of `_.every`. It should be used to verify a condition, not too loop on elements and set a var name. No? Am I missing anything?

jbaudoux: It is also a way to loop and break on a condition

simahawk: you mean an hack? :smile: 
You could achieve the same w/ any other looping func, no?

simahawk (approved):

PR #347, opened 2021-12-29
[13.0] shopfloor: refactor design of packaging qty picker

jbaudoux:
New design of the quantity display and picker

* Takes less space on the screen
* The total quantity is now also editable
* Total done quantity is colored in comparison with the initial quantity to do (orangered more, pink less).
* Packaging detail can be collapsed
* Open a number virtual keyboard on editing values (instead of a text keyboard)
* Auto empty packaging quantity input when edited, no need to empty the old value before inserting the new one
* Added readonly version that shows only packaging having a value.
* tech vuejs: removed the init and use of watch

Moreover, scenarios have been updated to hide the move line quantity in the product detail when the quantity picker is also visible. 

The cluster picking now shows the picker from the beginning of the scenario. Before, you had to scan the source location/product on the first step before seeing the quantity block.

| New look | Before |
| ------------- | ------------- |
| PickerNew | PickerBefore |

@simahawk @sebalix @lmignon @jgrandguillaume @hparfr

### DIFF
--- a/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
+++ b/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
@@ -2,6 +2,7 @@
  * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
  * @author Simone Orsi <simahawk@gmail.com>
  * Copyright 2021 Jacques-Etienne Baudoux (BCIM)
+ * @author Jacques-Etienne Baudoux <je@bcim.be>
  * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
  */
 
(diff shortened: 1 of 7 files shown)


jbaudoux:
## Cluster Picking

ClusterPickingNewPicker

sebalix (approved):

lmignon (approved):

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -212,67 +182,95 @@ export var PackagingQtyPickerMixin = {
...
-                <div v-if="contained_packaging[pkg.id]" class="pkg-qty">({{ contained_packaging[pkg.id].qty }} {{ contained_packaging[pkg.id].pkg.name }})</div>
-            </div>
-        </v-col>
-    </v-row>
+                    </v-col>
+                    <v-col cols="3" md="2" :class="readonly ? 'd-none' : ''">
+                        <span class="qty-todo">/ {{ qty_todo }}</span>
+                    </v-col>
+                    <v-col>
+                        Pieces
```
jbaudoux: The unit of measure should be displayed here instead

jbaudoux (commented):
As the total quantity is now editable, the packaging picker should not be displayed if the product has no packaging otherwise we see twice the same info

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
jbaudoux: Fixed. It is now displaying the UOM and the packagings are not displayed if they are not defined on the product

JuMiSanAr (approved):

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -212,67 +182,95 @@ export var PackagingQtyPickerMixin = {
...
     updated: function() {
-        this.$root.trigger("qty_edit", this.value);
+        // Put in updated to prevent triggerring when the component is initialized
+        this.$root.trigger("qty_edit", this.qty_done);
+    },
+    computed: {
+        qty_color: function() {
+            if (this.qty_done == this.qty_todo) {
+                if (this.readonly) return "";
+                return "background-color: rgb(143, 191, 68)";
```
simahawk, line 222: would be better to toggle a class

simahawk:
you can fix the class issue later

/ocabot merge minor

Thanks for this nice work! :)

simahawk:
ah, sorry, I blocked the build as the other PR is still merging

simahawk:
@jbaudoux pls rebase

jbaudoux:
@simahawk rebase done

PR #348, opened 2022-01-04
[13.0][IMP] shopfloor_delivery_shipment: restrict dock to current WH

sebalix:
Depends on:
- https://github.com/OCA/stock-logistics-transport/pull/76

Ref. 3051


mmequignon (approved):
LG

PR #355, opened 2022-01-11
[14.0][IMP] stock_storage_type: Allow to restrict some putaway sequence

### LINKED ISSUE #343: stock_storage_type: allow to restrict some putaway sequence
### Use case 1: Choose destination according to demand on the location
In the pack put-away sequence, I have 2 rules:
* Stock <= I want to add here a condition to only send in Stock if there is a demand in Stock
* Over-stock

When we store a product after a reception, I have to prevent to consume to much pickable locations in Stock. If the location storage type is configured on "only empty" or "do not mix lot", you don't want to send the operator to a empty location in Stock if there is already enough stock. And you don't want to send all in over-stock if there is a demand in Stock.

### Use case 2: Choose destination according to stock (existing lot) on the location
* Stock <= I want to add here a condition to only send in Stock if the lot is already in stock
* Over-stock

When we store a product after reception, I have to prevent to consume to much locations in over-stock if I can fill a location in Stock with additional product. In this case, even if there is no qualified demand for Stock, I still want to put-away the product in Stock if there is enough space in the current location. This is in case of "do not mix lot" and slow rotating products (class C). If the lot is already in stock on a location, I want first to fill that location and not consume a new over-stock location.

## Proposal
Add a model to express a stock storage location sequence *condition* as an eval python code (like sale exception).
In the "stock.storage.location.sequence", add an optional condition field (m2m_tags).
In the loop evaluating each pack putaway sequence (https://github.com/OCA/wms/blob/14.0/stock_storage_type/models/stock_location.py#L359), check if all the conditions are met to elect the destination location.

Side note: To support fixed locations put-away, if the put-away strategy is None, we also need to compute the odoo standard put-away on that destination.

cc @lmignon @sebalix @jgrandguillaume

lmignon:
fixes #343


lmignon:
ping @jbaudoux

stock_storage_type/models/stock_storage_location_sequence_cond.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def evaluate(self, storage_location_sequence, putaway_location, quant, product):
+        self.ensure_one()
+        if self.condition_type == "code":
+            return self._exec_code(
+                storage_location_sequence, putaway_location, quant, product
+            )
+        condition_type = self.condition_type
+        raise exceptions.UserError(
+            _(f"Not able to evaluate condtion of type {condition_type}")
```
jbaudoux:
```suggestion
            _(f"Not able to evaluate condition of type {condition_type}")
```

jbaudoux (approved):
LGTM. Thanks for the implementation.

Do you make a separate PR for the last remark on the issue?
_To support fixed locations put-away, if the put-away strategy is None, we also need to compute the odoo standard put-away on that destination._

jbaudoux:
cc @sebalix @jgrandguillaume

sebalix (approved):
Code review and functional need seems legit so :+1:

jbaudoux:
@sebalix Can it be merged?

sebalix:
@lmignon can you rebase this PR?

lmignon:
@sebalix done

sebalix:
A check is failing because of another module
```sale_stock_available_to_promise_release_cutoff (Beta) depends on sale_delivery_date (Alpha)```
We need to get merged https://github.com/OCA/sale-workflow/pull/1968/ and update it to `Beta` as it is used since months now and is working as expected.

PR #356, opened 2022-01-12
[14.0] apply product putaway for strategy none

lmignon:
If the put-away strategy is None, we also need to compute the odoo standard put-away on that destination.

refs  #343 

ping @jbaudoux  @jgrandguillaume @sebalix

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -380,10 +380,36 @@ class StockLocation(models.Model):
         locations = self.browse()
         if self.pack_putaway_strategy == "none":
             locations = self
+            if products and len(products) == 1:
+                locations = self._get_product_putaway(products) or self
         else:
             products = products or self.env["product.product"]
             locations = self._get_sorted_leaf_child_locations(products)
         return locations
+
+    def _get_product_putaway(self, product):
+        """Returns the direct location where the product has to be put if
+        found Otherwise returns an empty recordset.."""
+        self.ensure_one()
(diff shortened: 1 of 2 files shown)


stock_storage_type/models/stock_location.py
```diff
@@ -357,18 +357,6 @@ def _get_pack_putaway_strategy(self, putaway_location, quant, product):
...
-                and pref_loc.select_allowed_locations(
-                    package_storage_type, quant, product
-                )
-            ):
-                _logger.debug(
-                    "No putaway strategy defined on location %s and package "
-                    "storage type %s allowed."
-                    % (pref_loc.complete_name, package_storage_type.name)
-                )
-                return pref_loc
```
jbaudoux, line 371: After reviewing this part of code with Laurent, we could not find any use as this is the same as what is happening after the if

jbaudoux (approved):
LGTM

jbaudoux:
This PR is to support Dynamic shelving (or flow racking) where you use the package storage type to put-away to the right area of the warehouse and then use the product put-away to specify which final location in the area.

This allows you to define different put-away sequences with different strategies for a same package and provide a product put-away location in the area where the strategy is none, while letting the storage type compute a final location where the strategy is children location or abc.
Before this PR, you had to define the product put-away on a top level location causing the package storage type put-away to be completely skipped.

sebalix (approved):
I'm still not very familiar with this addon but the use case is well explained and code LG :+1:

PR #357, opened 2022-01-12
[14.0][[IMP] Add translations FR/EN to scenarii cluster_picking and location…

lmignon:
…_content_transfer

This PR also add an helper method to load translations from module that extends `shopfloor_mobile_base`

ping @simahawk


shopfloor_mobile_base/static/wms/src/services/translation_registry.js
```diff
@@ -3,6 +3,17 @@
  * @author Simone Orsi <simahawk@gmail.com>
  * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
  */
+export function loadTranslation(callback, url) {
```
simahawk: Not really limited to translations, it could be moved elsewhere too. I'd leave it here but change the name:
```suggestion
export function loadJSON(callback, url) {
```

shopfloor_mobile_base/static/wms/src/services/translation_registry.js
```diff
@@ -36,6 +47,15 @@ export class TranslationRegistry {
     get_default_lang() {
         return this._default_lang;
     }
+
+    load(lang, path) {
+        loadTranslation((json) => {
+            const messages = JSON.parse(json);
```
simahawk: I'd call `JSON.parse` in `loadJSON`

simahawk (changes requested):
@lmignon tnx for your addition. Can you please split the commits:

```
* shopfloor_mobile_base: add translation_registry.load

Allow to load translations from JSON files.

* shopfloor_mobile: add missing translations
```

Side note: from a certain point of view it would be better to keep translation in the same place where they are used (like [here](https://github.com/OCA/wms/blob/14.0/shopfloor_mobile_base_auth_api_key/static/wms/src/login.js#L75)).
We'd load one file less too.
OTOH for translators it's also handy to have them all in one place.

lmignon:
> 
> Side note: from a certain point of view it would be better to keep translation in the same place where they are used (like [here](https://github.com/OCA/wms/blob/14.0/shopfloor_mobile_base_auth_api_key/static/wms/src/login.js#L75)). We'd load one file less too. OTOH for translators it's also handy to have them all in one place.

@simahawk with this approach we can load translation from files generated by [webpack-i18n-generate-json](http://vuejscomponent.com/webpack-i18n-generate-json)

simahawk:
> > Side note: from a certain point of view it would be better to keep translation in the same place where they are used (like [here](https://github.com/OCA/wms/blob/14.0/shopfloor_mobile_base_auth_api_key/static/wms/src/login.js#L75)). We'd load one file less too. OTOH for translators it's also handy to have them all in one place.
> 
> @simahawk with this approach we can load translation from files generated by [webpack-i18n-generate-json](http://vuejscomponent.com/webpack-i18n-generate-json)

That's great, I love it :+1:

simahawk (approved):

sebalix (approved):

PR #358, opened 2022-01-18
[14.0] Switch tests to GH actions

simahawk:


simahawk:
@sbidoul thanks for manifestoo! We didn't know about this problem :handshake: 

image

sbidoul:
Yep... wait until you run it on your private repos :)

sebalix:
Nice! We should set the AGPL-3 on `stock_reception_screen_qty_by_packaging` then.

sebalix (approved):
Pre-approving

mmequignon (approved):
pre-approving as well

simahawk:
base_jsonify https://github.com/OCA/server-tools/pull/2257
stock_reception_screen_qty_by_packaging will be changed in this PR

.t2d.yml
```diff
@@ -12,8 +12,8 @@ addons:
   apt:
     packages:
       - expect-dev # provides unbuffer utility
-      - cups
-      - libcups2-dev
+      - "cups"
+      - "libcups2-dev"
```
sbidoul, line 16: Note

These are part of the [oca/oca-ci Dockerfile](https://github.com/OCA/oca-ci/blob/9eff1d7235a52bde738b6ed1503e766a38f6dd70/Dockerfile#L71), as it is the only place we can declare *build* dependencies for third party python libs that don't publish manylinux wheels.

*runtime* deb dependencies, OTOH have to be declared in manifests, in a `deb` key of the `external_dependencies` section.

simahawk: @sbidoul it comes from a dependency on `base_report_to_printer` but that module declares only `pycups` as a dep indeed https://github.com/OCA/report-print-send/blob/14.0/base_report_to_printer/__manifest__.py#L32
Shall I update that?

simahawk:
https://github.com/OCA/stock-logistics-warehouse/pull/1352
https://github.com/OCA/wms/pull/361

simahawk:
https://github.com/OCA/stock-logistics-workflow/pull/955

simahawk:
https://github.com/OCA/product-attribute/pull/988

simahawk:
:green_circle: :man_dancing:

PR #360, opened 2022-01-18
[14.0][FIX] shopfloor_mobile_base: Ensure the right menu id is initialized …

### LINKED ISSUE #359: shopfloor_mobile_base: Wrong menu id when odoo is called into the init state on first click on a menu
## Module
shopfloor_mobile_base.

## Describe the bug
2 menus to the same scenario with an init state where odoo is called(ex location_content_transfer) with != pircking types
```js
states: {
                init: {
                    enter: () => {
                        this.wait_call(this.odoo.call("start_or_recover"));
                    },
                },
....
```
Click on the fist menu. When you click on the second menu, the call to odoo is made with the menu_id of the fist menu.

## To Reproduce
**13.0, 14.0**:

**Expected behavior**
Call to odoo into the init state should be made with the right menu id

**Additional context**
The `init` state is called by the router into [beforeRouteUpdate](https://github.com/OCA/wms/blob/14.0/shopfloor_mobile_base/static/wms/src/scenario/mixins.js#L37). At this stage, the menu_id is not yet changed to the one for the requested new route.

lmignon:
…when state is loaded

The init state is called by the  bedoreRouteUpdate method of the router. At this stage, the new route is not yet set into the contexte and therefore, the menu item id is still the one from the previous route. To ensure that the new menu id is properly set before the init state is entered, we force the menu id to the one of the new route. Call to odoo into the enter method of the init state will be made with the right menu id into the request headers

fixes #359

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/scenario/mixins.js
+++ b/shopfloor_mobile_base/static/wms/src/scenario/mixins.js
@@ -35,6 +35,11 @@ export var ScenarioBaseMixin = {
         this.menu_item_id = this._get_menu_item_id();
     },
     beforeRouteUpdate(to, from, next) {
+        // force the menu id to the one from the menu since at this stage
+        // the new route is not yet applied but a call to odoo could
+        // be done when entering into the init state and this call must be
+        // done with the right menu_item_id
+        this.menu_item_id = to.params.menu_id;
         // Load initial state
         this._state_load(to.params.state);
         next();


lmignon:
ping @jbaudoux @simahawk

shopfloor_mobile_base/static/wms/src/scenario/mixins.js
```diff
@@ -35,6 +35,11 @@ export var ScenarioBaseMixin = {
         this.menu_item_id = this._get_menu_item_id();
     },
     beforeRouteUpdate(to, from, next) {
+        // force the menu id to the one from the menu since at this stage
+        // the new route is not yet applied but a call to odoo could
+        // be done when entering into the init state and this call must be
+        // done with the right menu_item_id
+        this.menu_item_id = to.params.menu_id;
```
simahawk, line 42: Would be nice to improve `_get_menu_item_id` to receive a route (`to` in this case) and fallback to the current one if none is passed.
Not a blocker tho.

simahawk (approved):
Seems legit. Thanks for fixing it.

simahawk:
tip for rewriting the commit : `shopfloor_mobile_base: fix scenario state load` :wink:

PR #363, opened 2022-01-19
[14.0][MIG] stock_reception_screen_mrp_subcontracting

sebalix:


stock_reception_screen_mrp_subcontracting/models/stock_picking.py
```diff
@@ -0,0 +1,75 @@
...
+        # context key and we rewrite the business logic.
+        new_self = self
+        if any(picking._is_subcontract() for picking in self):
+            new_self = self.with_context(subcontracting_skip_action_done=True)
+        res = super(StockPicking, new_self)._action_done()
+        for picking in self.filtered(lambda p: p._is_subcontract()):
+            productions = picking.move_lines.move_orig_ids.production_id.filtered(
+                lambda p: p.state not in ("done", "cancel")
+            )[-1:]
+            for production in productions:
```
mmequignon, line 30: you should only have the last on of the recordset, right? If so, we don't need the `for` loop. Or am I missing smth?

mmequignon (approved):
Didn't get everything. Code looks good

stock_reception_screen_mrp_subcontracting/models/stock_picking.py
sebalix: Indeed, but I don't know if I should remove the for loop or the `[-1:]` (copied from Odoo std code), Odoo is acting as only one MO could be linked to a receipt, but I don't know what is blocking to have a receipt with 2 differents subcontracted product, with each of them having its MO... for this case we would need to loop.

To check.

sebalix:
Build is failing because of a conflict between `shopfloor` and `mrp_subcontracting` (reproduced locally with only these two addons installed), not related to this addon especially. Checking...

i-vyshnevska (approved):

simahawk:
@sebalix any news?

sebalix:
This PR is working fine on production so I'll merge it. But if we could get rid of this wobbly screen for further versions (and switch to something based on Shopfloor) I'm OK with that.

sebalix:
Something is broken... to check later

flachica:
Hello @sebalix. I can't see the migrations issue to 15.0. I'm working to migrate shopfloor to 15.0. Where is the correct Issue to say this?

PR #364, opened 2022-01-20
[14.0][FIX] fixing .gitignore lib

mt-software-de:
The line
lib/
in the file .gitignore
ignores all directories with the name lib
That is not useful because of the directory 
shopfloor_mobile_base/static/wms/src/lib

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@ build/
 develop-eggs/
 dist/
 eggs/
-lib/
 lib64/
 parts/
 sdist/


simahawk (changes requested):
thanks for your contrib, this should be changed here https://github.com/OCA/oca-addons-repo-template

mt-software-de:
> thanks for your contrib, this should be changed here https://github.com/OCA/oca-addons-repo-template

This is now done. See https://github.com/OCA/oca-addons-repo-template/pull/139
Should this pr also than be merged?

simahawk (approved):

PR #365, opened 2022-01-20
[13.0] shopfloor: refactor the packaging quantity picker - fixup!

jbaudoux:
When the component is initialized, the shown quantity must be propagated to the root component otherwise it doesn't know the validated quantity.

In the location content transfer, the first line works properly but when the second line is displayed, while the quantity is shown properly, the quantity of the previous line was send in the endpoint call. On refresh, the quantity was also lost.

cc @lmignon @simahawk @sebalix 

Already fixed the V14.0 migration PR

### DIFF
--- a/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
+++ b/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
@@ -211,8 +211,11 @@ export var PackagingQtyPicker = Vue.component("packaging-qty-picker", {
             },
         },
     },
+    created: function() {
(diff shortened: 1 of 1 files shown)


lmignon (approved):

simahawk (approved):

simahawk:
@jbaudoux I aborted the build. Please, amend your commit msg. Can't be a `!fixup` :)

Just say `shopfloor: fix quantity picker`, thanks.

jbaudoux:
@simahawk Done

PR #368, opened 2022-02-02
[14.0]Fix shopfloor edit picking carrier form

TDu:
There is an issue with the get url on the form although there is two url
options one with `/get` at the end and one without. Only the former works.
The one not working returns a 405.
This is why the call is changed in the js file.

Another problem for both url is their route are recorded with <int:id>
instead of <int:_id> wich is the default for a get route to work with
the `base_rest` module.
The workaround for this is to add a validator for the `id` parameter.

### DIFF
--- a/shopfloor/services/forms/picking_form.py
+++ b/shopfloor/services/forms/picking_form.py
@@ -46,7 +46,10 @@ class ShopfloorPickingFormValidator(Component):
     _usage = "form_edit_stock_picking.validator"
 
     def get(self):
-        return {}
+        return {
+            "id": {"type": "integer", "rename": "_id"},
+            "_id": {"type": "integer"},
+        }
 
     def update(self):
         return {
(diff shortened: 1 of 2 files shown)


shopfloor/services/forms/picking_form.py
```diff
@@ -46,7 +46,10 @@ class ShopfloorPickingFormValidator(Component):
     _usage = "form_edit_stock_picking.validator"
 
     def get(self):
-        return {}
+        return {
```
simahawk, line 49: I'd move this to the mixin.

simahawk (changes requested):
build is :red_circle:

shopfloor/services/forms/picking_form.py
TDu: When I moved this to the mixin I get a NotImplementedError and I don't know why ?

TDu:
In that fixup I tried to get travis green.
And also I removed the option to change the carrier if it is not set, because it breaks in the form after.

shopfloor/services/forms/picking_form.py
TDu: It is :green_apple: now

mmequignon (approved):

simahawk (approved):

PR #369, opened 2022-02-10
[14.0][FIX] stock_available_to_promise_release: priority propagation on release

TDu:
The management of priorities on stock moves and stock picking has changed
between on v14 see

* https://github.com/odoo/odoo/commit/e40553723b893bcd2244e9b9fcc5c1a7e1bf27d3

This make the priority field that was nicely flowing between stock move
through the procurement values, not working any more.

That change on Odoo core omits the inverse function on the priority
field in stock move, but it would not help when going through the
procurement mechanism, because moves are created before they are
assigned to picking.

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -371,6 +371,10 @@ class StockMove(models.Model):
             moves = moves.mapped("move_orig_ids")
         pickings = self.env["stock.picking"].browse(picking_ids)
         pickings._after_release_update_chain()
+        # Set the highest priority on all pickings in the chain
+        priorities = pickings.mapped("priority")
+        if priorities:
+            pickings.write({"priority": max(priorities)})
 
     def _after_release_assign_moves(self):
         moves = self
@@ -403,3 +407,9 @@ class StockMove(models.Model):
         new_move._assign_picking()
 
         return new_move.with_context(context)
+
+    def _assign_picking_post_process(self, new=False):
+        super()._assign_picking_post_process(new)
+        priorities = self.mapped("move_dest_ids.picking_id.priority")
+        if priorities:
+            self.picking_id.write({"priority": max(priorities)})
--- a/stock_available_to_promise_release/tests/test_reservation.py
+++ b/stock_available_to_promise_release/tests/test_reservation.py
@@ -936,3 +936,52 @@ class TestAvailableToPromiseRelease(PromiseReleaseCommonCase):
         # TODO a MTO picking should work normally
 
     # TODO: test w/ multiple orders by priority
+
+    def test_picking_priority(self):
+        self.pick_type = self.env.ref("stock.picking_type_out")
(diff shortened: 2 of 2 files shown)


stock_available_to_promise_release/models/stock_move.py
```diff
@@ -324,6 +324,7 @@ def _run_stock_rule(self):
                 backorder_links[new_move.picking_id] = move.picking_id
 
             values = move._prepare_procurement_values()
+
```
sebalix: New line to remove (to reduce the diff)

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -403,3 +409,10 @@ def _release_split(self, remaining_qty):
         new_move._assign_picking()
 
         return new_move.with_context(context)
+
+    def _assign_picking_post_process(self, new=False):
+        super()._assign_picking_post_process(new)
+        highest_priorities = self.mapped("move_dest_ids.picking_id.priority")
+        if highest_priorities:
+            highest_priorities.sort()
+            self.picking_id.write({"priority": highest_priorities[-1]})
```
mmequignon:
```suggestion
            self.picking_id.write({"priority": max(highest_priorities)})
```
I suppose… If you agree, you can apply the same everywhere else.

mmequignon (approved):

stock_available_to_promise_release/models/stock_move.py
mmequignon:
```python
>>> max(["1", "12", "42", "0"])
'42'
```

stock_available_to_promise_release/models/stock_move.py
TDu: Removed

stock_available_to_promise_release/models/stock_move.py
TDu: Thanks, I updated the pr

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -371,6 +371,10 @@ def _after_release_update_chain(self):
             moves = moves.mapped("move_orig_ids")
         pickings = self.env["stock.picking"].browse(picking_ids)
         pickings._after_release_update_chain()
+        # Set the highest priority on all pickings in the chain
+        highest_priorities = pickings.mapped("priority")
```
sebalix:
```suggestion
        priorities = pickings.mapped("priority")
```
The highest is then given by the call to `max`, not before :p

TDu: True, renamed.

sebalix: It remains one place where it should be renamed ;)

TDu: Should be good now.

simahawk:
@sebalix good for you?

sebalix (approved):
LG, would be good to have this in a dedicated method but not blocking.

PR #370, opened 2022-02-10
[14.0][FIX] shopfloor: fetch lot related to current product(s)

sebalix:
In Odoo a lot number can be shared by multiple lots as soon as the product is different.
This has to be taken into account by Shopfloor each time we scan a lot number by restricting the search to relevant products, or it could result unexpected result (a "Barcode not found" for instance).

**TODO:**
- [x] fix the delivery shipment scenario

Ref. 3261

### DIFF
--- a/shopfloor/actions/search.py
+++ b/shopfloor/actions/search.py
@@ -46,11 +46,17 @@ class SearchAction(Component):
             product = packaging.product_id
         return product
 
-    def lot_from_scan(self, barcode):
+    def lot_from_scan(self, barcode, products=None, limit=1):
         model = self.env["stock.production.lot"]
         if not barcode:
             return model.browse()
-        return model.search([("name", "=", barcode)], limit=1)
+        domain = [
+            ("company_id", "=", self.env.company.id),
+            ("name", "=", barcode),
+        ]
+        if products:
+            domain.append(("product_id", "in", products.ids))
+        return model.search(domain, limit=limit)
 
     def generic_packaging_from_scan(self, barcode):
         model = self.env["product.packaging"]
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -556,15 +556,17 @@ class ZonePicking(Component):
         message = None
         response = None
         search = self._actions_for("search")
-        lot = search.lot_from_scan(barcode)
-        if not lot:
+        # Could get several lots from different products, check each of them
+        lots = search.lot_from_scan(barcode, limit=None)
+        if not lots:
             return response, message
-        move_lines = self._find_location_move_lines(lot=lot)
-        if move_lines:
-            response = self._response_for_set_line_destination(first(move_lines))
-        else:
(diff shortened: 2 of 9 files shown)


shopfloor/actions/search.py
```diff
@@ -46,11 +46,17 @@ def product_from_scan(self, barcode):
...
         return product
 
-    def lot_from_scan(self, barcode):
+    def lot_from_scan(self, barcode, products=None, limit=1):
         model = self.env["stock.production.lot"]
         if not barcode:
             return model.browse()
-        return model.search([("name", "=", barcode)], limit=1)
+        domain = [
+            ("company_id", "=", self.env.company.id),
```
sebalix, line 54: I don't know if the company criteria if it's required or not: maybe already handled by `search`, to check

shopfloor/services/zone_picking.py
```diff
@@ -556,15 +556,17 @@ def _scan_source_lot(self, barcode, confirmation=False):
         message = None
         response = None
         search = self._actions_for("search")
-        lot = search.lot_from_scan(barcode)
-        if not lot:
+        # Could get several lots from different products, check each of them
+        lots = search.lot_from_scan(barcode, limit=None)
```
mmequignon, line 560: why do you specify the limit, as it defaults as None already?

mmequignon: Or you could pass 1 and remove the loop to restore the previous behavior.
I have no idea if it could break smth…

mmequignon (approved):
LG, one nitpick

shopfloor/services/zone_picking.py
sebalix: That's the goal: we have no product to filter lots here, so we could find several lots of different products.
From there we iterate on each lot to search for existing move line linked to this lot:

- if there is one, we return the next screen with `_response_for_set_line_destination()` (success)
- if not, we check the next lot
- once all lots have been checked and no move line has been found, we return `_list_move_lines()` with an error message "Lot not found".

Like before in fact, but before we were expecting one lot corresponding to our barcode... which is not true.

sebalix: And the default value for `limit` here is `1` on purpose (it represents the majority of our cases, to get the lot corresponding to the current product, there is only one case where we need to check all lots - this one):
```def lot_from_scan(self, barcode, products=None, limit=1):```

shopfloor/actions/search.py
sebalix: Checked: it is

simahawk (approved):
can you squash?

shopfloor/actions/search.py
sebalix, line 54: I don't know what is the best option: filtering on current user's company or on the picking type company, I would favor the second...

simahawk:
@sebalix any update?

sebalix:
Update: it is running well now.

/ocabot merge patch

PR #372, opened 2022-02-14
[14.0][FIX] s_storage_type: fix computation of 'max_height_in_m'

sebalix:
The computation of `max_height_in_m` depends also on `length_uom_id` field.

Updating `lengh_uom_id` with a new UoM like `mm` doesn't recompute the max height in meters, making the computation of the putaway allowing wrong locations (e.g. a pallet of 1.1m in a location of 0.8m maximum).

Ref. 3263

### DIFF
--- a/stock_storage_type/models/stock_location_storage_type.py
+++ b/stock_storage_type/models/stock_location_storage_type.py
@@ -174,7 +174,7 @@ class StockLocationStorageType(models.Model):
                 ]
             )
 
-    @api.depends("max_height")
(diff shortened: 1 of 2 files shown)


ivantodorovich (approved):

ivantodorovich:
It's probably a good idea to add an unit test for this case

sebalix:
@ivantodorovich test added

sebalix:
Fast track it
/ocabot merge patch

PR #339, opened 2021-12-13
[14.0][FIX]  shopfloor_packing_info is not an application

andreampiovesana:
fix manifest, is not an application

PR #338, opened 2021-12-13
[14.0][FIX] shopfloor_base is not an application

andreampiovesana:
fix manifest, is not an application

PR #319, opened 2021-11-03
[14.0] shopfloor_base: add app menu icon

petrus-v:

PR #373, opened 2022-02-15
[14.0][FIX] Fix 'application' flag for some Shopfloor addons + Add icon

sebalix:
I'm wondering if addons like:

- `shopfloor_delivery_shipment`
- `shopfloor_manual_product_transfer`

should be considered as applications or not? And if they should be applications, it would be probably better to set the `_mobile` ones as applications as the ones listed above are useless from a user POV without the UI...

ATM `shopfloor_delivery_shipment` isn't an app while `shopfloor_manual_product_transfer` is, it's not consistent.

NOTE: I cherry-picked commits done by @andreampiovesana from #338 and #339 and the icon from @petrus-v in #319 + change the commit messages.

ping @simahawk


simahawk:
Ok, I needed a bit of time to think about this story.
With #316 coming soon I'm not convinced anymore that shopfloor_base shouldn't be an app.
It's an app that allows you to define mobile apps. To complete this we need an icon as per #319 (I was waiting for the svg icon, but we can already merge it probably. I have to look again at it).

Regarding all the other modules: they are satellites, they shouldn't be "apps".
One special case could be `shopfloor` which is WMS specific (and for this reason I'd like to rename it to `shopfloor_wms` in the future). This is a very specific full scope implementation of shopfloor app and it probably deserves its own menu in the home (not 100% sure tho).

sebalix:
So for now this PR is good enough, no need to do more.

simahawk (approved):

PR #374, opened 2022-02-15
[14.0][FIX] shopfloor_mobile_base: fix state events binding

sebalix:
Some buttons were not working at all because of this.

Ref. 3313


simahawk (changes requested):
cannot work here :)
It's already on #316

PR #375, opened 2022-02-16
[14.0] shopfloor_base: add svg icon

petrus-v:
as mentioned here: https://github.com/OCA/wms/pull/319#issuecomment-1040430137 convert the PNG image to SVG


petrus-v:
@simahawk @sebalix you give me the energie to use inskape to convert the image to SVG ! I'haven't test in odoo this PR using this svg image should probably test with the oca responsive module I suppose before merging !

petrus-v:
also I'm not sure how to properly reference in app kanban view

image

(I'm loving new runboat, so fast :heart: )

I'should probably let the .png as well ?

simahawk:
SVG should be there but not used in the menu AFAIK.
I can test.

simahawk:
I confirm. We need png in the menu. Having the SVG is still handy when we need to edit the image or scale it.

simahawk (changes requested):
restore png

pedrobaeza:
Well, Odoo is doing the reverse move, finally using svg for menus.

petrus-v:
> Well, Odoo is doing the reverse move, finally using svg for menus.

Thanks for update. As this is true in master branch, while this PR is addressed to odoo v14.0 I would arg to keep the .png in menu entry. As both file will be present in the repo after merging this branch moving to svg file will be easy in v16.0 !

petrus-v:
> restore png

I've regenerate the png from the svg file and squashed my commits to avoid extra file in the history (first one was 400px per 400px)...

petrus-v:
Just realized that the shadow should be to the bottom left (not right)...

likes other icons:
* https://github.com/odoo/odoo/blob/14.0/addons/website/static/description/icon.png
* https://github.com/odoo/odoo/blob/14.0/addons/purchase/static/description/icon.png
* https://github.com/odoo/odoo/blob/14.0/addons/sale/static/description/icon.png
* ...

pedrobaeza:
You can check this icon generator:

https://github.com/OCA/maintainer-tools/pull/491

although SVG is a bit more tricky

petrus-v:
here you are, inspired from odoo icon I've also add another shadow about 5px (in 400x400 image) to the bottom

simahawk (approved):
LG, thanks!

flotho (approved):
nice Icon ;-)

PR #379, opened 2022-03-09
[14.0] Improve shopfloor line priority search

TDu:
Yes in Odoo core since v14 there is only "0" and "1" priority, but for some
implementation this may change.
The fact that all priority higher than zero should be included seems
logical.

For reference here is the implementation on v13

https://github.com/OCA/wms/blob/9e5e9c97c85f905868221ad47c2d97577ffdfe76/shopfloor/actions/move_line_search.py#L105

### DIFF
--- a/shopfloor/actions/move_line_search.py
+++ b/shopfloor/actions/move_line_search.py
@@ -104,7 +104,7 @@ class MoveLineSearch(Component):
 
     def counters_for_lines(self, lines):
         # Not using mapped/filtered to support simple lists and generators
-        priority_lines = [x for x in lines if x.picking_id.priority == "1"]
+        priority_lines = [x for x in lines if int(x.picking_id.priority or "0") > 0]
         return {
             "lines_count": len(lines),
             "picking_count": len({x.picking_id.id for x in lines}),


shopfloor/actions/move_line_search.py
```diff
@@ -104,7 +104,9 @@ def _sort_key_move_lines(order):
 
     def counters_for_lines(self, lines):
         # Not using mapped/filtered to support simple lists and generators
-        priority_lines = [x for x in lines if x.picking_id.priority == "1"]
+        priority_lines = [
+            x for x in lines if ord((x.picking_id.priority or "0")[0]) > 48
+        ]
```
simahawk:
```suggestion
        priority_lines = [
            x for x in lines if int(x.picking_id.priority or "0") > 0
        ]
```
isn't this more clear?

TDu: It is, but less resilient if someone decides to create a priority with the first char not being a number.

mmequignon: just compare the strings then ? They are ordered:
```python
>>> "1" > "0"
True
>>> "a" > "0"
True
```
```python
        priority_lines = [
            x for x in lines if (x.picking_id.priority or "0") > "0"
        ]

sebalix: @TDu I'm not sure it's something that will happen, it is expected to have a number here, at least a negative one but `int` will still handle it.

TDu: Updated the code with @simahawk solution.

simahawk (approved):

sebalix (approved):

PR #386, opened 2022-04-07
[14.0] shopfloor_mobile_base: add event on drawer transition

JuMiSanAr:
The motivation for this PR is to have a way to identify whether the navigation drawer is open or closed in the app, and to apply any logic on transition.

To do so:
- A dynamic class is added to the body.
- An event is triggered.


simahawk (approved):
LG, tnx

PR #387, opened 2022-04-07
shopfloor_mobile_base: make home customizable

simahawk:


JuMiSanAr:
Updated so that the event on successful login now redirects to the page set as home.

shopfloor_mobile_base/static/wms/src/router.js
```diff
@@ -32,9 +26,16 @@ const routes = [
     // TODO Fix this it needs to be the last route, but I think it is not anymore with the dynamic one added.
     // { path: '*', component: NotFound },
 ];
+const default_home_route = {
+    path: "/",
+    component: HomePage,
+    name: "home",
+    meta: {requiresAuth: true, requiresProfile: true},
```
simahawk: @JuMiSanAr you must set `isHomeRoute` here otherwise your filter on login won't work

JuMiSanAr (approved):

simahawk:
this is not over: we have to take care of all the places where the route "home" is used directly. EG: https://github.com/OCA/wms/blob/f4a3d433ee65d279726902eb762300cbd0ffbba0/shopfloor_mobile_base/static/wms/src/settings/profile.js#L54

simahawk:
> this is not over: we have to take care of all the places where the route "home" is used directly. EG:
> 
> https://github.com/OCA/wms/blob/f4a3d433ee65d279726902eb762300cbd0ffbba0/shopfloor_mobile_base/static/wms/src/settings/profile.js#L54

fixed

JuMiSanAr:
Ended up using a redirect in the router so that whenever the user lands on "/", they're redirected to the home route of their choice.

PR #388, opened 2022-04-08
Add default shopfloor profile on res users

bguillot:
Hello,

This PR adds a default profile mechanism for shopfloor.

It adds a field on res_users to store the default profile.
And select it automaticaly if filled.

There is a config parameter to select automaticaly the last profile as default.

It allows 3 ways of working :
- no automation : ir.config.parameter to False and don't set default profile manualy on user. The user will have to select his profile every time he logged in (it is currently working like this)
- force default profile : ir.config.parameter to False and set a profile manualy on user : the user will have the same profile by default at the login
- force last selected profile : ir.config.parameter to True : the user will have the last selected profile automaticaly at the login

### DIFF
--- a/shopfloor_base/data/config_parameter.xml
+++ b/shopfloor_base/data/config_parameter.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data noupdate="1">
+        <record
+            id="shopfloor_user_default_profile"
+            model="ir.config_parameter"
+            forcecreate="True"
+        >
+            <field name="key">shopfloor.default.profile</field>
+            <field name="value">False</field>
+        </record>
+    </data>
+</odoo>
--- a/shopfloor_mobile_base/static/wms/src/main.js
+++ b/shopfloor_mobile_base/static/wms/src/main.js
@@ -118,6 +118,8 @@ new Vue({
         });
         event_hub.$on("profile:selected", function (profile) {
             self.profile = profile;
+            const odoo = self.getOdoo({usage: "profile"});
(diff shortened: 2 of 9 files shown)


shopfloor_base/data/config_parameter.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
simahawk, line 2: Now we have shopfloor.app model. This should be configure per app IMO.

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -92,6 +92,8 @@ new Vue({
         });
         event_hub.$on("profile:selected", function (profile) {
             self.profile = profile;
+            const odoo = self.getOdoo({usage: "profile"});
```
simahawk, line 121: We shouldn't do any call if not required IMO.
You could check for a flag on `shopfloor_app_info`.

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -128,6 +130,14 @@ new Vue({
             ];
         },
         has_profile: function () {
+            if (_.isEmpty(this.profile) && this.authenticated) {
```
simahawk, line 141: you are setting the profile in a function that is meant to do a check...

simahawk (changes requested):
I would move this feature to a separated module `shopfloor_base_default_profile`.
This is now much easier. You can hook to the event on the event_hub from outside.

brian10048:
@bguillot Thank you for your work on this feature.  I was looking to have the same option.

Are you available to address the review comments?

PR #390, opened 2022-04-08
[14-0] shopfloor_mobile_base: fix load method in translation registry

JuMiSanAr:
This method now allows the possibility to load entire translation files without overriding any existing translations


simahawk (approved):
@lmignon this could be important for you ;)

simahawk (approved):

PR #398, opened 2022-04-26
[14.0] shopfloor, delivery: new option 'Process as pre-packaged' products

sebalix:
This option allows to scan product packaging for bulk products and process the packaging qty.

Ref. rau-38

### DIFF
--- a/shopfloor/actions/search.py
+++ b/shopfloor/actions/search.py
@@ -34,15 +34,13 @@ class SearchAction(Component):
             return model.browse()
         return model.search([("name", "=", barcode)], limit=1)
 
-    def product_from_scan(self, barcode):
+    def product_from_scan(self, barcode, use_packaging=True):
         model = self.env["product.product"]
         if not barcode:
             return model.browse()
         product = model.search([("barcode", "=", barcode)], limit=1)
-        if not product:
-            packaging = self.env["product.packaging"].search(
-                [("product_id", "!=", False), ("barcode", "=", barcode)], limit=1
-            )
+        if not product and use_packaging:
+            packaging = self.packaging_from_scan(barcode)
             product = packaging.product_id
         return product
 
@@ -57,6 +55,14 @@ class SearchAction(Component):
         if products:
             domain.append(("product_id", "in", products.ids))
         return model.search(domain, limit=limit)
+
+    def packaging_from_scan(self, barcode):
+        model = self.env["product.packaging"]
+        if not barcode:
+            return model.browse()
+        return model.search(
+            [("barcode", "=", barcode), ("product_id", "!=", False)], limit=1
+        )
 
     def generic_packaging_from_scan(self, barcode):
         model = self.env["product.packaging"]
--- a/shopfloor/services/delivery.py
+++ b/shopfloor/services/delivery.py
@@ -94,6 +94,11 @@ class Delivery(Component):
         than one package, a package barcode is requested, and if the product is
         tracked by lot/serial, a lot is asked.
 
+        If the option 'Process as pre-packaged' is enabled on the menu, then
+        when a product's packaging is scanned, the first move line without
+        a source package (bulk line) corresponding to the quantity of the
+        packaging will be set to done.
+
         If the barcode is a lot, the lines for this lot are set to
         done. However, if the lot is in more than one package, a package
         barcode is requested.
@@ -112,6 +117,7 @@ class Delivery(Component):
         * deliver: always return here with the data for the last touched
         picking or no picking if the picking has been set to done
         """
+        allow_prepackaged_product = self.work.menu.allow_prepackaged_product
         search = self._actions_for("search")
         picking = search.picking_from_scan(barcode)
         barcode_valid = bool(picking)
@@ -130,7 +136,9 @@ class Delivery(Component):
                 return self._deliver_package(picking, package)
 
         if not barcode_valid:
-            product = search.product_from_scan(barcode)
+            product = search.product_from_scan(
+                barcode, use_packaging=(not allow_prepackaged_product)
+            )
             if product:
                 return self._deliver_product(picking, product)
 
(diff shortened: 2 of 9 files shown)


shopfloor/services/delivery.py
```diff
@@ -112,6 +112,7 @@ def scan_deliver(self, barcode, picking_id=None):
         * deliver: always return here with the data for the last touched
         picking or no picking if the picking has been set to done
         """
+        allow_prepackaged_product = self.work.menu.allow_prepackaged_product
```
simahawk, line 120: this should probably come from https://github.com/OCA/wms/blob/14.0/shopfloor_base/services/service.py#L282
but now that I look again at it, we are not overriding the value from the scenario w/ the value from the menu....
If you have time you could maybe have a look at this.

shopfloor/services/delivery.py
```diff
@@ -139,16 +142,34 @@ def scan_deliver(self, barcode, picking_id=None):
             if lot:
                 return self._deliver_lot(picking, lot)
 
+        if not barcode_valid and allow_prepackaged_product:
+            packaging = search.packaging_from_scan(barcode)
+            if packaging:
+                packaging_qty = packaging.product_uom_id._compute_quantity(
```
simahawk, line 155: add a comment like "Thus we use the full qty of the packaging" or similar

simahawk (approved):
LG overall

shopfloor/services/delivery.py
sebalix: done, and docstring of the `scan_deliver` method updated as well

shopfloor/services/delivery.py
sebalix: I didn't know about this `options` property. I'll check probably later if I have time.

shopfloor/actions/search.py
```diff
@@ -34,12 +34,12 @@ def picking_from_scan(self, barcode):
...
-    def product_from_scan(self, barcode):
+    def product_from_scan(self, barcode, use_packaging=True):
         model = self.env["product.product"]
         if not barcode:
             return model.browse()
         product = model.search([("barcode", "=", barcode)], limit=1)
-        if not product:
+        if not product and use_packaging:
             packaging = self.env["product.packaging"].search(
                 [("product_id", "!=", False), ("barcode", "=", barcode)], limit=1
```
jbaudoux: call packaging_from_scan

shopfloor/actions/search.py
```diff
@@ -58,6 +58,14 @@ def lot_from_scan(self, barcode, products=None, limit=1):
             domain.append(("product_id", "in", products.ids))
         return model.search(domain, limit=limit)
 
+    def packaging_from_scan(self, barcode):
+        model = self.env["product.packaging"]
+        if not barcode:
+            return model.browse()
+        return model.search(
+            [("barcode", "=", barcode), ("product_id", "!=", False)], limit=1
```
jbaudoux, line 64: and qty > 0 ?

jbaudoux (changes requested):
Reviewing the behavior when scanning a product barcode, so no prepacked product (product.packaging) or no packed product (stock.quant.package), the outcome is quite dangerous as it will deliver completely a matched move line.

We need to be able to disable that product scan in the scenario.

shopfloor/services/delivery.py
```diff
@@ -130,7 +136,9 @@ def scan_deliver(self, barcode, picking_id=None):
                 return self._deliver_package(picking, package)
 
         if not barcode_valid:
-            product = search.product_from_scan(barcode)
+            product = search.product_from_scan(
+                barcode, use_packaging=(not allow_prepackaged_product)
+            )
             if product:
                 return self._deliver_product(picking, product)
```
jbaudoux, line 143: The delivered quantity must always be 1. Same for lot.
Usually, when the package contains a single unit, it is the same barcode on the product and package.

shopfloor/services/delivery.py
```diff
@@ -206,24 +234,37 @@ def _lines_from_lot_domain(self, lot, no_qty_done=True):
...
         if product.tracking in ("lot", "serial"):
             return self._response_for_deliver(
                 picking, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
             )
 
+        limit = 1 if packaging_qty else None
         lines = self.env["stock.move.line"].search(
-            self._lines_from_product_domain(product)
+            self._lines_from_product_domain(product, packaging_qty=packaging_qty),
+            limit=limit,
```
jbaudoux, line 268: It is missing a sort by "picking scheduled date, picking id".

jbaudoux (changes requested):

ivantodorovich (approved):

shopfloor/actions/search.py
jbaudoux: I remove my comment, it's better to match a product through its barcode packaging even if we don't know the quantity

shopfloor/services/delivery.py
sebalix: You mean we should do the call like this?
```suggestion
                return self._deliver_product(picking, product, packaging_qty=1)
```

shopfloor/services/delivery.py
```diff
@@ -206,24 +234,37 @@ def _lines_from_lot_domain(self, lot, no_qty_done=True):
             [self._lines_base_domain(no_qty_done), [("lot_id", "=", lot.id)]]
         )
 
-    def _lines_from_product_domain(self, product, no_qty_done=True):
-        return expression.AND(
+    def _lines_from_product_domain(self, product, no_qty_done=True, packaging_qty=None):
```
jbaudoux: Similar code is done in https://github.com/OCA/wms/pull/410/files#diff-0ba2c9ccb67d9c1b060b30e26ff7ba99ef0b4733a3c5a6bfeebe64677cc60417R226
Can this be moved outside of the scenario as an action as this is useful in several scenario
cc @mmequignon

shopfloor/services/delivery.py
jbaudoux: yes but I would call it product_qty

shopfloor/services/delivery.py
sebalix: This means storing the `scheduled_date` field on the move line... I'll check.

shopfloor/services/delivery.py
sebalix: it is breaking existing tests, I'm doing something wrong probably.

shopfloor/services/delivery.py
sebalix: We now store the `move_id.date` field at the line level in a new `date_planned` (filled by a migration script for upcoming releases).

shopfloor/services/delivery.py
sebalix: tests have to be adapted for that use case.

sebalix: This require some changes, I prefer to postpone that in another PR for later (draft PR created here: https://github.com/OCA/wms/pull/418 )

shopfloor/services/delivery.py
sebalix: We will have to postpone this for later, probably addressed when we will write new scenarios to refactor a bit this code.

jbaudoux (approved):

PR #399, opened 2022-04-27
[14.0] shopfloor: Cluster picking / Location content transfer - Fix error messages in case of wrong thing scanned

jbaudoux:
If you don't scan the right package/product/lot/location, raise the correct error to the operator that he scanned the wrong package/product/lot/location instead of raising a "Barcode not found"

In location content transfer, initial step where you have to scan a location:
* add a check that the location is valid for the menu's picking types
* raise a "Location empty" error message if the location is empty instead of a "no pack found in location" (the pack error is for single_pack_transfer)

I changed the "Location empty" from info to error. I checked the scenario using this message and seems to me to always be an error and not an info.

cc @sebalix @simahawk @lmignon @Lmarion-source

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -1,8 +1,12 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
+
 from odoo import _
 
 from odoo.addons.component.core import Component
+
+_logger = logging.getLogger(__name__)
 
 
 class MessageAction(Component):
@@ -200,10 +204,23 @@ class MessageAction(Component):
         }
 
     def wrong_product(self):
+        # Method to drop in v15
+        _logger.warning("`wrong_product` is deprecated, use `wrong_record` instead")
         return {
             "message_type": "error",
-            "body": _("Wrong product."),
+            "body": self._wrong_record_msg("product.product"),
         }
+
+    def _wrong_record_msg(self, model_name):
+        return {
+            "product.product": _("Wrong product."),
+            "stock.production.lot": _("Wrong lot."),
+            "stock.location": _("Wrong location."),
+            "stock.quant.package": _("Wrong pack."),
+        }.get(model_name, _("Wrong."))
+
+    def wrong_record(self, record):
+        return {"message_type": "error", "body": self._wrong_record_msg(record._name)}
 
     def no_lot_for_barcode(self, barcode):
         return {
@@ -218,9 +235,11 @@ class MessageAction(Component):
         }
 
     def wrong_lot(self):
+        # Method to drop in v15
+        _logger.warning("`wrong_product` is deprecated, use `wrong_record` instead")
         return {
             "message_type": "error",
-            "body": _("Wrong lot."),
+            "body": self._wrong_record_msg("stock.production.lot"),
         }
 
     def several_lots_in_location(self, location):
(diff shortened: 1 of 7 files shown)


shopfloor/actions/message.py
```diff
@@ -199,12 +199,24 @@ def no_product_for_barcode(self, barcode):
             "body": _("No product found for {}").format(barcode),
         }
 
+    def wrong_package(self):
+        return {
+            "message_type": "error",
+            "body": _("Wrong pack."),
+        }
+
```
```suggestion
    def _wrong_record_msg(self, model_name):
        return {
            "product.product": _("Wrong product."),
            "stock.location": _("Wrong location."),
            "stock.quant.package": _("Wrong pack."),
        }.get(model_name, _("Wrong record"))
        
    def wrong_record(self, record):
        return {
            "message_type": "error",
            "body": self._wrong_record_msg(record._name)
        }

```

shopfloor/actions/message.py
```diff
@@ -199,12 +199,24 @@ def no_product_for_barcode(self, barcode):
...
             "message_type": "error",
             "body": _("Wrong product."),
         }
 
+    def wrong_location(self):
+        return {
+            "message_type": "error",
+            "body": _("Wrong location."),
+        }
+
```
```suggestion
```

shopfloor/services/cluster_picking.py
```diff
@@ -464,6 +464,23 @@ def scan_line(self, picking_batch_id, move_line_id, barcode):
...
+                move_line, message=self.msg_store.wrong_product()
+            )
+        if lot:
+            return self._response_for_start_line(
+                move_line, message=self.msg_store.wrong_lot()
+            )
+        if location:
+            return self._response_for_start_line(
+                move_line, message=self.msg_store.wrong_location()
+            )
```
simahawk:
```suggestion
        for rec in (package, product, lot, location):
            if rec:
                return self._response_for_start_line(
                    move_line, message=self.msg_store.wrong_record(rec)
                )
```

shopfloor/actions/message.py
jbaudoux: :+1:

shopfloor/actions/message.py
```diff
@@ -199,11 +199,16 @@ def no_product_for_barcode(self, barcode):
             "body": _("No product found for {}").format(barcode),
         }
 
-    def wrong_product(self):
```
simahawk, line 202: this way you break https://github.com/OCA/wms/blob/b84a3ef9d4d3f9667707316a1fb7afa76a8dc796/shopfloor_manual_product_transfer/services/manual_product_transfer.py#L483

Please:
1. do not remove this method
2. make it use the new one
3. log a deprecation warning if we expect to drop it completely for v15
4. adapt shopfloor_manual_product_transfer to use the new method

simahawk (changes requested):
@jbaudoux any reason to keep this in draft?

shopfloor/actions/message.py
jbaudoux: Thanks, done

sebalix (approved):

shopfloor/actions/message.py
```diff
@@ -218,9 +235,11 @@ def lot_on_wrong_product(self, barcode):
         }
 
     def wrong_lot(self):
+        # Method to drop in v15
+        _logger.warning("`wrong_product` is deprecated, use `wrong_record` instead")
```
```suggestion
        _logger.warning("`wrong_lot` is deprecated, use `wrong_record` instead")
```

simahawk (approved):

shopfloor/actions/message.py
jbaudoux: :+1:

PR #403, opened 2022-04-28
[14.0] shopfloor: Fix extract move lines

jbaudoux:
The location content transfer and zone picking are now aligned in the way lines are processed. Everything that you start to process is extracted in a split order and the user is set on it. This eases the tracking of what is being processed.

Refactor to reduce method complexity introduced by https://github.com/OCA/wms/pull/399

While reviewing the code, it appeared in the location content transfer that some pickings were lost when some gets their move lines extracted causing to move only split moves and not the other pickings having also content on the location.
I changed the tests in start in order to have the failing use case covered. The tests were already processing 2 different pickings. Now the second picking has stock on 2 different source locations. This causes a split on the second picking only.

In the zone picking, I added the extraction of processed lines / qty in a split order like in the location content transfer.
This solves a bug processing partial quantities. The last processed line was moving all quantity to its destination causing an automatic stock correction. 
This solves a bug when lines are reserved on multiple locations. It was only possible to process one of the lines.
When a line is put on a destination package, you can now find back in odoo the processed picking as it has the user set on it.
As the extraction in a split order is done at the beginning of the process, this has an impact on how you process backorder if you use the module `stock_picking_backorder_strategy` to automatically cancel backorders. A new glue module is required to cancel the initial order after extraction.

### DIFF
--- a/shopfloor/services/location_content_transfer.py
+++ b/shopfloor/services/location_content_transfer.py
@@ -319,36 +319,21 @@ class LocationContentTransfer(Component):
                 return self._response_for_start(
                     message=self.msg_store.cannot_move_something_in_picking_type()
                 )
-        # If there are different source locations, we put the move lines we are
-        # interested in in a separate picking.
-        # This is required as we can only deal within this scenario with pickings
-        # that share the same source location.
-        pickings = move_lines._split_pickings_from_source_location()
-        # Ensure we process move lines related to transfers having only one source
-        # location among all their move lines.
-        # We need to put the unreserved qty into separate moves as a new move
-        # line could be created in the middle of the process.
-        new_picking_ids = []
-        for picking in pickings:
-            #   -> put move lines to process in their own move/transfer
-            new_picking_id = picking.split_assigned_move_lines(move_lines)
-            new_picking_ids.append(new_picking_id)
-        if new_picking_ids != pickings.ids:
-            pickings = pickings.browse(new_picking_ids)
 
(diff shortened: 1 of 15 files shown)


sebalix:
> As the extraction in a split order is done at the beginning of the process, this has an impact on how you process backorder if you use the module stock_picking_backorder_strategy to automatically cancel backorders. A new glue module is required to cancel the initial order after extraction.

Indeed, that's why the extraction was pushed at the end if possible to easily rollback a started process for any reason.
That's a lot to understand for now to me but overall LG.

shopfloor/services/location_content_transfer.py
```diff
@@ -318,36 +318,21 @@ def scan_location(self, barcode):
...
+            # This is required as we can only deal within this scenario with pickings
+            # that share the same source location.
+            # We also need to put the unreserved qty into separate moves as a new move
+            # line could be created in the middle of the process.
+            move_lines._extract_in_split_order()
+        elif not self.is_allow_move_create():
+            savepoint.rollback()
+            return self._response_for_start(
+                message=self.msg_store.location_empty(location)
+            )
```
mmequignon: Why don't we check this before the split?

mmequignon (approved):

sebalix:
Rebased on top of 14.0 to resolve the conflict on `shopfloor/services/location_content_transfer.py`.

sebalix (approved):

PR #404, opened 2022-04-29
[14.0] shopfloor: add options on cluster and zone picking

mmequignon:
Based on https://github.com/OCA/wms/pull/398 to avoid conflicts.

Add a two options:
- `Unload package at destination`, which allows to unload goods as bulk products in the destination location.
- `Collect multiple moves on a same destination package`, which allows to set a destination package that was already used for the other move lines.

Also add some constrains, because `Collect multiple moves on a same destination package` should be only be enabled when `Unload package at destination` is enabled, and `Unload package at destination` cannot be enabled when `Pick and pack at the same time` is.

### DIFF
--- a/shopfloor/models/stock_quant_package.py
+++ b/shopfloor/models/stock_quant_package.py
@@ -15,6 +15,7 @@ class StockQuantPackage(models.Model):
     planned_move_line_ids = fields.One2many(
         comodel_name="stock.move.line",
         inverse_name="result_package_id",
+        domain=[("state", "not in", ("done", "cancel"))],
         readonly=True,
         help="Technical field. Move lines for which destination is this package.",
     )
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -1162,6 +1162,8 @@ class ZonePicking(Component):
         self._lock_lines(lines)
         lines.location_dest_id = location
         lines.package_level_id.location_dest_id = location
+        if self.work.menu.unload_package_at_destination:
+            lines.result_package_id = False
 
     def unload_split(self):
         """Indicates that now the buffer must be treated line per line
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -661,14 +661,8 @@ class ZonePicking(Component):
         return not bool(package.quant_ids)
 
     def _is_package_already_used(self, package):
-        return bool(
-            self.env["stock.move.line"].search_count(
-                [
-                    ("state", "not in", ("done", "cancel")),
-                    ("result_package_id", "=", package.id),
-                ]
-            )
-        )
+        # Deprecated, use planned_move_line_ids instead
+        return bool(package.planned_move_line_ids)
 
     def _move_line_compare_qty(self, move_line, qty):
         rounding = move_line.product_uom_id.rounding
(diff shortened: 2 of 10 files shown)


shopfloor/services/zone_picking.py
```diff
@@ -700,7 +700,9 @@ def _set_destination_package(self, move_line, quantity, package):
                 message=self.msg_store.package_not_empty(package),
             )
             return (package_changed, response)
-        if self._is_package_already_used(package):
+        package_already_used = self._is_package_already_used(package)
```
jbaudoux: Align cluster & zone picking.
Replace with `planned_move_line_ids`. See if you can move the domain (not in cancel/done) in the field definition.
`_is_package_already_used` can use `planned_move_line_ids` and marked as deprecated

jbaudoux (approved):
Just a small remark.

Can you also squash the commits and keep commit title short

shopfloor/actions/search.py
```diff
@@ -34,12 +34,12 @@ def picking_from_scan(self, barcode):
...
+    def product_from_scan(self, barcode, use_packaging=True):
         model = self.env["product.product"]
         if not barcode:
             return model.browse()
         product = model.search([("barcode", "=", barcode)], limit=1)
-        if not product:
+        if not product and use_packaging:
             packaging = self.env["product.packaging"].search(
                 [("product_id", "!=", False), ("barcode", "=", barcode)], limit=1
             )
```
sebalix: Note for later: need to address comment done there: https://github.com/OCA/wms/pull/398#discussion_r865988306=

shopfloor/actions/search.py
```diff
@@ -58,6 +58,14 @@ def lot_from_scan(self, barcode, products=None, limit=1):
             domain.append(("product_id", "in", products.ids))
         return model.search(domain, limit=limit)
 
+    def packaging_from_scan(self, barcode):
+        model = self.env["product.packaging"]
+        if not barcode:
+            return model.browse()
+        return model.search(
+            [("barcode", "=", barcode), ("product_id", "!=", False)], limit=1
```
sebalix: Note for later: need to address comment done there: https://github.com/OCA/wms/pull/398#discussion_r866036519=

sebalix (approved):

ivantodorovich (approved):

sebalix:
Rebased on top of 14.0 now that #398 is merged

shopfloor/models/stock_quant_package.py
```diff
@@ -15,6 +15,7 @@ class StockQuantPackage(models.Model):
     planned_move_line_ids = fields.One2many(
         comodel_name="stock.move.line",
         inverse_name="result_package_id",
+        domain=[("state", "not in", ("done", "cancel"))],
```
sebalix, line 18: For performance we should compare the state to open ones (`confirmed`, `assigned`, etc...) as the number of move lines `done` and `cancel` will increase over time and is higher than the others, that doesn't help PostgreSQL.

sebalix: It was already like this before, I won't fix this for now.

sebalix:
@jbaudoux commits squashed & messages updated. I kept two commits as this is two well defined features.

PR #407, opened 2022-05-04
[14.0] shopfloor_mobile_base: dynamic storage type

JuMiSanAr:
The config_registry can now make use of localStorage for data that must persist for a longer time than what's allowed by sessionStorage.


shopfloor_mobile_base/static/wms/src/services/config_registry.js
```diff
@@ -131,12 +148,26 @@ export class ConfigRegistry {
...
+        this._set_storage_val(config, v);
+    }
+
+    _switch_storage_driver(driver) {
+        // The app uses sessionStorage by default (currently set on app creation).
+        // If any piece of data needs to be handled by localStorage,
+        // we switch the vue2storage driver option to "local", we store / retrieve the value,
+        // and then we revert to "session" for further use.
+        // See example in documentation: https://github.com/yarkovaleksei/vue2-storage/blob/master/docs/en/started.md.
+        const permitted_types = ["local", "session", "memory"];
```
simahawk: please use https://github.com/yarkovaleksei/vue2-storage/blob/d48b1321bb97359129820e41d2ec3856edd05758/src/storage.ts#L10

simahawk (changes requested):
pls update the ui tests too

shopfloor_mobile_base/static/wms/src/services/config_registry.js
```diff
@@ -20,6 +20,7 @@ class StoredConfig {
         this.key = key;
         this.default = meta.default;
         this.reset_on_clear = meta.reset_on_clear;
+        this.storage_driver = meta.storage_driver;
```
simahawk: what if here we look for a dict?
We could have different TTL  per key.
```suggestion
        this.storage = meta.storage || {};
        [...]
        this.storage.driver | ttl | whatever
```

simahawk: and in the app init we could consider getting these defaults from `shopfloor_app_info`

simahawk (changes requested):

shopfloor_mobile_base/static/wms/src/services/config_registry.js
JuMiSanAr: Cannot be used, as the current latest version of the library doesn't export it:
https://unpkg.com/vue2-storage@6.1.3/dist/vue2-storage.js

shopfloor_base/models/shopfloor_app.py
```diff
@@ -86,6 +86,17 @@ class ShopfloorApp(models.Model):
         help="If set, the app will be first loaded with this lang.",
     )
     lang_ids = fields.Many2many("res.lang", string="Available languages")
+    storage_driver = fields.Selection(
```
simahawk: this is very low level stuff, pls drop it for now (all the backend part)

simahawk (changes requested):

shopfloor_base/models/shopfloor_app.py
JuMiSanAr: Dropped.

shopfloor_mobile_base/static/wms/src/services/config_registry.js
simahawk: ok, move this check to config key init

simahawk (changes requested):
rewrite last commit as `sf_mobile_base: ease override default storage conf`

shopfloor_mobile_base/static/wms/src/services/config_registry.js
```diff
@@ -20,10 +20,23 @@ class StoredConfig {
         this.key = key;
         this.default = meta.default;
         this.reset_on_clear = meta.reset_on_clear;
+        if (!_.isEmpty(meta.storage)) {
```
simahawk, line 23: Do not change the original value. Just raise an error if you try to register a wrong type.
This should not pass silently

JuMiSanAr: Added.

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -188,7 +195,10 @@ new Vue({
                     self.$storage.set("appconfig", self.appconfig);
                 }
                 event_hub.$emit("app.sync:update", {root: self, sync_data: result});
-                if (self.user.lang) {
+                const stored_lang = config_registry.get_value("current_language");
```
simahawk: why not `this.current_language`?

shopfloor_mobile_base/static/wms/src/services/config_registry.js
```diff
@@ -20,10 +20,30 @@ class StoredConfig {
...
         return v === null ? this.default : v;
     }
+    _is_storage_driver_allowed(driver) {
+        // TODO: these permitted types are hardcoded, matching the driver types of Vue2Storage.
+        // If possible, we should be able to get them from the library class instead.
+        const permitted_types = ["local", "session", "memory"];
+        if (permitted_types.includes(driver)) {
+            return true;
+        }
+        return false;
```
simahawk:
```suggestion
        return ["local", "session", "memory"].includes(driver);
```

shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -78,6 +84,7 @@ new Vue({
     },
     beforeCreate: function () {
         config_registry._set_root(this);
+        config_registry._set_storage_defaults();
```
simahawk: wouldn't be convenient to call this in `_set_root`?

simahawk (changes requested):

simahawk (approved):

PR #411, opened 2022-05-10
[14.0] shopfloor_mobile_base: manifest & app info improvements

JuMiSanAr:
The motivation of this PR is to fix and finalize shopfloor.app installation as a PWA:

- The entire url of the app is exposed in shopfloor_app_info.
- In the manifest:
    - "start_url" now gives the entire url of the app, and not only the path.
    - "scope" is added to the manifest, and it matches start_url.
    -  "id" is added but only as a relative path (as it is only used to identify unique PWAs).

### DIFF
--- a/shopfloor_mobile_base/controllers/main.py
+++ b/shopfloor_mobile_base/controllers/main.py
@@ -91,10 +91,18 @@ class ShopfloorMobileAppMixin(object):
         return all_icons
 
     def _get_manifest(self, shopfloor_app):
+        param = (
+            http.request.env["ir.config_parameter"]
+            .sudo()
+            .get_param("web.base.url", "")
+            .rstrip("/")
+        )
         return {
             "name": shopfloor_app.name,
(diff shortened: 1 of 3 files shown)


shopfloor_mobile_base/controllers/main.py
```diff
@@ -91,10 +91,18 @@ def _get_app_icons(self, shopfloor_app):
...
+            http.request.env["ir.config_parameter"]
+            .sudo()
+            .get_param("web.base.url", "")
+            .rstrip("/")
+        )
         return {
             "name": shopfloor_app.name,
             "short_name": shopfloor_app.short_name,
-            "start_url": shopfloor_app.url + "#",
+            "start_url": param + shopfloor_app.url + "/",
```
simahawk: do we need this trailing slash?

JuMiSanAr:
Updated test_make_app_info to include the "url" key that's added in this PR.

simahawk (approved):

PR #415, opened 2022-05-25
[14.0][WIP] stock_dynamic_routing & release: merge moves

jbaudoux:
## Merging pulled moves
In standard, when _action_confirm is called, moves can be merged.
With the dynamic routing, if a move is re-assigned to another picking, we should also try to merge it with another one.
A side effect is that the initial move is deleted when merged, so we need to ensure the move still exists in any further call after action_assign.

## Force new procurement group
Allow to alter the procurement group of a pulled move.

### Use case
10 SO for a same product. Routing configured in 2 steps. Set a fixed procurement group on the rerouted pick move.
This will create a single pick transfer for all the SO's shipments. The pick will contain a single move line with the total quantity to pick.

## Fix the scheduled date and deadline
When releasing a move:
*  on the released move:
    * the deadline remains the one coming from the SO
    * the scheduled date is set to now + the max prep time
* on the pulled moves:
    * the deadline is set to now + the max prep time <= CHANGED: before it was the deadline of the SO. Required for the merging to occur as the deadline is part of the merge key
    * the scheduled date is set to now + max prep time (it cannot be now otherwise it will immediately be counted as a late picking)

cc @jgrandguillaume

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -1,5 +1,5 @@
 # Copyright 2019-2022 Camptocamp (https://www.camptocamp.com)
-# Copyright 2020-2022 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# Copyright 2021-2022 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 import logging
@@ -304,6 +304,8 @@ class StockMove(models.Model):
         procurement_requests = []
         pulled_moves = self.env["stock.move"]
         backorder_links = {}
+
+        new_date = fields.Datetime.now()
         for move in self:
             if not move.need_release:
                 continue
@@ -322,9 +324,11 @@ class StockMove(models.Model):
                     # once
                     continue
                 new_move = move._release_split(remaining)
+                new_move._release_split_set_date(new_date)
                 backorder_links[new_move.picking_id] = move.picking_id
 
             values = move._prepare_procurement_values()
+            move._release_set_procurement_date(values, new_date)
             procurement_requests.append(
                 self.env["procurement.group"].Procurement(
                     move.product_id,
@@ -349,6 +353,7 @@ class StockMove(models.Model):
             # will be assigned to a new stock.picking
             original_picking = unreleased_move.picking_id
             unreleased_move._release_split(unreleased_move.product_qty)
+            unreleased_move._release_split_set_date(new_date)
             backorder_links[unreleased_move.picking_id] = original_picking
 
         for backorder, origin in backorder_links.items():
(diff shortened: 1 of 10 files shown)


stock_dynamic_routing/models/stock_routing_rule.py
```diff
@@ -38,12 +39,19 @@ class StockRoutingRule(models.Model):
         required=True,
         help="Operation type that will be applied on the move.",
     )
+    group_id = fields.Many2one(
+        comodel_name="procurement.group",
+        string="Fixed Procurement Group",
+        help="Operation type that will be applied on the move.",
```
jgrandguillaume: ?! I guess you mean "Procurement group that will be applied on the move" ?

jgrandguillaume (commented):
Heavy one, need to discuss, but overall the approach seems good to me.
This features will definitely be a killer one !

stock_dynamic_routing/models/stock_routing_rule.py
jbaudoux: fixed

stock_move_source_relocate_dynamic_routing/tests/test_dynamic_destination.py
```diff
@@ -12,9 +14,11 @@ def test_pickship_with_routing_and_relocate(self):
         destination location and one not assigned relocated and then re-routed
         to that same destination location. As both pick moves are going to the
         same destination location, ensure the ship move is not split.
+        As the pick moves share the same characteristics, they are merged back
+        together.
```
sebalix, line 18: On which event this is occurring? If the second pick move becomes available then it is merged back in the second?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -363,6 +369,23 @@ def _run_stock_rule(self):
 
         return True
 
+    def _release_get_deadline(self, date_planned):
+        prep_time = self.env.company.stock_release_max_prep_time
+        date_deadline = fields.Datetime.add(date_planned, minutes=prep_time)
+        return date_deadline
```
sebalix, line 375: Here it is not a date deadline I guess, but a scheduled/planned date which is computed:
```suggestion
    def _release_get_date_planned(self, date_planned):
        prep_time = self.env.company.stock_release_max_prep_time
        new_date_planned = fields.Datetime.add(date_planned, minutes=prep_time)
        return new_date_deadline
```
The date deadline is when the goods are shipped to the destination location (with the time needed by the carrier let's say). If a commitment date is set on a SO, this is the date deadline of the moves.
Note: we worked a lot on this PR to handle a lot of use cases regarding time windows etc: https://github.com/OCA/sale-workflow/pull/1968/

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -363,6 +369,23 @@ def _run_stock_rule(self):
...
+
+    def _release_split_set_date(self, date_planned):
+        """Set released move planned date to pulled moves planned date"""
+        self.date = self._release_get_deadline(date_planned)
```
sebalix, line 379:
```suggestion
        self.date = self._release_get_date_planned(date_planned)
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -363,6 +369,23 @@ def _run_stock_rule(self):
...
+
+    def _release_set_procurement_date(self, values, date_planned):
+        # We must set the deadline on the move before pulling to allow pulled
+        # moves to be merged
+        values["date_deadline"] = self._release_get_deadline(date_planned)
+        # It cannot be scheduled for now otherwise it will immediately appear
+        # in the late picking count. So it is also set to the deadline.
+        values["date_planned"] = self._release_get_deadline(date_planned)
```
sebalix, line 387: We should not change the `date_deadline` of moves (especially OUT ones), as this is the date which could have been communicated to the customer when confirming the SO (it is even enforced if we set a commitment date on the SO).
On std Odoo 14.0 the computation is the following:
```
PROC_VALUES.date_deadline = SO.commitment_date or (SO.date_order + SOL.customer_lead)
PROC_VALUES.date_planned = PROC['date_deadline'] - COMPANY.security_lead
MOVE.date = PROC_VALUES.date_planned - RULE.delay
MOVE.date_deadline = PROC_VALUES.date_deadline - RULE.delay
```
This computation has been overridden by `sale_delivery_date` (see https://github.com/OCA/sale-workflow/pull/1968/) to take into account the WH working hours, the WH cutoff time, the time windows of the customer etc.

jbaudoux:
Replaced by https://github.com/OCA/wms/pull/431

PR #420, opened 2022-06-15
[14.0] shopfloor : Raise excetion instead dict message

bguillot:
Hi,

This PR replace error messages returned to the app by raises.
It allows  to handle rollback more easily and avoid to take care of the return of methods in case of nested methods. 
For instance if a method returns a recordset but sometimes if error appends it returns dict of error message, so you have to check if the return is dict, and be sure that it will follow all the way down.

I did on one scenario to show the differences. We can change for the others after that.


sebastienbeau:
@lmignon

lmignon (approved):
IMO it's a must have... It restore 'normal' behavior when a process encounter an error into a python program.

shopfloor_base/__init__.py
```diff
@@ -2,3 +2,4 @@
 from . import models
 from . import actions
 from . import services
+from . import exceptions
```
simahawk, line 5: there's no need to import this file

shopfloor_base/exceptions.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2022 Akretion (https://www.akretion.com).
+# @author Sébastien BEAU <sebastien.beau@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.exceptions import UserError
+
+
+class ShopfloorDispatchError(UserError):
+    def __init__(self, payload):
+        payload["no_wrap"] = True
```
simahawk, line 9: can you document this flag?

shopfloor_rest_log/components/service.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def _dispatch_exception(
+        self, method_name, exception_klass, orig_exception, *args, params=None
+    ):
+        try:
+            super()._dispatch_exception(
+                method_name, exception_klass, orig_exception, *args, params=params
+            )
+        except Exception as e:
+            if isinstance(orig_exception, ShopfloorError):
```
simahawk, line 22: can you add a test for this?

simahawk (changes requested):
@bguillot can you rebase?

PR #421, opened 2022-06-22
[14.0] Shopfloor: Add no_prefill_qty option

mmequignon:
ref: rau-30

### DIFF
--- a/shopfloor/actions/search.py
+++ b/shopfloor/actions/search.py
@@ -34,15 +34,11 @@ class SearchAction(Component):
             return model.browse()
         return model.search([("name", "=", barcode)], limit=1)
 
-    def product_from_scan(self, barcode, use_packaging=True):
+    def product_from_scan(self, barcode):
         model = self.env["product.product"]
         if not barcode:
             return model.browse()
-        product = model.search([("barcode", "=", barcode)], limit=1)
-        if not product and use_packaging:
-            packaging = self.packaging_from_scan(barcode)
-            product = packaging.product_id
-        return product
+        return model.search([("barcode", "=", barcode)], limit=1)
 
     def lot_from_scan(self, barcode, products=None, limit=1):
         model = self.env["stock.production.lot"]
--- a/shopfloor/data/shopfloor_scenario_data.xml
+++ b/shopfloor/data/shopfloor_scenario_data.xml
@@ -1,4 +1,4 @@
-<odoo noupdate="1">
+<odoo noupdate="0">
     <record id="scenario_single_pack_transfer" model="shopfloor.scenario">
         <field name="name">Single Pack Transfer</field>
         <field name="key">single_pack_transfer</field>
@@ -18,7 +18,8 @@
 {
     "pick_pack_same_time": true,
     "unload_package_at_destination": true,
-    "multiple_move_single_pack": true
+    "multiple_move_single_pack": true,
+    "no_prefill_qty": true
 }
         </field>
     </record>
@@ -28,13 +29,19 @@
         <field name="options_edit">
 {
     "unload_package_at_destination": true,
-    "multiple_move_single_pack": true
+    "multiple_move_single_pack": true,
+    "no_prefill_qty": true
 }
         </field>
     </record>
     <record id="scenario_checkout" model="shopfloor.scenario">
         <field name="name">Checkout</field>
         <field name="key">checkout</field>
+        <field name="options_edit">
+{
+    "no_prefill_qty": true
+}
+        </field>
     </record>
     <record id="scenario_delivery" model="shopfloor.scenario">
         <field name="name">Delivery</field>
@@ -53,7 +60,8 @@
 {
     "allow_create_moves": true,
     "allow_unreserve_other_moves": true,
-    "allow_ignore_no_putaway_available": true
+    "allow_ignore_no_putaway_available": true,
+    "no_prefill_qty": true
 }
         </field>
     </record>
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -222,7 +222,7 @@ class Checkout(Component):
                     picking = pickings
         if not picking:
             # Try to find the product first
-            product = search.product_from_scan(barcode, use_packaging=False)
+            product = search.product_from_scan(barcode)
             # TODO Filter lines on picking_type before
             # line_domain = [
             #     ("move_id.picking_id.picking_type_id", "in", self.picking_types.ids)
@@ -403,6 +403,10 @@ class Checkout(Component):
             return self._select_lines_from_package(picking, selection_lines, package)
 
         product = search.product_from_scan(barcode)
+        if not product:
+            packaging = search.packaging_from_scan(barcode)
+            if packaging:
+                product = packaging.product_id
         if product:
             return self._select_lines_from_product(picking, selection_lines, product)
 
@@ -760,6 +764,9 @@ class Checkout(Component):
         selected_lines = self.env["stock.move.line"].browse(selected_line_ids).exists()
 
         product = search.product_from_scan(barcode)
+        if not product:
+            packaging = search.packaging_from_scan(barcode)
+            product = packaging.product_id
         if product:
             if product.tracking in ("lot", "serial"):
                 return self._response_for_select_package(
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -352,11 +352,16 @@ class Checkout(Component):
             return self._response_for_manual_selection(message=message)
         return self._select_picking(picking, "manual_selection")
 
-    def _select_lines(self, lines):
+    def _select_lines(self, lines, prefill_qty=0):
         for line in lines:
             if line.shopfloor_checkout_done:
                 continue
-            line.qty_done = line.product_uom_qty
+            if self.work.menu.no_prefill_qty:
+                line.qty_done += prefill_qty
+                # For prefill quantity we only want to increment one line
+                prefill_qty = 0
+            else:
+                line.qty_done = line.product_uom_qty
             line.shopfloor_user_id = self.env.user
 
         picking = lines.mapped("picking_id")
@@ -403,12 +408,15 @@ class Checkout(Component):
             return self._select_lines_from_package(picking, selection_lines, package)
 
         product = search.product_from_scan(barcode)
+        if product:
+            return self._select_lines_from_product(picking, selection_lines, product)
+
         if not product:
             packaging = search.packaging_from_scan(barcode)
             if packaging:
-                product = packaging.product_id
-        if product:
-            return self._select_lines_from_product(picking, selection_lines, product)
+                return self._select_lines_from_packaging(
+                    picking, selection_lines, packaging
+                )
 
         lot = search.lot_from_scan(barcode, products=picking.move_lines.product_id)
         if lot:
@@ -431,6 +439,8 @@ class Checkout(Component):
                 },
             )
         self._select_lines(lines)
+        if self.work.menu.no_prefill_qty:
+            lines = picking.move_line_ids
         return self._response_for_select_package(picking, lines)
 
     def _select_lines_from_product(self, picking, selection_lines, product):
@@ -442,11 +452,7 @@ class Checkout(Component):
         lines = selection_lines.filtered(lambda l: l.product_id == product)
         if not lines:
             return self._response_for_select_line(
-                picking,
-                message={
-                    "message_type": "error",
-                    "body": _("Product is not in the current transfer."),
-                },
+                picking, message=self.msg_store.product_not_found_in_current_picking()
             )
 
         # When products are as units outside of packages, we can select them for
@@ -467,7 +473,41 @@ class Checkout(Component):
             # package and we have only one.
             return self._select_lines_from_package(picking, selection_lines, packages)
 
-        self._select_lines(lines)
+        self._select_lines(lines, prefill_qty=1)
+        return self._response_for_select_package(picking, lines)
+
+    def _select_lines_from_packaging(self, picking, selection_lines, packaging):
+        product = packaging.product_id
+        if product.tracking in ("lot", "serial"):
+            return self._response_for_select_line(
+                picking, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
+            )
+
+        lines = selection_lines.filtered(lambda l: l.product_id == product)
+        if not lines:
+            return self._response_for_select_line(
+                picking, message=self.msg_store.product_not_found_in_current_picking()
+            )
+
+        # When products are as units outside of packages, we can select them for
+        # packing, but if they are in a package, we want the user to scan the packages.
+        # If the product is only in one package though, scanning the product selects
+        # the package.
+        packages = lines.mapped("package_id")
+        # Do not use mapped here: we want to see if we have more than one package,
+        # but also if we have one product as a package and the same product as
+        # a unit in another line. In both cases, we want the user to scan the
+        # package.
+        if packages and len({line.package_id for line in lines}) > 1:
+            return self._response_for_select_line(
+                picking, message=self.msg_store.product_multiple_packages_scan_package()
+            )
+        elif packages:
+            # Select all the lines of the package when we scan a product in a
+            # package and we have only one.
+            return self._select_lines_from_package(picking, selection_lines, packages)
+
+        self._select_lines(lines, prefill_qty=packaging.qty)
         return self._response_for_select_package(picking, lines)
 
     def _select_lines_from_lot(self, picking, selection_lines, lot):
@@ -499,7 +539,7 @@ class Checkout(Component):
             # package and we have only one.
             return self._select_lines_from_package(picking, selection_lines, packages)
 
-        self._select_lines(lines)
+        self._select_lines(lines, prefill_qty=1)
         return self._response_for_select_package(picking, lines)
 
     def _select_line_package(self, picking, selection_lines, package):
@@ -666,6 +706,25 @@ class Checkout(Component):
                 lambda l: l.product_uom_qty,
             )
 
+    def _increment_custom_qty(
+        self, picking, selected_line, increment_lines, qty_increment
+    ):
+        """Increment the  qty_done of a move line with a custom value
+
+        The selected_line_ids parameter is used to keep the selection of lines
+        stateless.
+
+        Transitions:
+        * select_package: goes back to this screen showing all the lines after
+          we changed the qty
+        """
+        return self._change_line_qty(
+            picking.id,
+            selected_line.ids,
+            increment_lines.ids,
+            lambda line: line.qty_done + qty_increment,
+        )
+
     @staticmethod
     def _filter_lines_unpacked(move_line):
         return (
@@ -763,6 +822,7 @@ class Checkout(Component):
 
         selected_lines = self.env["stock.move.line"].browse(selected_line_ids).exists()
 
+        packaging = None
         product = search.product_from_scan(barcode)
         if not product:
             packaging = search.packaging_from_scan(barcode)
@@ -775,12 +835,23 @@ class Checkout(Component):
                     message=self.msg_store.scan_lot_on_product_tracked_by_lot(),
                 )
             product_lines = selected_lines.filtered(lambda l: l.product_id == product)
-            return self._switch_line_qty_done(picking, selected_lines, product_lines)
+            if self.work.menu.no_prefill_qty:
+                quantity_increment = packaging.qty if packaging else 1
+                return self._increment_custom_qty(
+                    picking, selected_lines, product_lines, quantity_increment
+                )
+            else:
+                return self._switch_line_qty_done(
+                    picking, selected_lines, product_lines
+                )
 
         lot = search.lot_from_scan(barcode, products=selected_lines.product_id)
         if lot:
             lot_lines = selected_lines.filtered(lambda l: l.lot_id == lot)
-            return self._switch_line_qty_done(picking, selected_lines, lot_lines)
+            if self.work.menu.no_prefill_qty:
+                return self._increment_custom_qty(picking, selected_lines, lot_lines, 1)
+            else:
+                return self._switch_line_qty_done(picking, selected_lines, lot_lines)
 
         package = search.package_from_scan(barcode)
         if package:
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -453,6 +453,9 @@ class ClusterPicking(Component):
 
         # use the common search method so we search by packaging too
         product = search.product_from_scan(barcode)
+        if not product:
+            packaging = search.packaging_from_scan(barcode)
+            product = packaging.product_id
         if product and move_line.product_id == product:
             return self._scan_line_by_product(picking, move_line, product)
 
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -100,8 +100,11 @@ class ClusterPicking(Component):
             popup=popup,
         )
 
-    def _response_for_scan_destination(self, move_line, message=None):
-        data = self._data_move_line(move_line)
+    def _response_for_scan_destination(self, move_line, message=None, qty_done=None):
+        if qty_done is None:
+            data = self._data_move_line(move_line)
+        else:
+            data = self._data_move_line(move_line, qty_done=qty_done)
         last_picked_line = self._last_picked_line(move_line.picking_id)
         if last_picked_line:
             # suggest pack to be used for the next line
@@ -451,13 +454,13 @@ class ClusterPicking(Component):
         if package and move_line.package_id == package:
             return self._scan_line_by_package(picking, move_line, package, batch)
 
-        # use the common search method so we search by packaging too
         product = search.product_from_scan(barcode)
-        if not product:
-            packaging = search.packaging_from_scan(barcode)
-            product = packaging.product_id
         if product and move_line.product_id == product:
             return self._scan_line_by_product(picking, move_line, product)
+
+        packaging = search.packaging_from_scan(barcode)
+        if move_line.product_id == packaging.product_id:
+            return self._scan_line_by_packaging(picking, move_line, packaging)
 
         lot = search.lot_from_scan(barcode, products=move_line.product_id)
         if lot and move_line.lot_id == lot:
@@ -477,6 +480,14 @@ class ClusterPicking(Component):
             move_line, message=self.msg_store.barcode_not_found()
         )
 
+    def _get_prefill_qty(self, move_line, qty=False):
+        """Returns the increment qty or qty_done depending on the no_prefill_qty
+        option value.
+        """
+        if self.work.menu.no_prefill_qty:
+            return qty
+        return move_line.product_uom_qty
+
     def _scan_line_by_package(self, picking, move_line, package, batch):
         """Package scanned, just work with it."""
         packaging = self._actions_for("packaging")
@@ -487,7 +498,8 @@ class ClusterPicking(Component):
             message = self.msg_store.several_lots_in_package(package)
         if message:
             return self._pick_next_line(batch, message=message)
-        return self._response_for_scan_destination(move_line)
+        quantity = self._get_prefill_qty(move_line)
+        return self._response_for_scan_destination(move_line, qty_done=quantity)
 
     def _scan_line_by_product(self, picking, move_line, product):
         """Product scanned, check if we can work with it.
@@ -522,7 +534,35 @@ class ClusterPicking(Component):
                 move_line,
                 message=self.msg_store.product_multiple_packages_scan_package(),
             )
-        return self._response_for_scan_destination(move_line)
+        quantity = self._get_prefill_qty(move_line, 1.0)
+        return self._response_for_scan_destination(move_line, qty_done=quantity)
+
+    def _scan_line_by_packaging(self, picking, move_line, packaging):
+        """Packaging scanned, check if we can work with it.
+
+        If the packaging related product is part of several packages in the same location,
+        we can't be sure it's the correct one, in such case, ask to scan a package
+        """
+        product = packaging.product_id
+        if move_line.product_id.tracking in ("lot", "serial"):
+            return self._response_for_start_line(
+                move_line, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
+            )
+        other_product_lines = picking.move_line_ids.filtered(
+            lambda l: l.product_id == product and l.location_id == move_line.location_id
+        )
+        packages = other_product_lines.mapped("package_id")
+        # Do not use mapped here: we want to see if we have more than one package,
+        # but also if we have one product as a package and the same product as
+        # a unit in another line. In both cases, we want the user to scan the
+        # package.
+        if packages and len({line.package_id for line in other_product_lines}) > 1:
+            return self._response_for_start_line(
+                move_line,
+                message=self.msg_store.product_multiple_packages_scan_package(),
+            )
+        quantity = self._get_prefill_qty(move_line, packaging.qty)
+        return self._response_for_scan_destination(move_line, qty_done=quantity)
 
     def _scan_line_by_lot(self, picking, move_line, lot):
         """Lot scanned, check if we can work with it.
@@ -543,7 +583,8 @@ class ClusterPicking(Component):
             return self._response_for_start_line(
                 move_line, message=self.msg_store.lot_multiple_packages_scan_package()
             )
-        return self._response_for_scan_destination(move_line)
+        quantity = self._get_prefill_qty(move_line, 1.0)
+        return self._response_for_scan_destination(move_line, qty_done=quantity)
 
     def _scan_line_by_location(self, picking, move_line, location):
         """Location scanned, check if we can work on goods contained into it.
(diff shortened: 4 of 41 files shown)


shopfloor/services/checkout.py
```diff
@@ -352,11 +352,14 @@ def select(self, picking_id):
...
 
-    def _select_lines(self, lines):
+    def _select_lines(self, lines, no_prefill_qty=None):
         for line in lines:
             if line.shopfloor_checkout_done:
                 continue
-            line.qty_done = line.product_uom_qty
+            if no_prefill_qty is None:
+                no_prefill_qty = line.product_uom_qty
+            line.qty_done = self._get_prefill_qty(line, no_prefill_qty)
```
sebalix: I'm not sure to understand this parameter `no_prefill_qty`, is it a boolean or a qty? Just talking about the name, should it be renamed?

shopfloor/services/checkout.py
```diff
@@ -403,12 +411,15 @@ def scan_line(self, picking_id, barcode):
...
+
         if not product:
             packaging = search.packaging_from_scan(barcode)
             if packaging:
-                product = packaging.product_id
-        if product:
-            return self._select_lines_from_product(picking, selection_lines, product)
+                return self._select_lines_from_packaging(
+                    picking, selection_lines, packaging
+                )
```
sebalix: As you put a `if product` above, we could remove the `if not product` right after to get only:
```python
            packaging = search.packaging_from_scan(barcode)
            if packaging:
                return self._select_lines_from_packaging(
                    picking, selection_lines, packaging
                )
```

shopfloor/services/checkout.py
mmequignon: It is a quantity that represents the quantity to add to `qty_done` when the `no_prefill_qty` option is set to False
Do you think it needs to be renamed? Do you have a suggestion?

shopfloor/services/zone_picking.py
```diff
@@ -765,6 +797,14 @@ def set_destination(
         * set_line_destination+confirmation_required: the scanned location is not
           in the expected one but is valid (in picking type's default destination)
         """
+        if self.work.menu.no_prefill_qty:
+            # While a product, a lot, a packaging …  is scanned, increment the
+            # quantity_done.
+            # Otherwise, fallback to the destination package
+            res = self.scan_source(barcode, quantity, confirmation)
```
sebalix: Why do we call the previous endpoint `scan_source` from `set_destination` here? I do not understand the comment.

shopfloor/services/cluster_picking.py
```diff
@@ -564,6 +617,9 @@ def _scan_line_by_location(self, picking, move_line, location):
         return self._response_for_scan_destination(move_line)
 
     def scan_destination_pack(self, picking_batch_id, move_line_id, barcode, quantity):
+        # TODO rewrite
```
jbaudoux: Like something is still lacking here?

shopfloor/services/location_content_transfer.py
```diff
@@ -542,6 +545,21 @@ def scan_package(self, location_id, package_level_id, barcode):
...
+        """Either increments qty_done or sets it depending on the `no_prefill_qty`
+        option value.
+        """
+        qty = self._get_prefill_qty(move_line, qty)
+        if self.work.menu.no_prefill_qty:
+            move_line.qty_done += qty
+        else:
+            move_line.qty_done = qty
+
+    def _get_prefill_qty(self, move_line, qty=False):
```
jbaudoux: It should not be supported by this scenario as the qty done is always set on all moves at the beginning.
Quantity should be managed when unloading and not when loading, so completely different from what was done on the other scenario.
Should be dropped from this scenario

shopfloor/services/zone_picking.py
jbaudoux: The step is calling set_destination on scan but you can continue to scan the source to increment the qty of the scanned source instead of scanning a destination

jbaudoux (changes requested):

shopfloor/actions/search.py
```diff
@@ -34,15 +34,11 @@ def picking_from_scan(self, barcode):
             return model.browse()
         return model.search([("name", "=", barcode)], limit=1)
 
-    def product_from_scan(self, barcode, use_packaging=True):
+    def product_from_scan(self, barcode):
```
simahawk, line 37: I would not remove this on 14.0 to not risk to break things out there.
Please restore this key and log a deprecation msg.

simahawk: Although I don't see why we should drop it... can you mention a reason?

shopfloor/services/checkout.py
```diff
@@ -352,22 +352,42 @@ def select(self, picking_id):
...
+    def _select_lines(self, lines, no_prefill_qty=None):
         for line in lines:
             if line.shopfloor_checkout_done:
                 continue
-            line.qty_done = line.product_uom_qty
+            if no_prefill_qty is None:
+                qty = line.product_uom_qty
+            else:
+                qty = no_prefill_qty
+            self._set_or_increment_qty_done(line, qty)
```
simahawk: `no_prefill_qty` seems to be a flag but is used below as a value.... I would rename it to `prefill_qty` and default it to -1, then
```suggestion
            qty = line.product_uom_qty if prefill_qty > 0 else prefill_qty
            self._set_or_increment_qty_done(line, qty)
```

shopfloor/services/checkout.py
```diff
@@ -352,22 +352,42 @@ def select(self, picking_id):
...
 
     def _deselect_lines(self, lines):
         lines.filtered(lambda l: not l.shopfloor_checkout_done).write(
             {"qty_done": 0, "shopfloor_user_id": False}
         )
 
+    def _set_or_increment_qty_done(self, move_line, qty=False):
+        """Either increments qty_done or sets it depending on the `no_prefill_qty`
+        option value.
+        """
```
```suggestion
    def _set_qty_done(self, move_line, qty=False):
        """Update qty done on give move line.

        Increase the qty or sets it depending on the `no_prefill_qty` option value.
        """
```

shopfloor/services/checkout.py
```diff
@@ -427,6 +454,8 @@ def _select_lines_from_package(self, picking, selection_lines, package):
                 },
             )
         self._select_lines(lines)
+        if self.work.menu.no_prefill_qty:
+            return self._response_for_select_package(picking, picking.move_line_ids)
```
```suggestion
        if self.work.menu.no_prefill_qty:
           lines = picking.move_line_ids
```

shopfloor/services/checkout.py
```diff
@@ -463,7 +492,49 @@ def _select_lines_from_product(self, picking, selection_lines, product):
             # package and we have only one.
             return self._select_lines_from_package(picking, selection_lines, packages)
 
-        self._select_lines(lines)
+        self._select_lines(lines, no_prefill_qty=1.0)
+        if self.work.menu.no_prefill_qty:
+            return self._response_for_select_package(picking, picking.move_line_ids)
+        return self._response_for_select_package(picking, lines)
```
```suggestion
        if self.work.menu.no_prefill_qty:
            lines = picking.move_line_ids
        return self._response_for_select_package(picking, lines)
```

shopfloor/services/checkout.py
```diff
@@ -463,7 +492,49 @@ def _select_lines_from_product(self, picking, selection_lines, product):
...
+                picking, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
+            )
+
+        lines = selection_lines.filtered(lambda l: l.product_id == product)
+        if not lines:
+            return self._response_for_select_line(
+                picking,
+                message={
+                    "message_type": "error",
+                    "body": _("Product is not in the current transfer."),
```
simahawk: don't we have an exiting message for this?

shopfloor/services/checkout.py
```diff
@@ -463,7 +492,49 @@ def _select_lines_from_product(self, picking, selection_lines, product):
...
+                picking, message=self.msg_store.product_multiple_packages_scan_package()
+            )
+        elif packages:
+            # Select all the lines of the package when we scan a product in a
+            # package and we have only one.
+            return self._select_lines_from_package(picking, selection_lines, packages)
+
+        self._select_lines(lines, no_prefill_qty=packaging.qty)
+        if self.work.menu.no_prefill_qty:
+            return self._response_for_select_package(picking, picking.move_line_ids)
```
```suggestion
        if self.work.menu.no_prefill_qty:
            lines = picking.move_line_ids
```

shopfloor/services/checkout.py
```diff
@@ -495,7 +566,9 @@ def _select_lines_from_lot(self, picking, selection_lines, lot):
             # package and we have only one.
             return self._select_lines_from_package(picking, selection_lines, packages)
 
-        self._select_lines(lines)
+        self._select_lines(lines, no_prefill_qty=1.0)
+        if self.work.menu.no_prefill_qty:
+            return self._response_for_select_package(picking, picking.move_line_ids)
```
```suggestion
            lines = picking.move_line_ids
```

shopfloor/services/checkout.py
```diff
@@ -751,6 +824,11 @@ def scan_package_action(self, picking_id, selected_line_ids, barcode):
         * summary: if there is no other lines, go to the summary screen to be able
         to close the stock picking
         """
+        if self.work.menu.no_prefill_qty:
+            res = self.scan_line(picking_id, barcode)
```
simahawk: comment pls

shopfloor/services/cluster_picking.py
```diff
@@ -456,6 +456,11 @@ def scan_line(self, picking_batch_id, move_line_id, barcode):
         if product and move_line.product_id == product:
             return self._scan_line_by_product(picking, move_line, product)
 
+        if not product:
+            packaging = search.packaging_from_scan(barcode)
+            if packaging and move_line.product_id == product:
```
simahawk: meaning that `move_line.product_id` is not set? I don't think it can happen... Maybe better w/ `packaging.product_id`. Also, it tells me this case is not tested -> please add a test!

shopfloor/services/cluster_picking.py
```diff
@@ -474,8 +479,27 @@ def scan_line(self, picking_batch_id, move_line_id, barcode):
             move_line, message=self.msg_store.barcode_not_found()
         )
 
+    def _set_or_increment_qty_done(self, move_line, qty=False):
```
simahawk: rewrite it as suggested above. Also, move these methods to an helper or to a mixin instead of copy/pasting it around everywhere.

shopfloor/services/cluster_picking.py
simahawk: :man_shrugging:

shopfloor/services/zone_picking.py
```diff
@@ -477,7 +477,11 @@ def _scan_source_location(self, barcode, confirmation=False):
             match_user=True,
         )
         if move_lines:
+            move_line = first(move_lines)
             response = self._response_for_set_line_destination(first(move_lines))
+            response["data"]["set_line_destination"]["move_line"][
```
simahawk: this is quite ugly. Please find a better way (eg: pass to all `self._response_for*` a `data_for_move_line` optional arg to extend update the value returned by `self._data_for_move_line`

shopfloor/services/zone_picking.py
```diff
@@ -507,7 +511,11 @@ def _scan_source_package(self, barcode, confirmation=False):
             return response, message
         move_lines = self._find_location_move_lines(package=package)
         if move_lines:
-            response = self._response_for_set_line_destination(first(move_lines))
+            move_line = first(move_lines)
+            response = self._response_for_set_line_destination(move_line)
```
simahawk: ugly as above

shopfloor/services/zone_picking.py
```diff
@@ -535,23 +543,43 @@ def _scan_source_package(self, barcode, confirmation=False):
...
+            # Since there's no recover feature in this scenario, do not write
+            # qty_done on the backend, but send it to the frontend in the response
+            # instead
+            move_line = first(move_lines)
+            response = self._response_for_set_line_destination(move_line)
+            if packaging:
+                qty = packaging.qty
+            else:
+                qty = 1.0
+            response["data"]["set_line_destination"]["move_line"][
```
simahawk: ugly as above

shopfloor/services/zone_picking.py
```diff
@@ -535,23 +543,43 @@ def _scan_source_package(self, barcode, confirmation=False):
...
-            response = self._response_for_set_line_destination(first(move_lines))
+            # Since there's no recover feature in this scenario, do not write
+            # qty_done on the backend, but send it to the frontend in the response
+            # instead
+            move_line = first(move_lines)
+            response = self._response_for_set_line_destination(move_line)
+            if packaging:
+                qty = packaging.qty
+            else:
+                qty = 1.0
```
simahawk:
```suggestion
            qty = packaging.qty or 1.0
```

shopfloor/services/zone_picking.py
```diff
@@ -535,23 +543,43 @@ def _scan_source_package(self, barcode, confirmation=False):
...
+            response["data"]["set_line_destination"]["move_line"][
+                "qty_done"
+            ] = qty_done + self._get_prefill_qty(move_line, qty=qty)
         else:
             response = self._list_move_lines(self.zone_location)
             message = self.msg_store.product_not_found()
         return response, message
 
-    def _scan_source_lot(self, barcode, confirmation=False):
+    def _scan_source_lot(self, barcode, qty_done, confirmation=False):
```
simahawk: if someone has overridden these methods you'll break their code. Make it optional in all methods alike.

shopfloor/services/zone_picking.py
simahawk: After seeing this for way too many times I think is ugly as the other hack. 
The comment does not explain what you typed here.

shopfloor/tests/test_zone_picking_select_line.py
```diff
@@ -526,3 +532,102 @@ def test_list_move_lines_empty_location(self):
...
+        self.assert_response_set_line_destination(
+            response,
+            zone_location=self.zone_location,
+            picking_type=self.picking_type,
+            move_line=move_line,
+            qty_done=1.0,
+        )
+
+    def test_scan_source_barcode_packaging_no_prefill(self):
+        """When a packaging is scanned, qty_done in response is
```
simahawk: shorten this

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -112,10 +113,35 @@ export var PackagingQtyPickerMixin = {
...
+        },
     },
     created: function () {
+        this.qty_done = parseInt(this.opts.qty_done, 10);
         this.qty_todo = parseInt(this.opts.init_value, 10);
-        this.qty_done = parseInt(this.opts.init_value, 10);
+        this.qty_todo_by_pkg = this.compute_qty_todo_by_pkg();
+    },
+    updated: function () {
+        if (this.$route.params.state !== "change_quantity") {
```
simahawk: this is specific to a scenario :red_circle:

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -266,6 +294,9 @@ export var PackagingQtyPicker = Vue.component("packaging-qty-picker", {
                             @blur="$event.target.value=qty_by_pkg[pkg.id]"
                             />
                     </v-col>
+                    <v-col cols="1" md="2">
+                        <span class="qty-todo">/ {{ get_qty_todo_by_pkg(pkg.id) }}</span>
```
simahawk: I'd like to see a screenshot of the result

shopfloor_mobile/static/wms/src/scenario/checkout.js
```diff
@@ -65,6 +65,7 @@ const Checkout = {
                     :select_records="state.data.picking.move_lines"
                     :select_records_grouped="utils.wms.group_lines_by_location(state.data.picking.move_lines, {'prepare_records': utils.wms.only_one_package})"
                     :select_options="select_line_manual_select_opts()"
+                    :use_qty_todo_by_pkg="true"
```
simahawk: as per select options I would encapsulate this setting for the qty picker into `:qty_picker_options`

shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
```diff
@@ -34,8 +34,11 @@ const ClusterPicking = {
                 v-if="state_in(['start_line', 'scan_destination', 'change_pack_lot', 'stock_issue'])"
                 :line="state.data"
                 :article-scanned="state_is('scan_destination')"
-                :show-qty-picker="state_is('scan_destination')"
+                :show_qty_picker="state_is('scan_destination')"
+                :show_qty_picker_display="state_is('start_line')"
+                :use_qty_todo_by_pkg="true"
```
simahawk: `:qty_picker_options`

simahawk (changes requested):

jbaudoux:
@JuMiSanAr

## Zone picking

### 1/ Scan location - wrong packaging qty initialized
- [x] The quantity for the packaging is not initialized when you start by scanning a location:
image
The packaging is btw also wronly displayed. It should be 3 transport boxes & 0 units

### 2/ Location, then product Internal error
- [x] Scan the location to open the line
Scan the product to increase the quantity
```
  File "wms/shopfloor/services/zone_picking.py", line 813, in set_destination
    accept_only_package = not self._move_line_full_qty(move_line, quantity)
  File "wms/shopfloor/services/zone_picking.py", line 707, in _move_line_full_qty
    move_line.product_uom_qty - qty, precision_rounding=rounding
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
```

### 3/ Wrong package lost quantity + internal error
- [x] Have a line with qty > 1
Scan the right product to open the line.
Scan a wrong product (a barcode of another existing product)
You get a message but the previous quantity is lost
Scan again the right product, you get:
```
  File "wms/shopfloor/services/zone_picking.py", line 636, in scan_source
    response, message = handler(barcode, qty_done, confirmation=confirmation)
  File "wms/shopfloor/services/zone_picking.py", line 575, in _scan_source_product
    ] = qty_done + self._get_prefill_qty(move_line, qty=qty)
TypeError: unsupported operand type(s) for +: 'NoneType' and 'float'
```

### 4/ Edit quantity issue
- [ ] When you are processing a line, I cannot change anymore the content of any quantity fields

### 5/ Location scan should not increase quantity
- [ ] When you scan the location to open the product to process, the quantity should be initialized to 0

### 6/ Scan product update the quantity
- [ ] When you scan the product, the quantity is increased in the backend but not on the front-end


## Cluster picking
### 1/ Readonly issue

- [x] When the line to process is shown with quantity widget in readonly, the packaging info has been lost.
Before:
image
With this PR, no prefill option inactive
image
With this PR, no prefill option active
image

### 2/ Edit quantity issue
- [x] On the next screen once you have selected the product, I cannot change anymore the content of any quantity fields

### 3/ Display issue
- [ ] Can you improve the display to prevent line return in the packaging quantity
image

## Checkout
### 1/ Increased scanned quantity
- [ ] When the product is in a source package (e.g. 2 units in a PACK01), each time I scan the product the quantity is increased by the total quantity (i.e. 2) instead of 1

### 2/ Coloring
- [x] Lines that are not processed are white, lines that are properly processed are green.
Now that you increase step by step the quantity, the line should be colored differently is taken < requested or taken > requested, like on the quantity widget where you have a light red or strong red on the quantity input

### 3/ Edit quantity and split
- [ ] If I edit the quantity with the pencil and decrease the quantity, the line is split in 2 while only the taken quantity (done quantity) should be decreased. So if the source & destination pack is the same, no split should occur.
Before:
ProdA | src pkg: PACK01 | dest pck: PACK01 | reserved: 2 | done: 2
After quantity edit, expected:
ProdA | src pkg: PACK01 | dest pck: PACK01 | reserved: 2 | done: 1
and not:
ProdA | src pkg: PACK01 | dest pck: | reserved: 1 | done: 0
ProdA | src pkg: PACK01 | dest pck: PACK01 | reserved: 1 | done: 1

shopfloor/services/checkout.py
TDu: It has been changed into prefill_qty

shopfloor/services/checkout.py
TDu: done

shopfloor/services/checkout.py
TDu: done

shopfloor/services/checkout.py
TDu: done

shopfloor/services/checkout.py
TDu: done

shopfloor/services/checkout.py
TDu: done

shopfloor/services/checkout.py
TDu: Not exactly, so I created a new one

shopfloor/services/checkout.py
TDu: done

shopfloor/services/checkout.py
TDu: done

TDu:
Rebased on the 14.0 branch

shopfloor/services/zone_picking.py
TDu: done

shopfloor/services/zone_picking.py
TDu: done

shopfloor/services/zone_picking.py
TDu: It should be the job of the front end to call the correct endpoint, IMO.

shopfloor/actions/search.py
TDu: It is dropped because now it is handled by the `packaging_from_scan` in the specific cases.

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -404,3 +404,49 @@ def test_scan_line_error_not_found(self):
...
+        )
+        qty_done = response["data"]["scan_destination"]["qty_done"]
+        self.assertEqual(qty_done, expected_qty_done)
+
+    def test_scan_line_pack_no_prefill_ok(self):
+        """Check qty done when pack is scanned and no_prefill_qty is enabled"""
+        self._enable_no_prefill()
+        self._simulate_batch_selected(self.batch, in_package=True)
+        line = self.batch.picking_ids.move_line_ids
+        self._assert_qty_done(line, line.package_id.name, 0.0)
```
TDu: Shouldn't it be ?
```suggestion
        self._assert_qty_done(line, line.package_id.name, 1.0)
```

shopfloor/services/cluster_picking.py
```diff
@@ -456,6 +456,11 @@ def scan_line(self, picking_batch_id, move_line_id, barcode):
         if product and move_line.product_id == product:
             return self._scan_line_by_product(picking, move_line, product)
 
+        if not product:
+            packaging = search.packaging_from_scan(barcode)
```
TDu, line 460: Can we have a packaging without a package (search for on line 450) because if the answer is no this will line will never be reached !

jbaudoux: yes, that's the product packaging, like the barcode on the box (from the manufacturer and set and product master data)

TDu: oops, sorry I deleted my comment before seeing your answer. Which I thank you for.
it was (can we have a packaging without a package searched 10 lines above)

shopfloor/services/cluster_picking.py
TDu: Fixed

shopfloor/services/cluster_picking.py
TDu: fixed

shopfloor/services/location_content_transfer.py
TDu: It has been removed

shopfloor/services/zone_picking.py
TDu: fixed

shopfloor/services/zone_picking.py
TDu: done

shopfloor/services/cluster_picking.py
TDu: I think this is a left over comment because all the quantity increments are made in the _scan_line_by.... and it should be removed.

shopfloor/services/cluster_picking.py
```diff
@@ -583,6 +637,13 @@ def scan_destination_pack(self, picking_batch_id, move_line_id, barcode, quantit
         have the same destination.
         * start_line: to pick the next line if any.
         """
+        if self.work.menu.no_prefill_qty:
+            res = self.scan_line(picking_batch_id, move_line_id, barcode)
+            # If we got an error from scan line, fall back to finding
+            # the destination package
+            message_type = res.get("message", {}).get("message_type")
+            if message_type != "error":
+                return res
```
TDu, line 646: I see no relation to this change and the specs ?
Commenting this code does not make the tests fail ?

TDu: I see no relation to this change and the specs ?
Commenting this code does not make the tests fail ?

TDu: That is because the test are all based on `scan_line`

shopfloor/services/cluster_picking.py
```diff
@@ -501,6 +523,34 @@ def _scan_line_by_product(self, picking, move_line, product):
...
+        # Do not use mapped here: we want to see if we have more than one package,
+        # but also if we have one product as a package and the same product as
+        # a unit in another line. In both cases, we want the user to scan the
+        # package.
+        if packages and len({line.package_id for line in other_product_lines}) > 1:
+            return self._response_for_start_line(
+                move_line,
+                message=self.msg_store.product_multiple_packages_scan_package(),
+            )
+        self._set_or_increment_qty_done(move_line, packaging.qty)
```
TDu: This function will modify the qty_done on the move_line ! This is happening before any work is done by the user. This will lead to errors

jbaudoux:
I also included here the fix for the sorting of the packaging when you have a packaging of qty=1.
And I also fixed the qty reserved & available in the picking info display.

jbaudoux:
@TDu This test is failing test_scan_package_action_scan_product2_to_increment_qty
I have disabled it but needs fixing when you tackle checkout scenario issues

simahawk:
I don't have time to get back for a review. I trust JE in this case :wink: 
In any case, please, do not forget to squash and cleanup commits (eg: "removed pdb" :stuck_out_tongue_winking_eye: ).

shopfloor/actions/search.py
simahawk, line 37: why this change?

shopfloor/data/shopfloor_scenario_data.xml
```diff
@@ -1,4 +1,4 @@
-<odoo noupdate="1">
+<odoo noupdate="0">
```
simahawk, line 1: Do not change this pls.
Use a migration step if needed.

shopfloor/services/checkout.py
```diff
@@ -352,11 +352,16 @@ def select(self, picking_id):
...
 
-    def _select_lines(self, lines):
+    def _select_lines(self, lines, prefill_qty=0):
         for line in lines:
             if line.shopfloor_checkout_done:
                 continue
-            line.qty_done = line.product_uom_qty
+            if self.work.menu.no_prefill_qty:
+                line.qty_done += prefill_qty
+                # For prefill quantity we only want to increment one line
```
simahawk, line 361: I would not change the initial value in any case. I'd use `for i, line in enumerate(lines)` and do it only for i==0 when needed. 

Yet, I don't get why it has to be done on one line only but I might miss some implementation details.

shopfloor/services/checkout.py
```diff
@@ -662,6 +706,25 @@ def _switch_line_qty_done(self, picking, selected_lines, switch_lines):
                 lambda l: l.product_uom_qty,
             )
 
+    def _increment_custom_qty(
+        self, picking, selected_line, increment_lines, qty_increment
+    ):
+        """Increment the  qty_done of a move line with a custom value
+
+        The selected_line_ids parameter is used to keep the selection of lines
```
simahawk, line 714: `selected_line_ids` is nowhere to be seen

shopfloor/services/cluster_picking.py
```diff
@@ -100,8 +100,11 @@ def _response_for_start_line(self, move_line, message=None, popup=None):
             popup=popup,
         )
 
-    def _response_for_scan_destination(self, move_line, message=None):
-        data = self._data_move_line(move_line)
+    def _response_for_scan_destination(self, move_line, message=None, qty_done=None):
+        if qty_done is None:
+            data = self._data_move_line(move_line)
+        else:
+            data = self._data_move_line(move_line, qty_done=qty_done)
```
simahawk, line 107:
```suggestion
        data = self._data_move_line(move_line, qty_done=qty_done)
```

shopfloor/services/cluster_picking.py
```diff
@@ -474,6 +480,14 @@ def scan_line(self, picking_batch_id, move_line_id, barcode):
             move_line, message=self.msg_store.barcode_not_found()
         )
 
+    def _get_prefill_qty(self, move_line, qty=False):
+        """Returns the increment qty or qty_done depending on the no_prefill_qty
```
simahawk, line 484: pls follow this convention for multiline docstrings https://peps.python.org/pep-0257/#multi-line-docstrings

shopfloor/services/location_content_transfer.py
```diff
@@ -504,6 +504,9 @@ def scan_package(self, location_id, package_level_id, barcode):
         other_move_lines = move_lines - package_move_lines
 
         product = search.product_from_scan(barcode)
+        if not product:
```
simahawk, line 507: these changes should be done in different commits...

shopfloor/services/zone_picking.py
```diff
@@ -211,10 +211,12 @@ def _response_for_set_line_destination(
         move_line,
         message=None,
         confirmation_required=False,
+        **kw,
     ):
         if confirmation_required and not message:
             message = self.msg_store.need_confirmation()
         data = self._data_for_move_line(move_line)
+        data["move_line"].update(kw)
```
simahawk, line 219: why not passing `kw` to `_data_for_move_line`?

shopfloor/tests/test_checkout_scan_package_action_no_prefill_qty.py
```diff
@@ -0,0 +1,107 @@
...
+            move_line,
+            {move_line: origin_qty_done + 1},
+        )
+
+    def test_scan_package_action_scan_product2_to_increment_qty(self):
+        """
+        two lines with the same product
+
+        """
+        # FIXME - test to be fixed
```
simahawk, line 42: ?

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -1,28 +1,39 @@
 /**
  * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
  * @author Simone Orsi <simahawk@gmail.com>
- * Copyright 2021 Jacques-Etienne Baudoux (BCIM)
+ * Copyright 2021-2022 Jacques-Etienne Baudoux (BCIM)
```
simahawk, line 4: please move these changes to a specific commit, better, to a specific PR

simahawk (changes requested):

TDu:
To restore (kind of) atomic commits I reopened a new version of this **https://github.com/OCA/wms/pull/483**
Where I will also attend the last comments from @simahawk

shopfloor/actions/search.py
TDu: It is dropped because now it is handled by the packaging_from_scan in the specific cases.

shopfloor/services/checkout.py
TDu: Good point about not changing the initial value.

shopfloor/services/checkout.py
TDu: typo fixed

shopfloor/services/cluster_picking.py
TDu: If the parameter is present in the call the `qty_done` in the data will be changed. We don't want that

shopfloor/services/cluster_picking.py
TDu: done

shopfloor/tests/test_checkout_scan_package_action_no_prefill_qty.py
TDu: activated

TDu: If I am not mistaken this test is not relevant anymore here (maybe implemented next) @jbaudoux

shopfloor/actions/search.py
sebalix: The story is:

1. initially `product_from_scan` was searching both on product and packaging (without any choice)
2. when we introduced the scan of packaging as a dedicated flow in the `delivery` scenario, I added `packaging_from_scan` and needed to add the `use_packaging` option on `product_from_scan` to disable it without breaking other scenarios.

Now if we are sure we are not breaking anything by removing this parameter... OK to remove it and use `packaging_from_scan` in other scenarios to make things clear.

TDu: All instance of it have been changed.

sebalix:
I close this one then.

shopfloor/actions/search.py
sebalix: If the goal (not of this PR, but the general idea) is that all scenarios use this scan handler, we do not solve anything by scanning the product+packaging in the same place where some scenarios have a different flow regarding the packaging (like checkout).
We can keep it like this but we need to be aware of that.

PR #423, opened 2022-07-06
[WIP][14.0] stock_storage_type: set putaway dest. on package_levels

sebalix:
For packages containing several products, the putaway destination has to be set manually by clicking on the "Recompute putaway destination" button while it is done automatically for simple moves (or package levels with only one product/move).
This commit aims to set the putaway destination automatically when the underlying moves are assigned.

I am not sure about this PR yet, and it introduces regressions (a test has been added to raise it). Is it something we want? I don't know why the putaway of a package level has to be computed manually currently, I miss some info.

cc @jbaudoux


sebalix:
Closing in favor of #424 424

PR #427, opened 2022-07-21
Add sublocation scan on shopfloor delivery scenario

TDu:
When scanning a product barcode or a lot the scenario will search
for all related move lines and pick the first one by default.

For a shopfloor menu configuration with locations containing sub locations
this is not ideal. Because a user working at a sub location only wants
move lines related to that sub location to be selected.

To help with this situation, a user can first scan his working location
which is kept as long as no other location is scanned.
The sub location set can also be reset by the click of a button.

Also when the search for a move line to work on returns multiple move lines
related to different locations and no location has been preliminary scanned.
The user will be prompted to scann a location first.

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -421,6 +421,29 @@ class MessageAction(Component):
             ).format(location_dest.name),
         }
 
+    def product_in_multiple_sublocation(self, product):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "Product {} found in multiple locations. Scan your location first."
+            ).format(product.name),
+        }
+
+    def lot_in_multiple_sublocation(self, lot):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "Lot {lot} for product {product} found in multiple locations. "
+                "Scan your location first."
+            ).format(lot=lot.name, product=lot.product_id.name),
+        }
+
+    def location_src_set_to_sublocation(self, location_src):
+        return {
+            "message_type": "success",
+            "body": _("Working location changed to {}").format(location_src.name),
+        }
+
     def picking_already_started_in_location(self, pickings):
         return {
             "message_type": "error",
--- a/shopfloor/services/delivery.py
+++ b/shopfloor/services/delivery.py
@@ -41,7 +41,7 @@ class Delivery(Component):
     _usage = "delivery"
     _description = __doc__
 
-    def _response_for_deliver(self, picking=None, message=None):
+    def _response_for_deliver(self, picking=None, location=None, message=None):
         """Transition to the 'deliver' state
 
         If no picking is passed, the screen shows an empty screen
@@ -49,7 +49,10 @@ class Delivery(Component):
         return self._response(
             next_state="deliver",
             data={
-                "picking": self.data_detail.picking_detail(picking) if picking else None
+                "picking": self.data_detail.picking_detail(picking)
+                if picking
+                else None,
+                "sublocation": self.data.location(location) if location else None,
             },
             message=message,
         )
@@ -74,13 +77,15 @@ class Delivery(Component):
         return self._response(
             next_state="confirm_done",
             data={
-                "picking": self.data_detail.picking_detail(picking) if picking else None
+                "picking": self.data_detail.picking_detail(picking)
+                if picking
+                else None,
             },
             message=message,
         )
 
-    def scan_deliver(self, barcode, picking_id=None):
-        """Scan a stock picking or a package/product/lot
+    def scan_deliver(self, barcode, picking_id=None, location_id=None):
+        """Scan a stock picking, a package/product/lot or a stock location
 
         When a stock picking is scanned and is partially or fully available, it
         is returned to show its lines.
@@ -88,6 +93,10 @@ class Delivery(Component):
         When a package is scanned, and has an available move line part of the
         expected picking type, the package level is directly set to "done" and
         the stock picking of the line is returned to work on its other lines.
+
+        When a stock location is scanned and it is a sub-location of an operation
+        type allowed on the current shopfloor menu, the next delivery
+        operations will be put into that location.
 
         If the barcode is a product or a product's packaging, the move lines
         for this product are set to done. However, if the product is in more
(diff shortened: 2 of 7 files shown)


shopfloor/services/delivery.py
```diff
@@ -89,6 +94,10 @@ def scan_deliver(self, barcode, picking_id=None):
         expected picking type, the package level is directly set to "done" and
         the stock picking of the line is returned to work on its other lines.
 
+        When a stock location is scanned, and it is a sublocation of an operation
+        type set on the shopfloor current shopfloor menu. The next delivery
+        operations will be localized to that location.
```
simahawk:
```suggestion
        When a stock location is scanned and it is a sub-location of an operation
        type allowed on current shopfloor menu, the next delivery
        operations will be put into that location.
```

shopfloor/services/delivery.py
```diff
@@ -133,19 +148,19 @@ def scan_deliver(self, barcode, picking_id=None):
...
-                return self._deliver_package(picking, package)
+                return self._deliver_package(picking, package, location)
 
         if not barcode_valid:
             product = search.product_from_scan(
                 barcode, use_packaging=(not allow_prepackaged_product)
             )
             if product:
-                return self._deliver_product(picking, product)
+                return self._deliver_product(location, picking, product)
```
simahawk: I'd keep location at the end as for the `_deliver_package` and `_deliver_lot`.

shopfloor/services/delivery.py
```diff
@@ -156,11 +171,21 @@ def scan_deliver(self, barcode, picking_id=None):
...
                 )
                 return self._deliver_product(
-                    picking, packaging.product_id, product_qty=packaging_qty
+                    location, picking, packaging.product_id, product_qty=packaging_qty
                 )
 
+        if not barcode_valid:
+            sublocation = search.location_from_scan(barcode)
+            if sublocation and sublocation.is_sublocation_of(
+                self._get_process_picking_types().mapped("default_location_src_id")
```
simahawk:
```suggestion
                self.picking_types.mapped("default_location_src_id")
```

shopfloor/services/delivery.py
```diff
@@ -229,16 +255,27 @@ def _lines_base_domain(self, no_qty_done=True):
...
+            [
+                self._lines_base_domain(no_qty_done),
+                [("lot_id", "=", lot.id)],
+                location_domain,
+            ]
         )
 
-    def _lines_from_product_domain(self, product, no_qty_done=True, product_qty=None):
+    def _lines_from_product_domain(
+        self, location, product, no_qty_done=True, product_qty=None
```
simahawk: keep location optional as kw arg

shopfloor/services/delivery.py
```diff
@@ -256,21 +293,35 @@ def _lines_from_package_domain(self, package, no_qty_done=True):
             [self._lines_base_domain(no_qty_done), [("package_id", "=", package.id)]]
         )
 
-    def _deliver_product(self, picking, product, product_qty=None):
+    def _deliver_product(self, location, picking, product, product_qty=None):
         if product.tracking in ("lot", "serial"):
```
simahawk, line 300: move it to the end to respect other signature and - maybe - make it kw arg (also not to break existing overrides if any)

shopfloor/services/delivery.py
```diff
@@ -256,21 +293,35 @@ def _lines_from_package_domain(self, package, no_qty_done=True):
...
-                picking, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
+                picking,
+                location=location,
+                message=self.msg_store.scan_lot_on_product_tracked_by_lot(),
             )
 
         limit = 1 if product_qty else None
         lines = self.env["stock.move.line"].search(
-            self._lines_from_product_domain(product, product_qty=product_qty),
+            self._lines_from_product_domain(location, product, product_qty=product_qty),
```
simahawk: move arg, see above

shopfloor_mobile/static/wms/src/scenario/delivery.js
```diff
@@ -12,6 +12,7 @@ const Delivery = {
     template: `
         <Screen :screen_info="screen_info">
             <template v-slot:header>
+                <state-display-info :info="current_location_msg()" v-if="current_location()"/>
```
simahawk, line 15: this msg should be set on state.display_info likely. Any reason to keep it separated?

shopfloor_mobile/static/wms/src/scenario/delivery.js
```diff
@@ -37,6 +38,11 @@ const Delivery = {
                     />
             </div>
             <div class="button-list button-vertical-list full">
+                <v-row align="center" v-if="state_is('deliver') && is_current_location_set()">
+                    <v-col class="text-center" cols="12">
+                        <btn-action @click="this.reset_current_location">Reset location</btn-action>
```
simahawk:
```suggestion
                        <btn-action @click="reset_current_location">Reset location</btn-action>
```

simahawk (changes requested):

shopfloor_mobile/static/wms/src/scenario/delivery.js
TDu: The message is set on display_info for the answer of a location scan (`scan_deliver` endpoint)

image

But for all other calls it is not because it would require more changes on the back-end (each endpoint would need to receive the location id to send the information back).
But the front-end always knows on what sub-location it is working on by looking at the answer received by the last call to `scan_deliver`. And the sub-location is always displayed to the user.

image

simahawk: > But for all other calls it is not because it would require more changes on the back-end (each endpoint would need to receive the location id to send the information back).

That's not true, there's a trick ;)
For zone picking I used an header for all the info that were mostly the same across the whole scenario.
Check these lines:

https://github.com/OCA/wms/blob/14.0/shopfloor/services/zone_picking.py#L95-L154

Client side you need this https://github.com/OCA/wms/blob/14.0/shopfloor_mobile/static/wms/src/scenario/zone_picking.js#L225

i-vyshnevska (approved):

jbaudoux (changes requested):
Tested and works properly.
The input label needs to be changed into "Scan pack / product / picking / location" like in the checkout

jbaudoux (approved):

jbaudoux:
@simahawk Can this be merged now?

jbaudoux:
@TDu Can you squash?

simahawk (approved):

shopfloor/actions/message.py
```diff
@@ -421,6 +421,28 @@ def location_content_unable_to_transfer(self, location_dest):
...
+                "Product {} found in multiple locations. Scan your location first."
+            ).format(product.name),
+        }
+
+    def lot_in_multiple_sublocation(self, lot):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "Lot {} for product {} found in multiple locations. Scan your location first."
+            ).format(lot.name, lot.product_id.name),
```
sebalix: Not blocking, but term places could be swapped in some languages, better to use keys to place them:
```suggestion
            "body": _(
                "Lot {lot} for product {product} found in multiple locations. Scan your location first."
            ).format(lot=lot.name, product=lot.product_id.name),
```

sebalix (approved):

shopfloor/actions/message.py
TDu: You have a point, I added the change.
And rebased.

PR #428, opened 2022-07-21
[WIP][14.0][ADD] shopfloor_inventory: inventory scenario

bguillot:


TDu:
I think the idea now is to have new scenario in their own module like `shopfloor_delivery_shipment` and `shopfloor_delivery_shipment_mobile` for examples.

alexis-via:
This PR would deserve to be re-open. We need an inventory scenario for shopfloor !

rousseldenis:
@bguillot @alexis-via What's the status of this ? As indeed, we need an inventory scenario in shopfloor :smiley:

luke-stdev001:
Is there any way we can understand the status of the work on this PR, as we would like to use it to base our migration to 16.0 for our own use and submission back to the OCA after this is complete. It's the missing piece for us to replace the core barcode app with something more robust, but I would just like to understand if this PR is ready to merge in 14.0 so that we can gauge what needs to be done before migration to 16.0.

PR #430, opened 2022-07-26
[14.0] shopfloor_base: fix cross worker problem

simahawk:
Depends on:

- [ ] https://github.com/OCA/web-api/pull/3

With this change sf apps can be created and made available on the fly w/out restarting when working in cross worker envs.

TODO
- [ ] add mig to enable existing apps on upgrade
- [ ] remove issue from roadmap


shopfloor_base/models/shopfloor_app.py
```diff
@@ -109,9 +89,9 @@ def _compute_url(self):
     @api.depends("tech_name")
     def _compute_registered_routes(self):
         for rec in self:
-            routes = sorted(rec._registered_routes())
+            routes = sorted(rec._registered_routes(), key=lambda x: x.route)
```
mmequignon, line 92: I have the feeling that `_registered_routes()` should return routes, not rules, or be renamed to `_registered_rules()` maybe.
What do you think ?

shopfloor_base/tests/test_shopfloor_app.py
```diff
@@ -49,17 +49,21 @@ def test_app_create(self):
...
-        rec._register_endpoints()
+        rec._register_controllers(init=True)
         routes = rec._registered_routes()
         _check = {}
-        endpoint_func_repr = (
-            "bound method _process_endpoint "
-            "of <odoo.addons.base_rest.controllers.main.RestController"
-        )
-        for __, rule in routes:
+        for rule in routes:
```
mmequignon, line 55:
```python
for rule in rules:
```
related to my previous suggestion

mmequignon (approved):
just nitpicks

shopfloor_base/models/shopfloor_app.py
simahawk: Yeah, TBH I'm still puzzled sometimes regarding the names because all in all the border is quite thin -> we are talking about "routing rules". However, given that this particular method returns `EndpointRule` objects, seems fair.

simahawk:
replaced by https://github.com/OCA/wms/pull/574

PR #431, opened 2022-07-26
[14.0][ADD] stock_warehouse_flow[_release]: manage multiple delivery routes for a single warehouse

sebalix:
New feature allowing to re-qualify the operation type of a stock move at confirmation.
The main criteria to requalify a move is the carrier, as this one often implies a dedicated routing (we do not have the same flow, nor the same carrier, depending if we ship a pallet or a bunch of products for instance).

This job was previously done by configuring the `stock_dynamic_routing` with a domain on the carrier, but the dynamic routing was applied on reservation, while we could already qualify moves at confirmation, or when releasing an operation.

This PR adds two modules:
- `stock_warehouse_flow`: update the operation type of a stock move when this one is confirmed
- `stock_warehouse_flow_release`: update the operation type of a stock move when this one is released

```mermaid
  graph LR;
  PRODUCT(Goods to ship)-->|Pre-packaged pallet?|PICK[Pick]
  subgraph sg1["Full Pallet Routing Flow"]
    PICK-->SHIP[Ship];
  end
  PRODUCT(Goods to ship)-->|Bunch of goods to pack?|PICK2[Pick]
  subgraph sg2["Box/Package Routing Flow"]
    PICK2-->PACK[Pack];
    PACK-->SHIP2[Ship];
  end
  PRODUCT(Goods to ship)-->|Cold chain goods?|PICK3[Pick]
  subgraph sg3["Cold-Chain Routing Flow"]
    PICK3-->PACK2[Pack];
    PACK2-->SHIP3[Ship];
  end
  style sg1 fill:lightgray,stroke:#333
  style sg3 fill:#6495ED,stroke:#333,color:white
```

image

### TODO
- <s>write README</s>
- <s>fix Travis build (issue with `delivery_carrier_preference`)</s>

Ref: 47


stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,195 @@
...
+    from_location_dest_id = fields.Many2one(
+        comodel_name="stock.location",
+        compute="_compute_from_location_dest_id",
+    )
+    to_picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type",
+        ondelete="restrict",
+        string="To operation type",
+        required=True,
+        domain="[('default_location_dest_id', '=', from_location_dest_id)]",
```
jbaudoux, line 50: We need a to_location_src_id, could be a child of the from_location_src_id. This is in order to be able to pull the rule of the flow (and not the default one from the WH, as the procurement is searching a rule according to the location_src)
Then the picking type source location has to be that to_location_src_id.
Ideally, we should be able to generate that picking type as you generate the route. Like the WH config provides you the picking types. And we need one for each delivery step.

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,195 @@
...
+                "name": f"{self.warehouse_id.name}: {self.name}",
+                "active": True,
+                "flow_id": self.id,
+                # Give the priority to routes/rules not tied to a flow
+                # when the initial move is created by the procurement
+                "sequence": 1000,
+            }
+        )
+        for rule_vals in vals["rule_ids"]:
+            # Replace the initial outgoing picking type by the new one
```
jbaudoux: The complete flow needs dedicated picking types for each step

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -323,6 +323,8 @@ def _run_stock_rule(self):
                 new_move = move._release_split(remaining)
                 backorder_links[new_move.picking_id] = move.picking_id
 
+            move._before_release()
```
jbaudoux, line 326: As discussed and in a second improvement phase, investigate if this stock_available_to_promise_release could call again action_confirm on the updated move instead of manually creating a Procurement for the pull chain and run it (which is exactly what action_confirm does). If this is the case, then this new module would not require a glue module anymore and that would reduce the code and complexity

stock_warehouse_flow/models/stock_warehouse_flow.py
sebalix: Is it something we want every time or it's something we need for our project? Why can't we re-use an existing picking type for a pick or pack operation if the work to do is the same than another route or flow?

jbaudoux: Mostly every time. Especially that the number of steps can be different from the warehouse config. So the location src/dest of the warehouse default picking types won't fit anyway.

jbaudoux: In the rare case you would not want a new picking type, you can select an existing one before the route is generated

sebalix: got it

stock_warehouse_flow/models/stock_warehouse_flow.py
sebalix: Operation types are now automatically generated.

stock_warehouse_flow/models/stock_location_route.py
```diff
@@ -0,0 +1,29 @@
...
+
+class StockLocationRoute(models.Model):
+    _inherit = "stock.location.route"
+
+    flow_id = fields.Many2one(
+        comodel_name="stock.warehouse.flow",
+        ondelete="set null",
+        string="Routing Flow",
+    )
+    applicable_flow_ids = fields.One2many(
```
jbaudoux, line 15: What is/was the purpose of this?
The flow defines new picking types for the route's rules, so the search does not make sense to me

sebalix: This is to inform the user that one route could be replaced by another flow (if the criterias are matching), it's not used for other purpose.

jbaudoux: got it :)

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,292 @@
...
+        if self.delivery_route_id:
+            return False
+        # Re-generate the WH configuration with the chosen delivery steps in a
+        # savepoint to get a whole route+rules+picking types configured almost
+        # automatically (to not re-invent the wheel).
+        # Delivery route values are then copied to generate a new one.
+        # Copy some picking types with new sublocations first so the new
+        # delivery route will use them.
+        pick_type, pack_type, out_type = False, False, False
+        out_type = self.warehouse_id.out_type_id.copy(
```
jbaudoux: The new picking types should be generated only if they are not provided in the configuration on the warehouse flow. If they are generated, they should be accessible on the warehouse flow (like you have the picking types on the tech info of a warehouse)

sebalix: Indeed, this still need to be implemented...

stock_available_to_promise_release/models/stock_move.py
sebalix: Checked: `stock_available_to_promise_release` can't call `action_confirm` to do the work because

- releasing a move implies some split of qties
- `action_confirm` accepts only `draft` moves, and the one to release is already `confirmed`

So better to keep the current implementation.

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,292 @@
...
+        compute="_compute_from_location_dest_id",
+    )
+    to_picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type",
+        ondelete="restrict",
+        string="To operation type",
+        readonly=True,
+        domain="[('default_location_dest_id', '=', from_location_dest_id)]",
+    )
+    carrier_ids = fields.Many2many(
```
jbaudoux, line 59: @sebalix Can you add a smart button on the carrier to see/jump to the flow?

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,154 @@
...
+                                type="object"
+                                string="Generate route"
+                                class="oe_highlight"
+                                attrs="{'invisible': [('delivery_route_id', '!=', False)]}"
+                            />
+            </group>
+          </group>
+          <notebook>
+            <page string="Rules" name="rules">
+              <field name="rule_ids" nolabel="1" />
```
jbaudoux: Display from the rules:
* action
* operation type
* location from
* location to
* fixed procurement group (group_id)

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,292 @@
...
+            output_loc = self.warehouse_id.wh_output_stock_loc_id.copy(
+                {
+                    "name": self.sequence_prefix,
+                    "location_id": self.warehouse_id.wh_output_stock_loc_id.id,
+                }
+            )
+            pick_type.default_location_dest_id = output_loc
+            out_type.default_location_src_id = output_loc
+            if "pack_" in self.delivery_steps:
+                pack_type = self.warehouse_id.pack_type_id.copy(
```
jbaudoux: The warehouse picking type could be archived. When you copy, you need to ensure it is not archived anymore

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,292 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class StockWarehouseFlow(models.Model):
+    _name = "stock.warehouse.flow"
+    _description = "Stock Warehouse Routing Flow"
+
+    name = fields.Char(required=True)
+    active = fields.Boolean(default=True)
```
jbaudoux, line 19: need also a sequence for evaluating the flows in the right order if multiple are valid

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,292 @@
...
+        return [
+            ("from_picking_type_id", "=", move.picking_type_id.id),
+            ("carrier_ids", "in", move.group_id.carrier_id.ids),
+        ]
+
+    @api.model
+    def _search_for_move(self, move):
+        """Return the first flow matching the move."""
+        domain = self._search_for_move_domain(move)
+        flow = self.search(domain, limit=1)
```
jbaudoux: We need to loop on each valid flow by sequence
```
+    def _is_valid_for_move(self, move):
+        if not self.move_domain:
+            return self
+        domain = safe_eval(self.move_domain or "[]")
+        if not domain:
+            return self
+        return move.filtered_domain(domain)
+
     @api.model
     def _search_for_move(self, move):
         """Return the first flow matching the move."""
         domain = self._search_for_move_domain(move)
-        flow = self.search(domain, limit=1)
-        if flow and move.filtered_domain(safe_eval(flow.move_domain or "[]")):
-            return flow
+        for flow in self.search(domain):
+            if flow._is_valid_for_move(move):
+                return flow

```

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,292 @@
...
+        if flow and move.filtered_domain(safe_eval(flow.move_domain or "[]")):
+            return flow
+        return self.browse()
+
+    def _apply_on_move(self, move):
+        """Apply the flow configuration on the move."""
+        if not self:
+            return False
+        logger.info("Applying flow '%s' on '%s'", self.name, move)
+        rule = self.delivery_route_id.rule_ids.filtered(
```
jbaudoux: If no rule is found, an error must be raised
A warning on the flow form would be also helpful in this case

stock_warehouse_flow/models/stock_move.py
```diff
@@ -0,0 +1,21 @@
...
+        # its chained moves (if any)
+        for move in self:
+            if not move._apply_flow_on_action_confirm():
+                continue
+            flow = self.env["stock.warehouse.flow"]._search_for_move(move)
+            flow._apply_on_move(move)
+        return super()._action_confirm(merge=merge, merge_into=merge_into)
+
+    def _apply_flow_on_action_confirm(self):
+        return True
```
jbaudoux: Shouldn't we limit that to only deliveries?
```suggestion
        return self.picking_type_id.code == "outgoing"
```

sebalix: Well, when coding the whole module I had in mind that it could also handle reception flows, not only delivery ones (with some new code of course).

sebalix:
@jbaudoux some of your comments have been addressed, here is the form with the changes (sequence field, rule fields and the warning if no rule is found):
image

sebalix:
@jbaudoux technical informations have been added on the flow, like on the WH:
image
I should merge the "To operation type" and "Out type" fields I guess, they mean the same currently... but this could change if one day we include the "Reception steps" configuration, the out type is the one used by default when we return the goods to a supplier.

stock_warehouse_flow/models/delivery_carrier.py
```diff
@@ -0,0 +1,26 @@
...
+    flow_ids = fields.Many2many(
+        comodel_name="stock.warehouse.flow",
+        relation="delivery_carrier_stock_warehouse_flow_rel",
+        column1="delivery_carrier_id",
+        column2="stock_warehouse_flow_id",
+        string="Routing flows",
+        copy=False,
+    )
+
+    def action_view_flows(self):
```
simahawk, line 19: pls add a test, at least w/ a simple call

stock_warehouse_flow/models/stock_location_route.py
```diff
@@ -0,0 +1,29 @@
...
+        comodel_name="stock.warehouse.flow",
+        compute="_compute_relate_to_flow_ids",
```
simahawk:
```suggestion
        compute="_compute_applicable_flow_ids",
```

stock_warehouse_flow/models/stock_warehouse.py
```diff
@@ -0,0 +1,23 @@
...
+class StockWarehouse(models.Model):
+    _inherit = "stock.warehouse"
+
+    flow_ids = fields.One2many(
+        comodel_name="stock.warehouse.flow",
+        inverse_name="warehouse_id",
+        string="Flows",
+    )
+
+    def action_view_all_flows(self):
```
simahawk, line 16: test

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+        string="Warehouse",
+        default=lambda o: o._default_warehouse_id(),
+        required=True,
+    )
+    from_picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type",
+        ondelete="restrict",
+        string="From operation type",
+        required=True,
+        domain="[('code', '=', 'outgoing')]",
```
simahawk, line 35: you said above we support incoming too, seems not the case here, no?

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+    delivery_steps = fields.Selection(
+        [
+            ("ship_only", "Deliver goods directly (1 step)"),
+            ("pick_ship", "Send goods in output and then deliver (2 steps)"),
+            (
+                "pick_pack_ship",
+                "Pack goods, send goods in output and then deliver (3 steps)",
+            ),
+        ],
+        "Outgoing Shipments",
```
simahawk:
```suggestion
        selection=[
            ("ship_only", "Deliver goods directly (1 step)"),
            ("pick_ship", "Send goods in output and then deliver (2 steps)"),
            (
                "pick_pack_ship",
                "Pack goods, send goods in output and then deliver (3 steps)",
            ),
        ],
        string="Outgoing Shipments",
```

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+        help="Outgoing route to follow",
+    )
+    sequence_prefix = fields.Char(
+        help=(
+            "Used to generate the default prefix of new operation types. "
+            "(e.g. 'DG' => 'WH/OUT/DG/')"
+        )
+    )
+    delivery_route_id = fields.Many2one(
+        "stock.location.route",
```
simahawk:
```suggestion
        comodel_name="stock.location.route",
```

stock_warehouse_flow/models/stock_warehouse_flow.py
simahawk: `index=True`?

stock_warehouse_flow/models/stock_warehouse_flow.py
simahawk: `index=True`

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,189 @@
...
+                </tree>
+              </field>
+            </page>
+            <page
+                            string="Technical Information"
+                            name="technical_info"
+                            groups="base.group_no_one"
+                        >
+              <group>
+                <group string="Locations">
```
simahawk: name these groups

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+            return self
+        domain = safe_eval(self.move_domain or "[]")
+        if not domain:
+            return self
+        return move.filtered_domain(domain)
+
+    @api.model
+    def _search_for_move(self, move):
+        """Return the first flow matching the move."""
+        domain = self._search_for_move_domain(move)
```
simahawk, line 348: Wondering: could you make this work w/ a list of moves? They could be grouped by `(picking_type_id, carrier_id)` after search. No?

simahawk (changes requested):

stock_warehouse_flow/models/stock_move.py
jbaudoux: The from_picking_type_id is restricted currently on "outgoing", so I think it is better to be consistent here

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,189 @@
...
+                  <field name="pack_stock_loc_id" readonly="1" />
```
jbaudoux: Can you make locations & picking types readonly only when route is generated (route_id != False) ?

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,189 @@
...
+              <h1><field name="name" placeholder="Flow Name" /></h1>
+          </div>
+          <group>
+            <group name="info" colspan="4" col="4">
+              <field
+                                name="warehouse_id"
+                                attrs="{'readonly': [('delivery_route_id', '!=', False)]}"
+                            />
+              <field name="sequence" />
+              <field name="active" invisible="1" />
```
jbaudoux, line 29: company_id field is missing. It is required to be part of the view as you defined check_company on some fields

Error while validating view:

Field company_id used in field out_type_id default domain ([('company_id', 'in', [company_id, False])]) must be present in view but is missing.

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def _apply_on_move(self, move):
+        """Apply the flow configuration on the move."""
+        if not self:
+            return False
+        logger.info("Applying flow '%s' on '%s'", self.name, move)
+        rule = self._get_rule_from_delivery_route()
+        move.picking_id = False
+        move.picking_type_id = self.to_picking_type_id
+        move.location_id = self.to_picking_type_id.default_location_src_id
```
jbaudoux: As we pull the chain from the output location and that is it now on the flow itself, if we can put a sub-location, we could achieve even more advanced routing with this little change
```suggestion
        move.location_id = self.output_stock_loc_id
```
The location output_stock_loc_id should then never be readonly

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+                    {
+                        "name": self.sequence_prefix,
+                        "location_id": self.warehouse_id.wh_pack_stock_loc_id.id,
+                    }
+                )
+                self.pick_type_id.default_location_dest_id = self.pack_stock_loc_id
+                self.pack_type_id.default_location_src_id = self.pack_stock_loc_id
+                self.pack_type_id.default_location_dest_id = self.output_stock_loc_id
+        # Enable relevant locations (pack/output) based on the delivery steps
+        self.warehouse_id._update_location_delivery(self.delivery_steps)
```
jbaudoux: The warehouse locations should not be modified as the flow has its own intermediate locations
```suggestion
```

stock_warehouse_flow/views/stock_warehouse_flow.xml
jbaudoux: And also based on the number of steps:
1 step, I can define out
2 steps, I can define pick & out + out location
3 steps, I can define pick, pack & out + pack & out locations

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+    output_stock_loc_id = fields.Many2one(
+        "stock.location", "Output Location", check_company=True
+    )
+    pick_type_id = fields.Many2one(
+        "stock.picking.type", "Pick Type", check_company=True
+    )
+    pack_type_id = fields.Many2one(
+        "stock.picking.type", "Pack Type", check_company=True
+    )
+    out_type_id = fields.Many2one("stock.picking.type", "Out Type", check_company=True)
```
jbaudoux: Looks redundant with `to_picking_type_id`
I would drop or replace with a related field to prevent inconsistency

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,189 @@
...
+      </form>
+    </field>
+  </record>
+
+  <record id="stock_warehouse_flow_view_tree" model="ir.ui.view">
+    <field name="name">stock.warehouse.flow.tree</field>
+    <field name="model">stock.warehouse.flow</field>
+    <field name="arch" type="xml">
+      <tree string="Routing Flows" decoration-warning="warning">
+        <field name="name" />
```
jbaudoux, line 161: add warehouse in list view
```suggestion
        <field name="warehouse_id" />
        <field name="name" />
```

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,189 @@
...
+        <field name="from_picking_type_id" />
+        <field name="carrier_ids" widget="many2many_tags" />
```
jbaudoux, line 164: add domain in list view
```suggestion
        <field name="move_domain" />
        <field name="carrier_ids" widget="many2many_tags" />
```

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+                            "name": (
+                                f"{self.warehouse_id.pack_type_id.name} "
+                                f"{self.sequence_prefix}"
+                            ),
+                            "active": True,
+                        }
+                    )
+                    self.pack_type_id.sequence_id.prefix += f"{self.sequence_prefix}/"
+                # Create a dedicated 'Packing Zone/SUB' location
+                self.pack_stock_loc_id = self.warehouse_id.wh_pack_stock_loc_id.copy(
```
jbaudoux: In 3 steps delivery, when you set a pack_stock_loc_id but not a pick type, the pack_stock_loc_id location is overridden by a newly generated one.
This should not occur.
So there is something weird being the scene after doing https://github.com/OCA/wms/pull/431/files#diff-ec16197c7c77f54e4580085fc7fa4bfe23625085a978e8658cd0835e253423a1R270

stock_warehouse_flow/models/stock_warehouse_flow.py
sebalix: Incoming is not supported yet (we do not need it), but this is something the module could support later so I mention this in the ROADMAP file.

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,367 @@
...
+        help=(
+            "Used to generate the default prefix of new operation types. "
+            "(e.g. 'DG' => 'WH/OUT/DG/')"
+        )
+    )
+    delivery_route_id = fields.Many2one(
+        "stock.location.route",
+        string="Delivery Route",
+        ondelete="set null",
+        readonly=True,
```
jbaudoux, line 98: Add copy=False

stock_warehouse_flow/models/stock_warehouse_flow.py
jbaudoux: This needs to be tackled. Changing the output location is the key concept of pulling the new chain of moves (not changing the out picking type). I could keep the warehouse out picking type and only change the output location to alter the pick/pack flow

jbaudoux: reopen https://github.com/OCA/wms/pull/431/files/6a426befaa5569da758340982e48cb4aacb75310#r964760022

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,213 @@
...
+              <group>
+                <group string="Locations" name="locations">
+                  <field
+                                        name="pack_stock_loc_id"
+                                        attrs="{
+                      'readonly': [('delivery_route_id', '!=', False)],
+                      'invisible': [('delivery_steps', '!=', 'pick_pack_ship')],
+                    }"
+                                    />
+                  <field name="output_stock_loc_id" />
```
jbaudoux: maybe move it above inside the "To" section as it is now used as `to_picking_type_id` in `apply_on_moves`

stock_warehouse_flow/models/stock_warehouse_flow.py
sebalix: I like the idea of making it a related+readonly field on `to_picking_type_id` so we get a good picture of the flow configuration in the `Technical Information` tab, like we have on a warehouse form.

stock_warehouse_flow/models/stock_warehouse_flow.py
sebalix: Not sure how to do that, I prefer to keep it like this right now. Let's see if it becomes an issue later on...

jbaudoux (approved):
Thanks for the implementation This is a great feature.
Tested and happy with the result and fixes.

sebalix:
I won't enable the flows by default in the demo data, this makes tests of other modules failing.
EDIT: done

TDu (approved):
If the flow is only related to the delivery. Shouldn't the module have `delivery` in its name ?
Otherwise quick review not functional.

sebalix:
@TDu well, it could support the reception as well later (by adding a type on the flow) without too much change. I don't know if having two modules is good in this case.

PR #432, opened 2022-08-22
[14.0] shopfloor: change search location action

TDu:
Change the search to first look on the barcode field and then on
location name.

### DIFF
--- a/shopfloor/actions/search.py
+++ b/shopfloor/actions/search.py
@@ -14,13 +14,16 @@ class SearchAction(Component):
 
     # TODO: these methods shall be probably replaced by scan anything handlers
 
-    def location_from_scan(self, barcode):
+    def location_from_scan(self, barcode, limit=1):
         model = self.env["stock.location"]
         if not barcode:
             return model.browse()
-        return model.search(
-            ["|", ("barcode", "=", barcode), ("name", "=", barcode)], limit=1
-        )
+        # First search location by barcode
+        res = model.search([("barcode", "=", barcode)], limit=limit)
+        # And only if we have not found through barcode search on the location name
+        if len(res) < limit:
+            res |= model.search([("name", "=", barcode)], limit=(limit - len(res)))
+        return res
 
     def package_from_scan(self, barcode):
         model = self.env["stock.quant.package"]
--- a/shopfloor/tests/test_actions_search.py
+++ b/shopfloor/tests/test_actions_search.py
@@ -22,6 +22,13 @@ class TestSearchCase(TestSearchBaseCase):
         self.assertEqual(handler(False), rec.browse())
         self.assertEqual(handler("NONE"), rec.browse())
 
+    def test_search_location_with_limit(self):
+        rec = self.customer_location
+        rec2 = self.customer_location.sudo().copy({"barcode": "CUSTOMERS2"})
+        handler = self.search.location_from_scan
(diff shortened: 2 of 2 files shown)


shopfloor/actions/search.py
```diff
@@ -14,13 +14,14 @@ class SearchAction(Component):
...
-    def location_from_scan(self, barcode):
+    def location_from_scan(self, barcode, limit=1):
         model = self.env["stock.location"]
         if not barcode:
             return model.browse()
-        return model.search(
-            ["|", ("barcode", "=", barcode), ("name", "=", barcode)], limit=1
-        )
+        res = model.search([("barcode", "=", barcode)], limit=limit)
+        if len(res) < limit:
```
simahawk, line 24: I don't get why you need to play w/ limit :thinking:

shopfloor/actions/search.py
```diff
@@ -14,13 +14,14 @@ class SearchAction(Component):
...
+            res += model.search([("name", "=", barcode)], limit=(limit - len(res)))
```
jbaudoux: I think with += you could return twice the  same location
```suggestion
            res |= model.search([("name", "=", barcode)], limit=(limit - len(res)))
```

TDu: I don't know about having multiple record in a set ? But better use the one listed in the docs https://www.odoo.com/documentation/14.0/developer/reference/addons/orm.html#operations.
Changed

shopfloor/actions/search.py
TDu: This was a requirement, @jbaudoux should know the underlying reason.

jbaudoux (approved):

shopfloor/actions/search.py
simahawk: Indeed, the docs talk about using `set` operators on recordsets :wink: 
```
>>> rset1
res.partner(14, 26)
>>> rset2
res.partner(14, 26, 33, 27)
>>> rset1 | rset2
res.partner(14, 26, 33, 27)
>>> rset1 + rset2
res.partner(14, 26, 14, 26, 33, 27)
```
and if you use `+=`
```
>>> rset3
res.partner(14, 26)
>>> rset3 += rset2
>>> rset3
res.partner(14, 26, 14, 26, 33, 27)
```
while if you use `|=`
```
>>> rset3
res.partner(14, 26)
>>> rset3 |= rset2
>>> rset3
res.partner(14, 26, 33, 27)
```

shopfloor/actions/search.py
simahawk: @jbaudoux ?
I guess the idea is: if we have a limit of 10 and we find only 3 results, we do another search by barcode to find the rest.
Having a comment on top would be nice.

jbaudoux: The concern is that the search is not scoped to scenario/menu context. So to have a proper match, you must ensure location uniqueness which is not a standard odoo requirement.
Ideally we should be able to pass a domain & limit in the search. Having the limit is a first good improvement, you can then apply scope on the returned recordset

sebalix: Having a comment on top is not only nice, it's a requirement I would say :)

TDu: Comment added

sebalix (approved):

PR #350, opened 2022-01-10
[14.0] Shopfloor: Control the display of the virtual keyboard (Gboard) + SECURITY FIX

jbaudoux:
Fwd port of https://github.com/OCA/wms/pull/345

PR #433, opened 2022-08-23
[14.0] Add shopfloor_package_measurement

TDu:
Depends on:
- [x] https://github.com/OCA/delivery-carrier/pull/522

ref: rau-43


shopfloor_checkout_package_measurement/actions/schema.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from odoo.addons.component.core import Component
+
+
+class ShopfloorSchemaAction(Component):
+    _inherit = "shopfloor.schema.action"
+
+    def package(self, with_packaging=False):
+        res = super().package(with_packaging)
```
simahawk:
```suggestion
        res = super().package(with_packaging=with_packaging)
```

shopfloor_checkout_package_measurement/services/checkout.py
```diff
@@ -0,0 +1,116 @@
...
+        picking = self.env["stock.picking"].browse(picking_id)
+        message = self._check_picking_status(picking)
+        if message:
+            return self._response_for_select_document(message=message)
+        package = self.env["stock.quant.package"].browse(package_id).exists()
+        if not package:
+            return self._response_for_summary(
+                picking, message=self.msg_store.record_not_found()
+            )
+        if height:
```
simahawk, line 49: cant we get a single dict called `measures` instead of single values? Not important tho

shopfloor_checkout_package_measurement/services/checkout.py
```diff
@@ -0,0 +1,116 @@
...
+            package.pack_length = length
+        if width:
+            package.width = width
+        if weight:
+            package.pack_weight = weight
+        return self._response_for_summary(
+            picking,
+            message={
+                "message_type": "success",
+                "body": _("Package measurement changed."),
```
simahawk: to be moved to action messages

shopfloor_checkout_package_measurement/tests/test_checkout_package_measurement.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+        cls.picking.carrier_id = cls.carrier
+        cls.pack1_moves = cls.picking.move_lines[:3]
+        cls.pack2_moves = cls.picking.move_lines[3:]
+        cls._fill_stock_for_moves(cls.pack1_moves, in_package=True)
+        cls._fill_stock_for_moves(cls.pack2_moves, in_package=True)
+        cls.picking.action_assign()
+
+    def test_requested_measurement_for_new_package(self):
+        """Check package measurement are requested fro new package."""
```
simahawk:
```suggestion
        """Check package measurement are requested for new package."""
```

shopfloor_checkout_package_measurement_mobile/static/wms/src/scenario/checkout_package_measurement.js
```diff
@@ -0,0 +1,100 @@
...
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+ */
+
+import {process_registry} from "/shopfloor_mobile_base/static/wms/src/services/process_registry.js";
+
+import {checkout_states} from "/shopfloor_mobile/static/wms/src/scenario/checkout_states.js";
+
+const checkout_scenario = process_registry.get("checkout");
+const template = checkout_scenario.component.template;
+const pos = template.indexOf("</Screen>");
```
simahawk: A comment explaining what you do and why would be nice.

simahawk:
@TDu remember to add the dependency in the description please.

shopfloor_checkout_package_measurement_mobile/static/wms/src/scenario/checkout_package_measurement.js
```diff
@@ -0,0 +1,100 @@
...
+    template: new_template,
+    "methods._get_states": function () {
+        let states = checkout_states(this);
+        states["package_measurement"] = {
+            display_info: {
+                title: "Set packaging measurement",
+            },
+            events: {
+                go_back: "on_back",
+                done: "on_done",
```
simahawk: This is useless if you don't trigger the event `done`.
You are using `@click` above -> no need to bind again here AFAIS.

simahawk (changes requested):
Also, please, put core changes in their own commits.

shopfloor_checkout_package_measurement/actions/schema.py
TDu: Changed

shopfloor_checkout_package_measurement/services/checkout.py
TDu: done

shopfloor_checkout_package_measurement/tests/test_checkout_package_measurement.py
TDu: fixed

shopfloor_checkout_package_measurement_mobile/static/wms/src/scenario/checkout_package_measurement.js
TDu: Added

shopfloor_checkout_package_measurement_mobile/static/wms/src/scenario/checkout_package_measurement.js
TDu: Removed

TDu:
I have split the commit for the change in the base addons.
Also the fixups have the change requested as well as some new improvements (only display the required measurement, add the uom and use the shipping_weight field.

shopfloor_checkout_package_measurement/actions/message.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo.addons.component.core import Component
+
+
+class MessageAction(Component):
+    _inherit = "shopfloor.message.action"
+
+    def package_measurement_changed(self):
+        return {
+            "message_type": "success",
+            "body": _("Package measurement changed."),
```
simahawk:
```suggestion
            "body": _("Package measure(s) changed."),
```

shopfloor_checkout_package_measurement_mobile/static/wms/src/scenario/checkout_package_measurement.js
```diff
@@ -0,0 +1,114 @@
+/**
```
simahawk: can you remove `wms` from the path pls?

shopfloor_mobile/static/wms/src/scenario/checkout_states.js
```diff
@@ -0,0 +1,366 @@
+/**
+ * Copyright 2022 Camptocamp SA (http://www.camptocamp.com)
```
simahawk:
```suggestion
 * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
```

shopfloor_mobile/static/wms/src/scenario/checkout_states.js
```diff
@@ -0,0 +1,366 @@
...
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+ */
+
+export const checkout_states = function ($instance) {
```
simahawk, line 10:
```suggestion

/*
    Define states for checkout scenario.
    @param $instance VueJS component instance  
*/
export const checkout_states = function ($instance) {
```

simahawk (commented):
LG, keeping blocked till dep is merged

shopfloor_checkout_package_measurement/actions/message.py
TDu: done

shopfloor_checkout_package_measurement_mobile/static/wms/src/scenario/checkout_package_measurement.js
TDu: done

shopfloor_mobile/static/wms/src/scenario/checkout_states.js
TDu: done

shopfloor_mobile/static/wms/src/scenario/checkout_states.js
TDu: done

shopfloor_checkout_package_measurement/actions/message.py
simahawk: you didn't update tests ;)

shopfloor_checkout_package_measurement_mobile/static/src/scenario/checkout_package_measurement.js
```diff
@@ -0,0 +1,114 @@
...
+        :card_color="utils.colors.color_for('screen_step_done')"
+    />
+    <detail-package
+        :record="state.data.package"
+    />
+
+    <v-form>
+        <v-container>
+            <v-row>
+                <v-text-field
```
jbaudoux, line 32: If I remember properly, v-text-field does not manage properly the display of the virtual keyboard on a mobile device. This should likely be replaced by the searchbar or something alike and you should set the expected type, in this case number, to show the right keyboard

jbaudoux (changes requested):
Mobile device Virtual keyboard (Gboard) display issue. Can you test from an android device?

shopfloor_checkout_package_measurement_mobile/static/src/scenario/checkout_package_measurement.js
TDu: The searchbar uses a v-text-field as well. Oh no wait, are you refering to this PR https://github.com/OCA/wms/pull/350/files I see it is not yet merged in 14.0 and the `v-text-field` has been replaced by an `input`element ?

jbaudoux:
Yes indeed. Should be merged by the way

jbaudoux:
Add some space on the screen between the fields:
image

shopfloor_checkout_package_measurement_mobile/static/src/scenario/checkout_package_measurement.js
TDu: The problem is by switching to the `input` element we will loose all the nice thing of Material Design and vuetify and the rendering of the field will not  be nice at all.
Do we know why both are not compatible ?

TDu:
> Add some space on the screen between the fields: image

Done

shopfloor_checkout_package_measurement_mobile/static/src/scenario/checkout_package_measurement.js
jbaudoux: @TDu As we have to type something in the field, we don't have to hide the virtual keyboard. So fine to use v-text-field.
Still need to put type number to display right keyboard and prevent to enter chars

TDu: Done

jbaudoux (approved):

jbaudoux:

shopfloor_checkout_package_measurement/services/checkout.py
```diff
@@ -0,0 +1,125 @@
...
+        if shipping_weight:
+            package.shipping_weight = shipping_weight
+        return self._response_for_summary(
+            picking,
+            message=self.msg_store.package_measurement_changed(),
+        )
+
+    def _put_lines_in_allowed_package(self, picking, selected_lines, package):
+        res = super()._put_lines_in_allowed_package(picking, selected_lines, package)
+        if not picking._check_required_package_measurement(package=package):
```
simahawk: this line has to be adapted to changes in https://github.com/OCA/delivery-carrier/pull/522

TDu: Yes. But actually the latest changes done in the related PR actually break this one in more ways...
Now that we have the required measurements checked by a constrain this code does not work anymore, because we first save the stock.quant.package (the constrain stops us here) and then check if some of the measurements are required,
One way to get around this would be 
```
    @api.constrains("length", "width", "height", "weight")
    def _check_required_dimension(self, ignore_picking_state=False):
        for package in self:
            domain = [('result_package_id', '=', package.id)]
            pickings = self.env['stock.move.line'].search(domain).mapped('picking_id')  
            if pickings.state != "done" and not ignore_picking_state:
                continue
            required_dimension = []
            if package.length_required and not package.pack_length:
                required_dimension.append(_("length"))
...
```
It can be improve with a read_group but still not that nice...

TDu: Here is the PR https://github.com/OCA/delivery-carrier/pull/546 to fix that issue...

TDu:
I squashed all the reviewed fixes. And updated the pr to handle the new changes on https://github.com/OCA/delivery-carrier/pull/546

shopfloor_checkout_package_measurement/services/checkout.py
```diff
@@ -61,7 +63,11 @@ def set_package_measurement(
 
     def _put_lines_in_allowed_package(self, picking, selected_lines, package):
         res = super()._put_lines_in_allowed_package(picking, selected_lines, package)
-        if not picking._check_required_package_measurement(package=package):
+        try:
+            package.with_context(
```
simahawk: See my comment on the other PR.
Also, never override the whole ctx unless you really want to. Use kw arg instead.

TDu: fixed

TDu:
rebased

TDu:
> LG, keeping blocked till dep is merged

@simahawk dependency has been merged...

shopfloor_checkout_package_measurement/actions/schema.py
```diff
@@ -0,0 +1,37 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from odoo.addons.component.core import Component
+
+
+class ShopfloorSchemaAction(Component):
+    _inherit = "shopfloor.schema.action"
+
+    def package(self, with_packaging=False):
+        res = super().package(with_packaging=with_packaging)
+        res.update(
```
simahawk, line 11: doesn't make sense only if `with_packaging` is True?

simahawk (approved):
Minor question

shopfloor_checkout_package_measurement/actions/schema.py
TDu: I would say no, because the measurements are for packages and not for packaging.

sebalix (approved):
LG, fixup to squash and we can merge

PR #434, opened 2022-08-24
[14] shopfloor_mobile: single_pack_transfer display pack

JuMiSanAr:
When showing the pack being transferred, the screen shows the pack with its content and not the first product.


simahawk (approved):
LG

PR #435, opened 2022-08-25
[14.0] Add shopfloor_reception

mmequignon:
Shopfloor implementation of the reception scenario.
Allows to receive products and create the proper packs for each logistic unit.

Depends on this PR https://github.com/OCA/stock-logistics-workflow/pull/1116

ref: rau-49 + rau-57


shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+            ("state", "=", "assigned"),
+            ("picking_type_id", "in", self.picking_types.ids),
+            ("user_id", "=", False),
+        ]
+
+    def _filter_move_lines(self, move_lines):
+        filtered_lines = move_lines.filtered(
+            lambda l: l.move_id.picking_id.state == "assigned"
+        )
+        filtered_lines = filtered_lines.filtered(
```
TDu: Grouping all filtering conditions in one call to the `filtered` function is probably better for performance.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+        return self._response(next_state="select_line", data=data, message=message)
+
+    def _select_document_from_product(self, product):
+        """Select the document by product
+
+        next states:
+            - set_lot: a single picking has been found for this packaging
+            - select_document: A single or no pickings has been found for this packaging
+        """
+        move_lines = self._filter_move_lines(self._move_line_by_product(product))
```
TDu: This will return all existing move lines for a product without any other restriction (that is a lot of lines)
It should filter on the picking types from the shopfloor menu and their state.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+                    "message_type": "error",
+                    "body": _(
+                        "Several transfers found, please select a transfer manually."
+                    ),
+                },
+            )
+        # If no available picking with the right state has been found,
+        # return an error
+        return self._response_for_select_document(
+            message={
```
TDu: Messages should be added to `.../actions/message.py`

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+        )
+
+    def _select_document_from_packaging(self, packaging):
+        """Select the document by packaging
+
+        next states:
+            - set_lot: a single picking has been found for this packaging
+            - select_document: A single or no pickings has been found for this packaging
+        """
+        move_lines = self._move_line_by_packaging(packaging)
```
TDu, line 146: Same problem than before with the products.

shopfloor_reception/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "name": "Shopfloor Reception",
+    "summary": "Reception scenario for shopfloor",
+    "version": "14.0.1.0.0",
+    "category": "Inventory",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "maintainers": ["mmequignon"],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": True,
```
sebalix: `"auto_install": False` no?

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+
+
+class Reception(Component):
+
+    _inherit = "base.shopfloor.process"
+    _name = "shopfloor.reception"
+    _usage = "reception"
+    _description = __doc__
```
sebalix, line 44: A docstring explaining this scenario is welcome, as it is done in others.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+                        "Several transfers found, please select a transfer manually."
+                    ),
+                },
+            )
+        # If no available picking with the right state has been found,
+        # return a barcode not found error message
+        return self._response_for_select_document(
+            message={
+                "message_type": "error",
+                "body": _("No picking found for the scanned packaging."),
```
sebalix: As said, messages have to be moved to `actions/message.py`, or re-use if possible the existing messages in `shopfloor` module.
Also, I don't think the term `picking` should be used for user messages, we use operation or transfer (to check, I don't remember) in other places. Better to stick with Odoo naming when we share info with the user.

shopfloor_reception/docs/reception_sequence_graph.mermaid
```diff
@@ -0,0 +1,38 @@
+%%{init: {'theme': 'neutral' } }%%
```
sebalix, line 1: Can you mention the name of the methods to call to switch from one state to another?
Like it is done in https://github.com/OCA/wms/blob/14.0/shopfloor/docs/delivery_diag_seq.png for instance

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+                    lot = lot_model.create(
+                        {
+                            "name": lot_name,
+                            "product_id": line.product_id.id,
+                            "company_id": self.env.company.id,
+                            "use_expiration_date": line.product_id.use_expiration_date,
+                        }
+                    )
+                line.lot_id = lot.id
+                # TODO: Dunno why, I have to explicitely call the onchange method
```
sebalix: Not sure to understand this comment?
It's not a good thing to call the onchange methods here anyway? Looking at the code of `product_expiry`, this onchange updates the `expiration_date` field of the line, that seems important.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+        transitions:
+          - select_line: User clicked on back
+          - set_lot: Barcode not found. Ask user to create one from barcode
+          - set_lot: expiration_date has been set on the selected line
+          - set_lot: lot_it has been set on the selected line
+          - set_lot: Error: expiration_date is required
+          - set_quantity: User clicked on the confirm button
+        """
+        picking = self.env["stock.picking"].browse(picking_id)
+        selected_lines = self.env["stock.move.line"].browse(selected_line_ids)
```
sebalix: Do not hesitate to check if the ID of the records you are receiving really exist (with `exists()` method), like here: https://github.com/OCA/wms/blob/14.0/shopfloor/services/zone_picking.py#L768-L770

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+            if package:
+                pack_location = package.location_id
+                if pack_location:
+                    if pack_location not in child_locations:
+                        # If the scanned package has a location that isn't a child
+                        # of the move dest, return an error
+                        message = {
+                            "message_type": "error",
+                            "body": _("You cannot place it here."),
+                        }
```
sebalix:
```suggestion
                        message = self.msg_store.dest_location_not_allowed()
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+                selected_lines.result_package_id = package
+                return self._response_for_set_destination(picking, selected_lines)
+            location = search.location_from_scan(barcode)
+            if location:
+                if location not in child_locations:
+                    # Scanned location isn't a child of the move's dest location
+                    message = {
+                        "message_type": "error",
+                        "body": _("You cannot place it here."),
+                    }
```
sebalix:
```suggestion
                    message = self.msg_store.dest_location_not_allowed()
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+            if not confirmation:
+                return self._response_for_set_quantity(
+                    picking,
+                    selected_lines,
+                    message={
+                        "message_type": "warning",
+                        "body": _("Put in this new PACK {}?").format(barcode),
+                    },
+                )
+            package = self.env["stock.quant.package"].create({"name": barcode})
```
sebalix, line 648: The creation of a package wasn't optional (reading the specs)?

> Option on menu item:
> - Always put in Pack (SSCC)
>

mmequignon: Well, regarding the specs, I should have put that in the shopfloor module.
I thought it would be smarter to create a new module. If the module is installed, then I suppose that this option is enabled by default.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+        return {"set_destination"}
+
+    def _process_without_pack_next_states(self):
+        return {"set_destination"}
+
+    # SCHEMAS
+
+    @property
+    def _schema_select_document(self):
+        return {"pickings": self.schemas._schema_list_of(self.schemas.picking())}
```
JuMiSanAr: We need the lines in the response, so this needs to make use of _schema_stock_picking as discussed.

shopfloor_reception/services/reception.py
sebalix: It has its place in this module I think as we generally create packages when receiving the goods. And the option is enabled on the menu, so we could have two different menus to receive goods, with the option enabled only on one.
cc @jbaudoux

shopfloor_reception/__manifest__.py
simahawk: yes pls

shopfloor_reception/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "Shopfloor Reception",
+    "summary": "Reception scenario for shopfloor",
+    "version": "14.0.1.0.0",
+    "category": "Inventory",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
```
simahawk:
```suggestion
    "author": "Camptocamp, Odoo Community Association (OCA)",
```

shopfloor_reception/services/reception.py
simahawk: A long one pls :)

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+            self._domain_move_line_by_packaging(packaging)
+        )
+
+    # DOMAIN METHODS
+
+    def _domain_move_line_by_packaging(self, packaging):
+        return [("package_id.product_packaging_id", "=", packaging.id)]
+
+    def _domain_move_line_by_product(self, product):
+        return [("product_id", "=", product.id)]
```
simahawk: must filter by state and by self.picking_types at least

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+        return self._response_for_set_lot(picking, lines)
+
+    def _order_stock_picking(self):
+        return "scheduled_date ASC"
+
+    # DATA METHODS
+
+    def _data_for_stock_picking(self, picking):
+        data = self.data.picking(picking)
+        data.update({"move_lines": self.data.move_lines(picking.move_line_ids)})
```
```suggestion
        data.update({"move_lines": self._data_for_move_lines(picking.move_line_ids)})
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+        return self._response(
+            next_state="select_dest_package",
+            data={
+                "selected_move_lines": self._data_for_move_lines(lines),
+                "picking": self.data.picking(picking),
+            },
+            message=message,
+        )
+
+    # ENDPOINTS
```
simahawk, line 427: I find it weird to have all the endpoints a the bottom... but yeah...

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+          - select_line: Error: barcode not found
+          - set_lot: Packaging / Product has been scanned. Tracked product
+          - set_quantity: Packaging / Product has been scanned. Not tracked product
+        """
+        picking = self.env["stock.picking"].browse(picking_id)
+        message = self._check_picking_status(picking)
+        if message:
+            return self._response_for_select_line(picking, message=message)
+
+        search = self._actions_for("search")
```
simahawk: not a blocker but you could do it like this https://github.com/OCA/wms/blob/14.0/shopfloor/services/zone_picking.py#L594

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+        packaging = search.packaging_from_scan(barcode)
+        if packaging:
+            return self._select_line_from_packaging(picking, packaging)
+        # Nothing has been found, return an error
+        return self._response_for_select_line(
+            picking, message=self.msg_store.barcode_not_found()
+        )
+
+    def done_action(self, picking_id, confirmation=False):
+        picking = self.env["stock.picking"].browse(picking_id)
```
simahawk, line 489: please use `_check_picking_status` when you lookup for pickings https://github.com/OCA/wms/blob/40f1e785a313f38f85b27871f9e81a3e2a4ae72a/shopfloor/services/service.py#L49

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+                # Not all lines are fully done, ask the user to confirm the
+                # backorder creation
+                return self._response_for_select_line(
+                    picking, message=self.msg_store.transfer_confirm_done()
+                )
+            # all lines are done, ask the user to confirm anyway
+            return self._response_for_select_line(
+                picking, message=self.msg_store.need_confirmation()
+            )
+        backorders_before = picking.backorder_ids
```
simahawk, line 517: maybe delegate to `_handle_backorders`?

shopfloor_reception/services/reception.py
mmequignon: Yeah this is important. The thing is that changing the lot doesn't update the expiration date.
We have to force it, despite what suggests this piece of code https://github.com/odoo/odoo/blob/14.0/addons/product_expiry/models/stock_move_line.py#L29-L36

shopfloor_reception/services/reception.py
mmequignon: @jbaudoux I'll need help for that.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,951 @@
...
+
+    def _use_handlers(self, handlers, *args, **kwargs):
+        for handler in handlers:
+            response = handler(*args, **kwargs)
+            if response:
+                return response
+
+    # DATA METHODS
+
+    def _data_for_stock_picking(self, picking):
```
simahawk: This seems sub optimal. We should apply the same parser to the whole recordset.
You should probably use `data_detail.pickings_detail()` which already contains data for the move lines

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,951 @@
...
+            )
+        picking.user_id = self.env.user
+        if packaging.product_id.tracking not in ("lot", "serial"):
+            return self._response_for_set_quantity(picking, lines)
+        return self._response_for_set_lot(picking, lines)
+
+    def _order_stock_picking(self):
+        return "scheduled_date ASC"
+
+    def _scan_document__scan_picking(self, barcode):
```
simahawk: maybe replace "scan" w/ "by" in all these methods?
```suggestion
    def _scan_document__by_picking(self, barcode):
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,951 @@
...
+          - set_lot: Packaging / Product has been scanned. Tracked product
+          - set_quantity: Packaging / Product has been scanned. Not tracked product
+        """
+        picking = self.env["stock.picking"].browse(picking_id)
+        message = self._check_picking_status(picking)
+        if message:
+            return self._response_for_select_line(picking, message=message)
+        message = self._check_picking_status(picking)
+        if message:
+            return self._response_for_select_line(picking, message=message)
```
```suggestion
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,951 @@
...
+          - set_lot: expiration_date has been set on the selected line
+          - set_lot: lot_it has been set on the selected line
+          - set_lot: Error: expiration_date is required
+          - set_quantity: User clicked on the confirm button
+        """
+        picking = self.env["stock.picking"].browse(picking_id)
+        selected_lines = self.env["stock.move.line"].browse(selected_line_ids)
+        message = self._check_picking_status(picking)
+        if message:
+            return self._response_for_set_lot(picking, selected_lines, message=message)
```
```suggestion
        message = self._check_picking_status(picking)
        if message:
            return self._response_for_set_lot(picking, selected_lines, message=message)
        selected_lines = self.env["stock.move.line"].browse(selected_line_ids).exists()
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,951 @@
...
+        if not selected_lines.exists():
+            message = self.msg_store.record_not_found()
+            return self._response_for_set_lot(picking, selected_lines, message=message)
+        lot_model = self.env["stock.production.lot"]
+        search = self._actions_for("search")
+        if lot_name:
+            for line in selected_lines:
+                lot = search.lot_from_scan(
+                    lot_name, products=selected_lines.mapped("product_id")
```
simahawk: For each line you search for all the products of all lines.
I would collect all the line w/out a lot, collect all values for missing lots and create them all at once.
You can map line ids and lots in a dict.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,951 @@
...
+          - select_line: User clicked on back
+          - set_lot: Barcode not found. Ask user to create one from barcode
+          - set_lot: expiration_date has been set on the selected line
+          - set_lot: lot_it has been set on the selected line
+          - set_lot: Error: expiration_date is required
+          - set_quantity: User clicked on the confirm button
+        """
+        picking = self.env["stock.picking"].browse(picking_id)
+        selected_lines = self.env["stock.move.line"].browse(selected_line_ids)
+        message = self._check_picking_status(picking)
```
simahawk, line 614: always check the status before doing anything else

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,951 @@
...
+    @property
+    def _schema_select_document(self):
+        return {
+            "pickings": self.schemas._schema_list_of(
+                self._schema_stock_picking(), required=True
+            )
+        }
+
+    def _schema_stock_picking(self, lines_with_packaging=False):
+        schema = self.schemas.picking()
```
simahawk, line 1009: use picking_detail schema

shopfloor_reception/tests/common.py
```diff
@@ -0,0 +1,83 @@
...
+        for line in picking.move_line_ids:
+            product = line.product_id
+            packaging = packagings.filtered(lambda p: p.product_id == product)
+            package = cls.env["stock.quant.package"].create(
+                {"product_packaging_id": packaging.id}
+            )
+            line.package_id = package
+
+    @classmethod
+    def setUpClassVars(cls, *args, **kwargs):
```
simahawk, line 44: please keep setup methods on top

shopfloor_reception/services/reception.py
jbaudoux: Here is what you can add in the docstring:

Process a picking and track progress by product.

Once a transfer is opened, you first need to select a product (you can scan its barcode or one of its packaging barcode). Then set the processed quantity. Put it in a internal PACK (this is optional but can be made mandatory by menu configuration), this PACK can be a new one (like an empty pallet) or an existing one you add products to (like a pallet you continue to fill in). And finally, the location where you put the product (iow. the location where is the transport trolley or pallet), unless you fill an existing PACK as its location was already defined when its first product was put on it.

In case of product tracked by lot, you will have to enter the lot number and its expiry date (unless it is already known by the system).

Moves are not validated as they are processed. It is the responsibility of the user to decide when to mark as done already processed lines.Any remaining lines will be pushed to a backorder.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,877 @@
...
+                        selected_lines.result_package_id = package
+                        selected_lines.location_dest_id = pack_location
+                        return self._response_for_select_line(picking)
+                # Scanned package has no location, move to the location selection
+                # screen
+                selected_lines.result_package_id = package
+                return self._response_for_set_destination(picking, selected_lines)
+            location = search.location_from_scan(barcode)
+            if location:
+                if location not in child_locations:
```
jbaudoux: I think its lacking a test if the pack is mandatory and already set

shopfloor_reception/services/reception.py
jbaudoux: At this stage, it looks good, all the tests have been done and the pack can now be created

jbaudoux (changes requested):

shopfloor_reception/services/reception.py
mmequignon: can't do that, selected lines is sent in the response.

shopfloor_reception/services/reception.py
mmequignon: In all cases, I return a response with the selected lines, whether there's a message or not.

shopfloor/__manifest__.py
```diff
@@ -27,6 +27,7 @@
         #  OCA / stock-logistics-workflow
         "stock_quant_package_dimension",
         "stock_quant_package_product_packaging",
+        "stock_picking_completion",
```
jbaudoux: Where is this module?
Rename to stock_picking_progress

mmequignon: done.

shopfloor_reception/services/reception.py
sebalix: Onchange methods are a UI feature, they are never called when fields are updated on server-side, so it's normal to call them manually. This will be fixed on 16.0 with the new way to write "onchange" methods.

shopfloor_reception/services/reception.py
```diff
@@ -72,6 +72,11 @@ def _response_for_select_line(self, picking, message=None):
         data = {"picking": self._data_for_stock_picking(picking, with_lines=True)}
         return self._response(next_state="select_line", data=data, message=message)
 
+    def _response_for_confirm_done(self, picking, message=None):
+        picking.user_id = self.env.user
+        data = {"picking": self._data_for_stock_picking(picking, with_lines=True)}
+        return self._response(next_state="confirm_done", data=data, message=message)
+
```
mmequignon: @JuMiSanAr could you add tests for that?

shopfloor/actions/search.py
```diff
@@ -28,11 +30,19 @@ def package_from_scan(self, barcode):
...
             return model.browse()
-        return model.search([("name", "=", barcode)], limit=1)
+        domain = [("name", "=", barcode)]
+        if accept_source_document:
+            source_document_domain = [
+                "&",
+                ("origin", "=", barcode),
+                ("state", "=", "assigned"),
+            ]
+            domain = expression.OR([domain, source_document_domain])
```
mmequignon: Needs to be tested too

shopfloor_reception/services/reception.py
```diff
@@ -319,10 +319,27 @@ def _response_for_set_destination(self, picking, lines, message=None):
...
+            return self._response_for_set_quantity(
+                picking,
+                lines,
+                message=self.msg_store.no_valid_package_to_select(),
+            )
+        packages_data = self.data.packages(
+            packages.with_context(picking_id=picking.id).sorted(),
+            picking=picking,
+            with_packaging=True,
+        )
```
mmequignon: we need tests here too

shopfloor_reception/services/reception.py
```diff
@@ -158,8 +158,15 @@ def _order_stock_picking(self):
 
     def _scan_document__scan_picking(self, barcode):
         search = self._actions_for("search")
-        picking = search.picking_from_scan(barcode)
+        picking = search.picking_from_scan(barcode, True)
```
mmequignon: Since this is a named argument, add the key arg, it will make this easier to read
```python
        picking = search.picking_from_scan(barcode, accept_source_document=True)
```
Anyway, why making this argument optional? Can't see the point

shopfloor/actions/search.py
```diff
@@ -28,11 +30,20 @@ def package_from_scan(self, barcode):
             return model.browse()
         return model.search([("name", "=", barcode)], limit=1)
 
-    def picking_from_scan(self, barcode):
+    def picking_from_scan(self, barcode, accept_source_document=False):
```
sebalix: We have the same kind of parameter named `use_packaging` when scanning a product, so I would name it the same, and based on the technical field name, not its label:
```suggestion
    def picking_from_scan(self, barcode, use_origin=False):
```

shopfloor_reception/services/reception.py
sebalix: That's a good thing to make it optional by default, we already had done this on product scan in the past to also accept product packaging barcode, and with time we had to refactor it to make it optional.
Now we can debate: should the default value on the method be `True` or `False`...

JuMiSanAr: Since at this stage the only scenario that accepts SO is reception (last to arrive), I set it to false by default in all other cases. Not sure if all other scenarios will have to accept it as well in the future? @jbaudoux

shopfloor_reception/services/reception.py
JuMiSanAr: We only deal with 1 line at a time now so this is no longer the case.

shopfloor/actions/data.py
```diff
@@ -49,6 +49,7 @@ def _picking_parser(self):
             "bulk_line_count",
             "total_weight:weight",
             "scheduled_date",
+            "progress",
```
simahawk: Are we sure we need to load this information for ALL the pickings and for all the cases?
AFAIS on https://github.com/OCA/stock-logistics-workflow/pull/1116 it can be heavy (unless you improve perf)

shopfloor/actions/data.py
```diff
@@ -64,8 +65,10 @@ def package(self, record, picking=None, with_packaging=False, **kw):
         data = self._jsonify(record, parser, **kw)
         # handle special cases
         if data and picking:
-            # TODO: exclude canceled and done?
-            lines = picking.move_line_ids.filtered(lambda l: l.package_id == record)
+            lines = picking.move_line_ids.filtered(
+                lambda l: l.result_package_id == record
+                and l.state in ["partially_available", "assigned", "done"]
```
simahawk, line 76: @JuMiSanAr this is the change the required the tests update, not the reception scenario introduction

shopfloor/actions/data.py
```diff
@@ -131,7 +134,7 @@ def lots(self, record, **kw):
 
     @property
     def _lot_parser(self):
-        return self._simple_record_parser() + ["ref"]
+        return self._simple_record_parser() + ["ref", "expiration_date"]
```
simahawk, line 143: and this one

shopfloor/actions/data.py
```diff
@@ -176,6 +179,7 @@ def _move_line_parser(self):
                 "move_id:priority",
                 lambda rec, fname: rec.move_id.priority or "",
             ),
+            "progress",
```
simahawk, line 188: and this one

shopfloor_mobile/static/wms/src/components/scenario_picking_detail/picking_summary.js
```diff
@@ -114,53 +116,87 @@ Vue.component("picking-summary-content", {
                     <span class="item-counter">
                         <span>{{ index + 1 }} / {{ count }}</span>
                     </span>
-                    {{ record.title }}
+                    <strong>
+                        {{ options.group_header_title_key ? pkg_type.records[0].product[options.group_header_title_key] : record.title }}
```
simahawk:
```suggestion
                        {{ get_group_title(rec, pkg_type) }}
```

shopfloor_mobile/static/wms/src/components/scenario_picking_detail/picking_summary.js
```diff
@@ -114,53 +116,87 @@ Vue.component("picking-summary-content", {
...
     props: {
         record: Object,
         index: Number,
         count: Number,
+        options: Object,
     },
     template: `
         <div class="summary-content-item">
-            <v-list-item-title>
+            <v-list-item-title v-if="options.show_title">
```
simahawk, line 175: this should default to true and is not the case now, no?

shopfloor_mobile/static/wms/src/wms_utils.js
```diff
@@ -86,6 +86,47 @@ export class WMSUtils {
         return res;
     }
 
+    group_lines_by_product(lines, options) {
+        const self = this;
+        // {'key': 'no-group', 'title': '', 'records': []}
+        options = _.defaults(options || {}, {
+            group_key: "product",
```
simahawk: what's the point of making the group key configurable?

simahawk: Do you need all these options?

shopfloor_mobile_base/static/wms/src/components/datepicker/date_picker.js
```diff
@@ -0,0 +1,48 @@
...
+
+export var DatePicker = Vue.component("date-picker-input", {
+    data: function () {
+        return {
+            date: "",
+        };
+    },
+    watch: {
+        date: function () {
+            this.$root.trigger("date_picker_selected", this.date);
```
simahawk: Either you use `event_hub.$emit` or `this.$root.trigger`.
I don't think you need an event which has a global state here -> you don't need `$root.trigger`.

shopfloor_mobile_base/static/wms/src/components/datepicker/date_picker.js
```diff
@@ -0,0 +1,48 @@
...
+        },
+    },
+    mounted() {
+        event_hub.$on("datepicker:lotselected", (lot) => {
```
simahawk: :fire: 

This is a specific feature that should be implemented by the consumer of this widget.
You should have something like

`<date-picker-input v-on:date_selected="on_date_selected"...`

shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
```diff
@@ -52,6 +52,11 @@ export var Searchbar = Vue.component("searchbar", {
             type: Number,
             default: 0,
         },
+        // uses the clearable input type logic (using an X icon to clear the input)
+        clearable_input_type: {
+            type: Boolean,
+            default: false,
```
simahawk, line 58: can be true likely?

shopfloor_mobile_base/static/wms/src/components/datepicker/date_picker.js
simahawk: Also, this change should be move to its own commit and separated

shopfloor_reception/data/shopfloor_scenario_data.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2022 Camptocamp SA
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
```
simahawk:
```suggestion
<odoo noupdate="1">
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+    Methods for the Reception Process
+
+    You can find a sequence diagram describing states and endpoints relationships
+    [here](../docs/reception_sequence_graph.png)
+    Keep [the sequence diagram](../docs/reception_sequence_graph.mermaid)
+    up-to-date if you change endpoints.
+
+    Process a picking and track progress by product.
+
+    Once a transfer is opened, you first need to select a product (you can scan its
```
simahawk: can you rephrase this and use an ordered list?

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+
+    def _response_for_confirm_done(self, picking, message=None):
+        picking.user_id = self.env.user
+        data = {"picking": self._data_for_stock_picking(picking, with_lines=True)}
+        return self._response(next_state="confirm_done", data=data, message=message)
+
+    def _response_for_confirm_new_package(
+        self, picking, line, new_package_name, message=None
+    ):
+        picking.user_id = self.env.user
```
simahawk: I would not set the user here. This should be one explicitly by the endpoint, not by the method collecting data for the response.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        """Select the document by packaging
+
+        next states:
+            - set_lot: a single picking has been found for this packaging
+            - select_document: A single or no pickings has been found for this packaging
+        """
+        move_lines = self._move_line_by_packaging(packaging)
+        pickings = move_lines.mapped("move_id.picking_id")
+        if len(pickings) == 1:
+            pickings.user_id = self.env.user
```
```suggestion
            self._assign_user_to_picking(picking)
```
you know where to use it ;)

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+                     single correspondance. Tracked product
+          - set_quantity: Packaging / Product has been scanned,
+                          single correspondance. Not tracked product
+        """
+        self._actions_for("search")
+        handlers = [
+            self._scan_document__by_picking,
+            self._scan_document__by_product,
+            self._scan_document__by_packaging,
+        ]
```
```suggestion
        handlers = (
            self._scan_document__by_picking,
            self._scan_document__by_product,
            self._scan_document__by_packaging,
        )
```
always use tuples when you don't expect modifications

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        transitions:
+          - select_line: Error: barcode not found
+          - set_lot: Packaging / Product has been scanned. Tracked product
+          - set_quantity: Packaging / Product has been scanned. Not tracked product
+        """
+        picking = self.env["stock.picking"].browse(picking_id)
+        message = self._check_picking_status(picking)
+        if message:
+            return self._response_for_select_line(picking, message=message)
+        handlers = [
```
simahawk: same

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+            # process without pack, set destination location, and go back to
+            # `select_line`
+            selected_line.location_dest_id = location
+            return self._response_for_select_line(picking)
+
+    def _use_handlers(self, handlers, *args, **kwargs):
+        for handler in handlers:
+            response = handler(*args, **kwargs)
+            if response:
+                return response
```
simahawk, line 330:
```suggestion
            return handler(*args, **kwargs)
```

shopfloor_reception/services/reception.py
simahawk: still not done :arrow_up:

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+            lot = search.lot_from_scan(lot_name, products=product)
+            if not lot:
+                lot = self.env["stock.production.lot"].create(
+                    {
+                        "name": lot_name,
+                        "product_id": product.id,
+                        "company_id": self.env.company.id,
+                        "use_expiration_date": product.use_expiration_date,
+                    }
+                )
```
```suggestion
                lot = self.env["stock.production.lot"].create(
                    self._create_lot_values(product, lot_name)
                )
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+            selected_line.lot_id = lot.id
+            selected_line._onchange_lot_id()
```
simahawk, line 557: use onchange_helper instead?

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        elif expiration_date:
+            selected_line.write({"expiration_date": expiration_date})
+            selected_line.lot_id.write({"expiration_date": expiration_date})
+        return self._response_for_set_lot(picking, selected_line)
+
+    def set_lot_confirm_action(self, picking_id, selected_line_id):
+        picking = self.env["stock.picking"].browse(picking_id)
```
simahawk, line 572: I repeat myself: every time you lookup for a picking, check its status.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        selected_line = self.env["stock.move.line"].browse(selected_line_id)
+        message = self._check_expiry_date(selected_line)
```
simahawk, line 577: this is a check, is not meant to give you a msg IMO

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        if quantity:
+            selected_line.qty_done = quantity
+            return self._response_for_set_quantity(picking, selected_line)
+        if barcode:
+            handlers = [
+                self._set_quantity__by_product,
+                self._set_quantity__by_packaging,
+                self._set_quantity__by_package,
+                self._set_quantity__by_location,
+            ]
```
```suggestion
            handlers = (
                self._set_quantity__by_product,
                self._set_quantity__by_packaging,
                self._set_quantity__by_package,
                self._set_quantity__by_location,
            )
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        pick_type_dest_location = picking.picking_type_id.default_location_dest_id
+        pick_type_child_locations = self.env["stock.location"].search(
+            [("id", "child_of", pick_type_dest_location.id)]
+        )
+        if location in move_child_locations:
+            # If location is a child of move's dest lcoation, assign it without asking
+            selected_line.location_dest_id = location
+            return self._response_for_select_line(picking)
+        elif location in pick_type_child_locations:
+            # If location is a child of picking types's dest lcoation,
```
```suggestion
            # If location is a child of picking types's dest location,
```

shopfloor_reception/services/reception.py
simahawk: ?

shopfloor_reception/tests/common.py
simahawk: ???

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -0,0 +1,723 @@
...
+                prev.group_color > next.group_color ? -1 : 1
+            );
+        },
+        on_search: function (input) {
+            this.filtered_pickings = this.state.data.pickings.filter((picking) =>
+                this._apply_search_filter(picking, input.text)
+            );
+        },
+        on_select_document: function (selected) {
+            this.$root.trigger("scan", selected.name);
```
simahawk: why do you trigger an event from another event?

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -0,0 +1,723 @@
...
+        },
+        on_mark_done: function () {
```
simahawk: same

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -0,0 +1,723 @@
...
+            return !_.isEmpty(expiry_date);
+        },
+        get_next_line_id_to_handle: function () {
+            // The enpoints in the backend accept multiple selected line ids.
+            // However, in this particular scenario, we only want to deal with one line at a time.
+            // For that, we select the first line we find that hasn't been completely dealt with yet.
+            const next_unhandled_line = this.state.data.selected_move_lines.find(
+                (line) => line.qty_done < line.quantity
+            );
+            return this._get_selected_line_ids([next_unhandled_line]);
```
simahawk: `_get_selected_line_ids` is used only here and only w/ one record, why do you need a list and loop on it to get the id?

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -0,0 +1,723 @@
...
+                    },
+                    on_mark_as_done: () => {
+                        this.wait_call(
+                            this.odoo.call("done_action", {
+                                picking_id: this.state.data.picking.id,
+                            })
+                        );
+                    },
+                    on_cancel: () => {
+                        // TODO: endpoint missing in backend
```
simahawk: ??

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -0,0 +1,723 @@
...
+                                picking_id: this.state.data.picking.id,
+                                selected_line_id: this.line_being_handled.id,
+                                quantity: this.scan_destination_qty,
+                                barcode: barcode.text,
+                                confirmation: true,
+                            })
+                        );
+                    },
+                    on_cancel: () => {
+                        // TODO: endpoint missing in backend
```
simahawk: ???

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -0,0 +1,723 @@
...
+            filtered_pickings: [],
+        };
+    },
+};
+
+process_registry.add("reception", Reception);
+
+export default Reception;
+
+// TODO: handle all confirmations in the scenario
```
simahawk: ???

simahawk (changes requested):

shopfloor/actions/search.py
```diff
@@ -28,11 +30,20 @@ def package_from_scan(self, barcode):
...
         if not barcode:
             return model.browse()
-        return model.search([("name", "=", barcode)], limit=1)
+        domain = [("name", "=", barcode)]
+        if not use_origin:
+            return model.search(domain, limit=1)
+        source_document_domain = [
+            "&",
+            ("origin", "=", barcode),
+            ("state", "=", "assigned"),
```
sebalix: Why are we checking the state when searching against `origin` and not with `name`? IMO we should not check the `state` of the transfer at all in this method, it's the job of the `_check_picking_status` general method. So we could write:
```python
    def picking_from_scan(self, barcode, use_origin=False):
        model = self.env["stock.picking"]
        if not barcode:
            return model.browse()
        domain = [("name", "=", barcode)]
        if use_origin:
            origin_domain = [("origin", "=", barcode)]
            domain = expression.OR([domain, origin_domain])
        return model.search(domain)
```

shopfloor_reception/docs/reception_sequence_graph.mermaid
```diff
@@ -0,0 +1,61 @@
...
+	set_destination ->> set_destination: Warning: User scanned a child location of the picking type. Ask for confirmation
+	set_destination ->> set_destination: Error: User tried to scan a non-valid location
+	set_destination ->> select_line: User scanned a child location of the move's dest location
+    end
+    rect rgb(250, 150, 150)
+    note left of select_dest_package: select_dest_package(picking_id, selected_line_ids, location_id, confirmation=False)
+	select_dest_package ->> select_line: User scanned a valid package
+	select_dest_package ->> select_dest_package: Warning: User scanned an unknown barcode. Confirm to create one.
+	select_dest_package ->> select_dest_package: Error: User scanned a non-empty package
+    end
```
sebalix, line 61: IMO I find the graph generated with mermaid not readable at all compared to what we were using in other scenarios (PlantUML)... need to zoom in and scroll a lot (and the image quality is not so good).
But don't spend time on this.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+class Reception(Component):
+    """
+    Methods for the Reception Process
+
+    You can find a sequence diagram describing states and endpoints relationships
+    [here](../docs/reception_sequence_graph.png)
+    Keep [the sequence diagram](../docs/reception_sequence_graph.mermaid)
+    up-to-date if you change endpoints.
+
+    Process a picking and track progress by product.
```
sebalix: Better to fit with what we are processing with this scenario, and re-use Odoo terms:
```suggestion
    Process a receipt transfer and track progress by product.
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        packaging = search.packaging_from_scan(barcode)
+        if packaging:
+            return self._select_line_from_packaging(picking, packaging)
+
+    def _set_quantity__by_product(self, picking, selected_line, barcode):
+        # needed values:
+        # selected_line
+        # picking
+        search = self._actions_for("search")
+        product = search.product_from_scan(barcode, use_packaging=False)
```
sebalix, line 254: NOTE for ourself: remove `use_packaging=False` once https://github.com/OCA/wms/pull/483/commits/f3cda915e9659a3cbadc66ee4c8781f76bb22281 is merged (#483)

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        if product:
+            selected_line.qty_done += 1
```
sebalix: We should check if the product scanned matches with the line, no? Right now I could scan product A while the one on the selected line is B.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+    def _scan_line__by_packaging(self, picking, barcode):
+        search = self._actions_for("search")
+        packaging = search.packaging_from_scan(barcode)
+        if packaging:
+            return self._select_line_from_packaging(picking, packaging)
+
+    def _set_quantity__by_product(self, picking, selected_line, barcode):
+        # needed values:
+        # selected_line
+        # picking
```
sebalix:
```suggestion
```

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        product = search.product_from_scan(barcode, use_packaging=False)
+        if product:
+            selected_line.qty_done += 1
+            return self._response_for_set_quantity(picking, selected_line)
+
+    def _set_quantity__by_packaging(self, picking, selected_line, barcode):
+        search = self._actions_for("search")
+        packaging = search.packaging_from_scan(barcode)
+        if packaging:
+            selected_line.qty_done += packaging.qty
```
sebalix: Same remark than :arrow_up: check that the packaging product is the same than the line.

shopfloor_reception/services/reception.py
```diff
@@ -0,0 +1,1079 @@
...
+        if not pickings:
+            pickings = self.env["stock.picking"].search(
+                self._domain_stock_picking(),
+                order=self._order_stock_picking(),
+            )
+        data = {"pickings": self._data_for_stock_pickings(pickings, with_lines=False)}
+        return self._response(next_state="select_document", data=data, message=message)
+
+    def _response_for_set_lot(self, picking, line, message=None):
+        # If the product isn't tracked, move to the next screen
```
sebalix:
```suggestion
```
As that's not what this method is doing.

shopfloor_reception/services/reception.py
JuMiSanAr: It was updated to _schema_stock_picking_with_lines

shopfloor_reception/services/reception.py
JuMiSanAr: Updated.
I'm assuming this scenario description is the only place we should update, as this is referred to as "picking" in the rest of the module right?

sebalix: Yes, only in the docstring.
I think the right fix would be to rename `stock.picking` to `stock.transfer`, as picking is a type of transfer, but we won't win this battle I guess :)

shopfloor/actions/data.py
JuMiSanAr: Not sure I understand what you mean here.

shopfloor/actions/data.py
JuMiSanAr: For the moment the performance is good. 

Should I create a different parser just for this?

shopfloor/actions/data.py
simahawk: I mean that the fix for the tests has been committed into the reception but it has nothing to deal with it

shopfloor_mobile/static/wms/src/components/scenario_picking_detail/picking_summary.js
JuMiSanAr: It does (in the List component).

shopfloor_reception/services/reception.py
JuMiSanAr: Hmmm wouldn't this always return whatever the first handler returns, and stop the loop there?

shopfloor_reception/services/reception.py
JuMiSanAr: it's the same as _check_picking_status isn't it?

shopfloor_mobile_base/static/wms/src/components/searchbar/searchbar.js
JuMiSanAr: This is a very specific thing of this filter searchbar (as opposed to the standard scan searchbar we use everywhere), I wouldn't set it as true by default as it can mess with the rest of the searchbars.

I've given it a try and it alters the behaviour of the rest of the searchbars in the app.

shopfloor_reception_mobile/static/src/scenario/reception.js
JuMiSanAr: It was decided with @jbaudoux that we'd leave this implementation in the backend for a second version of the scenario.
I left this in the frontend so that it's easier to implement later.

shopfloor_reception/services/reception.py
simahawk: you got my point ;)

shopfloor/actions/data.py
simahawk: as we said, pls make it optional

sebalix:
Closing in favor of #500

PR #437, opened 2022-09-01
shopfloor: zone picking scenario stay in zone

TDu:
When scanning a location out of the previously selected zone, instead of
sending the user back to the start, keep him on the line selection of
the zone.
ref: rau-60


simahawk (approved):
Seems legit

jbaudoux (approved):

PR #439, opened 2022-09-01
[14.0] shopfloor: zone picking add unloading on first screen

TDu:
The user may end up in the `scan_location` screen with some products to
unload. It can happen when clicking on the back button but also when
there is no lines left to process in the current zone.

This change adds an Unload At Destination button on the `scan_location`
screen which will be display only when the user has some product to
unload.

ref.: rau-41


shopfloor/services/zone_picking.py
```diff
@@ -160,9 +160,21 @@ def _response_for_start(self, message=None):
...
         )
+        data = {"zones": self._data_for_select_zone(zones)}
+        buffer = self._find_buffer_move_lines()
+        if buffer:
+            # Some lines can be unloaded, let the user know
+            # The call to the endpoint will need the location and picking id
+            line = first(buffer)
+            picking_type = line.picking_id.picking_type_id
+            zone = line.move_id.location_id
+            data["buffer_is_loaded"] = {
```
sebalix: having a buffer already means we loaded something
```suggestion
            data["buffer"] = {
```
could be enough?

TDu: could be :shrug:

simahawk: I agree, can you rename it to "buffer"?

shopfloor/services/zone_picking.py
```diff
@@ -1510,6 +1522,25 @@ def _schema_for_select_zone(self):
         zone_schema.update(self._schema_for_zone_line_counters)
         zone_schema = {
             "zones": self.schemas._schema_list_of(zone_schema),
+            "buffer_is_loaded": {
+                "type": "dict",
+                "nullable": False,
+                "required": False,
+                "schema": {
+                    "zone_location": {
```
simahawk: you can use `_schema_dict_of`

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -284,7 +290,11 @@ const ZonePicking = {
             }
             const data = this.state_get_data("select_line");
             if (_.isEmpty(data) || _.isEmpty(data.picking_type)) {
-                return {};
+                const data_sl = this.state_get_data("scan_location");
+                if (_.isEmpty(data_sl) || _.isEmpty(data_sl.buffer_is_loaded)) {
+                    return {};
+                }
+                return data_sl.buffer_is_loaded.picking_type;
```
```suggestion
                const buffer = this.state_get_data("scan_location").buffer;
                if (_.isEmpty(buffer)) {
                    return {};
                }
                return buffer.picking_type;
```
same for zone_location ;)

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -540,6 +554,16 @@ const ZonePicking = {
                             this.odoo.call("scan_location", {barcode: scanned.text})
                         );
                     },
+                    on_unload_at_destination: () => {
+                        const loaded_data = this.state.data.buffer_is_loaded;
+                        this.odoo._update_headers(
```
simahawk: IMO you don't need this call to `_update_headers` because the headers will be inject automatically w/ values from `current_*` which you already overridden. 
If they don't get refreshed because `odoo` computed property is not refreshed I would create a new one instead:
```
            const odoo_params = this._get_odoo_params();
            const odoo = this.$root.getOdoo(odoo_params);
```

simahawk (changes requested):

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
TDu: The problem is `_get_odoo_params` relies on `current_picking_type` and `current_zone_location` and they would both need some re factoring only for this specific case. Knowing that they are used in other places in this module I found safer to not change them and do the `_update_headers` here.

shopfloor/services/zone_picking.py
TDu: Changed.

shopfloor/services/zone_picking.py
TDu: done

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
TDu: done

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
simahawk: Then I prefer you go for
```
const odoo_params = this._get_odoo_params(); 
// set your custom params here
const odoo = this.$root.getOdoo(odoo_params);
```
w/out altering the other object

TDu: Done that.

simahawk (approved):
LG, can you squash?

TDu:
Squashed

jbaudoux:
@TDu I just tested and it is not working. You indeed now have the button and can open the unload screen but then you cannot unload, I get an error

TDu:
> @TDu I just tested and it is not working. You indeed now have the button and can open the unload screen but then you cannot unload, I get an error

Oops, there is a fix #442

PR #441, opened 2022-09-05
[14.0] stock_available_to_promise_release: multi-stages of release

jbaudoux:
In case of multi-stages releases, the priority date must be kept.
The new moves created have bigger ids, so the comparison sign for moves having same priority date has been inverted to propagate the allocation to the new moves

cc @sebalix @mt-software-de

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -76,6 +76,29 @@ class StockMove(models.Model):
             GROUP BY move.id;
         """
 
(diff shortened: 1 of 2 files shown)


mt-software-de (commented):
Would it be not also be possible, to add a priority field on the stock.picking.type?
Because then we can configure which moves are handled first?

rousseldenis (approved):
LGTM

lmignon (approved):
Makes sens. :+1:

sebalix (approved):

PR #443, opened 2022-09-16
[14.0] Shopfloor imp zone picking and cluster picking

TDu:
ref.: rau-44

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -254,6 +254,20 @@ class MessageAction(Component):
             "body": _("Several lots found in %s, please scan the lot." % package.name),
         }
 
+    def several_move_in_different_location(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "Several moves found on different locations, please scan first the location."
+            ),
+        }
+
+    def several_move_with_different_lot(self):
+        return {
+            "message_type": "warning",
+            "body": _("Several moves found for different lots, please scan the lot."),
+        }
+
     def several_products_in_location(self, location):
         return {
             "message_type": "warning",
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -248,12 +248,24 @@ class MessageAction(Component):
             "body": _("Several lots found in %s, please scan a lot.") % location.name,
         }
 
+    def several_lots_in_package(self, package):
+        return {
+            "message_type": "error",
+            "body": _("Several lots found in %s, please scan the lot." % package.name),
+        }
+
     def several_products_in_location(self, location):
         return {
             "message_type": "warning",
             "body": _(
                 "Several products found in %s, please scan a product." % location.name
             ),
+        }
+
+    def several_products_in_package(self, package):
+        return {
+            "message_type": "error",
+            "body": _("Several products found in %s, please scan the product." % package.name),
         }
 
     def no_product_in_location(self, location):
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -279,7 +279,9 @@ class MessageAction(Component):
     def several_products_in_package(self, package):
         return {
             "message_type": "error",
-            "body": _("Several products found in %s, please scan the product." % package.name),
+            "body": _(
+                "Several products found in %s, please scan the product." % package.name
+            ),
         }
 
     def no_product_in_location(self, location):
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -490,21 +490,31 @@ class ClusterPicking(Component):
         """Product scanned, check if we can work with it.
 
         If scanned product is part of several packages in the same location,
-        we can't be sure it's the correct one, in such case, ask to scan a package
+        we can't be sure it's the correct one, in such case, ask to scan a package.
+
+        If the product is tracked by lot and there is only one lot id in the location
+        not in a package. It can safely be picked up.
         """
-        if move_line.product_id.tracking in ("lot", "serial"):
-            return self._response_for_start_line(
-                move_line, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
-            )
-        other_product_lines = picking.move_line_ids.filtered(
-            lambda l: l.product_id == product and l.location_id == move_line.location_id
+        message = None
+        location_quants = move_line.location_id.quant_ids.filtered(
+            lambda quant: quant.quantity > 0 and quant.product_id == product
         )
-        packages = other_product_lines.mapped("package_id")
+        packages = location_quants.mapped("package_id")
+
+        if move_line.product_id.tracking == "lot":
+            lots_at_location = location_quants.mapped("lot_id")
+            if len(lots_at_location) > 1 or packages:
+                message = self.msg_store.scan_lot_on_product_tracked_by_lot()
+        elif move_line.product_id.tracking == "serial":
+            message = self.msg_store.scan_lot_on_product_tracked_by_lot()
+        if message:
+            return self._response_for_start_line(move_line, message=message)
+
         # Do not use mapped here: we want to see if we have more than one package,
         # but also if we have one product as a package and the same product as
         # a unit in another line. In both cases, we want the user to scan the
         # package.
-        if packages and len({line.package_id for line in other_product_lines}) > 1:
+        if packages and len({quant.package_id for quant in location_quants}) > 1:
             return self._response_for_start_line(
                 move_line,
                 message=self.msg_store.product_multiple_packages_scan_package(),
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -527,13 +527,16 @@ class ClusterPicking(Component):
         If we scanned a lot and it's part of several packages, we can't be
         sure the user scanned the correct one, in such case, ask to scan a package
         """
-        other_lot_lines = picking.move_line_ids.filtered(lambda l: l.lot_id == lot)
-        packages = other_lot_lines.mapped("package_id")
+        location_quants = move_line.location_id.quant_ids.filtered(
+            lambda quant: quant.quantity > 0 and quant.lot_id == lot
+        )
+        packages = location_quants.mapped("package_id")
+
         # Do not use mapped here: we want to see if we have more than one
         # package, but also if we have one lot as a package and the same lot as
-        # a unit in another line. In both cases, we want the user to scan the
+        # a unit in another quant. In both cases, we want the user to scan the
         # package.
-        if packages and len({line.package_id for line in other_lot_lines}) > 1:
+        if packages and len({quant.package_id for quant in location_quants}) > 1:
             return self._response_for_start_line(
                 move_line, message=self.msg_store.lot_multiple_packages_scan_package()
             )
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -530,7 +530,7 @@ class ClusterPicking(Component):
         location_quants = move_line.location_id.quant_ids.filtered(
             lambda quant: quant.quantity > 0 and quant.lot_id == lot
         )
-        packages = location_quants.mapped("package_id")
+        packages = location_quants.package_id
 
         # Do not use mapped here: we want to see if we have more than one
         # package, but also if we have one lot as a package and the same lot as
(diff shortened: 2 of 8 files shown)


shopfloor/actions/message.py
```diff
@@ -256,6 +276,12 @@ def several_products_in_location(self, location):
             ),
         }
 
+    def several_products_in_package(self, package):
+        return {
+            "message_type": "error",
+            "body": _("Several products found in %s, aborting." % package.name),
```
```suggestion
            "body": _("Several products found in %s, please scan the product." % package.name),
```

shopfloor/actions/message.py
```diff
@@ -248,6 +248,26 @@ def several_lots_in_location(self, location):
             "body": _("Several lots found in %s, please scan a lot.") % location.name,
         }
 
+    def several_lots_in_package(self, package):
+        return {
+            "message_type": "error",
+            "body": _("Several lots found in %s, aborting." % package.name),
```
```suggestion
            "body": _("Several lots found in %s, please scan the lot." % package.name),
```

shopfloor/actions/message.py
```diff
@@ -248,6 +248,26 @@ def several_lots_in_location(self, location):
...
+        }
+
+    def several_move_in_different_location(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "Several moves found on different locations, please scan first a location."
```
```suggestion
                "Several moves found on different locations, please scan first the location."
```

shopfloor/actions/message.py
```diff
@@ -248,6 +248,26 @@ def several_lots_in_location(self, location):
...
+            ),
+        }
+
+    def several_move_with_different_lot(self):
+        return {
+            "message_type": "warning",
+            "body": _("Several moves found for different lots, please scan a lot."),
```
```suggestion
            "body": _("Several moves found for different lots, please scan the lot."),
```

shopfloor/services/cluster_picking.py
```diff
@@ -509,13 +527,16 @@ def _scan_line_by_lot(self, picking, move_line, lot):
         If we scanned a lot and it's part of several packages, we can't be
         sure the user scanned the correct one, in such case, ask to scan a package
         """
-        other_lot_lines = picking.move_line_ids.filtered(lambda l: l.lot_id == lot)
-        packages = other_lot_lines.mapped("package_id")
+        location_quants = move_line.location_id.quant_ids.filtered(
+            lambda quant: quant.quantity > 0 and quant.lot_id == lot
+        )
+        packages = location_quants.mapped("package_id")
```
jbaudoux: No need for mapped in v14

shopfloor/services/cluster_picking.py
```diff
@@ -530,21 +551,17 @@ def _scan_line_by_location(self, picking, move_line, location):
...
-        location = move_line.location_id
-        ml_search = self.search_move_line
-        pending_lines = ml_search.search_move_lines_by_location(location)
-
-        lots = pending_lines.mapped("lot_id")
-
+        location_quants = move_line.location_id.quant_ids.filtered(
+            lambda quant: quant.quantity > 0
+        )
+        lots = location_quants.mapped("lot_id")
```
jbaudoux: No need for mapped in v14

shopfloor/services/cluster_picking.py
```diff
@@ -530,21 +551,17 @@ def _scan_line_by_location(self, picking, move_line, location):
...
         if len(lots) > 1:
             return self._response_for_start_line(
                 move_line,
                 message=self.msg_store.several_lots_in_location(move_line.location_id),
             )
-
-        packages = pending_lines.mapped("package_id")
-        products = pending_lines.mapped("product_id")
-
+        packages = location_quants.mapped("package_id")
```
jbaudoux: No need for mapped in v14

shopfloor/services/cluster_picking.py
```diff
@@ -530,21 +551,17 @@ def _scan_line_by_location(self, picking, move_line, location):
...
+        products = location_quants.mapped("product_id")
```
jbaudoux: No need for mapped in v14

shopfloor/services/zone_picking.py
```diff
@@ -495,15 +511,18 @@ def _scan_source_location(self, barcode, confirmation=False):
...
         product = quants.product_id
         lot = quants.lot_id
         package = quants.package_id
         return product, lot, package
 
-    def _scan_source_package(self, barcode, confirmation=False):
+    def _scan_source_package(self, barcode, confirmation=False, product_id=False):
         """Search a package and find available lines for it.
 
         Fist search for lines that have the specific package.
```
```suggestion
        First search for lines that have the specific package.
```

shopfloor/services/zone_picking.py
```diff
@@ -555,32 +581,44 @@ def _scan_source_product(self, barcode, confirmation=False):
         if not product:
             return response, message
         move_lines = self._find_location_move_lines(product=product)
-        if move_lines:
+        if len(move_lines.mapped("location_id")) > 1:
```
jbaudoux: No need for mapped in v14

shopfloor/services/zone_picking.py
```diff
@@ -555,32 +581,44 @@ def _scan_source_product(self, barcode, confirmation=False):
...
-        lots = search.lot_from_scan(barcode, limit=None)
+        lots = search.lot_from_scan(barcode, products=products, limit=None)
         if not lots:
             return response, message
         for lot in lots:
             move_lines = self._find_location_move_lines(lot=lot)
-            if move_lines:
+            if not move_lines:
+                continue
+            if len(move_lines.mapped("location_id")) > 1:
```
jbaudoux: No need for mapped in v14

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -72,10 +102,20 @@ def test_scan_line_serial_ok(self):
         self._scan_line_ok(line, line.lot_id.name)
 
     def test_scan_line_error_product_tracked(self):
-        """Scan a product tracked by lot, must scan the lot"""
+        """Scan a product tracked by lot, must scan the lot.
+
+        If for the same product there is multiple lot in the location.
+        The user need to scan the lot.
```
jbaudoux:
```suggestion
        If for the same product there is multiple lots in the location,
        the user must scan the lot.
```

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -260,7 +312,7 @@ def test_scan_line_location_error_several_products(self):
         line = self.batch.picking_ids.move_line_ids
         location = line.location_id
         # add a second product in the location
-        self._update_qty_in_location(location, self.product_b, 10)
+        # self._update_qty_in_location(location, self.product_b, 10)
```
jbaudoux: on purpose?

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -288,7 +340,9 @@ def test_scan_line_location_error_several_products(self):
         # Although, if one line is already processed,
         # we can work automatically on the next one
         line.qty_done = line.product_uom_qty
-        self._scan_line_ok(new_move.move_line_ids[0], line.location_id.barcode)
+        line._action_done()
+        # FIXME the previous action_done does not work !
```
jbaudoux: FIXME

jbaudoux (commented):
Thanks for the improvements. A few minor remarks and a test you disabled with a FIXME

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -249,6 +300,7 @@ def test_scan_line_location_error_several_package(self):
         # Although, if one line is already processed,
         # we can work automatically on the next one
         line.qty_done = line.product_uom_qty
+        line._action_done()
```
sebalix: Calling this method at the `stock.move.line` level seems wrong, what was the intent?

shopfloor/tests/test_cluster_picking_scan_line.py
sebalix: Same remark, we should not call `_action_done` on a line, this one is called by the `stock.move._action_done()`, itself called by `stock.picking._action_done()` (which is the only method we should deal with when it comes to validate a move).

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -326,6 +380,7 @@ def test_scan_line_location_error_several_lots(self):
...
```
sebalix: `/!\`

TDu:
@sebalix and @jbaudoux I have done the requested change and refactor the tests that caused problems.

jbaudoux (approved):
Thanks. LGTM

TDu:
I squashed the fixups.

shopfloor/tests/test_zone_picking_select_line.py
```diff
@@ -419,6 +426,34 @@ def test_scan_source_barcode_lot(self):
...
+        self._fill_stock_for_moves(
+            picking.move_lines, in_lot=lot, location=self.zone_sublocation3
+        )
+        picking.action_assign()
+        response = self.service.dispatch(
+            "scan_source",
+            params={"barcode": lot.name},
+        )
+        # FIX ME: need to filter lines only on lot scanned !!
+        move_lines = self.service._find_location_move_lines()
```
sebalix, line 446: Is it still an issue?

shopfloor/tests/test_zone_picking_select_line.py
```diff
@@ -404,6 +404,13 @@ def test_scan_source_barcode_lot(self):
         """Scan source: scanned lot has one related move line,
         next step 'set_line_destination' expected on it.
         """
+        # Product C has already one lot from test_zone_picking_base.py
+        # So lets add one more lot for that product in a different location.
+        pick = self._create_picking(lines=[(self.product_c, 10)])
+        self._fill_stock_for_moves(
+            pick.move_lines, in_lot=True, location=self.zone_sublocation1
+        )
+        pick.action_assign()
```
sebalix, line 413: Why do we need to add this in this test? It didn't run well without this change? Or it is to ensure that everything is working as expected by adding another lot on a different location?

shopfloor/services/cluster_picking.py
```diff
@@ -490,21 +490,31 @@ def _scan_line_by_product(self, picking, move_line, product):
...
-        if move_line.product_id.tracking in ("lot", "serial"):
-            return self._response_for_start_line(
-                move_line, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
-            )
-        other_product_lines = picking.move_line_ids.filtered(
-            lambda l: l.product_id == product and l.location_id == move_line.location_id
+        message = None
+        location_quants = move_line.location_id.quant_ids.filtered(
+            lambda quant: quant.quantity > 0 and quant.product_id == product
         )
```
sebalix, line 501: A `search` on `stock.quant` will be better I guess here, if you get thousands of quants in your location, it'll loop over all of them with `filtered`.

Also I'm wondering: we will grab quants that are not related to the current `picking`, even some reserved by other operations, is it what we want?! The packages and lots we are looking for should be those already reserved in `picking` IMO. cc @jbaudoux

sebalix: This commit and the [next one](https://github.com/OCA/wms/pull/443/commits/99a7cbdc2802f141fb285d5a10815a6f2aa1f55f) are titled

```
shopfloor: fix cluster picking _scan_line_by_product

Check the content of the location not the move lines of the picking.
```
```
shopfloor: fix cluster picking _scan_line_location

Fix the error raised if there are multiple packs, products or lot in the location.
Not based on move lines but based on actual stock at location.
```
So it has been done on purpose, but I do not understand why :thinking:

shopfloor/tests/test_cluster_picking_scan_line.py
sebalix: But if the only product available in the operation processed is in only one lot among reserved move lines, and in a second lot in another operation for the same location (but we don't care about), why should we block the scan? :thinking:

shopfloor/services/cluster_picking.py
jbaudoux: @sebalix You are in front of a location and we want to be sure you take the right product from that location. So it must look at what is on the location, not what is in the processed picking.
I think filtered is fine, there won't be much quants on a location bin.

jbaudoux: @sebalix Same for lot. If there are multiple lots on the location, you have to scan the right one. Scanning the product is only enough if there is only a single lot on the location. No matter what are the picking move lines

sebalix: This implies to use location bin, with a too big source location we'll fetch too many things. Perhaps it'll never happen, I haven't the big picture, because it's like we developed this feature to fit our customers config, while it could not work well in another context. I could be wrong.

sebalix (approved):

shopfloor/tests/test_zone_picking_select_line.py
TDu: Yes, it seems not to be implemented, see line below as well. Activating it fail.

shopfloor/tests/test_zone_picking_select_line.py
TDu: It ran well without it. It was to give the test a little more context.

shopfloor/services/cluster_picking.py
TDu: Are we not filtering on the quants related to that location and move line on line 499 ?

PR #460, opened 2022-10-02
[15.0][MIG] stock_available_to_promise_release

flachica:


rousseldenis:
@flachica Could you check at https://github.com/OCA/wms/tree/16.0/stock_available_to_promise_release

in commit migration, you'll have the fixes needed.

Moreover, there is an improvement PR there  : https://github.com/OCA/wms/pull/478

flachica:
I add changes from [this](https://github.com/OCA/wms/commit/7e8cec1225e8f79e517e8d620f379a863a093851) commit. I don't know if I'm doing in the correct way

rousseldenis:
> I add changes from [this](https://github.com/OCA/wms/commit/7e8cec1225e8f79e517e8d620f379a863a093851) commit. I don't know if I'm doing in the correct way

@flachica In fact, no. 

The good way is to cherry-pick with `git cherry-pick <commit sha>`  for each commit contained in it. This in order to keep authorship.

Thanks

flachica:
But by this way, are many changes that I can use in 15.0. In 16.0 picking_id.move_lines was renamed to picking.move_ids. What can I do to preserve authoship?

PR #470, opened 2022-10-18
[MIG] stock_storage_type: Migration to 16.0

rousseldenis:
Depends on :

- [x] https://github.com/OCA/product-attribute/pull/1160
- [x] https://github.com/OCA/product-attribute/pull/1161
- [x] https://github.com/OCA/server-tools/pull/2413
- [x] https://github.com/OCA/stock-logistics-workflow/pull/1102
- [x] https://github.com/OCA/stock-logistics-workflow/pull/1103
- [x] https://github.com/OCA/stock-logistics-workflow/pull/1104
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/1509
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/1655
- [x] https://github.com/OCA/stock-logistics-workflow/pull/1274

TODO: 

- [x] Remove test-requirements.txt entries


rousseldenis:
@jbaudoux @lmignon I think this is a first testable version.

stock_storage_type/tests/test_storage_type_move.py
```diff
@@ -346,29 +352,29 @@ def _levels_for(packages):
 
         # Cardbox with different product go into different cardbox location
         self.assertEqual(len(third_level), 1)
-        self.assertEqual(third_level.location_dest_id, self.cardboxes_bin_2_location)
+        self.assertEqual(third_level.location_dest_id, self.cardboxes_bin_3_location)
```
rousseldenis, line 349: @jbaudoux @lmignon This is the only test case that I've changed.

It seems that order has been changed

jbaudoux:
@rousseldenis Do you think we can replace `stock.location.storage.type` by the new standard `stock.storage.category` and add a tab on the "Storage Category" called "Capacity by Storage" where you show the list of available storage types?
Drawback I see is:
* the max_height won't be supported for the capacity by package type or product
* the location can only have a single storage category but that's not bad limitation

rousseldenis:
> @rousseldenis Do you think we can replace `stock.location.storage.type` by the new standard `stock.storage.category` and add a tab on the "Storage Category" called "Capacity by Storage" where you show the list of available storage types? Drawback I see is:
> 
> * the max_height won't be supported for the capacity by package type or product
> * the location can only have a single storage category but that's not bad limitation

Not yet tested but that can make sense to get back to standard.

jbaudoux:
@rousseldenis Coming back on this:
> * the location can only have a single storage category but that's not bad limitation

I checked more deeply the implications and that would mean that the `allow_new_product` needs to be also added on the relation between the storage category and the product/pack storage. It would be an optional setting overriding the default one from the category.

Today we are using this config: Location A has:
* location storage type LST-1 is empty for product/pack storage type PST-X
* location storage type LST-2 is do not mix for product/pack storage type PST-Y

That would become: location A has:
* location storage category LC-12 for product/pack storage type PST-X in mode is empty and for PST-Y in mode do not mix

lmignon:
> @rousseldenis Coming back on this:
> 
> > * the location can only have a single storage category but that's not bad limitation
> 
> I checked more deeply the implications and that would mean that the `allow_new_product` needs to be also added on the relation between the storage category and the product/pack storage. It would be an optional setting overriding the default one from the category.
> 
> Today we are using this config: Location A has:
> 
>     * location storage type LST-1 is empty for product/pack storage type PST-X
> 
>     * location storage type LST-2 is do not mix for product/pack storage type PST-Y
> 
> 
> That would become: location A has:
> 
>     * location storage category LC-12 for product/pack storage type PST-X in mode is empty and for PST-Y in mode do not mix

If we add information on the relation between storage category and pack/product storage, we could also add the max_weight at the same level

jbaudoux:
> If we add information on the relation between storage category and pack/product storage, we could also add the max_weight at the same level

The max weight/height is a location characteristic not depending on what you put in. According to me, no need to put that on the relation.

lmignon:
> The max weight/height is a location characteristic not depending on what you put in. According to me, no need to put that on the relation.

It makes sens. That means we can remove all the logic used to compute this max_height on the stock_location and declare the field on the stock_location if it doesn't exist.

lmignon:
@jbaudoux 

Since `stock.package.storage.type`is now a `stock.package.type`, we already have the relation on `stock.storage.category`. The only missing piece according to our last comment is that we should add an `allow_new_product` optional field on this relation. (Implemented by `stock.storage.category.capacity`)

jbaudoux:
> Since `stock.package.storage.type`is now a `stock.package.type`, we already have the relation on `stock.storage.category`. The only missing piece according to our last comment is that we should add an `allow_new_product` optional field on this relation. (Implemented by `stock.storage.category.capacity`)

No, you cannot replace `stock.package.storage.type`by `stock.package.type`. It is a complete different concept. The package type will say for example that it is a pallet. The storage type will tell the putaway configuration to apply. We need the 2 concepts as 2 different products on a pallet can require completely different putaway configuration.

jbaudoux:
> It makes sens. That means we can remove all the logic used to compute this max_height on the stock_location and declare the field on the stock_location if it doesn't exist.

It can be replaced by a related readonly to the storage category.

rousseldenis:
> No, you cannot replace `stock.package.storage.type`by `stock.package.type`. It is a complete different concept.

@jbaudoux Have you tested it on runboat ? I made it working with nothing changed at tests levels.

jbaudoux:
> > No, you cannot replace `stock.package.storage.type`by `stock.package.type`. It is a complete different concept.
> 
> @jbaudoux Have you tested it on runboat ? I made it working with nothing changed at tests levels.

The issue is that it is impossible to configure because it is 2 different concepts. We need multiple package storage types for a same package type. The package storage type represent the putaway rule you set for the package type. And we have many rules (package storage type) for a box or a pallet (package type).

jbaudoux (changes requested):
You cannot replace stock.package.storage.type by stock.package.type.
See previous comments

jbaudoux:
> The issue is that it is impossible to configure because it is 2 different concepts. We need multiple package storage types for a same package type. The package storage type represent the putaway rule you set for the package type. And we have many rules (package storage type) for a box or a pallet (package type).

After discussion, we can use the new `stock.package.type` instead of `stock.package.storage`. However, this is not to be confused by the package type OCA module (https://github.com/OCA/product-attribute/tree/14.0/product_packaging_type) which has then to not be replaced by the new `stock.package.type`

rousseldenis:
@jbaudoux I try to finish the refactor and I have one question about the combinations we could have on a particular location about strategies (do not mix products / mix - empty/not empty - same lot/not same lot).

Here, the first approach I took is to have a related field 'allow_new_product' on locations that comes from the storage category.

The problem is that I loose the values that are on capacities.

Do we want to keep that flexibility on strategies possible combinations ?

Like determined by an 'any()' : https://github.com/OCA/wms/blob/14.0/stock_storage_type/models/stock_location.py#L148

As we have now the information on two levels (category and capacity).

So, to summarize:

- Either we only have the 'alllow_new_product' value coming from the category
- Either we keep all 'do_not_....' fields on locations computed on (category or/and capacities)


It"s quite strategic as this is the way locations are filtered in order to determine the first good one.

Thanks for your input

jbaudoux:
This one:
> * Either we keep all 'do_not_....' fields on locations computed on (category or/and capacities)

rousseldenis:
> This one:
> 
> > * Either we keep all 'do_not_....' fields on locations computed on (category or/and capacities)

@jbaudoux And should it be computed only on capacities or on both capacities and category values (for each allow_new_product) ?

rousseldenis:
> This one:
> 
> > * Either we keep all 'do_not_....' fields on locations computed on (category or/and capacities)

And last but not least:

- As we have a 1 -> 1 relation between location > category and as capacities belong to one and only one category, we can have only one value of the 'allow_new_product' coming from the capacities. So, the fields computed on location side can have just one combination at one time. So, should we keep instead the 'do not mix lots' that I've added to 'allow_new_product' possible values on capacities side ?

jbaudoux:
> > This one:
> > > * Either we keep all 'do_not_....' fields on locations computed on (category or/and capacities)
> 
> And last but not least:
> 
> * As we have a 1 -> 1 relation between location > category and as capacities belong to one and only one category, we can have only one value of the 'allow_new_product' coming from the capacities. So, the fields computed on location side can have just one combination at one time. So, should we keep instead the 'do not mix lots' that I've added to 'allow_new_product' possible values on capacities side ?

On the category & capacity, we need to add that option "do not mix lots".
Then on the location, it serves only to know what to compute further. So you probably need to keep all the booleans on the location and compute them based on category and capacity. True if any is True. However, the boolean could be moved to the category.

rousseldenis:
> > > This one:
> > > > * Either we keep all 'do_not_....' fields on locations computed on (category or/and capacities)
> > 
> > 
> > And last but not least:
> > 
> > * As we have a 1 -> 1 relation between location > category and as capacities belong to one and only one category, we can have only one value of the 'allow_new_product' coming from the capacities. So, the fields computed on location side can have just one combination at one time. So, should we keep instead the 'do not mix lots' that I've added to 'allow_new_product' possible values on capacities side ?
> 
> On the category & capacity, we need to add that option "do not mix lots". Then on the location, it serves only to know what to compute further. So you probably need to keep all the booleans on the location and compute them based on category and capacity. True if any is True. However, the boolean could be moved to the category.

Geat! Let's go

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,652 @@
...
+    only_empty = fields.Boolean(
+        compute="_compute_only_empty", store=True, recursive=True
+    )
+
+    @api.depends(
+        "usage",
+        "computed_storage_category_id.do_not_mix_lots",
+        "computed_storage_category_id.capacity_ids.do_not_mix_lots",
+    )
+    def _compute_do_not_mix_lots(self):
```
rousseldenis, line 149: @jbaudoux So, the last thing is to move this to storage category.

As there is a manual field there too. Do you see an approach to allow setting a value on the category itself apart the capacities (colliding field names) ?

rousseldenis: @jbaudoux

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -0,0 +1,149 @@
...
+    allow_new_product = fields.Selection(
+        [
+            ("empty", "If the location is empty"),
+            ("same", "If all products are same"),
+            ("mixed", "Allow mixed products"),
+        ],
+        default="mixed",
+        required=True,
+    )
+    do_not_mix_lots = fields.Boolean(
```
jbaudoux: This needs to be replaced by an additional option in the selection `allow_new_product` that you can name `same_lot` = "If all lots are the same"

stock_storage_type/views/stock_storage_category.xml
```diff
@@ -0,0 +1,21 @@
...
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="stock_storage_category_form" model="ir.ui.view">
+        <field name="name">stock.storage.category.form (in stock_storage_type)</field>
+        <field name="model">stock.storage.category</field>
+        <field name="inherit_id" ref="stock.stock_storage_category_form" />
+        <field name="arch" type="xml">
+            <field name="package_type_id" position="after">
+                <field name="allow_new_product" />
+                <field
```
jbaudoux: include as an option of `allow_new_product`

stock_storage_type/models/stock_quant.py
```diff
@@ -0,0 +1,155 @@
...
+                ):
+                    capacity_fails.append(
+                        _(
+                            "Storage Capacity {storage_capacity} is flagged 'do not mix"
+                            " products' but there are other products in "
+                            "location."
+                        ).format(storage_capacity=capacity.display_name)
+                    )
+                    continue
+                if capacity.do_not_mix_lots and (
```
jbaudoux:
```suggestion
                if capacity.allow_new_product == "same_lot" and (
```

stock_storage_type/models/stock_storage_category.py
```diff
@@ -0,0 +1,94 @@
...
+
+class StockStorageCategory(models.Model):
+
+    _inherit = "stock.storage.category"
+
+    computed_location_ids = fields.One2many(
+        comodel_name="stock.location", inverse_name="computed_storage_category_id"
+    )
+
+    do_not_mix_lots = fields.Boolean(
```
jbaudoux: This needs to be replaced by an additional option in the selection `allow_new_product` that you can name `same_lot` = "If all lots are the same"

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -0,0 +1,149 @@
...
+        locations.
+        """
+        self.ensure_one()
+        location_domain = [
+            ("id", "in", candidate_locations.ids),
+            ("computed_storage_category_id.capacity_ids", "in", self.ids),
+        ]
+        # TODO this method and domain is applied once per storage type. If it's
+        # too slow at some point, we could group the storage types by similar
+        # configuration (only_empty, do_not_mix_products, do_not_mix_lots) and
```
jbaudoux: Comment to update with new selection options

stock_storage_type/models/stock_storage_category_capacity.py
rousseldenis: @jbaudoux I tried to do it but I don't remember the case where this does not work. I'll keep you in touch

rousseldenis:
@jbaudoux @lmignon Could you do a review pass on this one ?

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,653 @@
...
+        return self._get_pack_putaway_strategy(putaway_location, quants, product)
+
+    def _get_pack_putaway_strategy(self, putaway_location, quants, product):
+        package_type = False
+        if quants:
+            package_type = quants.package_id.package_type_id
+            _logger.debug(
+                "Computing putaway for pack %s (%s)"
+                % (quants.package_id, quants.package_id)
+            )
```
lmignon, line 356: @rousseldenis When no quant is provided you should look on the product to get the `package_type_id` . see https://github.com/acsone/wms/commit/48454dcb6731fc2c4bc805a882a82713e1a93ad4

lmignon (changes requested):
Thank you @rousseldenis. A little comment...

stock_storage_type/models/stock_location.py
rousseldenis: @lmignon This has been solved there https://github.com/acsone/wms/commit/48454dcb6731fc2c4bc805a882a82713e1a93ad4

rousseldenis:
@lmignon @jbaudoux I think all comments have been attended.

lmignon (commented):
@rousseldenis At first read, all this looks good to me. We need to make more tests to ensure that everything is working as expected on our database.

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+                    % allowed_location.complete_name
+                )
+                return allowed_location
+        _logger.debug(
+            "Could not find a valid putaway location, fallback to %s"
+            % putaway_location.complete_name
+        )
+        return putaway_location
+
+    def get_storage_locations(self, products=None):
```
rousseldenis, line 422: @jbaudoux We are questionning the 'products' attribute as there is no need of 'multiple' products anywhere.

The code calling this method is (here above):


image





We tend to say, we should change to 'product'.

If yes we could also remove the first() method there:

https://github.com/OCA/wms/blob/14.0/stock_storage_type_putaway_abc/models/stock_location.py#L40

But maybe you have a case where this is a recordset

rousseldenis: @jbaudoux

sbejaoui (approved):
LGTM (code review)

rousseldenis:
@jbaudoux @lmignon

simahawk:
There's plenty of changes here... Do you mind if we add the module as not installable in 16.0 to ease reviews?

lmignon:
> There's plenty of changes here... Do you mind if we add the module as not installable in 16.0 to ease reviews?

`not installable in 16.0 to ease reviews`.???? How can you review a module if you're not able to install-it to make functional tests?
For sure there are plenty of changes since some concepts into the initial implementation are now part of the odoo standard. The migration has been a complex process to partially reconcile the initial implementation with the odoo standard without loosing functionalities.  We use this branch in our deployment and it should remain installable to not break it.

simahawk:
@lmignon I have different opinion regarding the review, but ok. Maybe we can open another one... or I surrender and I can review commit by commit :)

simahawk:
>  How can you review a module if you're not able to install-it to make functional tests?

BTW I don't get this. Your PR here will still be installable and you can run tests and do all you want as normal...

jbaudoux:
> There's plenty of changes here... Do you mind if we add the module as not installable in 16.0 to ease reviews?

The adaptation to v16 had been finalized by Denis with required refactoring. I'm going to retest it and update my review.
This is about to be approved

stock_storage_type/models/stock_storage_location_sequence.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def button_show_locations(self):
+        xmlid = "stock.action_location_form"
+        action = self.env["ir.actions.act_window"]._for_xml_id(xmlid)
+        action["domain"] = [
+            ("parent_path", "=ilike", "{}%".format(self.location_id.parent_path)),
+            (
+                "allowed_location_storage_type_ids",
+                "in",
+                self.package_type_id.location_storage_type_ids.ids,
```
jbaudoux: AttributeError: 'stock.package.type' object has no attribute 'location_storage_type_ids'

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+                in ("same", "same_lot")
+            )
+
+    @api.depends(
+        "location_id", "storage_category_id", "location_id.computed_storage_category_id"
+    )
+    def _compute_computed_storage_category_id(self):
+        """
+        This computes the Storage Storage depending on:
+          - its own Storage Storage
```
```suggestion
        This computes the Storage Category depending on:
          - its own Storage Category
```

stock_storage_type/views/stock_location.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="view_location_form_inherit" model="ir.ui.view">
+        <field name="name">stock.location.form.inherit (in stock_storage_type)</field>
+        <field name="model">stock.location</field>
+        <field name="inherit_id" ref="stock.view_location_form" />
+        <field name="arch" type="xml">
+            <field name="storage_category_id" position="after">
+                <field name="computed_storage_category_id" readonly="1" />
```
jbaudoux:
```suggestion
                <field name="computed_storage_category_id" readonly="1" groups="stock.group_stock_storage_categories"/>
```

stock_storage_type/views/stock_location.xml
```diff
@@ -0,0 +1,22 @@
...
+            </field>
+            <field name="removal_strategy_id" position="after">
+                <field name="pack_putaway_strategy" />
+                <field name="pack_putaway_sequence" />
+                <field
+                    name="computed_storage_capacity_ids"
+                    widget="many2many_tags"
+                    readonly="1"
+                />
```
jbaudoux:
```suggestion
                <field name="pack_putaway_strategy" groups="stock.group_stock_storage_categories" />
                <field name="pack_putaway_sequence" groups="stock.group_stock_storage_categories" />
                <field
                    name="computed_storage_capacity_ids"
                    widget="many2many_tags"
                    readonly="1"
                    groups="stock.group_stock_storage_categories"
                />
```

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+        default="none",
+        string="Packs Put-Away Strategy",
+        help="This defines the storage strategy to use when packs are put "
+        "away in this location.\n"
+        "None: when a pack is moved to this location, it will not be put"
+        " away any further.\n"
+        "Ordered Children Locations: when a pack is moved to this "
+        "location, a suitable location will be searched in its children "
+        "locations according to the restrictions defined on their "
+        "respective location storage types.",
```
jbaudoux: As you added the package type on the product itself also, I would remove the restriction on "Packs"
```suggestion
        string="Put-Away Strategy",
        help="This defines the storage strategy based on package type to use when a product or package is put "
        "away in this location.\n"
        "None: when moved to this location, it will not be put"
        " away any further.\n"
        "Ordered Children Locations: when moved to this "
        "location, a suitable location will be searched in its children "
        "locations according to the restrictions defined on their "
        "respective location storage types.",
```

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+    )
+    pack_putaway_sequence = fields.Integer()
```
```suggestion
    pack_putaway_sequence = fields.Integer(
        string="Putaway Sequence",
        help="Allow to sort the valid locations by sequence for the storage strategy based on package type")
```
and even rename the field to package_type_putaway_sequence

stock_storage_type/models/stock_location.py
jbaudoux: resolved

stock_storage_type/models/stock_location.py
jbaudoux: This is required because a package can contain multiple products. This was clarified by adding a "s" to the product argument of the method call.
Can be marked as resolved.

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2022 ACSONE SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class StorageCategoryProductCapacity(models.Model):
+
+    _inherit = "stock.storage.category.capacity"
+
+    allow_new_product = fields.Selection(
```
jbaudoux, line 10: It is missing a _rec_name on allow_new_product.
On the location, it is currently displayed as:
image

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -0,0 +1,96 @@
...
+        comodel_name="stock.location",
+        related="storage_category_id.computed_location_ids",
+    )
+    has_restrictions = fields.Boolean(
+        compute="_compute_has_restrictions",
+        help="Technical: This is used to check if we need to display warning message",
+    )
+    # TODO: Check if this is convenient with the constraint on barcode field
+    # in core module
+    active = fields.Boolean(default=True)
```
jbaudoux: I would drop this active field

stock_storage_type/models/stock_storage_category.py
```diff
@@ -0,0 +1,94 @@
...
+    allow_new_product = fields.Selection(
+        selection_add=[("same_lot", "If lots are all the same")],
+        ondelete={"same_lot": "cascade"},
+    )
+
+    computed_location_ids = fields.One2many(
+        comodel_name="stock.location", inverse_name="computed_storage_category_id"
+    )
+
+    # TODO: Move these fields in another module ?
```
jbaudoux, line 19: TODO can be removed

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+            putaway_location, product, quantity, package, packaging, additional_qty
+        )
+        if package:
+            quants = package.quant_ids
+            product = package.single_product_id or product
+        else:
+            quants = self.env["stock.quant"].browse()
+        return self._get_pack_putaway_strategy(putaway_location, quants, product)
+
+    def _get_pack_putaway_strategy(self, putaway_location, quants, product):
```
jbaudoux: I would refactor a bit this and rename the method
```suggestion
        return self._get_package_type_putaway_strategy(putaway_location, product, quantity, package)

    def _get_package_type_putaway_strategy(self, putaway_location, product, quantity, package):
        ...
```

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+        package_type = False
+        if quants:
```
```suggestion
        if package.package_type_id:
```

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+            package_type = quants.package_id.package_type_id
+            _logger.debug(
+                "Computing putaway for pack %s (%s)"
```
```suggestion
                "Computing putaway for package % of package type %s"
                % (package, package.package_type_id)
```

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+                % (quants.package_id, quants.package_id)
+            )
+        else:
```
```
elif product.package_type_id:
  ....
else:
    return putaway_location

```

stock_storage_type/models/stock_location.py
```diff
@@ -0,0 +1,661 @@
...
+            # Get default package type on product if defined
+            package_type = product.package_type_id
+            if package_type:
+                _logger.debug(
+                    "Computing putaway for pack %s (%s - from product %s)"
+                    % (quants.package_id, quants.package_id, product)
```
```suggestion
                    "Computing putaway for product %s of package type %s"
                    % (product, product.package_type_id)
```

stock_storage_type/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+class ProductTemplate(models.Model):
+
+    _inherit = "product.template"
+
+    package_type_id = fields.Many2one(
+        "stock.package.type",
+        string="Package type",
+        help="Defines a 'default' package type for this product to be "
+        "applied on packages without product packagings for put-away "
+        "computations.",
```
jbaudoux:
```suggestion
        help="Defines a 'default' package type for this product to be "
        "applied on packages without product packagings and on put-away computation based on package type for product not in a package.",
```

stock_storage_type/models/stock_quant_package.py
```diff
@@ -0,0 +1,93 @@
...
+            if not package.package_type_id:
+                # if no storage type could be set by auto assign,
+                # fallback on the default product's storage type (if any)
+                package._sync_storage_type_from_single_product()
+        return res
+
+    @api.model_create_multi
+    def create(self, vals):
+        records = super().create(vals)
+        records._sync_storage_type_from_packaging()
```
jbaudoux: Rename in this file storage_type by package_type

jbaudoux (changes requested):

rousseldenis:
@jbaudoux Stay in touch, I will do the modifications today

stock_storage_type/models/stock_storage_category_capacity.py
rousseldenis: Did this:

https://github.com/OCA/stock-logistics-warehouse/pull/1655

image

stock_storage_type/models/stock_storage_category.py
rousseldenis: I've added it during migration as, for me, those fields can be put in a less specialized module.

stock_storage_type/models/stock_location.py
rousseldenis: @jbaudoux We need at least the quant as the package_type can come from it

jbaudoux: it is either on the package or on the product, you don't really need the quant

rousseldenis: @jbaudoux In fact, quant is needed further as we need to get some information for it for constraints in the select_allowed_locations(), at least the lot comes from it. So, I will let the function as is.

rousseldenis:
@jbaudoux I think this is ready

jbaudoux (changes requested):
The README needs to be rewritten.

I would also rename all `pack_putaway_strategy` into `package_type_putaway_strategy`

rousseldenis:
@jbaudoux @lmignon This one seems ready and cover now the flow without package (put in pack) for the allowed destinations

lmignon (approved):

jbaudoux (approved):
LG now

PR #474, opened 2022-10-26
[14.0][FIX] shopfloor, delivery: process lines of one delivery operation

sebalix:
If the same bulk product was present in several delivery operations (of the same picking type), the first call to `scan_deliver` was working, but with the side-effect to set the `qty_done` for the lines of ALL delivery operations related to this product.

This fix will process only the lines from one delivery operation at a time, leaving the other lines untouched so they can be processed by subsequent calls to `scan_deliver`, without raising the `"No product found among current transfers."` error anymore (product was not found because of the `qty_done` set during the first scan).

cc @jbaudoux 

Ref rau-69

### DIFF
--- a/shopfloor/services/delivery.py
+++ b/shopfloor/services/delivery.py
@@ -367,6 +367,10 @@ class Delivery(Component):
                     location=location,
                     message=self.msg_store.product_mixed_package_scan_package(),
                 )
+        # Filter lines to keep only ones from one delivery operation
+        # (we do not want to process lines of another delivery operation)
+        lines = lines._filter_on_picking(picking)
+        # Validate lines (this will validate the delivery if all lines are processed)
         if self._set_lines_done(lines, product_qty):
             return self._response_for_deliver(
                 location=location, message=self.msg_store.transfer_complete(new_picking)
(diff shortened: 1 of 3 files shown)


shopfloor/services/delivery.py
```diff
@@ -312,6 +312,11 @@ def _deliver_product(self, picking, product, product_qty=None, location=None):
             limit=limit,
             order="date_planned",
         )
+        # Filter lines to keep only ones from one delivery operation
+        # (we do not want to process lines of another delivery operation)
+        if len(lines.picking_id) > 1:
+            lines_picking = fields.first(picking or lines.picking_id)
+            lines = lines.filtered_domain([("picking_id", "=", lines_picking.id)])
```
jbaudoux: This should then be placed after the multiple_location check.

jbaudoux: And after the package check

sebalix: Done, I also moved this code in a dedicated method on `stock.move.line`.

jbaudoux (approved):
LGTM

PR #476, opened 2022-10-28
[14.0] shopfloor_manual_product_transfer: new warning about reserved qty

sebalix:
... and display processed quantity with the `packaging-qty-picker` in readonly mode + some minor improvements/fixes.

image

Ref: rau-67

### DIFF
--- a/shopfloor_manual_product_transfer/actions/message.py
+++ b/shopfloor_manual_product_transfer/actions/message.py
@@ -25,3 +25,12 @@ class MessageAction(Component):
             "message_type": "success",
             "body": _("Transfer {} canceled.").format(picking.name),
         }
+
+    def qty_assigned_to_preserve(self, product, qty_assigned):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "{qty_assigned} {uom} are reserved in this "
+                "location and should not be taken"
+            ).format(qty_assigned=qty_assigned, uom=product.uom_id.name),
+        }
--- a/shopfloor_manual_product_transfer/services/manual_product_transfer.py
+++ b/shopfloor_manual_product_transfer/services/manual_product_transfer.py
@@ -266,10 +266,8 @@ class ManualProductTransfer(Component):
         # Compute the initial quantity
         initial_qty = self._get_initial_qty(location, product, lot)
         # No product available quantity to move
-        if (
-            product
-            and lot
-            and float_is_zero(initial_qty, precision_rounding=product.uom_id.rounding)
+        if (product or lot) and float_is_zero(
+            initial_qty, precision_rounding=product.uom_id.rounding
         ):
             return self._response_for_scan_product(
                 location, message=self.msg_store.no_product_in_location(location)
--- a/shopfloor_manual_product_transfer/services/manual_product_transfer.py
+++ b/shopfloor_manual_product_transfer/services/manual_product_transfer.py
@@ -304,6 +304,24 @@ class ManualProductTransfer(Component):
                 message=self.msg_store.qty_exceeds_initial_qty(),
             )
 
+    def _check_quantity_in_stock(self, location, product, quantity, lot=None):
+        """Check if there is enough quantity of a product in the location."""
+        current_qty = self._get_product_qty(location, product, lot, free=True)
+        initial_qty = self._get_initial_qty(location, product, lot)
+        quantity_lt_current_qty = (
+            float_compare(
+                quantity, current_qty, precision_rounding=product.uom_id.rounding
+            )
+            == -1
+        )
+        quantity_gte_initial_qty = (
+            float_compare(
+                quantity, initial_qty, precision_rounding=product.uom_id.rounding
+            )
+            >= 0
+        )
+        return quantity_lt_current_qty and quantity_gte_initial_qty
+
     def set_quantity(self, location_id, product_id, quantity, lot_id=None):
         """Allows to change the initial quantity to move.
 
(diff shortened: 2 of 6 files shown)


shopfloor_manual_product_transfer/services/manual_product_transfer.py
```diff
@@ -511,9 +521,17 @@ def _unreserve_other_lines(self, location, move_lines):
...
+                # add 'product_id' in the domain here? We could block the entire
+                # process here (see check on 'extra_move_lines' below) because
+                # another move line for another product is found in the given
+                # location... is it what we want?
             ]
         )
         extra_move_lines = location_move_lines - move_lines
+        # TODO: question, why do we care about lines that could be related to
+        # other products here? Should it prevent us to unreserve the moves of
+        # the product we're interested in?
```
sebalix: ping @jbaudoux I would like your insight on this, I think `location_move_lines` could contain unrelated lines (of other products sharing the same location), maybe it doesn't happen IRL with location bins but to me it's a bug, unless I didn't understood what was the purpose of the unreserve process in this scenario.

jbaudoux: Indeed the method should take in argument an option product. For the location content transfer, it is correct to unreserve all others moves as we empty the location. For this scenario where we move only a product, we should limit to the given product.

sebalix: Fixed here:
- https://github.com/OCA/wms/pull/482

shopfloor_manual_product_transfer/services/manual_product_transfer.py
```diff
@@ -304,6 +324,17 @@ def _check_quantity(self, location, product, lot, quantity):
                 message=self.msg_store.qty_exceeds_initial_qty(),
             )
 
+    def _check_quantity_in_stock(self, location, product, quantity, lot=None):
+        """Check if there is enough quantity of a product in the location."""
+        current_qty = self._get_product_qty(location, product, lot, free=True)
+        initial_qty = self._get_initial_qty(location, product, lot)
+        return (
+            # FIXME use float_compare
```
TDu: Probably a good time to fix it.

shopfloor_manual_product_transfer_mobile/static/wms/src/scenario/manual_product_transfer.js
```diff
@@ -66,9 +76,11 @@ const ManualProductTransfer = {
 
             <div class="detail item-detail-card" v-if="state_is('scan_destination_location')">
                 <v-card :color="utils.colors.color_for('screen_step_done')">
-                    <v-card-title>
-                        <span class="label">Quantity = {{ quantity() }}</span>
-                    </v-card-title>
+                    <packaging-qty-picker
+                        :key="make_state_component_key(['packaging-qty-picker', 1232])"
```
simahawk: 1232?? this should be a uniq value. Use the id of the record if you have one

shopfloor_manual_product_transfer_mobile/static/wms/src/scenario/manual_product_transfer.js
```diff
@@ -130,6 +142,16 @@ const ManualProductTransfer = {
...
         },
+        warning: function () {
+            const data = this.state.data;
+            if (_.isEmpty(data)) {
+                return "";
+            }
+            if (!_.isEmpty(data.warning)) {
+                return data.warning;
+            }
+            return "";
```
simahawk:
```suggestion
            return _.result(this.state, "data.warning", "")
```

shopfloor_manual_product_transfer/actions/message.py
```diff
@@ -25,3 +25,12 @@ def transfer_canceled_success(self, picking):
             "message_type": "success",
             "body": _("Transfer {} canceled.").format(picking.name),
         }
+
+    def qty_assigned_to_preserve(self, product, qty_assigned):
+        return {
+            "message_type": "warning",
+            "body": _(
+                f"{qty_assigned} {product.uom_id.name} are reserved in this "
```
simahawk: Are you sure `f` strings work fine w/ translations?

shopfloor_manual_product_transfer_mobile/static/wms/src/scenario/manual_product_transfer.js
simahawk: BTW IMO this warning should be defined only frontend side and enabled w/ a flag instead.

simahawk (changes requested):

simahawk:
And please shorten the commits that are too long ;)

shopfloor_manual_product_transfer_mobile/static/wms/src/scenario/manual_product_transfer.js
sebalix: I think about it, and also think about the translations which are already integrated in the backend (not there are no translations in Shopfloor but it's easier on the backend).

shopfloor_manual_product_transfer/actions/message.py
sebalix: Not at all :D fixed

shopfloor_manual_product_transfer_mobile/static/wms/src/scenario/manual_product_transfer.js
sebalix: Fixed, everywhere + tested locally.

shopfloor_manual_product_transfer/services/manual_product_transfer.py
sebalix: Fixed

simahawk (approved):

TDu (approved):

PR #477, opened 2022-11-02
[14.0][IMP] shopfloor delivery: add scan carrier tracking ref

TDu:
When goods are delivered to the customer with a carrier. The carrier package will have a label with the carrier tracking reference.

This change improve the delivery scenario and allows to scan the carrier tracking reference to deliver the package.

ref: rau-72


TDu:
We don not want this anymore

PR #475, opened 2022-10-28
[16.0] stock_available_to_promise_release:  Migration to 16.0

lmignon:

PR #478, opened 2022-11-03
[16.0] stock_available_to_promise_release: manage backorders and allows unrelease

lmignon:
* [x] includes  #475 

By default, when an outgoing transfer is released, a backorder is created for
the remaining quantities for products not available at release time. This behaviour
can also be changed by stock route by setting the option "No backorder at release".
In such a case, moves created for unavailable products will remain in the original
outgoing picking and no backorder will be created. At the same time, outgoing
picking will not bet set to printed at release time if one move has been created
from a route whith the option "No backorder at release" set.

At the end of the picking process (when the picking is validated) of an outgoing
transfer, if a backorder is created, the chained moves can be automatically
unreleased. This behaviour is activated by setting the option "Unrelease on backorder"
on the stock picking type.

cc @jbaudoux

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -1,6 +1,6 @@
 # Copyright 2019-2020 Camptocamp (https://www.camptocamp.com)
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
-
+import itertools
 import logging
 import operator as py_operator
 
@@ -353,7 +353,8 @@ class StockMove(models.Model):
             backorder_links[unreleased_move.picking_id] = original_picking
 
         for backorder, origin in backorder_links.items():
-            backorder._release_link_backorder(origin)
+            if backorder != origin:
+                backorder._release_link_backorder(origin)
 
         self.env["procurement.group"].run_defer(procurement_requests)
 
@@ -382,24 +383,32 @@ class StockMove(models.Model):
             pickings.write({"priority": max(priorities)})
 
     def _after_release_assign_moves(self):
-        moves = self
-        while moves:
-            moves._action_assign()
-            moves = moves.mapped("move_orig_ids")
+        move_ids = []
+        for origin_moves in self._get_chained_moves_iterator("move_orig_ids"):
+            move_ids += origin_moves.ids
+        self.env["stock.move"].browse(move_ids)._action_assign()
 
     def _release_split(self, remaining_qty):
         """Split move and create a new picking for it.
 
-        Instead of splitting the move and leave remaining qty into the same picking
-        we move it to a new one so that we can release it later as soon as
-        the qty is available.
+        By default, when we split a move at release to isolate the remaining qty
+        into a new move, we also create a new picking for it so that we can
+        release it later as soon as the qty is available.
+        This behavior can be changed by setting the flag no_backorder_at_release
+        on the stock.route of the move. This will allow to create the backorder
+        at the end of the picking process and release the unreleased moves into
+        the same picking as long as the picking is not done. By doing so, we
+        can also cleanup the backorders of the linked pickings created when
+        a released move was not processed (no qty found, or no time to do it for
+         example).
         """
         context = self.env.context
         self = self.with_context(release_available_to_promise=True)
         # Rely on `printed` flag to make _assign_picking create a new picking.
         # See `stock.move._assign_picking` and
         # `stock.move._search_picking_for_assignation`.
-        if not self.picking_id.printed:
+        original_printed = self.picking_id.printed
+        if not self.picking_id.printed and not self.rule_id.no_backorder_at_release:
             self.picking_id.printed = True
         new_move = self  # Work on the current move if split doesn't occur
         new_move_vals = self._split(remaining_qty)
@@ -411,6 +420,9 @@ class StockMove(models.Model):
         # and the move is not assigned.
         new_move._assign_picking()
 
+        # restore the original value of the printed flag only used to ensure
+        # that a backorder is created if required
+        self.picking_id.printed = original_printed
         return new_move.with_context(**context)
 
     def _assign_picking_post_process(self, new=False):
@@ -419,3 +431,14 @@ class StockMove(models.Model):
         if priorities:
             self.picking_id.write({"priority": max(priorities)})
         return res
+
+    def _get_chained_moves_iterator(self, chain_field):
+        """Return an iterator on the moves of the chain.
+
+        The iterator returns the moves in the order of the chain.
+        The loop into the iterator is the current moves.
+        """
+        moves = self
+        while moves:
+            yield moves
+            moves = moves.mapped(chain_field)
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -514,8 +514,10 @@ class StockMove(models.Model):
         moves_to_unrelease._check_unrelease_allowed()
         moves_to_unrelease.write({"need_release": True})
         impacted_picking_ids = set()
+
         for move in moves_to_unrelease:
             iterator = move._get_chained_moves_iterator("move_orig_ids")
+            moves_to_cancel = self.env["stock.move"]
             next(iterator)  # skip the current move
             for origin_moves in iterator:
                 origin_moves = origin_moves.filtered(
@@ -526,7 +528,9 @@ class StockMove(models.Model):
                     impacted_picking_ids.update(origin_moves.mapped("picking_id").ids)
                     # avoid to propagate cancel to the original move
                     origin_moves.write({"propagate_cancel": False})
-                    origin_moves._action_cancel()
+                    # origin_moves._action_cancel()
+                    moves_to_cancel |= origin_moves
+            moves_to_cancel._action_cancel()
         moves_to_unrelease.write({"need_release": True})
         for picking, moves in itertools.groupby(
             moves_to_unrelease, lambda m: m.picking_id
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -427,9 +427,6 @@ class StockMove(models.Model):
 
         self.env["procurement.group"].run_defer(procurement_requests)
 
-        # Set all transfers released to "printed", consider the work has
-        # been planned and started and another "release" of moves should
-        # (for instance) merge new pickings with this "round of release".
         pulled_moves._after_release_assign_moves()
         pulled_moves._after_release_update_chain()
 
@@ -473,12 +470,6 @@ class StockMove(models.Model):
         """
         context = self.env.context
         self = self.with_context(release_available_to_promise=True)
-        # Rely on `printed` flag to make _assign_picking create a new picking.
-        # See `stock.move._assign_picking` and
-        # `stock.move._search_picking_for_assignation`.
-        original_printed = self.picking_id.printed
-        if not self.picking_id.printed and not self.rule_id.no_backorder_at_release:
-            self.picking_id.printed = True
         new_move = self  # Work on the current move if split doesn't occur
         new_move_vals = self._split(remaining_qty)
         if new_move_vals:
@@ -488,10 +479,6 @@ class StockMove(models.Model):
         # thus the `_should_be_assigned` condition is not satisfied
         # and the move is not assigned.
         new_move._assign_picking()
-
-        # restore the original value of the printed flag only used to ensure
-        # that a backorder is created if required
-        self.picking_id.printed = original_printed
         return new_move.with_context(**context)
 
     def _assign_picking_post_process(self, new=False):
@@ -569,3 +556,13 @@ class StockMove(models.Model):
                 break
             origins -= origin
         return new_origin_moves
+
+    def _search_picking_for_assignation_domain(self):
+        domain = super()._search_picking_for_assignation_domain()
+        if self.env.context.get("release_available_to_promise"):
+            force_new_picking = not self.rule_id.no_backorder_at_release
+            if force_new_picking:
+                domain = expression.AND([domain, [("id", "!=", self.picking_id.id)]])
+        if self.picking_type_id.prevent_new_move_after_release:
+            domain = expression.AND([domain, [("last_release_date", "=", False)]])
+        return domain
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -330,6 +330,9 @@ class StockMove(models.Model):
         )
 
     def _action_cancel(self):
+        # Unrelease moves that can be, before canceling them.
+        moves_to_unrelease = self.filtered(lambda m: m.unrelease_allowed)
+        moves_to_unrelease.unrelease()
         super()._action_cancel()
         self.write({"need_release": False})
         return True
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -49,8 +49,77 @@ class StockMove(models.Model):
         search="_search_release_ready",
     )
     need_release = fields.Boolean(index=True, copy=False)
+    unrelease_allowed = fields.Boolean(compute="_compute_unrelease_allowed")
     zip_code = fields.Char(related="partner_id.zip", store=True)
     city = fields.Char(related="partner_id.city", store=True)
+
+    @api.depends("rule_id", "rule_id.available_to_promise_defer_pull")
+    def _compute_unrelease_allowed(self):
+        for move in self:
+            unrelease_allowed = move._is_unreleaseable()
+            if unrelease_allowed:
+                iterator = move._get_chained_moves_iterator("move_orig_ids")
+                next(iterator)  # skip the current move
+                for origin_moves in iterator:
+                    unrelease_allowed = move._is_unrelease_allowed_on_origin_moves(
+                        origin_moves
+                    )
+                    if not unrelease_allowed:
+                        break
+            move.unrelease_allowed = unrelease_allowed
+
+    def _is_unreleaseable(self):
+        """Check if the move can be unrelease. At this stage we only check if
+        the move is at the end of a chain of moves and has the caracteristics
+        to be unrelease. We don't check the conditions on the origin moves.
+        The conditions on the origin moves are checked in the method
+        _is_unrelease_allowed_on_origin_moves.
+        """
+        self.ensure_one()
+        user_is_allowed = self.env.user.has_group("stock.group_stock_user")
+        return (
+            user_is_allowed
+            and not self.need_release
+            and self.state not in ("done", "cancel")
+            and self.picking_type_id.code == "outgoing"
+            and self.rule_id.available_to_promise_defer_pull
+        )
+
+    def _is_unrelease_allowed_on_origin_moves(self, origin_moves):
+        """We check that the origin moves are in a state that allows the unrelease
+        of the current move. At this stage, a move can't be unreleased if
+          * a picking is already printed. (The work on the picking is planed and
+            we don't want to change it)
+          * the processing of the origin moves is partially started.
+        """
+        self.ensure_one()
+        pickings = origin_moves.mapped("picking_id")
+        if pickings.filtered("printed"):
+            # The picking is printed, we can't unrelease the move
+            # because the processing of the origin moves is started.
+            return False
+        origin_moves = origin_moves.filtered(
+            lambda m: m.state not in ("done", "cancel")
+        )
+        origin_qty_todo = sum(origin_moves.mapped("product_qty"))
+        return (
+            float_compare(
+                self.product_qty,
+                origin_qty_todo,
+                precision_rounding=self.product_uom.rounding,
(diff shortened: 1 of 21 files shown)


stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -160,7 +160,13 @@ def _after_release_set_expected_date(self):
         new_expected_date = fields.Datetime.add(
             fields.Datetime.now(), minutes=prep_time
         )
-        self.scheduled_date = new_expected_date
+        move_to_update = []
+        for move in self.move_ids.filtered(lambda m: m.state == "assigned"):
```
jbaudoux: all released move of the chain must have the new_expected_date

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -383,10 +385,10 @@ def _after_release_update_chain(self):
             pickings.write({"priority": max(priorities)})
 
     def _after_release_assign_moves(self):
-        moves = self
-        while moves:
-            moves._action_assign()
-            moves = moves.mapped("move_orig_ids")
+        move_ids = self.ids
+        for ids in self._get_all_orig_ids().values():
```
jbaudoux: Need to check it's not having side effect on stock dynamic routing

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -427,3 +429,54 @@ def _assign_picking_post_process(self, new=False):
...
+        """
+        self.env.cr.execute(
+            SQL,
+            {
+                "select_col": AsIs(select_col),
+                "link_col": AsIs(link_col),
+                "move_ids": tuple(self.ids),
+            },
+        )
+        res.update(dict(self.env.cr.fetchall()))
```
jbaudoux: The chain of moves has just been created. Aren't all ids already in the cache? Do you have less queries with this change?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -51,9 +52,50 @@ class StockMove(models.Model):
...
 
+    @api.depends("rule_id", "rule_id.available_to_promise_defer_pull")
+    def _compute_unrelease_allowed(self):
+        user_is_allowed = self.env.user.has_group("stock.group_stock_user")
+        move_orig_ids_by_depth = self._get_all_orig_ids_by_depth(prefetch=True)
+        for move in self:
+            unrelease_allowed = (
+                user_is_allowed
+                and not move.need_release
+                and move.rule_id.available_to_promise_defer_pull
```
jbaudoux: and picking_type_id.code == 'outgoing' like done here https://github.com/OCA/wms/pull/478/files#diff-05db432ffe85dc4ee9d20893829c7d8c96d484cd2cb12c0be7e0a70a949be3d9R29

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -478,5 +542,35 @@ def _get_all_moves_from_chain(self, field):
...
+        impacted_picking_ids = set()
+        for move in self:
+            move_origs = self.browse(move_orig_ids[move.id]).filtered(
+                lambda m: m.state not in ("done", "cancel")
+            )
+            impacted_picking_ids.update(move_origs.mapped("picking_id").ids)
+            if move_origs:
+                # avoid to propagate cancel to the original move
+                move_origs.write({"propagate_cancel": False})
+                move_origs._action_cancel()
```
jbaudoux: Needs split before if `move_orig.product_qty` > `move.product_qty`.

Example use case - set a fixed procurement group on pick rule, pick moves are merged
* ship1 <- pack1 <- pickA
* ship2 <- pack2 <- pickA

Unreleasing ship1 needs to split the merged moves of pickA

stock_available_to_promise_release/models/stock_picking_type.py
```diff
@@ -14,6 +14,12 @@ class StockPickingType(models.Model):
         "backorder are unreleased if they come from a route configured "
         "to manually release moves",
     )
+    set_printed_at_release = fields.Boolean(
```
jbaudoux: This is required in case of backorder at release to ensure that 2 packs are not linked to a single ship.
Can you remind me why we would be interested by having this as an option?

stock_available_to_promise_release/models/stock_move.py
lmignon: The chained moves are not into the cache in case of unrelease and when computing if an unreleased is allowed. The goal is to be sure that with just 1 additional SQL query we've all the information required to compute in batch the `unrelease_allowed`field regardless of the moves chain depth. The benefit may not be enough compared to the complexity. Furthermore, it is true that such a query requires an explicit flush of the cache for the stock.move model and so, not sure if this is a real benefit in the end. I'll change the code.

stock_available_to_promise_release/models/stock_picking_type.py
lmignon: we need it since the backorder creation is specified on the route and a picking is linked to many routes... we could also set printed at release only if no route is configured with 'no_backorder_at_release'.

stock_available_to_promise_release/models/stock_move.py
lmignon: seems to be far more complex.... we could have more than one move as origin and one origin could have more than one move as destination :thinking:

stock_available_to_promise_release/models/stock_move.py
lmignon: :+1:

lmignon:
@jbaudoux All comments implemented

sebalix:
@lmignon #475 has been merged

lmignon:
Thank you @sebalix! Rebase done.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -49,9 +49,51 @@ class StockMove(models.Model):
...
     need_release = fields.Boolean(index=True, copy=False)
+    unrelease_allowed = fields.Boolean(compute="_compute_unrelease_allowed")
     zip_code = fields.Char(related="partner_id.zip", store=True)
     city = fields.Char(related="partner_id.city", store=True)
 
+    @api.depends("rule_id", "rule_id.available_to_promise_defer_pull")
+    def _compute_unrelease_allowed(self):
+        user_is_allowed = self.env.user.has_group("stock.group_stock_user")
+        for move in self:
+            unrelease_allowed = (
```
rousseldenis: @lmignon Shouldn't this kind of rule in a separate method in order allow override?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +460,73 @@ def _assign_picking_post_process(self, new=False):
         if priorities:
             self.picking_id.write({"priority": max(priorities)})
         return res
+
+    def _get_chained_moves_iterator(self, chain_field):
```
rousseldenis, line 494: Shouldn't this method go in stock_helper module as it's quite generic?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +460,73 @@ def _assign_picking_post_process(self, new=False):
...
+        move and the quantity in the origin moves.
+
+        Return the origins for the move's quantity.
+        """
+        self.ensure_one()
+        qty = self.product_qty
+        rounding = self.product_uom.rounding
+        new_origin_moves = self.env["stock.move"]
+        while float_compare(qty, 0, precision_rounding=rounding) > 0 and origins:
+            origin = origins[0]
```
rousseldenis: I'd prefer using fields.first()

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +460,73 @@ def _assign_picking_post_process(self, new=False):
...
+            if float_compare(qty, origin.product_qty, precision_rounding=rounding) >= 0:
+                qty -= origin.product_qty
+                new_origin_moves |= origin
+            else:
+                new_move_vals = origin._split(qty)
+                new_origin_moves |= self.create(new_move_vals)
```
rousseldenis, line 562: What about using a set of ids to avoid recordset operations that can be bulky ?

rousseldenis (commented):
LGTM. Some comments

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +460,73 @@ def _assign_picking_post_process(self, new=False):
...
+
+        The iterator returns the moves in the order of the chain.
+        The loop into the iterator is the current moves.
+        """
+        moves = self
+        while moves:
+            yield moves
+            moves = moves.mapped(chain_field)
+
+    def unrelease(self):
```
jbaudoux: Add an action (wizard) to call this from the UI on the move AND on the picking

jbaudoux (changes requested):

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +460,73 @@ def _assign_picking_post_process(self, new=False):
...
+        impacted_picking_ids = set()
+        for move in self:
+            iterator = move._get_chained_moves_iterator("move_orig_ids")
+            next(iterator)  # skip the current move
+            for origin_moves in iterator:
+                origin_moves = origin_moves.filtered(
+                    lambda m: m.state not in ("done", "cancel")
+                )
+                if origin_moves:
+                    origin_moves = self._split_origins(origin_moves)
```
```suggestion
                    origin_moves = move._split_origins(origin_moves)
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -49,9 +49,51 @@ class StockMove(models.Model):
...
+                        )
+                        <= 0
+                    )
+                    if not unrelease_allowed:
+                        break
+            move.unrelease_allowed = unrelease_allowed
+
+    def _check_unrelease_allowed(self):
+        for move in self:
+            if not move.unrelease_allowed:
```
jbaudoux, line 116: If one move of the picking chain is started, we should not be able to unrelease

stock_available_to_promise_release/models/stock_move.py
lmignon: hard to express this kind of check. I prefer to wait for the need of such a split...

stock_available_to_promise_release/models/stock_move.py
lmignon: because in this case, we don't expect a large recordset...

lmignon:
@jbaudoux @rousseldenis Changes done

lmignon:
@jbaudoux @rousseldenis @sebalix Can we move forward with this one. I need it to be merged to avoid some trouble between all my pending PRs around the `stock_release_channel` addon.
For example in #493 I includes the last changes from @mt-software-de https://github.com/OCA/wms/pull/506/commits/bfb4f9b1ba5d8ae0607c9b4c71166587a6399cec but since this commit modify also the code into the `stock_available_to_promise_release`, It becomes impossible to use the current PR in my own project without making a glue branch for the 2 PRs... At the same time, it become impossible to tests #508 without this glue branch...

lmignon:
As temp fix I removed the forward port of https://github.com/OCA/wms/commit/bfb4f9b1ba5d8ae0607c9b4c71166587a6399cec from the migration of `stock_release_channel`

rousseldenis (approved):
LGTM. Code review

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +487,82 @@ def _assign_picking_post_process(self, new=False):
...
+        The iterator returns the moves in the order of the chain.
+        The loop into the iterator is the current moves.
+        """
+        moves = self
+        while moves:
+            yield moves
+            moves = moves.mapped(chain_field)
+
+    def unrelease(self, safe_unrelease=False):
+        """Unrelease unreleasavbe moves
```
jbaudoux, line 506: typo

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +487,82 @@ def _assign_picking_post_process(self, new=False):
...
+                qty -= origin.product_qty
+                new_origin_moves |= origin
+            else:
+                new_move_vals = origin._split(qty)
+                new_origin_moves |= self.create(new_move_vals)
+                break
+            origins -= origin
+        return new_origin_moves
+
+    def _search_picking_for_assignation_domain(self):
```
jbaudoux, line 567: With the refactor https://github.com/OCA/wms/pull/533/files#diff-fca2e9865c2d75a9793dbe699bffdf976606d996b69d09cacd25660f2926e998R410 you don't need to hack this method anymore. There is a new method for moving the moves to a backorder that can be improved with the new flag

lmignon: @jbaudoux I would prefer to finish this and the following PRs before forward porting the latest changes from V14 as it becomes too complicated and creates too many conflicts to integrate new changes from 14 into the current PRs.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +487,82 @@ def _assign_picking_post_process(self, new=False):
...
+            for origin_moves in iterator:
+                origin_moves = origin_moves.filtered(
+                    lambda m: m.state not in ("done", "cancel")
+                )
+                if origin_moves:
+                    origin_moves = move._split_origins(origin_moves)
+                    impacted_picking_ids.update(origin_moves.mapped("picking_id").ids)
+                    # avoid to propagate cancel to the original move
+                    origin_moves.write({"propagate_cancel": False})
+                    origin_moves._action_cancel()
```
jbaudoux: You have to cancel starting of the first origin of the chain and go down the chain. Easier it to make a complete recordset and cancel them all together.
So in case of pick/pack/ship, you cannot cancel pack then pick. You have to cancel pick, then pack, or pick+pack together.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -419,3 +487,82 @@ def _assign_picking_post_process(self, new=False):
...
+        qty = self.product_qty
+        rounding = self.product_uom.rounding
+        new_origin_moves = self.env["stock.move"]
+        while float_compare(qty, 0, precision_rounding=rounding) > 0 and origins:
+            origin = fields.first(origins)
+            if float_compare(qty, origin.product_qty, precision_rounding=rounding) >= 0:
+                qty -= origin.product_qty
+                new_origin_moves |= origin
+            else:
+                new_move_vals = origin._split(qty)
```
jbaudoux, line 561: move_dest_ids on new_move_vals must be changed to be linked only on the next move that is concerned.
On the origin move, the next concerned move must be removed from move_dest_ids.

jbaudoux (changes requested):
I did some testing and had issues to unrelease. A fix proposal is on its way.

stock_available_to_promise_release/models/stock_move.py
jbaudoux: I see if we can merge first in v14 the backport of the new features you added in this migration. Then I'll rebase the refactor and it will be a lot easier to fwd port to this branch.

stock_available_to_promise_release/models/stock_move.py
jbaudoux: Note: This remark above is to be compatible with the module stock_picking_restrict_cancel_with_orig_move which ensures you don't cancel a move in the middle of a chain of moves (because it's bad)

stock_available_to_promise_release/views/stock_picking_type_views.xml
```diff
@@ -25,4 +25,22 @@
             </xpath>
         </field>
     </record>
+
+    <record id="stock_picking_type_form" model="ir.ui.view">
+        <field name="name">stock.picking.type.kanban</field>
```
jbaudoux: typo in the name. It's not the kanban

stock_available_to_promise_release/models/stock_picking_type.py
```diff
@@ -8,6 +8,17 @@ class StockPickingType(models.Model):
...
+    unrelease_on_backorder = fields.Boolean(
+        string="Unrelease on backorder",
+        help="If checked, when a backorder is created the moves into the "
+        "backorder are unreleased if they come from a route configured "
+        "to manually release moves",
+    )
+    prevent_new_move_after_release = fields.Boolean(
+        string="Prevent new move after release",
+        help="If checked, when a picking is released it's no more possible to "
+        "assign new moves to it.",
```
jbaudoux:
```suggestion
        help="If checked, once a delivery picking has been released, no more moves will be added to it. For example, if you add lines in the origin sales order, the new moves will be added to a new delivery.",
```

stock_available_to_promise_release/models/stock_picking_type.py
```diff
@@ -8,6 +8,17 @@ class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
     count_picking_need_release = fields.Integer(compute="_compute_picking_count")
+    unrelease_on_backorder = fields.Boolean(
+        string="Unrelease on backorder",
+        help="If checked, when a backorder is created the moves into the "
+        "backorder are unreleased if they come from a route configured "
+        "to manually release moves",
```
jbaudoux:
```suggestion
        help="If checked, when a backorder is created (i.e. at the validation of the delivery), the moves into the "
        "backorder are unreleased as long as they came from a route configured "
        "to manually release moves. This means that the moves that were not delivered are put back in need for release and the unprocessed internal pulled moves are canceled.",
```

stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -171,3 +179,20 @@ def action_open_move_need_release(self):
...
+            backorders_to_unrelease.unrelease()
+        return backorders
+
+    def unrelease(self, safe_unrelease=False):
+        """Unrelease the moves of the picking.
+
+        If safe_unrelease is True, the unreleasaable moves for which the
+        processing has already started will be ignored
+        """
+        self.mapped("move_ids").unrelease(safe_unrelease=safe_unrelease)
```
jbaudoux, line 198: no need for mapped

stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -171,3 +179,20 @@ def action_open_move_need_release(self):
...
+            lambda p: p.picking_type_id.unrelease_on_backorder
+        )
+        if backorders_to_unrelease:
+            backorders_to_unrelease.unrelease()
+        return backorders
+
+    def unrelease(self, safe_unrelease=False):
+        """Unrelease the moves of the picking.
+
+        If safe_unrelease is True, the unreleasaable moves for which the
```
jbaudoux: typo
```suggestion
        If safe_unrelease is True, the unreleasable moves for which the
```

jbaudoux:
@lmignon any reason why you always wrote "Un Release" instead of "Unrelease" ? (cc @TDu )

lmignon:
> @lmignon any reason why you always wrote "Un Release" instead of "Unrelease" ? (cc @TDu )

@jbaudoux my frenchglish...

lmignon:
@jbaudoux The management of the backorders introduce some side-effect into the computation of counters on the view.
With this last change, when some products are not available at release time we no more put the move into a new backorder pick (its done when the channel is put to sleep). Unfortunately some of the counters and fields computed from domains provided by the `_field_picking_domains` method (https://github.com/OCA/wms/pull/493/files#diff-a8bb5a8dfb388e49242b1d9fe2274b33dccde4e032d0983647f70fa52399ecbcR174) rely on the predicate that work is to be done as soon as a ship no more need to be released (the concept of partially released ship doesn't exist)

jbaudoux:
> I did some testing and had issues to unrelease. A fix proposal is on its way.

The fixes have been done in the backport on v14. They are in those commits that needs to be integrated here:

- https://github.com/OCA/wms/pull/556/commits/ea271a69769cd8fdab244562fdb955e3bfafc8b8
- https://github.com/OCA/wms/pull/556/commits/d3b09534492964a7ab6f22f6ca564b11d9a7b2e1
- https://github.com/OCA/wms/pull/556/commits/773f402d3c23d9a1f483a45613065b1f162060db

lmignon:
@jbaudoux commits included

jbaudoux (approved):

lmignon:
@jbaudoux @rousseldenis This one is ready to be merged...

lmignon:
@OCA/logistics-maintainers Someone to merge this one?

stock_available_to_promise_release/migrations/16.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,14 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info(
+        "stock_available_to_promise_release: Configure picking types"
+        " to prevent new moves into released pickings"
+    )
+    cr.execute("update stock_picking_type set prevent_new_move_after_release = true")
```
jbaudoux: This should be in a 16.0.2.0.0 migration script as this is also required to run on instances having already the module running in 16.0

jbaudoux (commented):
Can you change the version to 16.0.2.0.0 and make the migration script for that version?

lmignon:
@jbaudoux Changes done and commit history cleaned up

PR #480, opened 2022-11-04
[14.0] sf_manual_product_transfer: cancel button now cancels the created move

sebalix:
At the `confirm_quantity` step a move is created and we switch to the `scan_destination_location` step. From there, if we click on the `Cancel` button we want to cancel the move created in the previous step.

image

Depends on this one to fix `pre-commit` build:
- https://github.com/OCA/wms/pull/481


sebalix:
<s>NOTE: I still didn't test it from UI</s>
=> done, all good

TDu (approved):

PR #487, opened 2022-11-09
[15.0][MIG] stock_storage_type_putaway_abc

dreispt:


rousseldenis:
@dreispt I think you would look at https://github.com/OCA/wms/pull/470

A big refactor has been done to be as close as possible to Odoo core models

dreispt:
@rousseldenis Thank you for the pointer. We realized this module came with heavy dependencies that bring no value to our use case, so we gave up on it and just added the abc field as a customization.

PR #492, opened 2022-11-11
[14.0] shopfloor_mobile: checkout packaging list hide type

JuMiSanAr:
image

ref: rau-64

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/checkout.js
+++ b/shopfloor_mobile/static/wms/src/scenario/checkout.js
@@ -286,10 +286,6 @@ const Checkout = {
         select_delivery_packaging_manual_select_options: function () {
             return {
                 showActions: false,
-                list_item_options: {
-                    loud_title: true,
-                    fields: [{path: "packaging_type"}],
-                },
             };
         },
         select_line_manual_select_opts: function () {


sebalix (approved):

shopfloor_mobile/static/wms/src/scenario/checkout.js
```diff
@@ -283,15 +282,6 @@ const Checkout = {
                 },
             };
         },
-        select_delivery_packaging_manual_select_options: function () {
-            return {
-                showActions: false,
-                list_item_options: {
-                    loud_title: true,
-                    fields: [{path: "packaging_type"}],
```
simahawk, line 291: Can't you simply remove the fields from here instead of dropping the method?

simahawk (changes requested):
pls fix the commit msg being too long (see my update on PR title)

shopfloor_mobile/static/wms/src/scenario/checkout.js
JuMiSanAr: I've now removed the list_item_options entirely, as loud_title gave it an ugly look.

simahawk (approved):

PR #496, opened 2022-11-16
[MIG][16.0] stock_storage_type_putaway_abc

rousseldenis:
Depends on:
- [ ] #470


lmignon (approved):
LGTM (Code review)

rousseldenis:
After discussion, we need to add a hook on https://github.com/OCA/wms/pull/496/files#diff-6ef44c37f77d64daebdd91d3fc6f0aa468e31e5911f75a141e013a8b03609d8fR42

to add the possibility to change the order mechanism

stock_storage_type_putaway_abc/models/product_template.py
```diff
@@ -0,0 +1,12 @@
...
+from odoo import fields, models
+
+from .stock_location import ABC_SELECTION
+
+
+class ProductTemplate(models.Model):
+
+    _inherit = "product.template"
+
+    abc_storage = fields.Selection(ABC_SELECTION, default="b")
```
rousseldenis, line 12: This should be put in a new module. Too much dependencies pulled if we want just that one.

sbejaoui (approved):
LGTM, code review

PR #497, opened 2022-11-16
[16.0][ADD] stock_storage_type_putaway_abc_classification_sale_stock

rousseldenis:
Depends on:

- [ ] https://github.com/OCA/wms/pull/470
- [ ] https://github.com/OCA/product-attribute/pull/1192
- [ ] https://github.com/OCA/wms/pull/496
- [ ] https://github.com/OCA/product-attribute/pull/1196


stock_storage_type_putaway_abc_classification_sale_stock/models/stock_location.py
```diff
@@ -0,0 +1,24 @@
...
+    _inherit = "stock.location"
+
+    def _get_sorted_leaf_locations_orderby_abc(self, products):
+        """
+        'self' is the location we want to put products and apply the
+        putaway strategy.
+        Evaluate the abc classification profile ids in order to retrieve
+        the corresponding one.
+        """
+        location_level = products.abc_classification_product_level_ids.filtered(
```
lmignon, line 17: @rousseldenis product**s** means potentially a recordset with more than one product... Do we have to enforce that the method is always called with 1 product our do we have to take the first product into the recordset as into the initial implementation

lmignon (changes requested):

lmignon:
ping @rousseldenis can you reopen this one.

stock_storage_type_putaway_abc_classification_sale_stock/models/stock_location.py
rousseldenis: @jbaudoux Your input on this ? IMHO, we indeed could use ensure_one() as we must ensure location_level is 1.

jbaudoux: This can be called when computing the putaway of a stock.package.level for a mixed pallet containing multiple products.
Ideally we should get the main product of the set. `first` is current implementation but the product with the lowest (alphabetically) abc_storage would be a better implementation.

stock_storage_type_putaway_abc_classification_sale_stock/models/stock_location.py
```diff
@@ -0,0 +1,24 @@
...
+            lambda level, location=self: level.profile_type == "sale_stock"
+            and level.profile_id.warehouse_id == location.warehouse_id
```
jbaudoux, line 19: Where is `location.warehouse_id` coming from ? A location is not linked to a warehouse

rousseldenis: Of course it does: https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_location.py#L93

rousseldenis: @jbaudoux

PR #500, opened 2022-11-21
[14.0][ADD] shopfloor_reception

JuMiSanAr:
Shopfloor implementation of the reception scenario.
Allows to receive products and create the proper packs for each logistic unit.

Replaces PR #435 

Depends on:
- https://github.com/OCA/wms/pull/503
- https://github.com/OCA/stock-logistics-workflow/pull/1116
- https://github.com/OCA/stock-logistics-workflow/pull/1171

ref: rau-49 + rau-57 + rau-89 + rau-90

PR #503, opened 2022-11-23
[14.0][IMP]  Shopfloor improvements prior to reception scenario

JuMiSanAr:
This PR introduces new features in shopfloor to adapt it to the reception scenario introduced in #500 .

ref: rau-49 + rau-57

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -361,6 +361,16 @@ class MessageAction(Component):
             ),
         }
 
+    def source_document_multiple_pickings_scan_package(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                _(
+                    "This source document is part of multiple transfers, please scan a package."
+                )
+            ),
+        }
+
     def product_mixed_package_scan_package(self):
         return {
             "message_type": "warning",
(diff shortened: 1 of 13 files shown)


shopfloor/actions/message.py
```diff
@@ -361,6 +361,16 @@ def product_multiple_packages_scan_package(self):
             ),
         }
 
+    def source_document_multiple_pickings_scan_package(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                _(
+                    "This source document is part of multiple pickings, please scan a package."
```
```suggestion
                    "This source document is part of multiple transfers, please scan a package."
```

shopfloor/actions/message.py
```diff
@@ -382,6 +392,64 @@ def product_not_found_in_pickings(self):
...
+    def multiple_picks_found_select_manually(self):
+        return {
+            "message_type": "error",
+            "body": _("Several transfers found, please select a transfer manually."),
+        }
+
+    def no_transfer_for_packaging(self):
+        return {
+            "message_type": "error",
+            "body": _("No picking found for the scanned packaging."),
```
```suggestion
            "body": _("No transfer found for the scanned packaging."),
```

shopfloor/tests/test_actions_data.py
```diff
@@ -156,14 +161,14 @@ def test_data_move_line_package(self):
             "package_src": {
                 "id": move_line.package_id.id,
                 "name": move_line.package_id.name,
-                "move_line_count": 1,
+                "move_line_count": 0,
```
sebalix, line 165: That means `self.move_a` is `done`, `cancel` or `confirmed` here?

shopfloor/tests/test_actions_data.py
```diff
@@ -156,14 +161,14 @@ def test_data_move_line_package(self):
...
                 "weight": 20.0,
                 "storage_type": None,
             },
             "package_dest": {
                 "id": result_package.id,
                 "name": result_package.name,
-                "move_line_count": 0,
+                "move_line_count": 1,
```
sebalix, line 172: Same question, to be sure.

simahawk:
@JuMiSanAr pls rebase and answer @sebalix question :pray:

shopfloor/tests/test_actions_data.py
JuMiSanAr: Added a check for the state in these tests.

shopfloor/actions/search.py
```diff
@@ -31,11 +32,27 @@ def package_from_scan(self, barcode):
             return model.browse()
         return model.search([("name", "=", barcode)], limit=1)
 
-    def picking_from_scan(self, barcode):
+    def picking_from_scan(self, barcode, use_origin=False):
```
jbaudoux, line 35: This should be restricted to the picking types defined on the menu

jbaudoux (approved):
LGTM, just a small improvement remark related to picking search

shopfloor/actions/search.py
JuMiSanAr: This wasn't the case in the previous version either.
This check for picking types is done in each scenario's domain methods, as far as I can tell.

simahawk: Yes, this method is generic. You can validate the result where/when you need.

simahawk (approved):

simahawk:
@sebalix fine for you?

sebalix (approved):

sebalix:
Fine to me!

shopfloor/actions/search.py
jbaudoux: As it was before, it was fine because it was searching on the name and it's unique.
When searching on the origin, it may cause issue. The method was designed to return a single picking which is not the case anymore.
You can add a fixme for later but I would limit the last search to 1.

JuMiSanAr: This has been dealt with in the reception scenario as well (displaying the filtered list of pickings if there's more than 1 returned). That's why we allow more than 1 picking if we search by origin.

I think it should be each scenario that takes care of dealing with this case when filtering by origin.

PR #505, opened 2022-11-24
shopfloor_mobile: imp zone picking screen title

TDu:
If there is a current picking being worked on. Add it to the top of the screen. And remove redondant information.

ref.: rau-33


mmequignon (approved):

simahawk (approved):
LG

PR #493, opened 2022-11-14
[16.0][MIG] stock_release_channel: Migration to 16.0

lmignon:

PR #506, opened 2022-11-24
[16.0] stock_release_channel: Manage channel lifecycle

lmignon:
* [x] includes #493
* [x] depends on #478 

The availability of a release channel depends on its state. A release channel
can be in one of the following states:

  - Open: the channel is available and can be used to release transfers. New
    transfer are assigned automatically to this channel.
  - Locked: the channel is available but the release of transfers from the channel
    is no more possible. New transfers are still automatically assigned to this
  - Asleep: the channel is not available and cannot be used to release
    transfers. It is also no more possible to assign transfers to this channel.

New release channels are by default "Open". You can change its state by using
the "Lock" and "Sleep" buttons. When the "Sleep" button is used, in addition to
the state change, not completed transfers assigned to the channel are unassigned
from the channel. When the "Lock" button is used, only the state change is changed.
A locked channel can be unlocked by using the "Unlock" button.
A asleep channel can be waken up by using the "Wake up" button. When the "Wake up"
button is used, in addition to the state change, the system looks for pending
transfers requiring a release and try to assign them to a channel in the
"Open" or "Locked" state.

### DIFF
--- a/stock_release_channel/models/stock_release_channel.py
+++ b/stock_release_channel/models/stock_release_channel.py
@@ -170,6 +170,78 @@ class StockReleaseChannel(models.Model):
     )
     last_done_picking_name = fields.Char(compute="_compute_last_done_picking")
     last_done_picking_date_done = fields.Datetime(compute="_compute_last_done_picking")
+    state = fields.Selection(
+        selection=[("open", "Open"), ("locked", "Locked"), ("asleep", "Asleep")],
+        help="The state allows you to control the availability of the release channel.\n"
+        "* Open: Manual and automatic picking assignment to the release is effective "
+        "and release operations are allowed.\n "
+        "* Locked: Release operations are forbidden. (Assignement processes are "
+        "still working)\n"
+        "* Asleep: Assigned pickings not processed are unassigned from the release "
+        "channel.\n",
+        default="open",
+    )
+    is_action_lock_allowed = fields.Boolean(
+        compute="_compute_is_action_lock_allowed",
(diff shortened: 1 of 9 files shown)


stock_release_channel/models/stock_release_channel.py
```diff
@@ -621,3 +705,72 @@ def release_next_batch(self):
...
+    def action_unlock(self):
+        self._check_is_action_unlock_allowed()
+        self.write({"state": "open"})
+
+    def action_sleep(self):
+        self._check_is_action_sleep_allowed()
+        pickings_to_unassign = self.env["stock.picking"].search(
+            self._get_picking_to_unassign_domain()
+        )
+        pickings_to_unassign.write({"release_channel_id": False})
```
jbaudoux, line 788: Before unassigning, the picking must be "unreleased".
Could be some ongoing pickings cannot be unreleased and this action will be declined

lmignon: This PR doesn't depend on #478 I'll implement this requirement in #508

lmignon: @jbaudoux done

lmignon:
@simahawk @jbaudoux rebase done but  requires #478

rousseldenis (approved):
Code review.

@lmignon Could you remove the test-requirements commit ?

lmignon:
@rousseldenis test-requirements removed

jbaudoux (approved):
LG

PR #483, opened 2022-11-04
[14.0] Shopfloor: Add no_prefill_qty option v2

TDu:
Supersedes #421 

ref: rau-30

PR #507, opened 2022-11-25
[14.0][BUG]  shopfloor_mobile: fix highlights reactivity in checkout

JuMiSanAr:
This PR allows any scenario to trigger an event to update the manual_select highlights after an scan.

Depends on:
- https://github.com/OCA/wms/pull/483

ref: rau-81

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/checkout.js
+++ b/shopfloor_mobile/static/wms/src/scenario/checkout.js
@@ -7,6 +7,7 @@
 import {ScenarioBaseMixin} from "/shopfloor_mobile_base/static/wms/src/scenario/mixins.js";
 import {process_registry} from "/shopfloor_mobile_base/static/wms/src/services/process_registry.js";
 import {checkout_states} from "./checkout_states.js";
+import event_hub from "/shopfloor_mobile_base/static/wms/src/services/event_hub.js";
 
 const Checkout = {
     mixins: [ScenarioBaseMixin],
@@ -280,6 +281,19 @@ const Checkout = {
                     ],
                 },
             };
+        },
+        handle_manual_select_highlight_on_scan: function (res) {
+            const new_selected_package_data = res.data.select_package;
+            new_selected_package_data.selected_move_lines.forEach((line) => {
+                let checked = false;
+                if (line.qty_done > 0) {
+                    checked = true;
+                }
+                event_hub.$emit("manual_select:select_item", {
+                    rec_id: line.id,
+                    checked,
+                });
+            });
         },
         select_delivery_packaging_manual_select_options: function () {
             return {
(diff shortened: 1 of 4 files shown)


shopfloor_mobile_base/static/wms/src/components/manual_select.js
```diff
@@ -115,15 +121,27 @@ Vue.component("manual-select", {
...
+            this._updateElement(rec, event.target);
+            if (!this.opts.showActions) {
+                this._emitSelected(this._getSelected());
+            }
+        },
+        handleScan(new_data) {
+            new_data.selected_move_lines.forEach((line) => {
+                const query = $(`[input-value=${line.id}]`);
+                const elem = query[0];
+                if (line.qty_done > 0) {
```
simahawk: this cannot stay here, it's a generic component.

simahawk (changes requested):

shopfloor_mobile_base/static/wms/src/components/manual_select.js
JuMiSanAr: Moved it out of this component.

shopfloor_mobile/static/wms/src/scenario/checkout.js
```diff
@@ -281,6 +282,18 @@ const Checkout = {
                 },
             };
         },
+        handle_manual_select_highlight_on_scan: function (res) {
+            const new_data = res.data.select_package;
+            new_data.selected_move_lines.forEach((line) => {
+                //
```
TDu:
```suggestion
```

TDu (approved):

shopfloor_mobile/static/wms/src/scenario/checkout.js
```diff
@@ -281,6 +282,17 @@ const Checkout = {
                 },
             };
         },
+        handle_manual_select_highlight_on_scan: function (res) {
+            const new_data = res.data.select_package;
+            new_data.selected_move_lines.forEach((line) => {
+                const query = $(`[input-value=${line.id}]`);
```
simahawk: Do not rely on the markup. IMO you should simply get the ids of the records to select (or the records themselves) and trigger `manual_select:select_items` and the event handler in manual select will take care of selecting the right items

simahawk (changes requested):

shopfloor_mobile/static/wms/src/scenario/checkout.js
JuMiSanAr: Done.

shopfloor_mobile_base/static/wms/src/components/manual_select.js
```diff
@@ -45,6 +46,11 @@ Vue.component("manual-select", {
     mounted() {
         // Relies on properties
         this.selected = this._initSelected();
+
+        // Select item from outside this component.
+        event_hub.$on("manual_select:select_item", (data) => {
```
simahawk, line 51: Can you add a TODO? We should hook to the specific manual-select component via `refs` if we have multiple manual-select components in the same screen.

simahawk (approved):

PR #509, opened 2022-12-01
[ADD] shopfloor_base_default_profile

brian10048:
This PR supersedes #388 to add a default profile mechanism for shopfloor

Based on review comments from the previous PR:
- [x] Feature is a separate module
- [x] Configure default profile option per shopfloor app
- [x] Added flag for default profile to `shopfloor_app_info`
- [x] Apply default profile with hook to events on event_hub

@simahawk I'm having trouble with the JS side of this feature.  Would you be able to point me in the right direction to tie all of this together?

cc @bguillot


shopfloor_base_default_profile/models/shopfloor_app_user_profile.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2022 Brian McMaster <brian@mcmpest.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+from odoo import fields, models
+
+
+class ShopfloorAppUserProfile(models.Model):
+    _name = "shopfloor.app.user.profile"
```
simahawk, line 7: Do you really need this model?
Since profiles are linked to the app already https://github.com/OCA/wms/blob/14.0/shopfloor_base/models/shopfloor_app.py#L73 you can simply have a m2m on the user.

shopfloor_base_default_profile/models/res_user.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2022 Brian McMaster <brian@mcmpest.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    shopfloor_default_profile_ids = fields.One2many(
```
simahawk, line 9: Hmm I don't get why you can have multiple profiles on a user. 
If the whole point is to have "a default profile selected for the user" I'd expect to see only one here.

shopfloor_base_default_profile/services/user.py
```diff
@@ -0,0 +1,36 @@
...
+    @property
+    def _user_info_parser(self):
+        res = super()._user_info_parser
+        res.append(
+            ("shopfloor_default_profile_ids:default_profile", self._user_profile_parser)
+        )
+        return res
+    
+    def _user_profile_parser(self, rec, fname):
+        profile = rec[fname].search([("shopfloor_app_id", "=", self.collection.id)], limit=1).profile_id
```
simahawk: Do you really need this search?
IMO you should only expose the default profile for the current user w/ `self.env.user.shopfloor_default_profile_id`.
No?

simahawk (changes requested):
Regarding the front: you could hook to the event_hub event triggered after the config is loaded https://github.com/OCA/wms/blob/14.0/shopfloor_mobile_base/static/wms/src/main.js#L201 and then if you have a default profile, trigger the profile selected event https://github.com/OCA/wms/blob/14.0/shopfloor_mobile_base/static/wms/src/main.js#L119

brian10048:
Thank you @simahawk for taking the time to review this.  Your guidance is much appreciated.

I was able to get the default profile applied on the front in the latest commit.

The reason for multiple default profiles and why I made the new model `ShopfloorAppUserProfile` is in the case that we have multiple apps and a user needs a default profile for each app.

simahawk:
> Thank you @simahawk for taking the time to review this. Your guidance is much appreciated.
> 
> I was able to get the default profile applied on the front in the latest commit.
> 
> The reason for multiple default profiles and why I made the new model `ShopfloorAppUserProfile` is in the case that we have multiple apps and a user needs a default profile for each app.

Ok, I leave some more comments then.

shopfloor_base_default_profile/models/shopfloor_app_user_profile.py
```diff
@@ -0,0 +1,11 @@
...
+
+    user_id = fields.Many2one("res.users", string="User")
+    shopfloor_app_id = fields.Many2one("shopfloor.app", string="Shopfloor App")
```
simahawk, line 11: You must have a constraint here: you can have only one default profile per app per user.

shopfloor_base_default_profile/services/user.py
simahawk: There's no need to search IMO.
```suggestion
        profile = rec[fname].filtered(lambda x: x.shopfloor_app_id == self.collection.id).profile_id
```

simahawk (commented):
One more thing: would be nice to have tests ;)

shopfloor_base_default_profile/static/wms/src/main.js
```diff
@@ -0,0 +1,10 @@
+import event_hub from "/shopfloor_mobile_base/static/wms/src/services/event_hub.js";
```
simahawk: Can you please remove the `/wms/` level? It's  mistake we should have avoided...

simahawk (approved):
LG

jbaudoux (commented):
You can open a profile by scanning a workstation. This way the user do not need to choose a profile.

Considering this, I would have preferred to have a default workstation on the user than a default profile

PR #511, opened 2022-12-02
[14.0] shopfloor: fix zone picking scenario

TDu:
Doing a scan of a location to restrict the search for the next scan on a product, was not working.

ref.: rau-83

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -43,12 +43,15 @@ class ZonePicking(Component):
 
        * location, if only a single move line there; if a location is scanned
          and it contains several move lines, the view is updated to show only
-         them
+         them. The next scan (e.g. a product) will be based on the previous
+         scanned location.
        * package, if it is linked to a move line. If the package is not linked
          to an existing move line but can be a replacement for one, the view is
          updated to show only the fitting move lines. And the user can confirm
          the change of package by scanning it a second time.
-       * product
+       * product, if only a single move line matches. Otherwise the view is updated
+         to show only the matching move lines, The next scan (e.g. a location) will
+         be based on the previous product scanned.
        * lot
 
     5. The operator scans the destination for the line they scanned, this is where
@@ -194,10 +197,13 @@ class ZonePicking(Component):
         popup=None,
         confirmation_required=False,
         product=False,
+        sublocation=False,
     ):
         if confirmation_required and not message:
             message = self.msg_store.need_confirmation()
-        data = self._data_for_move_lines(move_lines, product=product)
+        data = self._data_for_move_lines(
+            move_lines, product=product, sublocation=sublocation
+        )
         data["confirmation_required"] = confirmation_required
         return self._response(
             next_state="select_line",
@@ -304,7 +310,12 @@ class ZonePicking(Component):
         }
 
     def _data_for_move_lines(
-        self, move_lines, zone_location=None, picking_type=None, product=None
+        self,
+        move_lines,
+        zone_location=None,
+        picking_type=None,
+        product=None,
+        sublocation=None,
     ):
         zone_location = zone_location or self.zone_location
         picking_type = picking_type or self.picking_type
@@ -315,6 +326,8 @@ class ZonePicking(Component):
         }
         if product:
             data["product"] = self.data.product(product)
+        if sublocation and sublocation != zone_location:
+            data["sublocation"] = self.data.location(sublocation)
         for data_move_line in data["move_lines"]:
             # TODO: this could be expensive, think about a better way
             # to retrieve if location will be empty.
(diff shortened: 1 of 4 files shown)


shopfloor/services/zone_picking.py
```diff
@@ -465,11 +475,15 @@ def list_move_lines(self):
...
+    def _list_move_lines(self, location, product=False, sublocation=False):
         move_lines = self._find_location_move_lines(location, product=product)
-        return self._response_for_select_line(move_lines, product=product)
+        return self._response_for_select_line(
+            move_lines, product=product, sublocation=sublocation
+        )
 
-    def _scan_source_location(self, barcode, confirmation=False, product_id=False):
+    def _scan_source_location(
+        self, barcode, confirmation=False, product_id=False, sublocation=False
```
jbaudoux, line 490: A location has been scanned, so we should not keep the previous location in the context.
I would not have changed this method

shopfloor/services/zone_picking.py
```diff
@@ -465,11 +475,15 @@ def list_move_lines(self):
         """
         return self._list_move_lines(self.zone_location)
 
-    def _list_move_lines(self, location, product=False):
+    def _list_move_lines(self, location, product=False, sublocation=False):
```
jbaudoux, line 481: no sublocation here, just location

shopfloor/services/zone_picking.py
```diff
@@ -572,15 +588,19 @@ def _scan_source_package(self, barcode, confirmation=False, product_id=False):
...
         message = None
         response = None
         search = self._actions_for("search")
         product = search.product_from_scan(barcode)
         if not product:
             return response, message
-        move_lines = self._find_location_move_lines(product=product)
+        move_lines = self._find_location_move_lines(
+            locations=sublocation, product=product
+        )
```
jbaudoux, line 608: first _list_move_lines below should get the location instead of the zone

shopfloor/services/zone_picking.py
```diff
@@ -465,11 +475,15 @@ def list_move_lines(self):
         """
         return self._list_move_lines(self.zone_location)
 
-    def _list_move_lines(self, location, product=False):
+    def _list_move_lines(self, location, product=False, sublocation=False):
         move_lines = self._find_location_move_lines(location, product=product)
-        return self._response_for_select_line(move_lines, product=product)
+        return self._response_for_select_line(
+            move_lines, product=product, sublocation=sublocation
```
jbaudoux, line 486:
```suggestion
            move_lines, product=product, sublocation=location
```

shopfloor/services/zone_picking.py
```diff
@@ -315,6 +323,8 @@ def _data_for_move_lines(
         }
         if product:
             data["product"] = self.data.product(product)
+        if sublocation:
```
```suggestion
        if sublocation and sublocation != zone_location:
```

jbaudoux (changes requested):
Thanks for the fix.
Reviewing the code, I had a small remark when scanning a location

shopfloor/services/zone_picking.py
TDu: This is because the `scan_source` implementation iterates over a list of handlers https://github.com/OCA/wms/blob/14.0/shopfloor/services/zone_picking.py#L657

shopfloor/services/zone_picking.py
TDu: yes same for the other handlers

shopfloor/services/zone_picking.py
TDu: ok

shopfloor/services/zone_picking.py
TDu: No because this is to pass the sublocation to the frontend

TDu: This is to pass the sublocation to the front end

shopfloor/services/zone_picking.py
TDu: This is to pass the sublocation to the frontend

jbaudoux (approved):
Tested and works properly
Thanks

TDu:
Squashed the fixups

shopfloor/services/zone_picking.py
```diff
@@ -637,12 +663,19 @@ def scan_source(self, barcode, confirmation=False, product_id=None):
...
+        to it.
+
         Transitions:
         * select_line: barcode not found or narrow the list on a location
         * set_line_destination: a line has been selected for picking
         """
-
         # select corresponding move line from barcode (location, package, product, lot)
+        sublocation = (
+            self.env["stock.location"].browse(sublocation_id)
```
sebalix: To protect against fake-ID
```suggestion
            self.env["stock.location"].browse(sublocation_id).exists()
```

sebalix (commented):
Having the docstring completed with this new behavior would be nice (section 4.?) :)

shopfloor/services/zone_picking.py
TDu: updated

sebalix (approved):

PR #513, opened 2022-12-05
[14.0][ADD] shopfloor_single_product_transfer

mmequignon:
Depends on:

- [x] https://github.com/OCA/wms/pull/483


sebalix (commented):
Can you generate a PlantUML document like other scenarios? This will help to understand the flow between endpoints.

shopfloor_single_product_transfer/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "Shopfloor Single Product Transfer",
+    "summary": "Brief Summary",
+    "version": "14.0.1.0.0",
+    "category": "Inventory",
```
JuMiSanAr, line 5: `    "development_status": "Beta",
`

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+        if product.tracking == "lot":
+            message = self.msg_store.scan_lot_on_product_tracked_by_lot()
+            return self._response_for_select_product(location, message=message)
+
+    def _scan_product__select_move_line_domain(self, product, location, lot=None):
+        domain = [
+            ("location_id", "=", location.id),
+            ("product_id", "=", product.id),
+            ("state", "in", ("assigned", "partially_available")),
+            ("shopfloor_user_id", "=", False),
```
jbaudoux:
```suggestion
            ("picking_id.user_id", "in", (False, self.env.uid)),
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+            if response:
+                return response
+
+    # Copied from manual_product_transfer
+    def _find_location_move_lines_domain(self, location, product, lot=None):
+        domain = [
+            ("location_id", "=", location.id),
+            ("product_id", "=", product.id),
+            ("state", "in", ("assigned", "partially_available")),
+            ("shopfloor_user_id", "=", False),
```
jbaudoux:
```suggestion
            ("picking_id.user_id", "in", (False, self.env.uid)),
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+        # TODO qty_done = 0: transfer_no_qty_done
+        # TODO qty done < product_qty: transfer_confirm_done
+        picking = move_line.picking_id
+        picking._action_done()
+        message = self.msg_store.transfer_done_success(picking)
+        return self._response_for_select_location(message=message)
+
+    def _start__find_ongoing_moves_for_user_domain(self, user):
+        return [
+            ("shopfloor_user_id", "=", user.id),
```
jbaudoux:
```suggestion
            ("picking_id.user_id", "in", (False, self.env.uid)),
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+            ("picking_id.picking_type_id", "in", self.picking_types.ids),
+            ("state", "in", ("assigned", "partially_available")),
+        ]
+
+    def _start__find_ongoing_moves_for_user(self):
+        user = self.env.user
+        domain = self._start__find_ongoing_moves_for_user_domain(user)
```
jbaudoux: Here I would search for ongoing picking and take the first move line

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+            handlers, move_line, barcode, confirmation=confirmation
+        )
+        if response:
+            return response
+        message = self.msg_store.barcode_not_found()
+        return self._response_for_set_quantity(move_line, message=message)
+
+    def set_quantity__action_cancel(self, selected_line_id):
+        move_line = self.env["stock.move.line"].browse(selected_line_id)
+        move_line.write({"qty_done": 0, "shopfloor_user_id": False})
```
jbaudoux: use `unmark_move_line_as_picked`

jbaudoux (changes requested):
shopfloor_user_id has been removed from all scenario except from the reception that support parallel work with multiple users.
In all other scenario, the user is set on the picking (which is visible in the backend) and shopfloor_user_id should never be used

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+            ...
+        elif location in valid_locations_for_menu:
+            # Considered valid if scan confirmed
+            if not confirmation:
+                # Ask for confirmation
+                orig_location = move_line.location_dest_id
+                message = self.msg_store.confirm_location_changed(
+                    orig_location, location
+                )
+            ...
```
jbaudoux:
```suggestion
            pass
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+
+    def _set_quantity__check_location(self, location, move_line, confirmation=False):
+        valid_locations_for_move_line = (
+            self._set_quantity__valid_dest_location_for_move_line(move_line)
+        )
+        valid_locations_for_menu = self._valid_dest_location_for_menu()
+        message = False
+        if location in valid_locations_for_move_line:
+            # scanned location is valid, return no response
+            ...
```
jbaudoux:
```suggestion
            pass
```

jbaudoux (commented):
Use pass for no-op

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+            message = self.msg_store.dest_location_not_allowed()
+        if message:
+            return self._response_for_set_quantity(move_line, message=message)
+
+    def _set_quantity__post_move(self, location, move_line, confirmation=False):
+        self._actions_for("stock")
+        # TODO qty_done = 0: transfer_no_qty_done
+        # TODO qty done < product_qty: transfer_confirm_done
+        picking = move_line.picking_id
+        picking._action_done()
```
jbaudoux: High: Before posting, the scanned location must be set on the move line.
Use `_write_destination_on_lines` as in location content transfer (this method should ideally be moved out of the scenario). Add support for this new scenario in `shopfloor_checkout_sync`

jbaudoux (changes requested):

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: Use `stock.validate_moves(move_line.move_id)`

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+            stock = self._actions_for("stock")
+            if self.work.menu.no_prefill_qty:
+                # First, mark move line as picked with qty_done = 0,
+                # so the move wont be split because 0 < qty_done < product_uom_qty
+                stock.mark_move_line_as_picked(move_line, quantity=0)
+                # Then, set the no prefill qty on the move line
+                if lot:
+                    qty_done = 1
+                else:
+                    qty_done = 1
```
jbaudoux: missing product packaging case where qty_done = packaging.qty

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+    @with_savepoint
+    def scan_product(self, location_id, barcode):
+        """Looks for a move line in the given location, from a barcode."""
+        location = self.env["stock.location"].browse(location_id)
+        if not location.exists():
+            return self._response_for_select_product(location)
+        handlers = [
+            self._scan_product__scan_product,
+            self._scan_product__scan_lot,
+        ]
```
jbaudoux: missing `_scan_product__scan_packaging` for scanning a product packaging

jbaudoux (changes requested):
Product packaging must also be supported

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+        move.with_context(
+            {"force_reservation": self.work.menu.allow_force_reservation}
+        )._action_assign()
+        assert move.state == "assigned", "The reservation of quantities has failed"
+        move_line = move.move_line_ids
+        stock = self._actions_for("stock")
+        if self.work.menu.no_prefill_qty:
+            stock.mark_move_line_as_picked(move_line, quantity=0)
+            # Just to be explicit
+            # TODO add if packaging
```
jbaudoux: idd packaging must be supported

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,588 @@
...
+
+    def _set_quantity__check_no_prefill_qty(self, move_line, product, lot=None):
+        if not self.work.menu.no_prefill_qty:
+            # If no_prefill_qty is False, then qty_done should have been prefilled
+            # with product_uom_qty in the select_product screen
+            message = self.msg_store.unable_to_pick_more(move_line.product_uom_qty)
+            return self._response_for_set_quantity(move_line, message=message)
+
+    def _set_quantity__increment_qty_done(self, move_line, product, lot=None):
+        # TODO if packaging
```
jbaudoux: idd product packaging must be supported

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: When starting the app, we should be redirected to the second screen if there's a picking with user_id == self.env.uid.
Therefore, I have the feeling that
`("picking_id.user_id", "=", self.env.uid)` is better than `("picking_id.user_id", "in", (False, self.env.uid))`, am I right?

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: hmmm, not sure to see the difference

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: > When starting the app, we should be redirected to the second screen if there's a picking with user_id == self.env.uid.

Only if there is a qty_done > 0 on the move line.
For non started lines, we should be able to select lines assigned to the user or any non assigned

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: indeed, but we should add qty_done > 0 in the domain

shopfloor_single_product_transfer/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
```
sebalix: `Camptocamp` only

sebalix:
@jbaudoux I don't think we removed all `shopfloor_user_id` references from other scenarios, it's still an ongoing work!

shopfloor_single_product_transfer/services/single_product_transfer.py
sebalix: `_write_destination_on_lines` is implemented in:
- `location_content_transfer`
- `zone_picking`
- `cluster_picking` (but it is called `_unload_write_destination_on_lines`)

And all of them has a different implementation... so for now I'll do the same: create our own method in the current scenario, add the support in `shopfloor_checkout_sync`. Refactoring this has to be planned, I prefer to not pass time on this right now.

sebalix: @jbaudoux would we need the `_lock_lines` method as well in this scenario? It is used in location content transfer, zone picking, cluster picking...

jbaudoux: using _lock_lines seems safer indeed

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,649 @@
...
+            "origin": self.work.menu.name,
+            "picking_type_id": picking_type.id,
+        }
+        move = self.env["stock.move"].create(move_vals)
+        move._action_confirm(merge=False)
+        move.with_context(
+            {"force_reservation": self.work.menu.allow_force_reservation}
+        )._action_assign()
+        assert move.state == "assigned", "The reservation of quantities has failed"
+        move_line = move.move_line_ids
```
sebalix: This won't work if the move reserved for instance 3 packages + a bunch of products w/o lot/package. In this case you'll got 4 move lines (1 line per package + 1 line for bulk products).

@jbaudoux should we check earlier the quantity we computed in the location for the given product/lot(/package)? I mean:

- if we scan a product and we have 1 PKG + some bulk products in the location, we move only the bulk products (so we can edit the qty to move on our created move line, and make use of the `no_prefill_qty` option)
- if we scan a package (to implement later), we move only this package

What we want to do is to get only one move line for the created move, so we can edit the qty to move. In the case of moving a package, we should even prevent to edit the qty to move so we move it entirely.

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,649 @@
...
+            response = unreserve.check_unreserve(location, move_lines, product, lot)
+            if response:
+                return response
+            unreserve.unreserve_moves(move_lines, self.picking_types)
+
+    def _scan_product__create_move_line_domain(self, product, location, lot=None):
+        domain = [
+            ("location_id", "=", location.id),
+            ("product_id", "=", product.id),
+            ("available_quantity", ">", 0),
```
sebalix: We'll add `("package_id", "=", False)` to not process such quants while we do not support packages. See my comment bellow.

shopfloor_single_product_transfer/services/single_product_transfer.py
sebalix: What I did/changed regarding packaging scanning:

- without `no_prefill_qty` option, when we scan an item (product/lot/packaging) on the `scan_product` state, the `qty_done` is always set to the `product_uom_qty`. If we have the `allow_move_create` option enabled and that we scan a packaging, the move created will have a qty == packaging.qty even if we have more available in the location.
- with `no_prefill_qty` option, when we scan a product/lot/packaging, the initual `qty_done` will be 0 (it was 1 before), so we wait the `set_quantity` state to increase the qty on the move line, but that means we have to scan the product/lot/packaging twice... which is maybe not what we want :thinking: 

I can revert the change on the initial value of `qty_done` with `no_prefill_qty` enabled, you tell me.

shopfloor_single_product_transfer/services/single_product_transfer.py
sebalix: done

shopfloor_single_product_transfer/services/single_product_transfer.py
sebalix: done

shopfloor_single_product_transfer/services/single_product_transfer.py
sebalix: I updated the PR to restore the initial behavior. Should be fixed.

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+from odoo.tools import float_compare
+
+from odoo.addons.base_rest.components.service import to_int
+from odoo.addons.component.core import Component
+from odoo.addons.component.exception import NoComponentError
+
+_logger = logging.getLogger("shopfloor.services.single_product_transfer")
+
+
+def with_savepoint(meth):
```
TDu:
```suggestion
def with_savepoint(method):
```
The abreviation is not needed imo, except if `method` is a reserved python word ?

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+    @wraps(meth)
+    def wrapper(self, *args, **kwargs):
+        savepoint = self._actions_for("savepoint").new()
+        response = meth(self, *args, **kwargs)
+        message_type = response.get("message", {}).get("message_type")
+        if message_type in ("error", "warning"):
+            _logger.info(
+                "%(method_name)s returned an error/warning. Transaction rollbacked.",
```
TDu, line 26:
```suggestion
                "%(method_name) has returned an error/warning. Transaction rollbacked.",
```

mmequignon: for string formatting, the pattern is `%()s`, hence the `%(method_name)s` here.

TDu: oh

jbaudoux:
While testing, when choosing an alternative destination location, you have to confirm it by scanning again that location, but it doesn't work, you stay on the screen asking you to confirm again and again that new destination location.
We need to block this feature by adding an option "Allow alternative destination". If the option is active, then we have current behavior with the confirmation. If the option is not active, then an error is display "Wrong destination location"

**Non blocking => Moved the issue to a new follow-up card so that we can merge this first version**

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+    _description = __doc__
+
+    # Responses
+
+    def _response_for_select_location(self, message=None):
+        return self._response(next_state="select_location", message=message)
+
+    def _response_for_select_product(self, location, message=None):
+        data = {"location": self.data.location(location)}
+        # import pdb; pdb.set_trace()
```
TDu:
```suggestion
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+        ]
+        lot = search.lot_from_scan(barcode)
+        if lot:
+            product = lot.product_id
+            product_response = self._use_handlers(handlers, product, location, lot=lot)
+            if product_response:
+                return product_response
+
+    def _use_handlers(self, handlers, *args, **kwargs):
+        # TODO: each handler should raise a Shopfloor dedicated exception
```
TDu, line 294: Why would that be better ? Not less complex anyway imo

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+                return response
+            unreserve.unreserve_moves(move_lines, self.picking_types)
+
+    def _scan_product__create_move_line_domain(
+        self, product, location, lot=None, packaging=None
+    ):
+        domain = [
+            ("location_id", "=", location.id),
+            ("product_id", "=", product.id),
+            ("available_quantity", ">", 0),
```
TDu: `available_quantity` Is it not a non stored computed field, and cannot be searched ?

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+        if lot:
+            lot_domain = [("lot_id", "=", lot.id)]
+            domain = AND([domain, lot_domain])
+        return domain
+
+    def _scan_product__select_move_line(
+        self, product, location, lot=None, packaging=None
+    ):
+        domain = self._scan_product__select_move_line_domain(product, location, lot=lot)
+        move_line = self.env["stock.move.line"].search(domain, limit=1)
```
jbaudoux: order by picking_id.user_id, picking_id.priority desc, picking_id.scheduled_date asc, id desc

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+        if not self.is_allow_move_create():
+            message = self.msg_store.no_operation_found()
+            return self._response_for_select_product(location, message=message)
+
+    def _scan_product__unreserve_move_line(
+        self, product, location, lot=None, packaging=None
+    ):
+        unreserve = self._actions_for("stock.unreserve")
+        if self.work.menu.allow_unreserve_other_moves:
+            move_lines = self._find_location_move_lines(location, product, lot=lot)
```
jbaudoux, line 217: The unreserve should occur at this stage only if there is no free quantity on the location. Otherwise, there is no reason to unreserve the other moves

```
free_qty = product.with_context(location=location.id, lot_id=lot.id if lot else None).free_qty
if float_compare(free_qty, packaging and packaging.qty or 1, precision_rounding=product.uom_id.rounding):
  .... 
```
After the unreserve, the new free_qty should be checked again as you can only unreserve moves from other picking types.


And before all this, I would also check the available quantity:
```
qty_available = product.with_context(location=location.id, lot_id=lot.id if lot else None).qty_available
if float_compare(qty_available, packaging and packaging.qty or 1, precision_rounding=product.uom_id.rounding):
   ...
```
otherwise an error should be raised. I would not raise an "No operation found" as we already allowed to create move but more a "Product not available".

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+            product, location, lot=lot, packaging=packaging
+        )
+        quants_in_location = self.env["stock.quant"].search(quant_domain)
+        available_quantity = sum(quants_in_location.mapped("available_quantity"))
+        if available_quantity:
+            if packaging:
+                # Limit the qty to move to the packaging qty
+                # FIXME use float_compare
+                if available_quantity >= packaging.qty:
+                    available_quantity = packaging.qty
```
jbaudoux: This is not necessary anymore if done before as commented on `_scan_product__unreserve_move_line`

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+        """
+        location = self.env["stock.location"].browse(location_id)
+        if not location.exists():
+            return self._response_for_select_product(location)
+        handlers = [
+            self._scan_product__scan_product,
+            self._scan_product__scan_packaging,
+            self._scan_product__scan_lot,
+        ]
+        response = self._use_handlers(handlers, location, barcode)
```
jbaudoux, line 629: I'm not fan of having handlers inside those handlers (and by the way, they can be all the same if you just add a `if not lot` in `_scan_product__check_tracking`).

I would have preferred to call:

* search.decode_barcode(barcode) returning a dict(product=..., lot=..., location=...) depending on what has been found
* get scan_product_handlers returning the list of handlers
* response = use_handlers calling the handlers

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: We don't want to move a package with this single_product_transfer, you should then use single_pack_transfer.

Regarding the reservation, it is true that, in theory, it can result in multiple move_lines. Would be a strange use case.

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,723 @@
...
+                qty_done = packaging.qty
+            move_line.qty_done = qty_done
+        else:
+            if packaging:
+                # We ensure the qty_done is 0 here, so we can set it manually after
+                # to avoid the split of the move line by 'mark_move_line_as_picked'.
+                stock.mark_move_line_as_picked(move_line, quantity=0)
+                move_line.qty_done = packaging.qty
+            else:
+                stock.mark_move_line_as_picked(move_line)
```
jbaudoux: As you just created the move for the quantity to process, I think you can always call 
```suggestion
        stock.mark_move_line_as_picked(move_line)
```

jbaudoux (changes requested):

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: After rebase, you can now use the new `find` action in search

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: You cannot override an ifbox. This you can.
You gain flexibility, for a few gotos, and by doing so, you also stick a lot more to the single responsibility principle.
It's worth it, IMO.

simahawk:
@JuMiSanAr ping
Also, can you rebase and squash?

shopfloor_single_product_transfer/services/single_product_transfer.py
TDu: IMO, this is a problem that needs to be fixed.
I see errors in the logs

mmequignon: will push that soon

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,744 @@
...
+        # '_write_destination_on_lines' is implemented in:
+        #
+        #   - 'location_content_transfer'
+        #   - 'zone_picking'
+        #   - 'cluster_picking' (but it is called '_unload_write_destination_on_lines')
+        #
+        # And all of them has a different implementation,
+        # To refactor later.
+        try:
+            # TODO loose dependency on 'shopfloor_checkout_sync' to avoid having
```
TDu, line 513: Why ? there must be a good reason to have a glue module between `shopfloor` and `checkout_sync`.
It does save on code repetition with the locking of related lines https://github.com/OCA/wms/blob/14.0/shopfloor_checkout_sync/actions/checkout_sync.py which is specific.
And separation of concern.

In the implementation of this PR, the locking of lines is missing and probably causes `Serialization Failure`

IMO we should keep here the same form of implementation used in other scenario. And propose a refactor in a distinct PR, if needs be.

TDu: Ah because this is a specific module and we would need another glue module for checkout sync !
I will fix the missing lock with the current implementation.

shopfloor_single_product_transfer/tests/test_set_quantity.py
```diff
@@ -0,0 +1,442 @@
...
+        expected_message = {"message_type": "error", "body": "Barcode not found"}
+        data = {
+            "move_line": self._data_for_move_line(move_line),
+            "asking_confirmation": False,
+        }
+        self.assert_response(
+            response, next_state="set_quantity", message=expected_message, data=data
+        )
+
+    def test_set_quantity_scan_product_prefill_qty_disabled(self):
```
TDu, line 61: has the option is named starting with 'no'. The name of the test function should also to stay coherent
```suggestion
    def test_set_quantity_scan_product_no_prefill_qty_disabled(self):
```

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -60,10 +60,12 @@ class ShopfloorSingleProductTransfer(Component):
     def _response_for_select_location(self, message=None):
         return self._response(next_state="select_location", message=message)
 
-    def _response_for_select_product(self, location, message=None):
+    def _response_for_select_product(self, location, message=None, popup=None):
         data = {"location": self.data.location(location)}
         # import pdb; pdb.set_trace()
-        return self._response(next_state="select_product", data=data, message=message)
+        return self._response(
+            next_state="select_product", data=data, message=message, popup=popup
```
mmequignon, line 67: à base de popopopup

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: We cannot use dotted notation in the order.
We cannot use a combination of asc/desc in the `sorted()` method.
Well… We cannot use the `minus` sign on dates for instance.
What do you suggest?

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: pfff, I don't even remember writing this…

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: Let's do this in a dedicated pr.

shopfloor_single_product_transfer/services/single_product_transfer.py
TDu: I just pushed a commit that adds the locking of the lines. And a little test that comes with it

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: I'm not sure to understand. What is to be done here?

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: does it mean we have to ignore no_prefill_qty and always pick what is available?

shopfloor_single_product_transfer/services/single_product_transfer.py
sebalix: Right now, at least add a comment above this line saying we expect to get only one move line as we are moving only bulk products w/o lot or package.

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,750 @@
...
+            unreserve.unreserve_moves(move_lines, self.picking_types)
+
+    def _scan_product__create_move_line_domain(
+        self, product, location, lot=None, packaging=None
+    ):
+        domain = [
+            ("location_id", "=", location.id),
+            ("product_id", "=", product.id),
+            ("available_quantity", ">", 0),
+            # FIXME: handle the scan of packages later
```
sebalix: We can remove this FIXME, and say that packages should be moved with the `single_pack_transfer` scenario.

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: look at _search, set query.order yourself

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -0,0 +1,760 @@
...
+            float_compare(
+                available_quantity,
+                packaging.qty if packaging else 1.0,
+                precision_rounding=product.uom_id.rounding,
+            )
+            >= 0
+        )
+        if is_product_available:
+            if packaging:
+                available_quantity = packaging.qty
```
jbaudoux: why?
you can create the move for the product available quantity. In create move, you will set the quantity done based on the packaging quantity

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: at second read, what you did looks right. Just in case of product packaging, the quantity todo looks wrong.

This comment can be resolved

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: Because the fact that `available_qty >= packaging.qty` when packaging is defined is now handled above.
Or this was wrong in the first place https://github.com/OCA/wms/compare/22a3e5b09df479f30fcd84e51940088f98e92183..9a0c0897de362d793687d87b9524ad938ccafc20#diff-f871450f795713d4ab1061b7117262fdee975730cb8f85f5963d06d339d7e39dL232-R236
cc @sebalix

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: This would be safer then
`move_line = move.move_line_ids[0]`

shopfloor_single_product_transfer/services/single_product_transfer.py
mmequignon: @sebalix 
We discussed with @jbaudoux 
We want to create moves with qty = qty_available, no matter what has been scanned.
This is the qty done that changes when you scan a product/lot/packaging, if `no_prefill_qty` is enabled.
I'll remove this block of code.

jbaudoux (changes requested):
If you change the quantity in the picker and then scan the destination, the new entered quantity is not taken into account in the processed move.

**Being fixed - last blocking issue**

jbaudoux (changes requested):
Once you scanned a product, the screen header must show the selected operation (like in the other scenario) and not the product name.
After the move is processed, the header must be cleaned up and not show anymore the done operation. I think in the other scenario, it shows the shopfloor menu name

**Non blocking => Moved the issue to a new follow-up card so that we can merge this first version**

shopfloor_single_product_transfer/services/single_product_transfer.py
jbaudoux: @mmequignon Do not forget this one if not yet resolved

simahawk:
status?

simahawk:
@mmequignon @JuMiSanAr we agreed w/ @jbaudoux that this PR and its companion for frontend must be merged w/ the current state unless there's something broken. Fix/imp can be done in the other PRs.
This will help w/ the depending PRs.
Can you fast track them?

mmequignon:
> If you change the quantity in the picker and then scan the destination, the new entered quantity is not taken into account in the processed move.
> 
> **Being fixed - last blocking issue**

@jbaudoux I fixed the missing picker qty sync.
It wil lhave to be added there as well https://github.com/OCA/wms/pull/520
ping @JuMiSanAr (set_quantity now required "quantity" as the current picker quantity)

jbaudoux (approved):

JuMiSanAr:
> > If you change the quantity in the picker and then scan the destination, the new entered quantity is not taken into account in the processed move.
> > **Being fixed - last blocking issue**
> 
> @jbaudoux I fixed the missing picker qty sync. It wil lhave to be added there as well #520 ping @JuMiSanAr (set_quantity now required "quantity" as the current picker quantity)

@mmequignon I've added the quantity in the front, I tested it and it works fine!

simahawk:
@sebalix good for you?

sebalix (approved):

simahawk:
NOTE: @JuMiSanAr  this commit here https://github.com/OCA/wms/pull/520/commits/4892cac10bb9161c88c83c9068b9f016d59861ee for the init / upgrade hook should have been done in this PR. 
:wink:

JuMiSanAr:
> NOTE: @JuMiSanAr this commit here [4892cac](https://github.com/OCA/wms/commit/4892cac10bb9161c88c83c9068b9f016d59861ee) for the init / upgrade hook should have been done in this PR. wink

@simahawk this was already discussed with @jbaudoux and @mmequignon, we agreed not to push any more commits to this PR to merge it asap.

PR #517, opened 2022-12-08
[14.0][IMP] shopfloor_workstation_label_printer: extract user config in another module

sebalix:
User configuration field has been moved to `base_report_to_printer_label` module to be used outside of Shopfloor:
- https://github.com/OCA/report-print-send/pull/309

Ref 3798

PR #518, opened 2022-12-09
[14.0][IMP] shopfloor_workstation_label_printer: leverage `base_report_to_label_printer`

sebalix:
Supersedes #517 

Module [`base_report_to_label_printer`](https://github.com/OCA/report-print-send/pull/313) allows to flag reports as labels and provides a method to get the expected label printer for a given user. So:

- Deprecate the `res.users.get_delivery_label_printer` method. A better alternative is to use `ir.actions.report._get_user_default_printer`.
- Remove the `res.users.default_label_printer_id` field definition from the current module as it already exists in `base_report_to_label_printer`.

Depends on:
- https://github.com/OCA/report-print-send/pull/313

Ref 3798

### DIFF
--- a/shopfloor_workstation_label_printer/__manifest__.py
+++ b/shopfloor_workstation_label_printer/__manifest__.py
@@ -10,7 +10,7 @@
     "website": "https://github.com/OCA/wms",
     "development_status": "Alpha",
     "license": "AGPL-3",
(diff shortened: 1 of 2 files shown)


jbaudoux (approved):

simahawk (approved):

simahawk:
@sebalix pls check the build

PR #521, opened 2022-12-15
[14.0] Add shopfloor_gs1

simahawk:
Depends on:

- [x] https://github.com/OCA/wms/pull/547

Add support for GS1 barcodes.

TODO 
- [ ] sf_gs1: improve test cov, more unit tests, remove scenario-dependent tests
- [ ] write docs and docstrings
- [ ] support non HRI barcodes
- [ ] update ROADMAP
- [x] pin biip / check changes

NOTE: based on 14a220944f630e073dd8dabff0e0c2b1fcbd825f


poiswe:
FYI @simahawk biip v3 just released, some breaking changes regarding parsing
https://github.com/jodal/biip/releases/tag/v3.0.0

simahawk:
> FYI @simahawk biip v3 just released, some breaking changes regarding parsing https://github.com/jodal/biip/releases/tag/v3.0.0

Thanks for the heads up! We'll pin the older version for now and then we'll see.

shopfloor/services/checkout.py
```diff
@@ -398,18 +398,19 @@ def scan_line(self, picking_id, barcode):
...
-            return self._select_lines_from_lot(picking, selection_lines, lot)
+        search_result = search.find(
+            barcode,
+            types=("package", "product", "packaging", "lot", "serial"),
+            handler_kw=dict(
+                lot=dict(products=picking.move_lines.product_id),
+                serial=dict(products=picking.move_lines.product_id),
+            ),
+        )
+        result_handler = getattr(self, "_select_lines_from_" + search_result.type)
```
mmequignon: by naming the handler methods `_{endpoint}_{search_result.type}`, you could create a `use_handler` decorator, and make this generic.
```python3
def use_handler(method):
    @wraps(method)
    def wrapper(self, *args, **kwargs):
        search_result = search.find(...)
        if search_result:
            return self.getattr(f"_{method.__name__}_{search_result.type}")()
        # fallback
        returrn method(self, *args, **kwargs)
    return wrapper
```

sebalix:
```
ERROR: Could not find a version that satisfies the requirement biip==2.3.0 (from versions: 0.6.0, 0.6.1, 0.6.2, 1.0.0, 1.1.0, 1.2.0, 2.0.0, 2.1.0)
ERROR: No matching distribution found for biip==2.3.0
```
Can't figure out why it is complaining, the archive is available there: https://pypi.org/project/biip/2.3.0/

EDIT: Oh, the Odoo container is probably using Python 3.6, and starting from `biip>=2.2.0` it requires Python >= 3.7

sebalix:
We can't merge this PR currently as we need `biip>=2.3.0` and `<3.0.0`, which requires Python >= 3.7 but 3.6 is used by Runboat.
I added a comment in the external dependencies section of the manifest file.

simahawk:
> We can't merge this PR currently as we need `biip>=2.3.0` and `<3.0.0`, which requires Python >= 3.7 but 3.6 is used by Runboat. I added a comment in the external dependencies section of the manifest file.

You mean OCA CI is using 3.6?
I wonder if we can upgrade to 3.7... :thinking:

simahawk:
@sebalix IMO we can switch this repo to 3.8 https://github.com/OCA/oca-ci/blob/master/.github/workflows/ci.yaml#L53
@sbidoul what's the policy - if any - for switching specific repos to different py versions on CI? (l10n-brazil is using it BTW)

sbidoul:
> what's the policy - if any - for switching specific repos to different py versions on CI?

The defacto policy is that all OCA addons require the same minimum python version.

If we need to deviate from that there are several thing to do, ideally:

- update the repo template to add a question to select the minimum python version used in GitHub workflows
- make sure the `setup.py` files have a `python_requires` keyword (to override the default that comes from [setuptools-odoo](https://github.com/acsone/setuptools-odoo/blob/1b5206cfdd0fff7773ca34a0be51b0bddf232b50/setuptools_odoo/core.py#L106)) [^1]
- update the runboat config to make sure the correct oca/oca-ci image is used for this repo

[^1]: this could also be done with a new option in the `setup.py` generator to be applied in the pre-commit config

sbidoul:
Of course, the ci and setup.py can be tweaked manually but it is a little bit more error prone and risk conflicts when updating the repo template.

simahawk:
I see. Thanks for the detailed info!
I don't think we want to go all-in on this... v14 is way in the past now and soon we'll migrate the proj requiring this module to v16. In our images we use 3.7 so we know everything works fine.
We can simply close this PR and keep using it via pending merge for a bit longer :man_shrugging:

PR #523, opened 2022-12-27
[14.0][IMP] shopfloor_batch_automatic_creation: make _create_batch return pickings with the same picking type

JuMiSanAr:
On the shopfloor menu, you can configure multiple picking types.

In case of automatic batch creation, a first picking is elected and then others are added to the batch. When this occurs, the other pickings must belong to the same picking type as the first added picking.

Ref: rau-88

### DIFF
--- a/shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
+++ b/shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
@@ -104,11 +104,14 @@ class PickingBatchAutoCreateAction(Component):
         # We can't use a limit in the SQL search because the 'priority' fields
         # is sometimes empty (it seems the inverse StockPicking.priority field
         # mess up with default on stock.move), we have to sort in Python.
-        return self.env["stock.picking"].search(
-            self._search_pickings_domain(
-                picking_types, user=user, commercial_partner=commercial_partner
-            )
+        domain = self._search_pickings_domain(
+            picking_types, user=user, commercial_partner=commercial_partner
         )
+        first_picking = self.env["stock.picking"].search(domain, limit=1)
+        if first_picking:
+            first_picking_type = first_picking.picking_type_id
+            domain.append(("picking_type_id", "=", first_picking_type.id))
+        return self.env["stock.picking"].search(domain)
 
     def _sort(self, pickings):
         return pickings.sorted(
--- a/shopfloor_batch_automatic_creation/tests/test_batch_create.py
+++ b/shopfloor_batch_automatic_creation/tests/test_batch_create.py
@@ -238,3 +238,33 @@ class TestBatchCreate(CommonCase):
             self.picking_type, group_by_commercial_partner=True
         )
         self.assertEqual(batch.picking_ids, self.picking2 | self.picking3)
+
+    def test_create_batch_only_one_picking_type(self):
+        # Test that the created batches only include pickings of the same picking type.
+        additional_picking_type = (
+            self.env["stock.picking.type"]
+            .sudo()
+            .create(
+                {
+                    "code": "internal",
+                    "name": "Cluster Picking 2",
+                    "sequence_id": 10,
+                    "sequence_code": "CPI",
+                    "warehouse_id": self.env.ref("stock.warehouse0").id,
+                    "default_location_src_id": self.env.ref(
+                        "stock.stock_location_stock"
+                    ).id,
+                    "default_location_dest_id": self.env.ref(
+                        "stock.location_pack_zone"
+                    ).id,
+                }
+            )
+        )
+        picking_types = self.picking_type | additional_picking_type
+        additional_picking = self._create_picking(
+            picking_type=additional_picking_type,
+            lines=[(self.product_a, 10), (self.product_b, 10)],
+        )
+        additional_picking.action_assign()
+        batch = self.auto_batch.create_batch(picking_types)
+        self.assertEqual(batch.picking_ids.picking_type_id, self.picking_type)


shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
```diff
@@ -104,11 +104,17 @@ def _search_pickings(self, picking_types, user=None, commercial_partner=None):
...
             self._search_pickings_domain(
                 picking_types, user=user, commercial_partner=commercial_partner
             )
         )
+        # The selected pickings need to match the type of the first picking found.
+        first_picking_type = fields.first(pickings).picking_type_id
+        filtered_pickings = pickings.filtered(
+            lambda p: p.picking_type_id == first_picking_type
+        )
+        return filtered_pickings
```
i-vyshnevska: picking_types is given as an argument to the method so you can add it to the domain in the search method 

another question what will be with other pickings? for me more reasonable is kept this method us is and modify the place where _search_pickings called, and create a batch per picking type

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
```diff
@@ -104,11 +104,17 @@ def _search_pickings(self, picking_types, user=None, commercial_partner=None):
...
+        pickings = self.env["stock.picking"].search(
             self._search_pickings_domain(
                 picking_types, user=user, commercial_partner=commercial_partner
             )
         )
+        # The selected pickings need to match the type of the first picking found.
+        first_picking_type = fields.first(pickings).picking_type_id
+        filtered_pickings = pickings.filtered(
+            lambda p: p.picking_type_id == first_picking_type
+        )
```
jbaudoux: The dataset can be large, you can't call filtered here. We need a lazy evaluation. Check filter.
In case of group_by_commercial_partner, the second search must respect your new condition also.

jbaudoux (changes requested):

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
jbaudoux: You can also check if 1 or 2 search needs to be done. Like multiple picking_types or group_by_commercial_partner would require 2 searches. Then you can perform a first one with limit 1 and a second one with the additional domain

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
JuMiSanAr: The search method already had picking_types as an argument before didn't it?
If you refer to the evaluation of first_picking_type, we want the picking type of the first filtered picking (and not the first picking type available).

Regarding the other pickings, the purpose of this PR is to simply make sure that the batch generated here contains pickings of the same type. The original implementation didn't handle the other pickings either.

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
JuMiSanAr: Can we return a recordset from filter()? From what I see, we can return lists but I'm not sure if there's an effective way to transform it to a recordset.

Regarding the number of checks, if we want to optimize this filter so that it doesn't run more than once in all cases, wouldn't it be better yo simply apply it after the if / else? That way we'd always have batches with one picking type and filter() would only run once.

jbaudoux: I would make a method that searches and returns the first picking
Then a method that builds the domain for the following pickings to find
And finally a method that searches and return the next pickings

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
i-vyshnevska: my idea was to modify returning domain, but the issue is at that moment we don't know which pickings type is found so, ok, the code is LG
but I don't get functional POV,
what is happening with picking which was filtered, they stayed not proceeded? the method called again and another picking type selected? what if we always will have 2 types in the set and another type never will proceed?

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
JuMiSanAr: But we'd need a domain for that first search already wouldn't we? What if the first picking doesn't match the user / the allowed picking types?

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
JuMiSanAr: @jbaudoux thoughts?

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
jbaudoux: Yes, you need to apply a first domain for the first search.
@lmignon did the same remark in the migration to v16 https://github.com/OCA/wms/pull/534#discussion_r1072404033

sebalix: Why we cannot do this:
```suggestion
        domain = self._search_pickings_domain(
            picking_types, user=user, commercial_partner=commercial_partner
        )
        first_picking = self.env["stock.picking"].search(domain, limit=1)
        if first_picking:
            first_picking_type = first_picking.picking_type_id
            domain.append(("picking_type_id", "=", first_picking_type.id))
        return self.env["stock.picking"].search(domain)
```
It's two search, but it should be quite fast and low on memory.

sebalix (approved):

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
```diff
@@ -104,11 +104,14 @@ def _search_pickings(self, picking_types, user=None, commercial_partner=None):
...
         # is sometimes empty (it seems the inverse StockPicking.priority field
         # mess up with default on stock.move), we have to sort in Python.
-        return self.env["stock.picking"].search(
-            self._search_pickings_domain(
-                picking_types, user=user, commercial_partner=commercial_partner
-            )
+        domain = self._search_pickings_domain(
+            picking_types, user=user, commercial_partner=commercial_partner
         )
+        first_picking = self.env["stock.picking"].search(domain, limit=1)
```
jbaudoux, line 110: sorting should be ensured on this first call:
`priority desc, scheduled_date asc, id asc`

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
```diff
@@ -104,11 +104,14 @@ def _search_pickings(self, picking_types, user=None, commercial_partner=None):
         # We can't use a limit in the SQL search because the 'priority' fields
         # is sometimes empty (it seems the inverse StockPicking.priority field
         # mess up with default on stock.move), we have to sort in Python.
-        return self.env["stock.picking"].search(
-            self._search_pickings_domain(
-                picking_types, user=user, commercial_partner=commercial_partner
-            )
+        domain = self._search_pickings_domain(
+            picking_types, user=user, commercial_partner=commercial_partner
```
jbaudoux, line 108: `create_batch` needs to be reworked to reduce the amount of calls to `_search_pickings` in the case of `group_by_commercial_partner`.

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
```diff
@@ -104,11 +104,14 @@ def _search_pickings(self, picking_types, user=None, commercial_partner=None):
...
         )
+        first_picking = self.env["stock.picking"].search(domain, limit=1)
+        if first_picking:
+            first_picking_type = first_picking.picking_type_id
+            domain.append(("picking_type_id", "=", first_picking_type.id))
```
jbaudoux, line 113: should be ideally in a method and use expression.AND instead of append

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
```diff
@@ -104,11 +104,14 @@ def _search_pickings(self, picking_types, user=None, commercial_partner=None):
...
+        return self.env["stock.picking"].search(domain)
 
     def _sort(self, pickings):
```
jbaudoux, line 116: I don't think this is still necessary now that you use search instead of filtered but you need to give the right orderby to search

shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py
```diff
@@ -104,11 +104,14 @@ def _search_pickings(self, picking_types, user=None, commercial_partner=None):
         # We can't use a limit in the SQL search because the 'priority' fields
         # is sometimes empty (it seems the inverse StockPicking.priority field
         # mess up with default on stock.move), we have to sort in Python.
```
jbaudoux, line 106: This should be dropped according to the last changes. Can't find such issue in v14 but have a check on your DBs.

jbaudoux (changes requested):
The code is still badly designed. You should regroup the concepts.
`create_batch` should do a call to search for pickings and then apply limits.
All the code for searching should be moved from `create_batch` to `_search_pickings`.

PR #525, opened 2023-01-04
[14.0] shopfloor: checkout prevent to take more than planned

JuMiSanAr:
For checkout, in the select_package screen, we now allow the user to scan articles and go beyond the maximum qty_done allowed (line will be highlighted in red and a warning message will be displayed).

However, assigning a package in that situation will not be allowed, and they will still receive an error message.

Depends on:
- [x] https://github.com/OCA/wms/pull/483#issuecomment-1370958786

ref: rau-92

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -693,18 +693,15 @@ class Checkout(Component):
                         split_partial=False,
                         result_package_id=False,
                     )
-                    if qty_check == "greater":
-                        qty_done = move_line.product_uom_qty
-                        message = {
-                            "body": _(
-                                "Not allowed to pack more than the quantity, "
-                                "the value has been changed to the maximum."
-                            ),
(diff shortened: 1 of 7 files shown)


shopfloor/services/checkout.py
```diff
@@ -926,6 +922,16 @@ def list_delivery_packaging(self, picking_id, selected_line_ids):
                 selected_lines,
                 message=self.msg_store.no_delivery_packaging_available(),
             )
+        for line in selected_lines:
```
simahawk: this loop is repeat 3 times: can we put it in its own method?

simahawk (changes requested):

jbaudoux (approved):
Tested

jbaudoux:
rebased

simahawk (approved):

PR #527, opened 2023-01-05
[14.0][IMP] stock_available_to_promise_release

mt-software-de:
own method for checking if a move is releasable

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -290,6 +290,37 @@ class StockMove(models.Model):
             vals.update({"procure_method": self.procure_method, "need_release": True})
         return vals
 
+    def _get_release_decimal_precision(self):
+        return self.env["decimal.precision"].precision_get("Product Unit of Measure")
+
+    def _get_release_remaining_qty(self):
+        self.ensure_one()
+        quantity = min(self.product_qty, self.ordered_available_to_promise_qty)
+        remaining = self.product_qty - quantity
+        precision = self._get_release_decimal_precision()
+        if not float_compare(remaining, 0, precision_digits=precision) > 0:
+            return
+        return remaining
+
+    def _is_releasable(self):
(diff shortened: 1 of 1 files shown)


mt-software-de:
fyi @jbaudoux

jbaudoux (approved):

jbaudoux:
@sebalix Can you review & merge this one?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -289,6 +290,36 @@ def _prepare_move_split_vals(self, qty):
...
+        if not self.need_release:
+            return
+        if self.state not in ("confirmed", "waiting", "done", "cancel"):
+            return
+        precision = self._get_release_decimal_precision()
+        available_quantity = self.ordered_available_to_promise_qty
+        if float_compare(available_quantity, 0, precision_digits=precision) <= 0:
+            return
+
+        if self._get_release_remaining_qty():
```
simahawk: The computation of the qty now happens 2 times on each iteration.
It would be better to make `_is_releasable` return the qty and use it right after at line 339.

simahawk (changes requested):
In addition can you please rewrite the commit msg to smth like:

```
[IMP] s_available_to_promise_release: add hook before release

Split _run_stock_rule to offer more hooks for override.
```

stock_available_to_promise_release/models/stock_move.py
mt-software-de: yes you are right, now _is_releaseable is returning the available_qty and the remaining_qty

sebalix:
@mt-software-de you can remove the `[14.0]` from commit message, as it has no value once ported to next versions (or backported).

mt-software-de:
> @mt-software-de you can remove the `[14.0]` from commit message, as it has no value once ported to next versions (or backported).

yes

sebalix (approved):

PR #530, opened 2023-01-11
shopfloor: zp use pack location as sublocation

TDu:
When scanning a package use the location of that package to filter on sub location.
So if the package has multiple products, the user will not be required to scan a location after scanning the product.

ref.: rau-93


simahawk (approved):
LG

jbaudoux (approved):

PR #534, opened 2023-01-13
[16.0][MIG] stock_picking_batch_creation

lmarion-source:
Depends on:
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/1626


lmignon:
cc @jbaudoux

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,305 @@
...
+                self.env["stock.picking"].browse(suitable_pick_ids),
+            )
+        return (
+            self.env["stock.device.type"].browse(),
+            self.env["stock.picking"].browse(),
+        )
+
+    def _create_batch(self, raise_if_not_possible=False):
+        user = self.user_id if self.user_id else self.env.user
+        candidates_pickings_to_batch = self._candidates_pickings_to_batch(user=user)
```
lmignon: @lmarion-source We should change the implementation to avoid to retrieve all the candidate picking in one query and making the filtering into odoo...
The idea is to:

1. Make one query to get the first picking suitable with one of our device
2. While the device is not full, make 1 query at a time to get a picking to add to the batch

With this kind of approach, we should limit the concurrent update since the result of a query should only return a picking that will be used into the batch.

To implement this approach we need:

1. A method used to get the base domain used to get a pciking candidate. This domain will be used to get the first picking and the followings.
2. A method used to get a domain to apply to filter a picking for a given device. Basically, a picking can be put into a specific device if the weight and the volume fit to the given device. The weight on the picking is already a computed stored field. If it's not case, the volume should also be a computed stored field. (The `_init_dimension_fields` methods must be removed and these fields should be declared as computed since the improvments into the orm in V16 not more justify this kind of hack)
3. A method to get the domain to apply to get a picking to add to the cluster. This method will recive as paramter the first picking retrieved, the actual cluster informations (current weight, number of lines, number of slots used, ....), the device used, ???) The challenge here is to be able to express all the criteria to apply to the picking candidates into a odoo domain...

The `_create_batch` method implementation will be therefore slightly different.

1. The first picking will retieved by querying the orm with the common domain for all pickings cancidates combinated using a *AND* operator with a domain for the device criteria. The domain for the device crieria will be a combination using a *OR* operator of the domain computed by the dedicated method for each device that can be used by the cluster.
2. Once we have the first picking and device, and while the cluster is not full, we can query for additional pickings by combining with an *AND* operator the common domain and the result to the call to the new method to get the domain to apply to add picking to the cluster...

lmignon (changes requested):
see comment

lmignon:
@JuMiSanAr @jbaudoux In the las commit I fully refactored the code to improves performances, avoid concurrency issues is the most efficient way and only rely on SQL queries to apply the constraints when looking for picking to add to a batch.

lmignon:
* [ ] depends on https://github.com/OCA/stock-logistics-warehouse/pull/1626

lmignon:
ping @rousseldenis

stock_picking_batch_creation/README.rst
```diff
@@ -0,0 +1,195 @@
...
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/285/16
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Odoo allows you to create batches of pickings by hand or automatically by
+specifying some criteria on the picking type definition.
+
+The approach in this addon is slightly different. Id doesn't depend on the
```
jbaudoux: typo
```suggestion
The approach in this addon is slightly different. It doesn't depend on the
```

stock_picking_batch_creation/README.rst
```diff
@@ -0,0 +1,195 @@
...
+for the batch creation. When all your criteria are set, click on the
+*Create Batch Picking* and lets the magic happen. A new window will open with
+the created batch picking.
+
+Behind the scene
+~~~~~~~~~~~~~~~~
+
+The creation of the picking batch is done in 3 steps:
+
+1. We search for a picking ready to be processed and that fit to the
```
jbaudoux:
```suggestion
1. We search for a picking ready to be processed and that fits the
```

stock_picking_batch_creation/README.rst
```diff
@@ -0,0 +1,195 @@
...
+   criteria specified in the wizard.
+2. We determine the device to use for this first picking.
+3. While the defined limits are not reached, we look for one picking at a
+   time that is ready to be processed and that fit the new criteria. The
```
jbaudoux:
```suggestion
   time that is ready to be processed and that fits the new criteria. The
```

stock_picking_batch_creation/README.rst
```diff
@@ -0,0 +1,195 @@
...
+
+At step2, since the first step will return a picking that can be handled by
+any of the device types specified, we process the list of device one by one
+in the order defined by the sequence field and we stop at the first device
+that can handle the picking.
+
+At step 3, now that we have a picking and a device, we have all the information
+to refine the search for the next picking to the volume, weight and number of
+lines remaining in the batch. While there is still some bins available and
+in the batch and the search returns result, we look for a picking that fit
```
jbaudoux:
```suggestion
lines remaining in the batch. While there are still some bins available and
?? in the batch and the search returns result, we look for a picking that fits
```

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+        if self.group_pickings_by_partner:
+            return "priority desc, scheduled_date asc, partner_id desc, id asc"
+        return "priority desc, scheduled_date asc, id asc"
+
+    def _get_picking_domain_common(self, user=None):
+        domain = [
+            ("picking_type_id", "in", self.picking_type_ids.ids),
+            ("state", "in", ("partially_available", "assigned")),
+            ("batch_id", "=", False),
+            ("user_id", "=", user.id if user else False),
```
jbaudoux: In all shopfloor scenario, what is assigned to a responsible/user can be processed by the user only, what is not assigned can be processed by anybody
```suggestion
            ("user_id", "in", [user.id, False] if user else [False]),
```

jbaudoux: Ok, I see, the call is done twice, once with the user and once without. If you do the change above and add user_id in the sorting, I think you should be able to drop the second call

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def _create_batch(self, raise_if_not_possible=False):
+        """Create a batch transfer."""
+        self._reset_counters()
+        # first we try to get the first picking for the user
+        first_picking = self._get_first_picking(user=self.user_id)
+        if not first_picking:
+            # if there is no picking for the user, we get the first picking
+            # without user
+            first_picking = self._get_first_picking()
```
jbaudoux: Add user_id in _get_picking_order_by and drop second call

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+                "nbr_picking_lines",
+                "<=",
+                self._remaining_nbr_picking_lines,
+            ),
+            ("id", "not in", excluded_ids),
+            ("weight", "<=", self._remaining_weight),
+        ]
+        user = self._first_picking.user_id
+        if user:
+            domain.append(("user_id", "=", user.id))
```
jbaudoux: Should be like above. If there is no user_id on the first picking, you cannot steal from another user
```suggestion
            domain.append(("user_id", "in", [user.id, False] if self.user_id or [False]))
```

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+            [
+                ("volume", "<=", self._get_remaining_volume()),
+            ]
+        ]
+        if self.group_pickings_by_partner:
+            previous_picking = self._previous_selected_picking
+            volume_domains.append(
+                [
+                    ("partner_id", "=", previous_picking.partner_id.id),
+                    ("scheduled_date", "=", previous_picking.scheduled_date),
```
jbaudoux: This looks very specific. Can you add a comment with the intention here? And why is the scheduled date necessary here?
Shouldn't this be favored compared to the other volume domain? Because you apply a OR between the 2 domains.

Can you also then keep the case with `group_pickings_by_commercial_partner` that is here https://github.com/OCA/wms/blob/14.0/shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py#L42

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def _get_picking_domain_common(self, user=None):
+        domain = [
+            ("picking_type_id", "in", self.picking_type_ids.ids),
+            ("state", "in", ("partially_available", "assigned")),
+            ("batch_id", "=", False),
+            ("user_id", "=", user.id if user else False),
+        ]
+        if not user:
+            domain.append(("printed", "=", False))
```
jbaudoux: Should not depend on the user
I would drop this

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+        domain = [
+            (
+                "nbr_picking_lines",
+                "<=",
+                self._remaining_nbr_picking_lines,
+            ),
+            ("id", "not in", excluded_ids),
+            ("weight", "<=", self._remaining_weight),
+        ]
+        user = self._first_picking.user_id
```
jbaudoux: It is missing `("picking_type_id", "=", self._first_picking.picking_type_id.id)` as you cannot mix pickings of different picking types in a same batch.
ref : https://github.com/OCA/wms/pull/523

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+                break
+            picking = self._get_additional_picking()
+            if picking:
+                self._add_picking(picking)
+
+    def _create_batch_values(self):
+        selected_pickings = self.env["stock.picking"].browse(self._selected_picking_ids)
+        return {
+            "picking_ids": [(6, 0, selected_pickings.ids)],
+            "user_id": self._first_picking.user_id.id,
```
jbaudoux: If we set a user_id in the creation wizard, the resulting batch should have that user
```suggestion
            "user_id": self.user_id.id,
```

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,404 @@
...
+                self._raise_create_batch_not_possible()
+            return self.env["stock.picking.batch"].browse()
+        device = self._compute_device_to_use(first_picking)
+        self._init_counters(first_picking, device)
+        self._apply_limits()
+        vals = self._create_batch_values()
+        batch = self.env["stock.picking.batch"].create(vals)
+        # Propagate user on selected pickings
+        user = self.user_id or self.env.user
+        batch.picking_ids.write({"user_id": user.id})
```
jbaudoux: As there is an explicit user on the wizard, only this one should be used. If you don't set a user on the wizard, then the batch will contain only pickings without responsible/user and the batch should not be assigned to a user
```suggestion
        if self.user_id:
            batch.picking_ids.write({"user_id": user.id})
```

jbaudoux (changes requested):

stock_picking_batch_creation/wizards/make_picking_batch.py
lmignon: @jbaudoux The implementation of the UC behind `group_pickings_by_commercial_partner` seems to ignore all the constrains... Do I miss something? In such a case we should introduce a kind of 'batch_mode' field into the wizard with 2 possibilities:
'based on limit' or 'by commercial partner'...

jbaudoux: @jgrandguillaume Do you remember why 2 users cannot prepare the same commercial partner for the rollercages process?
Can this be changed to always apply limits of the picking device but ensure commercial partners are not mixed in a batch?

lmignon:
@simahawk @jbaudoux The history is cleaned up

stock_picking_batch_creation/wizards/make_picking_batch.py
jbaudoux: I checked with Joel, the commercial partner is bypassing any limit due to a specific requirement in the process to ensure they all land on the same destination location. I would keep this case outside of this module as a project specific requirement

jbaudoux (approved):
Looks good now (besides a few useless `mapped`)
Not tested

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,415 @@
...
+                self._add_picking(picking)
+
+    def _create_batch_values(self):
+        selected_pickings = self.env["stock.picking"].browse(self._selected_picking_ids)
+        return {
+            "picking_ids": [(6, 0, selected_pickings.ids)],
+            "user_id": self.user_id.id,
+            "state": "draft",
+            "picking_device_id": self._device.id if self._device else None,
+            "is_wave": True,
```
rousseldenis: This is not correct as it will create ... a wave.

So, either, we delete this, either we should provide the choice to user as created records appear in different menus

@jbaudoux @lmignon

lmignon: @rousseldenis fixed

stock_picking_batch_creation/tests/test_clustering_conditions.py
```diff
@@ -0,0 +1,399 @@
...
+        )
+        # since the volume is computed at picking creation or update, force
+        # the recomputation
+        (self.pick1 | self.pick2 | self.pick3).mapped("move_ids")._compute_volume()
+        batch = self.make_picking_batch._create_batch(raise_if_not_possible=False)
+        self.assertFalse(batch)
+        message = "No device found for batch picking."
+        with self.assertRaises(NoSuitableDeviceError) as cm:
+            self.make_picking_batch._create_batch(raise_if_not_possible=True)
+            self.assertEqual(cm.exception.name, message)
```
rousseldenis:
```suggestion
            self.assertEqual(cm.exception, message)
```

As this provokes a warning

stock_picking_batch_creation/tests/test_clustering_conditions.py
```diff
@@ -0,0 +1,399 @@
...
+
+        # Display picking names in exception
+        self.make_picking_batch.add_picking_list_in_error = True
+
+        with self.assertRaises(NoSuitableDeviceError) as cm:
+            self.make_picking_batch._create_batch(raise_if_not_possible=True)
+        self.assertIn(
+            self.pick1.name,
+            cm.exception.name,
```
rousseldenis: Same as above

rousseldenis (commented):
You should do a copier update to mark this in rebel group as you pull sale modules (as delivery is in depends)

stock_picking_batch_creation/models/stock_picking_batch.py
```diff
@@ -24,10 +24,3 @@ class StockPickingBatch(models.Model):
         string="Number of picking lines",
         help="Indicates the picking lines ready for preparation.",
     )
-
```
lmignon: @sbejaoui This change breaks tests... When you write the user on the batch, it's not propagated to the pickings...

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -329,9 +329,6 @@ def _create_batch(self, raise_if_not_possible=False):
         self._apply_limits()
         vals = self._create_batch_values()
         batch = self.env["stock.picking.batch"].create(vals)
-        # Propagate user on selected pickings
```
lmignon: You're right here, this code is useless

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -0,0 +1,468 @@
...
+        self._reset_counters()
+        # first we try to get the first picking for the user
+        first_picking = self._get_first_picking()
+        if not first_picking and self.no_line_limit_if_no_candidate:
+            first_picking = self._get_first_picking(no_nbr_lines_limit=True)
+        if not first_picking:
+            if raise_if_not_possible:
+                self._raise_create_batch_not_possible()
+            return self.env["stock.picking.batch"].browse()
+        device = self._compute_device_to_use(first_picking)
```
jbaudoux, line 365: If first picking exceeds device's max_weight, create a split picking within limit.

In compute device, fallback on the last allowed device (as you will exclude them all if picking weight > all(device's max_weight)

jbaudoux: can be tackled as a further improvement in another PR

jbaudoux:
@lmignon @rousseldenis Can you update your review? I see all comments are outdated

lmignon (approved):
Works successfully in production for months

rousseldenis (approved):

lmignon:
po file fixed and branch rebased

PR #545, opened 2023-01-24
shopfloor: fix zp set destination qty zero

TDu:
Do not allow a user to pick a quantity of zero or less. When the `no_prefill_qty` is set. A quantity of zero is the default value and it will still work.

ref.: rau-109

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -972,6 +972,14 @@ class ZonePicking(Component):
         if response:
             return response
 
+        if quantity <= 0:
+            message = self.msg_store.picking_zero_quantity()
+            return self._response_for_set_line_destination(
+                move_line,
+                message=message,
+                qty_done=self._get_prefill_qty(move_line, qty=0),
+            )
+
         extra_message = ""
         if not accept_only_package:
             # When the barcode is a location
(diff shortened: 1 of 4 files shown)


jbaudoux (approved):

shopfloor/services/zone_picking.py
```diff
@@ -957,6 +957,12 @@ def set_destination(
         if response:
             return response
 
+        if quantity <= 0:
+            message = self.msg_store.picking_zero_quantity()
+            return self._response_for_set_line_destination(
+                move_line, message=message, qty_done=move_line.product_uom_qty
```
jbaudoux: The quantity done should not change while the error is raised

jbaudoux (changes requested):

jbaudoux (approved):

shopfloor/services/zone_picking.py
```diff
@@ -957,6 +957,14 @@ def set_destination(
         if response:
             return response
 
+        if quantity <= 0:
```
simahawk, line 975: no need for float_compare?

TDu: Nop, I don't think so.

simahawk:
@TDu can you rebase and squash?

TDu:
@simahawk  Rebased and squashed

PR #546, opened 2023-01-25
[14.0][FIX] shopfloor: fix flappy test

TDu:


simahawk (dismissed):

simahawk (changes requested):
merge failed, the error is still there

sebalix (approved):

shopfloor/tests/test_delivery_sublocation.py
```diff
@@ -34,24 +48,11 @@ def setUpClassBaseData(cls):
...
-                (cls.product_e, 10),  # E as raw product with a lot
-            ]
-        )
-        cls.raw_move = picking.move_lines[0]
-        cls.raw_lot_move = picking.move_lines[1]
-        cls._fill_stock_for_moves(cls.raw_move)
-        # Use the same lot on product from both picking
-        cls.lot = cls.raw_lot_move_sublocation.move_line_ids.lot_id
-        cls._fill_stock_for_moves(cls.raw_lot_move, in_lot=cls.lot)
-        picking.action_assign()
```
sebalix, line 54: Indeed this implementation was wrong, there is no move line available on `cls.raw_lot_move_sublocation.move_line_ids` until you call `picking.action_assign()`, no line, no lot.
It was the issue?

TDu: No, the lot was there. There is two pickings and the lot of the first one is reused on the 2nd one, so no problem there.
One picking and move lines is meant for the top location and the other one for the sublocation. But forcing the top location on the 2nd picking did not work what ever trick I used.
This change creates and assigns the top location picking before the sublocation is created to be sure that it used that top location.

TDu: Randomness was coming from this line https://github.com/OCA/wms/blob/d5628ed285e625c0c34a550d276a355a1d380965/shopfloor/services/delivery.py#L391
But now only one line should be found because the location is correct

simahawk (approved):

PR #548, opened 2023-01-29
[FIX] - stock_dynamic_routing: fix method params

sbejaoui:


sbejaoui:
cc/ @lmignon

lmignon (approved):
Thank you @sbejaoui Indeed, this override doesn't preserve the original method signature.

lmignon:
ping @OCA/logistics-maintainers

rousseldenis (approved):
Trivial

PR #547, opened 2023-01-26
[14.0] shopfloor: add search.find method

simahawk:
A more flexible way to search for records.

Extracted from #521

PR #549, opened 2023-01-30
[14.0] shopfloor: checkout allow pack all lines at once

simahawk:
Depends on:

- [x] https://github.com/OCA/wms/pull/547
- [x] https://github.com/OCA/wms/pull/562 

Main changes:

* [shopfloor: checkout allow pack all lines at once](https://github.com/OCA/wms/commit/a8baec01f83a87e77cdee3260482c7911fa2dfed)

On the select_line screen,
you can now skip the selection and scan a delivery package.

When this happens, the system asks for confirmation.

If you scan again the delivery package,
all lines will be selected and packed at the same time in the same package.

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -40,19 +40,24 @@ class Checkout(Component):
     _usage = "checkout"
     _description = __doc__
 
-    def _response_for_select_line(self, picking, message=None):
+    def _response_for_select_line(
+        self, picking, message=None, need_confirm_pack_all=False
+    ):
         if all(line.shopfloor_checkout_done for line in picking.move_line_ids):
             return self._response_for_summary(picking, message=message)
         return self._response(
             next_state="select_line",
-            data=self._data_for_select_line(picking),
+            data=self._data_for_select_line(
+                picking, need_confirm_pack_all=need_confirm_pack_all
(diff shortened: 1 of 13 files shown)


shopfloor/services/checkout.py
```diff
@@ -821,74 +893,101 @@ def scan_package_action(self, picking_id, selected_line_ids, barcode):
...
         selected_lines = self.env["stock.move.line"].browse(selected_line_ids).exists()
+        search_result = search.find(
+            barcode,
+            types=(
+                "package",
+                "product",
+                "packaging",
+                "lot",
+                "serial",
+                "delivery_packaging",
```
sebalix: If we want to preserve the previous implementation the package was handled later, but you probably put it first for some reasons. It could bring regressions only if the package names collide with other types (low chance). Ignore my comment if you think it's OK.
```suggestion
                "product",
                "packaging",
                "lot",
                "serial",
                "package",
                "delivery_packaging",
```

simahawk: package is usually more precise than product...

sebalix (approved):

simahawk:
moving out basic checkout changes to https://github.com/OCA/wms/pull/562

PR #550, opened 2023-01-30
[16.0][ADD] stock_release_channel_process_end_time

rousseldenis:
Depends on :

- https://github.com/OCA/wms/pull/493
- https://github.com/OCA/wms/pull/478
- https://github.com/OCA/partner-contact/pull/1464


stock_release_channel_process_end_date/models/stock_release_channel.py
```diff
@@ -0,0 +1,35 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class StockReleaseChannel(models.Model):
+
+    _inherit = "stock.release.channel"
+
+    process_end_date = fields.Datetime(
```
lmignon: @rousseldenis A release channel instance is not created every day. A release channel is a way to organize your pick to deliver process into your company. It's therefore useless to specify the date and time when the process of pickings should be completed. IMO, this process_end_date should be replaced by a process_end_time... 
@jbaudoux What do you thing about?

rousseldenis: > @rousseldenis A release channel instance is not created every day. A release channel is a way to organize your pick to deliver process into your company. It's therefore useless to specify the date and time when the process of pickings should be completed. IMO, this process_end_date should be replaced by a process_end_time...

@lmignon Ok, didn't catch that.

The only drawback I see is that is not flexible (but maybe it's wanted - I don't have the whole flow in mind) as one may want to have channel opened during the day, but pickings done during the night (so, overlap between two days)

test-requirements.txt
```diff
@@ -0,0 +1,2 @@
+odoo-addon-stock-release-channel @ git+https://github.com/OCA/wms.git@refs/pull/506/head#subdirectory=setup/stock_release_channel
+odoo-addon-stock-available-promise-release @ git+https://github.com/OCA/wms.git@refs/pull/478/head#subdirectory=setup/stock_available_promise_release
```
lmignon:
```suggestion
odoo-addon-stock-available-to-promise-release @ git+https://github.com/OCA/wms.git@refs/pull/478/head#subdirectory=setup/stock_available_to_promise_release
```

rousseldenis:
@lmignon This one is ready

stock_release_channel_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,79 @@
...
+        company_tz = self.env.company.partner_id.tz
+        for channel in self:
+            channel.process_end_time_tz = (
+                channel.warehouse_id.partner_id.tz or company_tz or False
+            )
+
+    @api.model
+    def assign_release_channel(self, picking):
+        res = super().assign_release_channel(picking)
+        # Check if a channel has been assigned to the picking and write scheduled_date
```
lmignon: @rousseldenis For my information, are you sure that a write is triggered if we assign the same value?

stock_release_channel_process_end_time/utils.py
```diff
@@ -0,0 +1,33 @@
...
+    return time_result
+
+
+def next_datetime(current: datetime, hours: time, **kwargs) -> datetime:
+    repl = current.replace(hour=hours.hour, minute=hours.minute, **kwargs)
+    if hours.tzinfo:
+        repl += hours.utcoffset()
+    while repl <= current:
+        repl = repl + timedelta(days=1)
+    return repl
```
lmignon, line 35: @rousseldenis the value must be rounded to the minute. Otherwise the schedule date will not be the same for all the pickings.

lmignon (changes requested):
A little change required but work as expected

stock_release_channel_process_end_time/utils.py
rousseldenis: Done

stock_release_channel_process_end_time/utils.py
```diff
@@ -0,0 +1,33 @@
...
+    fractional, integral = math.modf(hours)
+    time_result = time(
+        int(integral), int(float_round(60 * fractional, precision_digits=0)), 0
+    )
+    time_result = tz_to_utc_time(tz, time_result)
+    return time_result
+
+
+def next_datetime(current: datetime, hours: time, **kwargs) -> datetime:
+    repl = current.replace(hour=hours.hour, minute=hours.minute, second=0, **kwargs)
```
lmignon: @rousseldenis and now microsecond...
```suggestion
    repl = current.replace(hour=hours.hour, minute=hours.minute, second=0, microsecond=0 **kwargs)
```

rousseldenis: My god :-)

lmignon (approved):
LGTM (Code review, functional tests)

lmignon (changes requested):
~~@rousseldenis The `process_end_date` field is not taken into account into the computation of domains used to retrieve pickings to release | late | done
As consequence, if the process_end_date is computed to tomorrow, the pickings assigned to the channel will not be taken visible since the scheduled_date will be set at tomorrow and the date criteria into the domains is set to today... (see https://github.com/OCA/wms/pull/493/files#diff-a8bb5a8dfb388e49242b1d9fe2274b33dccde4e032d0983647f70fa52399ecbcR174)~~

implemented in https://github.com/OCA/wms/pull/550/commits/8f2038535412a6fc13f34c30793da926d9ee5a73

lmignon:
@rousseldenis Changes to the process_end_time is not applied to the process_end_date and pickings to process (maybe it's overkill???)

stock_release_channel_process_end_time/models/stock_release_channel.py
rousseldenis: Yes. Checked.

We pass always in that line:

https://github.com/odoo/odoo/blob/16.0/odoo/fields.py#L1320

In fact, I think this is done at purpose to avoid maybe a unneeded read if field value is not in the cache (to be able to compare) and to be able to update the write_uid/write_date as in fact, people try to update the record even if value is the same.

lmignon (approved):

stock_release_channel_process_end_time/models/stock_picking.py
```diff
@@ -0,0 +1,72 @@
...
+
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    schedule_date_prior_to_channel_process_end_date_search = fields.Boolean(
+        store=False,
+        search="_search_schedule_date_prior_to_channel_process_end_date",
+        help="Technical field to search on not processed pickings where the scheduled"
+        "date is prior to the process end date of available channels.",
```
jbaudoux:
```suggestion
        help="Technical field to search on not processed pickings where the scheduled "
        "date is prior to the process end date of available channels.",
```

stock_release_channel_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class StockReleaseChannel(models.Model):
+
+    _inherit = "stock.release.channel"
+
+    process_end_time = fields.Float(
+        help="Fill in this to indicates when this channel release process would"
+        "be ended. This information will be used to compute the channel pickings"
+        "scheduled date at channel awaking.",
```
jbaudoux:
```suggestion
        help="Fill in this to indicates when this channel release process would "
        "be ended. This information will be used to compute the channel pickings "
        "scheduled date at channel awaking.",
```

jbaudoux (commented):
Some typo in the helps

sbejaoui (approved):

stock_release_channel_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @api.depends("warehouse_id.partner_id.tz")
+    @api.depends_context("company")
+    def _compute_process_end_time_tz(self):
+        # As the time is timezone-agnostic, we use the channel warehouse adress timezone
+        # or the company one, either it will be considered as UTC
+        company_tz = self.env.company.partner_id.tz
+        for channel in self:
+            channel.process_end_time_tz = (
+                channel.warehouse_id.partner_id.tz or company_tz or False
```
jbaudoux: Otherwise float_to_time will fail if tz is a bool
```suggestion
                channel.warehouse_id.partner_id.tz or company_tz or "UTC"
```

jbaudoux (changes requested):

jbaudoux:
@rousseldenis @sbejaoui Can you fix the small remaining issues on this PR ?

lmignon:
ping @rousseldenis @sbejaoui

sbejaoui:
> @rousseldenis @sbejaoui Can you fix the small remaining issues on this PR ?

done

stock_release_channel_process_end_time/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import api, fields, models, SUPERUSER_ID, _
+from odoo.exceptions import UserError
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    update_scheduled_date = fields.Boolean(
```
jbaudoux: This config is a bit misleading as the picking scheduled date is anyway updated by the release data + delay.

stock_release_channel_process_end_time/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+        help="Will update scheduled date of picking based on process end date(time).",
```
jbaudoux:
```suggestion
        help="Will update scheduled date of picking based on process end date instead of release date + delay.",
```

jbaudoux (changes requested):
What's the reason behind this new config? In which situation you don't want to plan the work for the given end date?
In my opinion, we should have by default the option active.
If the option is inactive, we should still never have a scheduled date > process end date. So it should still be min(release date + delay, process end date)

jbaudoux:
> What's the reason behind this new config? In which situation you don't want to plan the work for the given end date? In my opinion, we should have by default the option active. If the option is inactive, we should still never have a scheduled date > process end date. So it should still be min(release date + delay, process end date)

So the option must remain inactive because it makes the test in `stock_available_to_promise_release` fail

stock_release_channel_process_end_time/models/res_config_settings.py
jbaudoux: Could be renamed to stock_release_use_channel_end_date

simahawk:
@rousseldenis ping (for the build)

jbaudoux (approved):

simahawk:
@rousseldenis do you have time to rebase and shorten/cleanup commits? I can do it otherwise

stock_release_channel_process_end_time/models/stock_picking.py
```diff
@@ -0,0 +1,82 @@
...
+            )
+        """
+        if value:
+            operator_inselect = "inselect" if operator == "=" else "not inselect"
+        else:
+            operator_inselect = "not inselect" if operator == "=" else "inselect"
+        return [("id", operator_inselect, (query, []))]
+
+    def _after_release_set_expected_date(self):
+        enabled_update_scheduled_date = bool(
```
simahawk, line 64: I'd use str2bool here

simahawk (approved):
LG

stock_release_channel_process_end_time/models/stock_picking.py
simahawk: FTR I tested w/ boolean field tied to a config param: when it sets to False the param is deleted complely -> nothing to fear here.

PR #554, opened 2023-02-02
[14.0] shopfloor_reception: exclude view locations from scan

JuMiSanAr:
View locations should not be accepted when scanning a location.

ref: rau-121


lmignon (approved):
Makes sense...

simahawk (approved):

PR #558, opened 2023-02-06
shopfloor: Add first scan location option on zp

TDu:
Add a new scenario option `First scan location or pack`. And implement it on the zone picking scenario.

ref.: rau-113

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -206,6 +206,7 @@ class ZonePicking(Component):
             move_lines, product=product, sublocation=sublocation
         )
         data["confirmation_required"] = confirmation_required
+        data["scan_location_or_pack_first"] = self.work.menu.scan_location_or_pack_first
         return self._response(
             next_state="select_line",
             data=data,
@@ -650,7 +651,9 @@ class ZonePicking(Component):
         elif len(move_lines.lot_id) > 1:
             message = self.msg_store.several_move_with_different_lot()
         if message:
-            response = self._list_move_lines(self.zone_location, product)
+            response = self._list_move_lines(
+                self.zone_location, product, sublocation=sublocation
+            )
         elif move_lines:
             move_line = first(move_lines)
             qty_done = self._get_prefill_qty(move_line, qty=(packaging.qty or 1.0))
@@ -658,7 +661,9 @@ class ZonePicking(Component):
                 move_line, qty_done=qty_done
             )
         else:
-            response = self._list_move_lines(sublocation or self.zone_location)
+            response = self._list_move_lines(
+                sublocation or self.zone_location, sublocation=sublocation
+            )
             message = self.msg_store.product_not_found()
         return response, message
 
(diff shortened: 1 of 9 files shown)


shopfloor/services/zone_picking.py
```diff
@@ -205,6 +206,12 @@ def _response_for_select_line(
             move_lines, product=product, sublocation=sublocation
         )
         data["confirmation_required"] = confirmation_required
+        placeholder = _("Scan location / pack / product / lot")
+        if self.work.menu.scan_location_or_pack_first and not sublocation:
+            placeholder = _("Scan location / pack")
+        elif sublocation:
+            placeholder = _("Scan product / lot")
+        data["searchbar_placeholder"] = placeholder
```
simahawk: I would expose the flag instead and change the message in the UI only. This is not an exception.

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -606,7 +606,13 @@ const ZonePicking = {
                 select_line: {
                     display_info: {
                         title: "Select move",
-                        scan_placeholder: "Scan location / pack / product / lot",
+                        scan_placeholder: () => {
+                            if (this.state.data.searchbar_placeholder) {
```
simahawk: see comment above

shopfloor/services/zone_picking.py
simahawk: hmm this flag shoud be used to check what you scan otherwise what's the point? :thinking:

simahawk (changes requested):

shopfloor/services/zone_picking.py
TDu: The `scan_location_or_pack_first` is used to configure the handler in `scan_source` below. https://github.com/OCA/wms/pull/558/files#diff-81fce66279454c4c0be2adf1f46eb4e629c63edfe409c4cf6bda95940686f9bcR753

TDu: > I would expose the flag instead and change the message in the UI only. This is not an exception.

To me it seems easier to pass the placeholder in the response, instead of generating a flag on the backend, that the frontend will need to decode to generate the placeholder.

simahawk: All the translations for the UI stay in the UI. This shouldn't be an exception.
This very same check you do here, can be done in the UI w/o complicating anything.
Moreover, if you want to display things differently or display another msg you have full control on the UI.

TDu: Well, it is a fine line between back and front. Like all error/warning messages are translated by the backend but displayed on the frontend.
Anyway I changed this PR see the fixup.

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -606,7 +606,19 @@ const ZonePicking = {
...
                     display_info: {
                         title: "Select move",
-                        scan_placeholder: "Scan location / pack / product / lot",
+                        scan_placeholder: () => {
+                            const sublocation = this.state.data.sublocation;
+                            if (
+                                this.state.data.scan_location_or_pack_first &&
+                                !sublocation
+                            ) {
+                                return "Scan location / pack 000";
```
```suggestion
                                return "Scan location / pack";
```

simahawk (approved):
LG

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
TDu: Oops, left over from my testing.
I fixed that. And squashed.

TDu: And rebased

jbaudoux (approved):
LG, tested

simahawk:
@TDu can you rebase?

PR #559, opened 2023-02-06
[14.0] shopfloor_reception: add return handling

mmequignon:
ref: rau-141

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -594,6 +594,12 @@ class MessageAction(Component):
             ),
         }
 
+    def move_already_returned(self):
+        return {
+            "message_type": "error",
+            "body": _("The product/packaging you selected has already been returned."),
+        }
+
     def return_line_invalid_qty(self):
         return {
             "message_type": "error",
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -594,6 +594,12 @@ class MessageAction(Component):
             ),
         }
 
+    def return_line_invalid_qty(self):
+        return {
+            "message_type": "error",
+            "body": _("You cannot return more quantity than what was initially sent."),
+        }
+
     def transfer_no_qty_done(self):
         return {
             "message_type": "warning",
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -565,6 +565,15 @@ class MessageAction(Component):
             ).format(lot=lot.name, product=lot.product_id.name),
         }
 
+    def no_default_location_on_picking_type(self):
+        return {
+            "message_type": "error",
+            "body": _(
+                "Operation types for this menu are missing "
+                "default source and destination locations."
+            ),
+        }
+
     def location_src_set_to_sublocation(self, location_src):
         return {
             "message_type": "success",
--- a/shopfloor/actions/search.py
+++ b/shopfloor/actions/search.py
@@ -1,6 +1,8 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from odoo.osv.expression import AND
+
 from odoo.addons.component.core import Component
 
 
@@ -53,6 +55,7 @@ class SearchAction(Component):
             "serial": self.lot_from_scan,
             "packaging": self.packaging_from_scan,
             "delivery_packaging": self.generic_packaging_from_scan,
+            "origin_move": self.origin_move_from_scan,
         }
 
     def _make_search_result(self, **kwargs):
@@ -164,3 +167,15 @@ class SearchAction(Component):
         return model.search(
             [("barcode", "=", barcode), ("product_id", "=", False)], limit=1
         )
+
+    def origin_move_from_scan(self, barcode, extra_domain=None):
+        model = self.env["stock.move"]
+        outgoing_move_domain = [
+            # We could have the same origin for multiple transfers
+            # but we're interested only in the "done" ones.
+            ("origin", "=", barcode),
+            ("state", "=", "done"),
+        ]
+        if extra_domain:
+            outgoing_move_domain = AND([outgoing_move_domain, extra_domain])
+        return model.search(outgoing_move_domain)
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-from odoo import _
+from odoo import _, fields
+from odoo.tools.float_utils import float_round
 
 from odoo.addons.component.core import Component
 
@@ -13,6 +14,98 @@ class StockAction(Component):
     _name = "shopfloor.stock.action"
     _inherit = "shopfloor.process.action"
     _usage = "stock"
+
+    def _create_return_move__get_max_qty(self, origin_move):
+        """Returns the max returneable qty."""
+        # The max returnable qty is the sent qty minus the already returned qties
+        quantity = origin_move.product_qty
+        for move in origin_move.move_dest_ids:
+            if (
+                move.origin_returned_move_id
+                and move.origin_returned_move_id != origin_move
+            ):
+                continue
+            if move.state in ("partially_available", "assigned"):
+                quantity -= sum(move.move_line_ids.mapped("product_qty"))
+            elif move.state in ("done"):
+                quantity -= move.product_qty
+        return float_round(
+            quantity, precision_rounding=origin_move.product_id.uom_id.rounding
+        )
+
+    def _create_return_move__get_vals(self, return_picking, origin_move):
+        product = origin_move.product_id
+        return_type = return_picking.picking_type_id
+        return {
+            "product_id": product.id,
+            "product_uom": product.uom_id.id,
+            "picking_id": return_picking.id,
+            "state": "draft",
+            "date": fields.Datetime.now(),
+            "location_id": return_picking.location_id.id,
+            "location_dest_id": return_picking.location_dest_id.id,
+            "picking_type_id": return_type.id,
+            "warehouse_id": return_type.warehouse_id.id,
+            "origin_returned_move_id": origin_move.id,
+            "procure_method": "make_to_stock",
+        }
+
+    def _create_return_move__link_to_origin(self, return_move, origin_move):
+        move_orig_to_link = origin_move.move_dest_ids.mapped("returned_move_ids")
+        move_orig_to_link |= origin_move
+        origin_move_dest = origin_move.move_dest_ids.filtered(
+            lambda m: m.state not in ("cancel")
+        )
+        move_orig_to_link |= origin_move_dest.move_orig_ids.filtered(
+            lambda m: m.state not in ("cancel")
+        )
+        move_dest_to_link = origin_move.move_orig_ids.mapped("returned_move_ids")
+        move_dest_orig = origin_move.returned_move_ids.move_orig_ids.filtered(
+            lambda m: m.state not in ("cancel")
+        )
+        move_dest_to_link |= move_dest_orig.move_dest_ids.filtered(
+            lambda m: m.state not in ("cancel")
+        )
+        write_vals = {
+            "move_orig_ids": [(4, m.id) for m in move_orig_to_link],
+            "move_dest_ids": [(4, m.id) for m in move_dest_to_link],
+        }
+        return_move.write(write_vals)
+
+    def create_return_move(self, return_picking, origin_moves):
+        """Creates a return move for a given return picking / move"""
+        # Logic has been copied from
+        # odoo_src/addons/stock/wizard/stock_picking_return.py
+        for origin_move in origin_moves:
+            # If max qty <= 0, it means that everything has been returned already.
+            # Try with the next one from the recordset.
+            max_qty = self._create_return_move__get_max_qty(origin_move)
+            if max_qty > 0:
+                return_move_vals = self._create_return_move__get_vals(
+                    return_picking, origin_move
+                )
+                return_move_vals.update(product_uom_qty=max_qty)
+                return_move = origin_move.copy(return_move_vals)
+                self._create_return_move__link_to_origin(return_move, origin_move)
+                return return_move
+
+    def _create_return_picking__get_vals(self, return_types, origin):
+        return_type = fields.first(return_types)
+        return {
+            "move_lines": [],
+            "picking_type_id": return_type.id,
+            "state": "draft",
+            "origin": origin,
+            "location_id": return_type.default_location_src_id.id,
+            "location_dest_id": return_type.default_location_dest_id.id,
+            "is_shopfloor_return": True,
+        }
+
+    def create_return_picking(self, picking, return_types, origin):
+        # Logic has been copied from
+        # odoo_src/addons/stock/wizard/stock_picking_return.py
+        return_values = self._create_return_picking__get_vals(return_types, origin)
+        return picking.copy(return_values)
 
     def mark_move_line_as_picked(
         self, move_lines, quantity=None, package=None, user=None, check_user=False
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -98,7 +98,7 @@ class StockAction(Component):
             "origin": origin,
             "location_id": return_type.default_location_src_id.id,
             "location_dest_id": return_type.default_location_dest_id.id,
-            "is_shopfloor_return": True,
+            "is_shopfloor_created": True,
         }
 
     def create_return_picking(self, picking, return_types, origin):
--- a/shopfloor/models/shopfloor_menu.py
+++ b/shopfloor/models/shopfloor_menu.py
@@ -39,6 +39,13 @@ automatically post the corresponding line
 if this option is checked.
 """
 
+RETURN_HELP = """
+When enabled, you can receive unplanned products that are returned
+from an existing delivery matched on the origin (SO name).
+A new move will be added as a return of the delivery,
+decreasing the delivered quantity of the related SO line.
+"""
+
 
 class ShopfloorMenu(models.Model):
     _inherit = "shopfloor.menu"
--- a/shopfloor/models/shopfloor_menu.py
+++ b/shopfloor/models/shopfloor_menu.py
@@ -186,6 +186,14 @@ class ShopfloorMenu(models.Model):
     allow_alternative_destination_is_possible = fields.Boolean(
         compute="_compute_allow_alternative_destination_is_possible"
     )
+    allow_return_is_possible = fields.Boolean(
+        compute="_compute_allow_return_is_possible"
+    )
+    allow_return = fields.Boolean(
+        string="Allow create returns",
+        default=False,
+        help=RETURN_HELP,
+    )
 
     auto_post_line = fields.Boolean(
         string="Automatically post line",
@@ -411,3 +419,8 @@ class ShopfloorMenu(models.Model):
             menu.allow_alternative_destination_is_possible = (
                 menu.scenario_id.has_option("allow_alternative_destination")
             )
+
+    @api.depends("scenario_id")
+    def _compute_allow_return_is_possible(self):
+        for menu in self:
+            menu.allow_return_is_possible = menu.scenario_id.has_option("allow_return")
--- a/shopfloor/services/service.py
+++ b/shopfloor/services/service.py
@@ -46,7 +46,7 @@ class BaseShopfloorProcess(AbstractComponent):
         # every time.
         return self._actions_for("search_move_line", picking_types=self.picking_types)
 
-    def _check_picking_status(self, pickings):
+    def _check_picking_status(self, pickings, states=("assigned",)):
         """Check if given pickings can be processed.
 
         If the picking is already done, canceled or didn't belong to the
@@ -57,7 +57,7 @@ class BaseShopfloorProcess(AbstractComponent):
                 return self.msg_store.stock_picking_not_found()
             if picking.state == "done":
                 return self.msg_store.already_done()
-            if picking.state != "assigned":  # the picking must be ready
+            if picking.state not in states:  # the picking must be ready
                 return self.msg_store.stock_picking_not_available(picking)
             if picking.picking_type_id not in self.picking_types:
                 return self.msg_store.cannot_move_something_in_picking_type()
--- a/shopfloor_reception/data/shopfloor_scenario_data.xml
+++ b/shopfloor_reception/data/shopfloor_scenario_data.xml
@@ -8,7 +8,8 @@
       <field name="key">reception</field>
       <field name="options_edit">
 {
-  "auto_post_line": true
+      "auto_post_line": true,
+      "allow_return": true
 }
       </field>
   </record>
--- a/shopfloor_reception/models/stock_picking.py
+++ b/shopfloor_reception/models/stock_picking.py
@@ -1,0 +1,10 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    is_shopfloor_return = fields.Boolean()
--- a/shopfloor_reception/models/stock_picking.py
+++ b/shopfloor_reception/models/stock_picking.py
@@ -7,4 +7,4 @@ from odoo import fields, models
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
-    is_shopfloor_return = fields.Boolean()
+    is_shopfloor_created = fields.Boolean()
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -629,14 +629,13 @@ class Reception(Component):
                     ),
                 )
             # If the scanned package has a valid destination,
-            # set both package and destination on the package,
-            # and go back to the selection line screen
+            # set both package and destination on the package.
             selected_line.result_package_id = package
             selected_line.location_dest_id = pack_location
-            if self.work.menu.auto_post_line:
-                # If option auto_post_line is active in the shopfloor menu,
-                # create a split order with this line.
-                self._auto_post_line(selected_line, picking)
+
+            response = self._post_line(selected_line)
+            if response:
+                return response
             return self._response_for_select_move(picking)
         # Scanned package has no location, move to the location selection
         # screen
@@ -1175,6 +1174,27 @@ class Reception(Component):
         selected_line.qty_done = quantity
         return self._response_for_set_destination(picking, selected_line)
 
+    def _post_line(self, selected_line):
+        if (
+            selected_line.picking_id.is_shopfloor_created
+            and self.work.menu.allow_return
+        ):
+            # If the transfer is not planned, and we allow unplanned returns,
+            # process the returned qty and mark it as done.
+            return self._post_shopfloor_created_line(selected_line)
+
+        if self.work.menu.auto_post_line:
+            # If option auto_post_line is active in the shopfloor menu,
+            # create a split order with this line.
+            self._auto_post_line(selected_line)
+
+    def _post_shopfloor_created_line(self, selected_line):
+        selected_line.product_uom_qty = selected_line.qty_done
+        selected_line.picking_id.with_context(cancel_backorder=True)._action_done()
+        return self._response_for_select_document(
+            message=self.msg_store.transfer_done_success(selected_line.picking_id)
+        )
+
     def _auto_post_line(self, selected_line):
         new_move = selected_line.move_id.split_other_move_lines(
             selected_line, intersection=True
@@ -1238,11 +1258,9 @@ class Reception(Component):
                     ),
                 )
             selected_line.location_dest_id = location
-
-        if self.work.menu.auto_post_line:
-            # If option auto_post_line is active in the shopfloor menu,
-            # create a split order with this line.
-            self._auto_post_line(selected_line)
+        response = self._post_line(selected_line)
+        if response:
+            return response
         return self._response_for_select_move(picking)
 
     def select_dest_package(
@@ -1284,10 +1302,9 @@ class Reception(Component):
                     message=self.msg_store.package_not_empty(package),
                 )
             selected_line.result_package_id = package
-            if self.work.menu.auto_post_line:
-                # If option auto_post_line is active in the shopfloor menu,
-                # create a split order with this line.
-                self._auto_post_line(selected_line)
+            response = self._post_line(selected_line)
+            if response:
+                return response
             return self._response_for_select_move(picking)
         message = self.msg_store.create_new_pack_ask_confirmation(barcode)
         self._assign_user_to_picking(picking)
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -233,8 +233,15 @@ class Reception(Component):
             message=self.msg_store.lot_not_found_in_pickings(),
         )
 
-    def _select_line(self, picking, line, move, increase_qty_done_by=1):
-        product = line.product_id
+    def _scan_line__find_or_create_line(self, picking, move, qty_done=1):
+        line = fields.first(
+            move.move_line_ids.filtered(
+                lambda l: (
+                    not l.result_package_id
+                    and l.shopfloor_user_id.id in [False, self.env.uid]
+                )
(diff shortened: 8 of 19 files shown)


shopfloor/actions/search.py
```diff
@@ -157,3 +158,20 @@ def generic_packaging_from_scan(self, barcode):
...
             [("barcode", "=", barcode), ("product_id", "=", False)], limit=1
         )
+
+    def delivered_picking_from_scan(self, barcode, use_origin=False):
+        model = self.env["stock.picking"]
+        if not barcode:
+            return model.browse()
+        picking = model.search([("name", "=", barcode)], limit=1)
+        if picking:
+            return picking
```
simahawk:
```suggestion
        picking = self.picking_from_scan(barcode)
        if picking:
            return picking
```

shopfloor_reception/services/reception.py
```diff
@@ -211,14 +243,34 @@ def _select_line_from_product(self, picking, move, product):
...
 
-    def _select_line_from_packaging(self, picking, move, packaging):
-        line = fields.first(
-            picking.move_line_ids.filtered(
+    def _select_line__filter_lines_by_packaging__rma(self, lines, packaging):
+        return_line = fields.first(
+            lines.filtered(
+                lambda l: not l.package_id.product_packaging_id
+                and not l.result_package_id
+                and l.shopfloor_user_id.id in [False, self.env.uid]
```
```suggestion
                and l.shopfloor_user_id.id in (False, self.env.uid)
```

shopfloor_reception/services/reception.py
```diff
@@ -275,45 +326,165 @@ def _scan_document__by_picking(self, barcode):
...
 
-    def _scan_line__by_packaging(self, picking, barcode):
-        search = self._actions_for("search")
-        packaging = search.packaging_from_scan(barcode)
-        if packaging:
-            move = picking.move_lines.filtered(
-                lambda m: packaging in m.product_id.packaging_ids
+    def _scan_line__create_return_move(self, return_picking, origin_move):
+        # copied from odoo/src/addons/stock/wizard/stock_picking_return.py
+        product = origin_move.product_id
```
simahawk: the whole method split in smaller methods could go to shopfloor/actions/stock IMO

simahawk (changes requested):

simahawk:
@mmequignon there are some TODOs: do you plan to address them here?

shopfloor/services/service.py
```diff
@@ -46,18 +46,19 @@ def search_move_line(self):
...
         return self._actions_for("search_move_line", picking_types=self.picking_types)
 
-    def _check_picking_status(self, pickings):
+    def _check_picking_status(self, pickings, states=None):
         """Check if given pickings can be processed.
 
         If the picking is already done, canceled or didn't belong to the
         expected picking type, a message is returned.
         """
+        valid_states = states or ["assigned"]
```
TDu: could be done in the function declaration ?
`def _check_picking_status(self, pickings, states=["assigned"]):`

shopfloor_reception/services/reception.py
```diff
@@ -134,6 +143,29 @@ def _response_for_confirm_new_package(
             next_state="confirm_new_package", data=data, message=message
         )
 
+    def _scan_document__create_return(self, outgoing_picking):
+        return_type = (
+            fields.first(self.picking_types)
+            or outgoing_picking.picking_type_id.return_picking_type_id
```
TDu: Why not use the `return_picking_type_id` by default ?

mmequignon: Because goods could be sent from a given warehouse and returned to another one.
Profiles are linked to the warehouse, so when you use profile's picking type's return type, you're sure the right warehouse is used for the process.
This is what I understood from @jbaudoux explaination.

shopfloor/services/service.py
mmequignon: this is bad practice to use mutable objects as default values:
```python
❯ python3
Python 3.8.10 (default, Nov 14 2022, 12:59:47) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def func(arg=[]):
...     arg.append("potato")
...     print(arg)
... 
>>> func()
['potato']
>>> func()
['potato', 'potato']
>>> func()
['potato', 'potato', 'potato']
```

mmequignon:
@simahawk finished to address your comments. Could you please make another pass ? :pray:

shopfloor/services/service.py
simahawk: yep: never use non hashable types as default args

shopfloor/models/shopfloor_menu.py
```diff
@@ -140,6 +140,8 @@ class ShopfloorMenu(models.Model):
     no_prefill_qty_is_possible = fields.Boolean(
         compute="_compute_no_prefill_qty_is_possible"
     )
+    allow_rma_is_possible = fields.Boolean(compute="_compute_allow_rma_is_possible")
+    allow_rma = fields.Boolean(string="Allow create returns", default=False, help="")
```
jbaudoux: Best help ever ;)

jbaudoux: rename rma by return. An RMA is an authorization to return.
```suggestion
    allow_create_return_is_possible = fields.Boolean(compute="_compute_allow_rma_is_possible")
    allow_create_return = fields.Boolean(string="Allow create returns", default=False, help="")
```

shopfloor_reception/services/reception.py
```diff
@@ -275,45 +310,113 @@ def _scan_document__by_picking(self, barcode):
...
+        return_picking.action_assign()
+        return return_move
+
+    def _scan_line__by_product__rma(self, picking, product):
+        origin_picking = self.env["stock.picking"].search(
+            [("name", "=", picking.origin)]
+        )
+        origin_move = origin_picking.move_lines.filtered(
+            lambda m: m.product_id == product
+        )
```
jbaudoux: This can return many moves

shopfloor_reception/services/reception.py
```diff
@@ -326,79 +429,120 @@ def _check_move_available(self, move, message_code="product"):
...
+            message_type = response.get("message", {}).get("message_type")
+            # If we have an error, return it, since this is also true for return lines
+            if message_type == "error":
+                return response
+            # We cannot set a qty_done superior to what has initally been sent
+            rounding = selected_line.product_uom_id.rounding
+            compare = float_compare(
+                selected_line.qty_done, max_qty_done, precision_rounding=rounding
+            )
+            # We cannot set a qty_done superior to what has initally been sent
```
jbaudoux, line 592:
```suggestion
            # We cannot set a qty_done **bigger than** what has **initially** been sent
```

shopfloor/actions/search.py
```diff
@@ -157,3 +158,18 @@ def generic_packaging_from_scan(self, barcode):
         return model.search(
             [("barcode", "=", barcode), ("product_id", "=", False)], limit=1
         )
+
+    def delivered_picking_from_scan(self, barcode, use_origin=False):
+        picking = self.packaging_from_scan(barcode)
+        if picking:
+            return picking
```
jbaudoux: I don't get why this here?

shopfloor/actions/search.py
```diff
@@ -157,3 +158,18 @@ def generic_packaging_from_scan(self, barcode):
...
+        model = self.env["stock.picking"]
+        if use_origin:
+            source_document_domain = [
+                # We could have the same origin for multiple transfers
+                # but we're interested only in the "done" ones.
+                ("origin", "=", barcode),
+                ("state", "=", "done"),
+            ]
+            return model.search(source_document_domain)
```
jbaudoux: Some remarks regarding the origin:
* we need to be able to pre-process the barcode. In our case, the origin is prefixed by the transporter and this prefix must be dropped to match a picking
* I would have added a domain to the method and include it in the document origin search. For picking, the domain would be type=incoming + state=assigned. For return, the domain would be type=outgoing + state=done. Then you would not need this new handler.
* a picking can have multiple origins separated by a comma. You could have a reception that concerns multiple POs. You can have a delivery to return that concerns multiple SOs. In those cases, the search on "=" won't work. I wonder if the origin must not be on moves instead of picking. This is because after matching the picking, you still have to match the move based on the scanned product/packaging/lot but you cannot mix the origins. You then would have an handler for the picking only on its name and another handler on the move origin

shopfloor/actions/stock.py
```diff
@@ -10,6 +13,93 @@ class StockAction(Component):
...
+        return_type = (
+            fields.first(return_types) or picking.picking_type_id.return_picking_type_id
+        )
+        return {
+            "move_lines": [],
+            "picking_type_id": return_type.id,
+            "state": "draft",
+            # This isn't std but helps in `scan_line`.
+            # f"Return of  {outgoing_picking.name}"
+            "origin": picking.name,
```
jbaudoux: I don't like this. Especially that you reach here after taking the first matching picking which could be wrong as you don't know yet which product.
We should not loose the origin and probably reuse it to collect back the complete set of valid pickings and not just the first one

jbaudoux (changes requested):
Not tested yet but here is a first bunch of remarks

shopfloor/actions/stock.py
```diff
@@ -10,6 +13,93 @@ class StockAction(Component):
...
+            "location_id": picking.location_dest_id.id,
```
jbaudoux:
```suggestion
            "location_id": return_type.location_id.id,
```

shopfloor/models/shopfloor_menu.py
jbaudoux: When enabled, you can receive unplanned products that are returned from an existing delivery matched on the origin (SO name). A new move will be added as a return of the delivery decreasing the delivered quantity of the related SO line.

mmequignon: This is true for the reception scenario.
I've put this option in `shopfloor` because at some point, we might want to manage returns for other scenarios.

mmequignon: I don't even know if it makes sense…

jbaudoux: Yes, it was a good idea. Nevertheless, whatever scenario implements it, the intent is that one

shopfloor_reception/data/shopfloor_scenario_data.xml
```diff
@@ -8,7 +8,8 @@
       <field name="key">reception</field>
       <field name="options_edit">
 {
-  "auto_post_line": true
+      "auto_post_line": true,
+      "allow_rma": true
```
jbaudoux:
```suggestion
      "allow_return": true
```

shopfloor/services/service.py
sebalix: Then why not using a tuple here? Auto-documented, no need to check the implementation.
```python
def _check_picking_status(self, pickings, states=("assigned",)):
```

simahawk:
@mmequignon pls rebase and squash

shopfloor/actions/stock.py
```diff
@@ -10,6 +13,96 @@ class StockAction(Component):
...
+        for move in origin_move.move_dest_ids:
+            if (
+                move.origin_returned_move_id
+                and move.origin_returned_move_id != origin_move
+            ):
+                continue
+            if move.state in ("partially_available", "assigned"):
+                quantity -= sum(move.move_line_ids.mapped("product_qty"))
+            elif move.state in ("done"):
+                quantity -= move.product_qty
```
jbaudoux, line 31: Some comments would be helpful understanding what is being done

shopfloor/actions/stock.py
```diff
@@ -10,6 +13,96 @@ class StockAction(Component):
...
+        product = origin_move.product_id
+        return_type = return_picking.picking_type_id
+        return {
+            "product_id": product.id,
+            "product_uom": product.uom_id.id,
+            "picking_id": return_picking.id,
+            "state": "draft",
+            "date": fields.Datetime.now(),
+            "location_id": return_picking.location_id.id,
+            "location_dest_id": return_picking.location_dest_id.id,
```
jbaudoux, line 46: The location on the picking type could be empty and this will cause the new picking creation to fail with an error 500. Add a check that the return picking is properly configured otherwise raise a proper error

shopfloor/actions/stock.py
```diff
@@ -10,6 +13,96 @@ class StockAction(Component):
...
+
+    def _create_return_picking__get_vals(self, return_types, origin):
+        return_type = fields.first(return_types)
+        return {
+            "move_lines": [],
+            "picking_type_id": return_type.id,
+            "state": "draft",
+            "origin": origin,
+            "location_id": return_type.default_location_src_id.id,
+            "location_dest_id": return_type.default_location_dest_id.id,
```
jbaudoux, line 100: The location on the picking type could be empty and this will cause the new picking creation to fail with an error 500. Add a check that the return picking is properly configured otherwise raise a proper error

jbaudoux, line 100: The location on the picking type could be empty and this will cause the new picking creation to fail with an error 500. Add a check that the return picking is properly configured otherwise raise a proper error

jbaudoux (commented):
Once the return picking is created in draft, if you cannot process the product, it would be nice to have a cancel button that deletes the draft picking and goes back to the start screen

shopfloor_reception/services/reception.py
```diff
@@ -302,91 +322,165 @@ def _scan_document__by_picking(self, barcode):
...
+    def _scan_line__by_lot(self, picking, lot):
+        lines = picking.move_line_ids.filtered(
+            lambda l: (
+                lot == l.lot_id
+                or (lot.name == l.lot_name and lot.product_id == l.product_id)
+                and not l.result_package_id
+            )
+        )
+        if not lines:
+            return self._scan_line__by_product(picking, lot.product_id)
```
jbaudoux, line 486: This should occur after the _scan_line__fallback that looks for a move with the lot_name. I think you are good to include the barcode in the arguments and move the fallback here before

jbaudoux (changes requested):
So far, very good. I like that it's not bound to SO line to count the returned quantity. Then it works also for bom products.

If you can fix the few remarks done here, then it should be ready to merge.

shopfloor_reception/services/reception.py
```diff
@@ -302,91 +322,165 @@ def _scan_document__by_picking(self, barcode):
...
-                    )
-                    == -1
-                )
+        origin_moves = search.origin_move_from_scan(picking.origin)
+        origin_moves = origin_moves.filtered(
+            lambda m: m.picking_id.picking_type_code == "outgoing"
+        )
+        origin_moves_for_product = origin_moves.filtered(
+            lambda m: m.product_id == product
+        )
```
jbaudoux: This is inefficient. Give a domain to origin_move_from_scan

shopfloor_reception/services/reception.py
```diff
@@ -302,91 +322,165 @@ def _scan_document__by_picking(self, barcode):
...
-        if not move:
-            line = picking.move_line_ids.filtered(
-                lambda l: not l.lot_id
-                and not l.lot_name
-                and l.product_id == lot.product_id
+        origin_moves = search.origin_move_from_scan(picking.origin)
+        origin_moves.filtered(lambda m: m.picking_id.picking_type_code == "outgoing")
+        origin_moves_for_packaging = origin_moves.filtered(
+            lambda m: packaging in m.product_id.packaging_ids
+        )
```
jbaudoux: inefficient, give a domain to the search

shopfloor_reception/services/reception.py
```diff
@@ -302,91 +322,165 @@ def _scan_document__by_picking(self, barcode):
...
+            # return_move._action_confirm()
+            return self._scan_line__find_or_create_line(picking, return_move)
+
+    def _scan_line__by_product(self, picking, product):
+        if self.work.menu.allow_return:
+            response = self._scan_line__by_product__return(picking, product)
+            if response:
+                return response
+        # Otherwise, the picking isn't a return, and should be a regular reception
+        move = picking.move_lines.filtered(lambda m: m.product_id == product)
```
jbaudoux: A move may already exist and should be used before creating a new return move. If you create a return in odoo backend, you need to be able to process the move even is the option is enabled. Otherwise, the move you create can get merged with the existing one and it will crash.

shopfloor_reception/services/reception.py
```diff
@@ -302,91 +322,165 @@ def _scan_document__by_picking(self, barcode):
...
+
+    def _scan_line__by_packaging(self, picking, packaging):
+        if self.work.menu.allow_return:
+            response = self._scan_line__by_packaging__return(picking, packaging)
+            if response:
+                return response
+        move = picking.move_lines.filtered(
+            lambda m: packaging in m.product_id.packaging_ids
+        )
+        message = self._check_move_available(move, "packaging")
```
jbaudoux: Same here.
A move may already exist and should be used before creating a new return move. If you create a return in odoo backend, you need to be able to process the move even is the option is enabled. Otherwise, the move you create can get merged with the existing one and it will crash.

jbaudoux (changes requested):

shopfloor/actions/stock.py
mmequignon: I added a check (not yet pushed) at return picking creation.
This will prevent to create a picking with empty locations (which are used in the rest of the flow).

mmequignon:
> Once the return picking is created in draft, if you cannot process the product, it would be nice to have a cancel button that deletes the draft picking and goes back to the start screen

question is, how do i detect the picking has to be unlinked?
allow_return = True, no move_lines, and state=draft?
ping @jbaudoux

shopfloor_reception/services/reception.py
mmequignon: We have to find origin mov,es, then filter origin_moves by product in order.
Otherwise we cannot know if picking is a return picking or a regular reception picking.
I used a domain to filter out non-outgoing moves though.

shopfloor_reception/services/reception.py
mmequignon: done

shopfloor_reception/services/reception.py
mmequignon: not sure to understand. let's discuss about it when you have time.

shopfloor/actions/message.py
```diff
@@ -550,6 +550,15 @@ def lot_in_multiple_sublocation(self, lot):
             ).format(lot=lot.name, product=lot.product_id.name),
         }
 
+    def no_default_location_on_picking_type(self):
+        return {
+            "message_type": "error",
+            "body": _(
+                "Picking types for this menu are missing "
```
```suggestion
                "Operation types for this menu are missing "
```

shopfloor/actions/message.py
```diff
@@ -550,6 +550,15 @@ def lot_in_multiple_sublocation(self, lot):
...
+                "default src and dest location."
```
```suggestion
                "default source and destination locations."
```

shopfloor/actions/stock.py
```diff
@@ -10,6 +13,98 @@ class StockAction(Component):
...
+            "location_dest_id": return_picking.location_dest_id.id,
+            "picking_type_id": return_type.id,
+            "warehouse_id": return_type.warehouse_id.id,
+            "origin_returned_move_id": origin_move.id,
+            "procure_method": "make_to_stock",
+        }
+
+    def _create_return_move__link_to_origin(self, return_move, origin_move):
+        move_orig_to_link = origin_move.move_dest_ids.mapped("returned_move_ids")
+        move_orig_to_link |= origin_move
```
sebalix, line 55: Questions, the first parameter is the returned move?
If the `origin_move` is the move going to Customers location (so the move to return), its `move_dest_ids` field should list the returned moves already (Customers -> Stock)? Why are we reading `returned_move_ids` from there?
A docstring explaining what we want to achieve here is welcome.
```suggestion
    def _create_return_move__link_to_origin(self, returned_move, origin_move):
        move_orig_to_link = origin_move.move_dest_ids.mapped("returned_move_ids")
        move_orig_to_link |= origin_move
```

shopfloor/services/service.py
sebalix: My comment still apply here.

shopfloor_reception/services/reception.py
```diff
@@ -48,6 +48,14 @@ class Reception(Component):
     _usage = "reception"
     _description = __doc__
 
+    def _check_picking_status(self, pickings):
```
sebalix, line 51: We should keep the original signature here:
```suggestion
    def _check_picking_status(self, pickings, states=None):
```
or
```suggestion
    def _check_picking_status(self, pickings, states=("assigned",)):
```

shopfloor_reception/services/reception.py
```diff
@@ -212,8 +226,15 @@ def _select_document_from_lot(self, lot):
...
         )
 
-    def _select_line(self, picking, line, move, increase_qty_done_by=1):
-        product = line.product_id
+    def _scan_line__find_or_create_line(self, picking, move, qty_done=1):
+        line = fields.first(
+            move.move_line_ids.filtered(
+                lambda l: (
+                    not l.result_package_id
+                    and l.shopfloor_user_id.id in [False, self.env.uid]
```
sebalix, line 255: I thought we had to ban the use of this deprecated `shopfloor_user_id` field?
cc @jbaudoux

shopfloor_reception/services/reception.py
```diff
@@ -222,54 +243,51 @@ def _select_line(self, picking, line, move, increase_qty_done_by=1):
...
     def _order_stock_picking(self):
         # We sort by scheduled date first. However, there might be a case
         # where two pickings have the exact same scheduled date.
         # In that case, we sort by id.
         return "scheduled_date ASC, id ASC"
 
-    def _scan_document__by_picking(self, barcode):
-        search = self._actions_for("search")
-        picking_filter_result = search.picking_from_scan(barcode, use_origin=True)
+    def _scan_document__by_picking(self, pickings, barcode):
```
sebalix, line 310: Seems that `barcode` is not used anymore in this method now. Is it expected as the method is named `_scan_document__[...]`?

shopfloor_reception/services/reception.py
```diff
@@ -630,6 +776,21 @@ def _response_for_select_dest_package(self, picking, line, message=None):
...
+        }
+
+    def _scan_document__get_find_kw(self):
+        return {
+            "picking": {"use_origin": True},
+            "delivered_picking": {"use_origin": True},
+        }
+
     def scan_document(self, barcode):
         """Scan a picking, a product or a packaging.
```
sebalix, line 821: I guess this docstring should be updated to explain the support of return flow.

sebalix (commented):
The support of return flow should be explained in the service class docstring (in addition to the current one to not mix them).
I also had difficulties to understand all these methods without docstring, their name doesn't always match what they are doing.

simahawk:
@mmequignon @jbaudoux what's the state here?

shopfloor_reception/services/reception.py
jbaudoux: This is the only scenario where we really need it :)

mmequignon:
> @mmequignon @jbaudoux what's the state here?

on it

shopfloor_reception/services/reception.py
mmequignon: yes, scan_document is the endpoint.
We already found a picking from the barcode.
To me it makes sense to name those methods `_<endpoint>__by_<model>`.
If we don't prefix by endpoint, we will end with a real mess, if for instance, a product can be scanned on every screen.

shopfloor/actions/stock.py
mmequignon: I see what you mean.
Origin move is the returned move. Its content has been returned.
It has generated a return move, which is return_move.
I can rename it "new_move" if you think it is clearer.

jbaudoux (approved):

shopfloor/actions/stock.py
mmequignon: Baaah, there's notes everywhere in this file about what it does, and where this has been copied from.
However, I can rename `return_move` to `new_move` if you think it's clearer.

shopfloor_reception/services/reception.py
mmequignon: no, this is a handler method, therefore, it is named `_<endpoint>__by_<model>`, but the scan is already done and we've found a record already.

jbaudoux:
@simahawk @sebalix This big one is now ready & green. Can you update your reviews?

shopfloor_reception/services/reception.py
```diff
@@ -324,91 +342,180 @@ def _scan_document__by_picking(self, barcode):
...
-                    lambda m: m.product_id == product
-                    and float_compare(
-                        m.quantity_done,
-                        m.product_uom_qty,
-                        precision_rounding=m.product_uom.rounding,
-                    )
-                    == -1
-                )
+    def _scan_document__by_origin_move(self, moves, barcode):
+        if self.work.menu.allow_return:
```
```suggestion
        if not self.work.menu.allow_return:
            return...
```
and you save indentation for the whole method.

shopfloor_reception/services/reception.py
```diff
@@ -846,6 +1014,51 @@ def _check_expiry_date(self, line):
...
+            "lot": self._set_quantity__by_lot,
+        }
+
+    def _set_quantity__by_barcode(
+        self, picking, selected_line, barcode, confirmation=False
+    ):
+        handlers_by_type = self._set_quantity__get_handlers_by_type()
+        search = self._actions_for("search")
+        search_result = search.find(barcode, handlers_by_type.keys())
+        # handler = handlers_by_type.get(search_result.type, self._set_quantity__fallback)
```
TDu: to remove

shopfloor_reception/services/reception.py
```diff
@@ -846,6 +1014,51 @@ def _check_expiry_date(self, line):
...
+        handler = handlers_by_type.get(search_result.type)
+        if handler:
```
TDu, line 1031: I don't understand if you inform `search.find` which models to search for with `handlers_by_type.keys()` how could the search return another type of record ?

shopfloor_reception/services/reception.py
TDu: But the barcode argument can be removed from the function declaration, no ?
Same question for the other following functions

shopfloor_reception/services/reception.py
```diff
@@ -324,91 +342,180 @@ def _scan_document__by_picking(self, barcode):
...
-        search = self._actions_for("search")
-        packaging = search.packaging_from_scan(barcode)
+    def _scan_document__by_packaging(self, packaging, barcode):
         if packaging:
             return self._select_document_from_packaging(packaging)
 
-    def _scan_document__by_lot(self, barcode):
-        return self._select_document_from_lot(barcode)
+    def _scan_document__by_lot(self, lot, barcode):
+        return self._select_document_from_lot(lot)
```
TDu, line 354: No need to check if there is a lot ? like the other `_scan_document__...` functions ?

shopfloor_reception/services/reception.py
```diff
@@ -324,91 +342,180 @@ def _scan_document__by_picking(self, barcode):
...
-            return self._select_line_from_packaging(picking, move, packaging)
+            if not return_move:
+                # It means that among all origin moves, none has been found with
+                # max qty to return being positive.
+                # Which means all lines have already been returned.
+                message = self.msg_store.move_already_returned()
+                return self._response_for_select_move(picking, message=message)
+            picking.action_confirm()
+            picking.action_assign()
+            # return_move._action_confirm()
```
TDu: to remove

shopfloor_reception/models/stock_picking.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    is_shopfloor_return = fields.Boolean()
```
jbaudoux: And move it to shopfloor

```suggestion
    is_shopfloor_create = fields.Boolean()
```

shopfloor/actions/stock.py
```diff
@@ -10,6 +13,98 @@ class StockAction(Component):
...
+    def _create_return_picking__get_vals(self, return_types, origin):
+        return_type = fields.first(return_types)
+        return {
+            "move_lines": [],
+            "picking_type_id": return_type.id,
+            "state": "draft",
+            "origin": origin,
+            "location_id": return_type.default_location_src_id.id,
+            "location_dest_id": return_type.default_location_dest_id.id,
+            "is_shopfloor_return": True,
```
jbaudoux:
```suggestion
            "is_shopfloor_create": True,
```

shopfloor_reception/services/reception.py
```diff
@@ -420,96 +527,137 @@ def _check_move_available(self, move, message_code="product"):
...
 
-    def _set_quantity__by_product(self, picking, selected_line, barcode):
-        search = self._actions_for("search")
-        product = search.product_from_scan(barcode)
-        if product:
-            if product.id != selected_line.product_id.id:
+    def _set_quantity__by_product(self, picking, selected_line, product):
+        # This is a general rule here. whether the return has been created from
+        # shopfloor or not, you cannot return more than what was shipped.
+        # Therefore, we cannot use the `is_shopfloor_return` here.
```
```suggestion
        # Therefore, we cannot use the `is_shopfloor_create` here.
```

shopfloor_reception/services/reception.py
```diff
@@ -420,96 +527,137 @@ def _check_move_available(self, move, message_code="product"):
...
-        search = self._actions_for("search")
-        packaging = search.packaging_from_scan(barcode)
-        if packaging:
-            if packaging.product_id.id != selected_line.product_id.id:
+        return response
+
+    def _set_quantity__by_packaging(self, picking, selected_line, packaging):
+        # This is a general rule here. whether the return has been created from
+        # shopfloor or not, you cannot return more than what was shipped.
+        # Therefore, we cannot use the `is_shopfloor_return` here.
```
```suggestion
        # Therefore, we cannot use the `is_shopfloor_create` here.
```

shopfloor_reception/services/reception.py
```diff
@@ -756,6 +918,12 @@ def done_action(self, picking_id, confirmation=False):
                 picking, message=self.msg_store.transfer_no_qty_done()
             )
         if not confirmation:
+            # Do not create a backorder if this is a shopfloor return.
+            if picking.is_shopfloor_return and self.work.menu.allow_return:
```
jbaudoux: We don't create backorder because it was created by shopfloor. We need this concept in other scenario where allow_move_create
```suggestion
            if picking.is_shopfloor_create and self.work.menu.allow_return:
```

jbaudoux (changes requested):
Behaving differently if the picking was created by shopfloor or not is a concept we need globally in shopfloor. Reading your comments on `is_shopfloor_return`, what you want to know is if the picking was created by shopfloor, not that it is a return. You put some warning about this. So let's rename this field to `is_shopfloor_create`.
We need this field for solving the same issue addressed here https://github.com/OCA/wms/pull/655

shopfloor_reception/services/reception.py
JuMiSanAr: I'm about to open a PR for reception where I rename this field, just in case.

jbaudoux (approved):
Approving as my remarks will be tackled by a follow-up PR

simahawk:
@mmequignon pls rebase and we can close this one

shopfloor_reception/models/stock_picking.py
simahawk: if you want to rename a field is better to do it before merging

jbaudoux:
@mmequignon What about integrating here the last 2 commits of https://github.com/OCA/wms/pull/658 as those are fixes of this PR?

shopfloor_reception/models/stock_picking.py
jbaudoux: Fixed in https://github.com/OCA/wms/pull/658
Idd, would be better to integrate here before merging

mmequignon:
> @mmequignon What about integrating here the last 2 commits of #658 as those are fixes of this PR?

I'll cherry pick those

shopfloor_reception/services/reception.py
mmequignon: it returns "none" when no record is found.
What you can see is the "fallback" behavior, when the barcode matches nothing valid for this `set_quantity` step

shopfloor_reception/models/stock_picking.py
jbaudoux: cherry-picked here, so resolving this

shopfloor_reception/services/reception.py
TDu: ok

TDu (approved):
LG about my question about the barcode that can be removed from a function declaration, it seems that it is used as origin on `create_return_picking`

mmequignon:
> /ocabot merge major

yeaaaah !

PR #561, opened 2023-02-07
shopfloor: fix location content transfer lines selected

TDu:
A move line can be in ready state, but the related picking not. This happens when the shipping policy on the pciking is set to `one` (When all products are ready).

After this change, move lines from a picking not ready, will not be offered to work on by the location content transfer scenario.


sebalix (approved):
LG

jbaudoux (approved):

PR #563, opened 2023-02-08
[14.0][FIX] shopfloor: fix zone picking package as substitue on lines

TDu:
Use the content of the package (products) and not the product in the packaging description to search for possible move lines.

ref.: rau-122

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -589,10 +589,14 @@ class ZonePicking(Component):
         pack_location = package.location_id
         if pack_location and pack_location.is_sublocation_of(self.zone_location):
             # Check if the package selected can be a substitute on a move line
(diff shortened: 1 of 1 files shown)


simahawk (approved):
LG

jbaudoux (approved):

jbaudoux (changes requested):
We need to collect all valid move lines, you cannot stop at the first product.

This is important as after this test, either you show a list of lines to choose from or you open the single valid one. So we need to know if there are many or one

TDu:
> We need to collect all valid move lines, you cannot stop at the first product.
> 
> This is important as after this test, either you show a list of lines to choose from or you open the single valid one. So we need to know if there are many or one

done

jbaudoux (approved):
Thanks. LG now

sebalix (approved):

PR #567, opened 2023-02-10
[14.0] shopfloor_reception: fix picking waiting if auto post with many users

JuMiSanAr:


jbaudoux (changes requested):
This is a FIXME on top of another FIXME.

Here is resolution of the underlying issue:
https://github.com/OCA/wms/pull/571

You can close this PR in favor of https://github.com/OCA/wms/pull/571

PR #569, opened 2023-02-13
[14.0] shopfloor: checkout scan change lot

simahawk:
[shopfloor, checkout: allow change lot on line select](https://github.com/OCA/wms/commit/0ff8111742482165230f5b3c9a1a3d45212c8bf3)

When it comes to select a line, scanning the wrong lot will invite the
user to scan it again to change it on the relevant line only if there is
one sharing the same product.

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -236,21 +236,19 @@ class Checkout(Component):
         return self._select_picking(pickings, "select_document")
 
     def _select_document_from_package(self, package, **kw):
-        pickings = package.move_line_ids.filtered(
+        picking = package.move_line_ids.filtered(
             lambda ml: ml.state not in ("cancel", "done")
         ).mapped("picking_id")
-        if len(pickings) > 1:
+        if len(picking) > 1:
             # Filter only if we find several pickings to narrow the
             # selection to one of the good type. If we have one picking
             # of the wrong type, it will be caught in _select_picking
             # with the proper error message.
             # Side note: rather unlikely to have several transfers ready
             # and moving the same things
-            pickings = pickings.filtered(
+            picking = picking.filtered(
                 lambda p: p.picking_type_id in self.picking_types
             )
-        if len(pickings) == 1:
-            picking = pickings
         return self._select_picking(picking, "select_document")
 
     def _select_document_from_product(self, product, line_domain=None, **kw):
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -537,10 +537,7 @@ class Checkout(Component):
         if not lines:
             return self._response_for_select_line(
                 picking,
-                message={
-                    "message_type": "error",
-                    "body": _("Lot is not in the current transfer."),
-                },
+                message=self.msg_store.lot_not_found_in_picking(),
             )
 
         # When lots are as units outside of packages, we can select them for
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -41,7 +41,7 @@ class Checkout(Component):
     _description = __doc__
 
     def _response_for_select_line(
-        self, picking, message=None, need_confirm_pack_all=False, need_confirm_lot=False
+        self, picking, message=None, need_confirm_pack_all=False, need_confirm_lot=None
     ):
         if all(line.shopfloor_checkout_done for line in picking.move_line_ids):
             return self._response_for_summary(picking, message=message)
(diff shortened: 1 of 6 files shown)


sebalix (approved):

sebalix:
NOTE: in a Pick/Pack/Ship this scenario is used normally during the pack step, as such this feature is not usable on Pack as Pick already reserved & moved the lot (no way to change the lot anymore, and Odoo is restricting the reservation of Pack moves based on the previous Pick move).
What we should do is to forbid to change a lot on a move line if there is an ancestor move linked to it. So we allow changing lot only on first moves to process (so always the Pick, being in a pick/pack/ship or pick/ship configuration or anything else).

For our own customer need it's OK as we are using this scenario and feature during the Pick.

simahawk:
@sebalix can you add this note as a TODO comment?

sebalix:
@simahawk I'm doing the patch in fact

shopfloor/services/checkout.py
```diff
@@ -231,21 +236,19 @@ def _select_document_from_location(self, location, **kw):
...
             # Filter only if we find several pickings to narrow the
             # selection to one of the good type. If we have one picking
             # of the wrong type, it will be caught in _select_picking
             # with the proper error message.
             # Side note: rather unlikely to have several transfers ready
             # and moving the same things
-            pickings = pickings.filtered(
+            picking = picking.filtered(
                 lambda p: p.picking_type_id in self.picking_types
             )
```
sebalix, line 251: IMO this selection of picking is broken here (even before these changes) if we use this checkout scenario on the first transfer to process (the Pick). This scenario has really been thought to be used on Pack, example:

- We have 2 Pick transfers assigned (same type), both reserving the same source package
- As these 2 Pick transfers reserve from the same source package, obviously they are not moving the whole pkg ("Move entire package" not enabled), so either there is no destination package (moving them as bulk), or a different destination package than the source one.
- The 2 Pack transfers that follow these 2 Picks get either no source package or the destination package set on the Pick.

Detecting the right picking record by scanning a package is not an issue if we are processing a Pack transfer, 3 possibilities:

- no package found (because no destination pkg was set on the Pick)
- package found because the whole package has been moved
- package found because a new destination package has bee created during the Pick

Now detecting the right package while working on the Pick isn't the same story: we have 2 Picks transfers reserving the same source package, which picking should be picked then? We should protect against this use case by taking the first I guess (I don't think we want to raise an error as we are at the beginning of the flow when processing the Pick :thinking: ):

```suggestion
            picking = fields.first(
                picking.filtered(
                    lambda p: p.picking_type_id in self.picking_types
                )
            )
```

jbaudoux: In the pickings scenario, we must always scan something that makes clear what is being processed.
In the checkout scenario, we decided to open the first matching. This was put in place when scanning raw products. So indeed, the suggestion of @sebalix to take the first is the right solution to be consistent.

However, the filtering on the picking type must always occur. This independently on the length of pickings. I would keep "pickings" instead of "picking" if you apply first in a later step, and removed the useless mapped.

shopfloor/services/checkout.py
```diff
@@ -533,15 +536,47 @@ def _select_lines_from_packaging(self, picking, selection_lines, packaging, **kw
...
+                lines_same_product = selection_lines.filtered(
+                    lambda l: (
+                        l.product_id == lot.product_id
+                        # We cannot change a lot on a move having ancestors.
+                        # That would mean we already picked up the wrong lot
+                        # on the previous move(s) and Odoo already restricts
+                        # the reservation based on the previous move(s).
+                        and not l.move_id.move_orig_ids
+                    )
+                )
```
jbaudoux, line 552: This deserves to be put out of the scenario. This is valid for all scenario

jbaudoux (changes requested):
_select_document_from_package needs fixing

shopfloor/services/checkout.py
sebalix: Indeed, this deserves a test too.

simahawk:
@sebalix @jbaudoux status?

jbaudoux:
> @sebalix @jbaudoux status?

My 2 remarks above are still valid:
* `_select_document_from_package` needs fixing
* the check on lot change for move ancestor is not specific to this scenario. We have several common checks in the service base model. Could it move there?

simahawk:
FTR I've just squashed commits, no rebase or other change

sebalix:
I rebased on top of 14.0 here: https://github.com/OCA/wms/pull/712

PR #571, opened 2023-02-13
shopfloor_reception: fix picking state after auto-posting

jbaudoux:
Picking was in waiting state with the remaining moves to receive instead of assigned state (available)

### DIFF
--- a/shopfloor/models/stock_move.py
+++ b/shopfloor/models/stock_move.py
@@ -39,8 +39,8 @@ class StockMove(models.Model):
                 )
             split_move_vals = self._split(qty_to_split)
             split_move = self.create(split_move_vals)
-            split_move._action_confirm(merge=False)
             split_move.move_line_ids = to_move
+            split_move._action_confirm(merge=False)
             split_move._recompute_state()
             split_move._action_assign()
             self._recompute_state()
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -975,19 +975,6 @@ class Reception(Component):
             selected_line, intersection=True
         )
         new_move.extract_and_action_done()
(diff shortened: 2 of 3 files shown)


shopfloor/models/stock_move.py
```diff
@@ -39,8 +39,8 @@ def split_other_move_lines(self, move_lines, intersection=False):
                 )
             split_move_vals = self._split(qty_to_split)
             split_move = self.create(split_move_vals)
-            split_move._action_confirm(merge=False)
             split_move.move_line_ids = to_move
+            split_move._action_confirm(merge=False)
```
simahawk, line 43: this is not a a fix in `shopfloor_reception`.  Can you split the commit pls?
Also, wondering: the change seems to make sense but... we didn't have any issue before?

simahawk (changes requested):

shopfloor/models/stock_move.py
jbaudoux: This is because the supplier location bypass the reservation and `action_assign` is called here https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_move.py#L1177
which causes also the creation of a new move line here
https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_move.py#L1324

jbaudoux: This issue is happening only in case of a reception

jbaudoux: I have done the split in 2 commits

jbaudoux:
@simahawk Can you update your review and merge it if now fine ?

simahawk:
@jbaudoux pls rebase

simahawk (approved):

PR #572, opened 2023-02-15
[16.0][ADD] stock_release_channel_shipment_advice_process_end_time

sbejaoui:
This module allows to set a delay time (in minutes) between the release channel process end time and the shipment advice arrival to the dock time.


sbejaoui:
cc/ @lmignon

stock_release_channel_shipment_advice_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,42 @@
...
+            return warehouse.release_channel_process_end_time_delay
+        return warehouse.company_id.release_channel_process_end_time_delay
+
+    def _get_shipment_advice_arrival_date(self, warehouse):
+        self.ensure_one()
+        process_end_time_delay = self._get_process_end_time_delay(warehouse)
+        hours = float_to_time(self.process_end_time)
+        return self.process_end_date.replace(
+            hour=hours.hour, minute=hours.minute
+        ) + timedelta(minutes=process_end_time_delay)
```
lmignon: @sbejaoui You should not modify the process_end_date... It's computed in the right way in `stock_release_channel_process_end_time`

stock_release_channel_shipment_advice_process_end_time/tests/test_stock_release_channel_shipment_advice_process_end_time.py
```diff
@@ -0,0 +1,72 @@
...
+class TestStockReleaseChannelShipmentAdviceProcessEndTime(ChannelReleaseCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        output_loc = cls.channel.picking_ids.move_ids.location_id
+        cls._update_qty_in_location(output_loc, cls.product1, 100)
+        cls._update_qty_in_location(output_loc, cls.product2, 100)
+        cls.channel.picking_ids.move_ids.write({"procure_method": "make_to_stock"})
+        cls.channel.picking_ids.action_assign()
+        cls.datetime_now = datetime.now()
```
lmignon: @sbejaoui you should use the `freezegun` python lib to freeze datetime and makes your tests result predictable. 
(This will avoid trouble for example if the tests is ran after the process_end_time)

stock_release_channel_shipment_advice_process_end_time/views/stock_release_channel.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <record model="ir.ui.view" id="stock_release_channel_form_view">
+        <field name="model">stock.release.channel</field>
+        <field
+            name="inherit_id"
+            ref="stock_release_channel_process_end_time.stock_release_channel_form_view"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='process_end_time']" position="after">
+                <field name="process_end_time_delay" />
```
lmignon: @sbejaoui It would be nice to display the word 'minutes' next to the input field ...
```suggestion
                <field name="process_end_time_delay" />
                <label for="process_end_time_delay" />
                    <div>
                        <field name="process_end_time_delay" class="oe_inline" />
                        <label string="minutes"  class="oe_inline" style="margin-left:5px" />
                   </div>
```

lmignon (changes requested):

stock_release_channel_shipment_advice_process_end_time/models/res_company.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    release_channel_process_end_time_delay = fields.Integer()
```
lmignon: @sbejaoui The name of this field doesn't reflect its usage....

stock_release_channel_shipment_advice_process_end_time/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+
+    _inherit = "res.config.settings"
+
+    release_channel_process_end_time_delay = fields.Integer(
```
lmignon: @sbejaoui Into the ui we've a text to describe the meaning of this field but the name doesn't reflect its usage

stock_release_channel_shipment_advice_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,37 @@
...
+
+from datetime import timedelta
+
+from odoo import api, fields, models
+
+
+class StockReleaseChannel(models.Model):
+
+    _inherit = "stock.release.channel"
+    process_end_time_delay = fields.Integer(
```
lmignon: As for the previous comment, this name doesn't reflect its usage... It should be renamed and a help text would be appreciated.

stock_release_channel_shipment_advice_process_end_time/models/stock_warehouse.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class StockWarehouse(models.Model):
+
+    _inherit = "stock.warehouse"
+
+    release_channel_process_end_time_delay = fields.Integer(
```
lmignon: @sbejaoui same as previous comment

stock_release_channel_shipment_advice_process_end_time/views/stock_release_channel.xml
```diff
@@ -0,0 +1,25 @@
...
+<odoo>
+
+    <record model="ir.ui.view" id="stock_release_channel_form_view">
+        <field name="model">stock.release.channel</field>
+        <field
+            name="inherit_id"
+            ref="stock_release_channel_process_end_time.stock_release_channel_form_view"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='process_end_time']" position="after">
```
lmignon: This field should be put after the shipment planning method and visible only if a method is selected... It could be useful to put fields related to the shipment advice planner into a dedicated group...

lmignon (changes requested):
@sbejaoui cosmetic changes are required to ensure a good UX

stock_release_channel_shipment_advice_process_end_time/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Release Channel Shipment Advice Process End Time",
+    "summary": """
+        c""",
```
jbaudoux: ?

sbejaoui: my bad

stock_release_channel_shipment_advice_process_end_time/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    release_channel_shipment_advice_arrival_delay = fields.Integer(
+        help="Global default value for the delay between the release channel process "
+        "end time and the arrival of shipments to the dock."
+    )
```
jbaudoux: I find this overkill. Having it configured on the warehouse is enough.

The stock_release_channel should be first made multi-company aware. It is useless to define a company setting while the model needing this is not bound to a company. There is a puzzle piece missing

I would drop this company setting

stock_release_channel_shipment_advice_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,43 @@
...
+        help="The delay between the release channel process end time and the arrival "
+        "of shipments to the dock.",
+    )
+
+    @api.depends("warehouse_id")
+    def _compute_shipment_advice_arrival_delay(self):
+        for rec in self:
+            if not rec.shipment_advice_arrival_delay:
+                rec.shipment_advice_arrival_delay = (
+                    rec.warehouse_id.release_channel_shipment_advice_arrival_delay
```
jbaudoux, line 68: or company setting if you want to keep it

stock_release_channel_shipment_advice_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _get_shipment_advice_arrival_delay(self, warehouse):
+        self.ensure_one()
+        if self.shipment_advice_arrival_delay:
+            return self.shipment_advice_arrival_delay
+        if not warehouse:
+            return 0
+        if warehouse.release_channel_shipment_advice_arrival_delay:
+            return warehouse.release_channel_shipment_advice_arrival_delay
+        return warehouse.company_id.release_channel_shipment_advice_arrival_delay
```
jbaudoux: There is a field set on this model. We must always use that value. If I configure 0, I should not fallback to default setting. This block should be removed

jbaudoux (changes requested):
Make it simpler

stock_release_channel_shipment_advice_process_end_time/models/res_company.py
jbaudoux: Nevertheless, ideally also add in another PR the missing company field on stock release channel + multi-company rules.
Even if not used here. This was just missing in the past.

stock_release_channel_shipment_advice_process_end_time/models/stock_release_channel.py
sbejaoui: thanks @jbaudoux , it make sense to keep this option at the warehouse level. I will make the changes.

stock_release_channel_shipment_advice_process_end_time/models/stock_release_channel.py
```diff
@@ -0,0 +1,41 @@
...
+                    rec.warehouse_id.release_channel_shipment_advice_arrival_delay
+                )
+
+    def _get_shipment_advice_arrival_delay(self, warehouse):
+        self.ensure_one()
+        if self.shipment_advice_arrival_delay:
+            return self.shipment_advice_arrival_delay
+        if not warehouse:
+            return 0
+        return warehouse.release_channel_shipment_advice_arrival_delay
```
jbaudoux: As I said before, I would drop this because shipment_advice_arrival_delay is computed store that already defaults on warehouse.release_channel_shipment_advice_arrival_delay.
So if the value was set intentionally to 0 on shipment_advice_arrival_delay, you have to use that value, and not the default value

stock_release_channel_shipment_advice_process_end_time/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+To set the delay time you have three options:
+
+1. Go to inventory settings and set a global delay time for all channels.
+2. Set a specific delay by warehouse.
+3. Set the delay at the release channel.
```
jbaudoux:
```suggestion
1. Set a specific default delay on the warehouse.
2. Set the delay at the release channel.
```

jbaudoux (changes requested):

lmignon (approved):
LGTM

jbaudoux (approved):
Thanks. Looks good now and much simpler :)

sbejaoui:
@lmignon , @jbaudoux 

Can you check #706, I have simplified the calculation of arrival and departure dates, it should replace this module.

jbaudoux:
@lmignon All depends have been merged, you should drop the DON'T MERGE commit

lmignon:
@jbaudoux "DON'T MERGE" commit dropped.

lmignon:
@sbejaoui Can you fix this one?

sbejaoui:
> @sbejaoui Can you fix this one?

done

jbaudoux (approved):

lmignon (approved):

PR #573, opened 2023-02-16
[14.0] shopfloor: improve test method '_fill_stock_for_moves'

sebalix:
Before this change, we could not get different package/lot on moves sharing the same product. By adding the move itself in the key, we ensure to get as many lot/serial number than moves.

This is useful especially if you tuned the key fields used to group moves (see https://github.com/OCA/sale-workflow/pull/2262 that avoid to merge moves sharing the same product in all the chained moves, so number of ordered lines == number of pick/pack/out moves)

### DIFF
--- a/shopfloor/tests/common.py
+++ b/shopfloor/tests/common.py
@@ -200,7 +200,7 @@ class CommonCase(BaseCommonCase):
             else:
                 package = cls.env["stock.quant.package"].create({})
         for move in moves:
-            key = (move.product_id, location or move.location_id)
+            key = (move, location or move.location_id)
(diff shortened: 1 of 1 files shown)


simahawk (approved):

sebalix:
I won't pass time to check why some tests are failing, I'll call this method on each move individually instead to get my test working as expected:
```python
        for move in picking.move_lines:
            # Call _fill_stock_for_moves on each move to generate a different
            # lot even for moves sharing the same product.
            cls._fill_stock_for_moves(move, in_lot=True)
```

PR #574, opened 2023-02-16
[14.0] shopfloor_base: cross worker fix

simahawk:
Depends on:

- [x] https://github.com/OCA/web-api/pull/3

With this change sf apps can be created and made available on the fly w/out restarting when working in cross worker envs.

TODO
- [x] add mig to enable existing apps on upgrade
- [x] remove issue from roadmap


sebalix (approved):
A bit too technical to understand everything, the code LG... now as it's working already on prod, approving :)

PR #575, opened 2023-02-18
[14.0][IMP] shopfloor: location content transfer find_work multiple user

mt-software-de:
Before:
Also pickings were found which are already assigned to a user This leaded to two problems:
- if the qty done was already set to the max qty by an other user you always got the error no work
- or you got the same location and the same move lines as the user before

Now:
You only get pickings whitout a assigned user

### DIFF
--- a/shopfloor/services/location_content_transfer.py
+++ b/shopfloor/services/location_content_transfer.py
@@ -1,5 +1,6 @@
 # Copyright 2020-2021 Camptocamp SA (http://www.camptocamp.com)
 # Copyright 2020-2022 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import _, fields
 
@@ -440,8 +441,7 @@ class LocationContentTransfer(Component):
 
     def _lock_lines(self, lines):
         """Lock move lines"""
-        sql = "SELECT id FROM %s WHERE ID IN %%s FOR UPDATE" % lines._table
-        self.env.cr.execute(sql, (tuple(lines.ids),), log_exceptions=False)
+        self._actions_for("lock").for_update(lines)
 
     def set_destination_all(self, location_id, barcode, confirmation=False):
         """Scan destination location for all the moves of the location
--- a/shopfloor/services/location_content_transfer.py
+++ b/shopfloor/services/location_content_transfer.py
@@ -51,6 +51,8 @@ class LocationContentTransfer(Component):
     _usage = "location_content_transfer"
     _description = __doc__
 
+    _advisory_lock_find_work = "location_content_transfer_find_work"
+
     def _response_for_start(self, message=None, popup=None):
         """Transition to the 'start' or 'get_work' state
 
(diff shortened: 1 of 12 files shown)


mt-software-de:
@jbaudoux this should solve the problem that the users getting always the same location until the previous picking isn't done

jbaudoux (approved):
Thanks for this fix.
To make it safe, there should also be a lock at the beginning of `_find_location_to_work_from` like here https://github.com/OCA/wms/blob/14.0/shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py#L59

mt-software-de:
> Thanks for this fix. To make it safe, there should also be a lock at the beginning of `_find_location_to_work_from` like here https://github.com/OCA/wms/blob/14.0/shopfloor_batch_automatic_creation/actions/picking_batch_auto_create.py#L59

done

shopfloor/services/location_content_transfer.py
```diff
@@ -245,18 +247,27 @@ def _create_moves_from_location(self, location):
...
-        )
-        move_lines = next_picking.move_line_ids.filtered(
-            lambda line: line.qty_done < line.product_uom_qty
-        )
-        return fields.first(move_lines).location_id
+        location = self.env["stock.location"]
+        offset = -1
+        while not location:
+            offset += 1
+            next_picking = self.env["stock.picking"].search(
```
simahawk: wouldn't be better to search them all just once and slice them w/ the offset?

shopfloor_base/utils.py
```diff
@@ -65,3 +68,17 @@ def _get_running_env():
 
 
 RUNNING_ENV = _get_running_env()
+
+
+def _lock(cr, name):
```
simahawk: I would move this to an action. We can have `self._actions_for("lock").advisory(name)`.

simahawk (changes requested):

shopfloor_base/utils.py
simahawk: we can have `for_update` too and replace `_lock_lines` in cluster_picking and location_content_transfer

mt-software-de: Thanks for the input i wasn't aware of  the actions, this a better place to put it in.
I created the new action and replaced the calls

shopfloor/services/location_content_transfer.py
```diff
@@ -273,6 +280,7 @@ def find_work(self):
         response = self._recover_started_picking()
         if response:
             return response
+        self._actions_for("lock").advisory("find_work")
```
jbaudoux: Can the advisory lock name move to a variable or method? And I would give it a more unique name like location_content_transfer_find_work

shopfloor_base/actions/lock.py
```diff
@@ -0,0 +1,32 @@
...
+        """
+        hasher = hashlib.sha1(str(name).encode())
+        # pg_lock accepts an int8 so we build an hash composed with
+        # contextual information and we throw away some bits
+        int_lock = struct.unpack("q", hasher.digest()[:8])
+
+        self.env.cr.execute("SELECT pg_advisory_xact_lock(%s);", (int_lock,))
+        self.env.cr.fetchone()[0]
+
+    def for_update(self, table, ids, log_exceptions=False):
```
jbaudoux: I would have given the recordset in argument instead of table & ids

shopfloor_base/actions/lock.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 Michael Tietz (MT Software) <mtietz@mt-software.de>
```
simahawk, line 1:
```suggestion
# Copyright 2020 Camptocamp
# Copyright 2022 Michael Tietz (MT Software) <mtietz@mt-software.de>
```

simahawk (commented):
LG really good, thanks.
One nitpicking thing: can you split the commit message pls?
Eg:
```
shopfloor_base: add actions for locking records

* advisory
* for_update
```

```
shopfloor: use lock actions
```

```
shopfloor: fix location content transfer find_work

Before:
Also pickings were found which are already assigned to a user
This leaded to two problems:
- if the qty done was already set to the max qty
  by an other user you always got the error no work
- or you got the same location and the same move lines as the user before

Now:
You only get pickings whithout a assigned user
```
The third one can contain the locking change for find_work, of course.

simahawk (approved):
Thanks for your improvements and your collaboration :+1:

PR #576, opened 2023-02-20
[14.0] shopfloor_reception: auto post fixes

JuMiSanAr:
Some fixes for the reception scenario:

* Auto posting wasn't working when scanning a package in set_quantity.
* "View" locations are not allowed anymore.
* Small change in the frontend for set_destination, we should display the location name.
* Prevent singleton error when multiple moves contain the same product / packaging.

ref: rau-133

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -358,7 +358,7 @@ class Reception(Component):
         if package:
             dest_location = selected_line.location_dest_id
             child_locations = self.env["stock.location"].search(
-                [("id", "child_of", dest_location.id)]
+                [("id", "child_of", dest_location.id), ("usage", "!=", "view")]
             )
             pack_location = package.location_id
             if pack_location:
@@ -889,11 +889,11 @@ class Reception(Component):
         location = search.location_from_scan(location_name)
         move_dest_location = selected_line.location_dest_id
         move_child_locations = self.env["stock.location"].search(
-            [("id", "child_of", move_dest_location.id)]
+            [("id", "child_of", move_dest_location.id), ("usage", "!=", "view")]
         )
         pick_type_dest_location = picking.picking_type_id.default_location_dest_id
         pick_type_child_locations = self.env["stock.location"].search(
-            [("id", "child_of", pick_type_dest_location.id)]
+            [("id", "child_of", pick_type_dest_location.id), ("usage", "!=", "view")]
         )
         if location not in move_child_locations | pick_type_child_locations:
             return self._response_for_set_destination(
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -5,6 +5,7 @@
 import pytz
 
 from odoo import fields
+from odoo.tools import float_compare
 
 from odoo.addons.base_rest.components.service import to_int
 from odoo.addons.component.core import Component
@@ -291,7 +292,17 @@ class Reception(Component):
         search = self._actions_for("search")
         product = search.product_from_scan(barcode)
         if product:
-            move = picking.move_lines.filtered(lambda m: m.product_id == product)
+            move = fields.first(
+                picking.move_lines.filtered(
+                    lambda m: m.product_id == product
+                    and float_compare(
+                        m.quantity_done,
+                        m.product_uom_qty,
+                        precision_rounding=m.product_uom.rounding,
+                    )
+                    == -1
+                )
+            )
             message = self._check_move_available(move, "product")
             if message:
                 return self._response_for_select_move(
@@ -304,8 +315,16 @@ class Reception(Component):
         search = self._actions_for("search")
         packaging = search.packaging_from_scan(barcode)
         if packaging:
-            move = picking.move_lines.filtered(
-                lambda m: packaging in m.product_id.packaging_ids
+            move = fields.first(
+                picking.move_lines.filtered(
+                    lambda m: packaging in m.product_id.packaging_ids
+                    and float_compare(
+                        m.quantity_done,
+                        m.product_uom_qty,
+                        precision_rounding=m.product_uom.rounding,
+                    )
+                    == -1
+                )
             )
             message = self._check_move_available(move, "packaging")
             if message:
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -958,6 +958,10 @@ class Reception(Component):
         search = self._actions_for("search")
 
         location = search.location_from_scan(location_name)
+        if not location:
+            return self._response_for_set_destination(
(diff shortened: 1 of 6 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -370,11 +370,30 @@ def _set_quantity__by_package(self, picking, selected_line, barcode):
                         picking, selected_line, message=message
                     )
                 else:
+                    quantity = selected_line.qty_done
+                    new_line, qty_check = selected_line._split_qty_to_be_done(
```
mmequignon: I cannot find where new_line is used?

shopfloor_reception/services/reception.py
```diff
@@ -888,23 +907,21 @@ def set_destination(
...
+            move_dest_location.parent_path.startswith(location.parent_path)
+            or pick_type_dest_location.parent_path.startswith(location.parent_path)
+        ):
             return self._response_for_set_destination(
                 picking,
                 selected_line,
                 message=self.msg_store.dest_location_not_allowed(),
             )
-        if location in move_child_locations:
+        if move_dest_location.parent_path.startswith(location.parent_path):
```
mmequignon: this is the other way around.
We want to check that location is a child of move_dest_location
```"a-b-c".startswith("a-b")```
In your case, you check that the selected location is a parent of the one set on the move.
To me those string methods doesn't makes things easier to understand.
I'd rather stick to the `child_of` search operator.

**Review by mmequignon — CHANGES_REQUESTED** [review]

File: shopfloor_reception/services/reception.py (hunk 2 above)
Comment (mmequignon) [reply shopfloor_reception/services/reception.py]: Also, parent_path is a simple field, which can be changed by hand by any user with the rights to do so.
I really wouldn't rely on this.
For this, I might have missed something, there might be a `_compute_parent_path` somewhere that I'm not aware of...

File: shopfloor_reception/services/reception.py (hunk 1 above)
Comment (JuMiSanAr) [reply shopfloor_reception/services/reception.py]: It's not.

File: shopfloor_reception/services/reception.py (hunk 2 above)
Comment (JuMiSanAr) [reply shopfloor_reception/services/reception.py]: I'm following the indications by @jbaudoux . It does make sense that it's the other way round, as you say. 

@jbaudoux thoughts?

File: shopfloor_reception/services/reception.py (hunk 1 above)
Comment (mmequignon) [reply shopfloor_reception/services/reception.py]: Then
```suggestion
                    __, qty_check = selected_line._split_qty_to_be_done(
```
No idea why pre-commit isn't saying anything about this.

JuMiSanAr: Hmmm I originally kept `new_line` because we were using _.("blabla") directly somewhere in the reception scenario, which caused issues with Python. I'm checking and it looks like we moved all those occurrences to the message.py file, so it should be ok to use the underscore. I'll update it.

shopfloor_reception/services/reception.py
jbaudoux: indeed should have been the other way around

shopfloor_reception/services/reception.py
mmequignon: it is a double underscore :)
Not the same thing as the single one that odoo uses for translating things :wink:

shopfloor_reception/services/reception.py
mmequignon: And it doesn't seems right to me, as this is a regular char field that might be edited.
Doesn't look safe to me.

shopfloor_reception/services/reception.py
JuMiSanAr: Well I remember there was a reason why we kept `new_line`. I've tested it with underscore and it seems to work fine so I'll keep it.

mmequignon: You mean double underscore?

JuMiSanAr: Yup

shopfloor_reception/services/reception.py
```diff
@@ -304,8 +309,11 @@ def _scan_line__by_packaging(self, picking, barcode):
         search = self._actions_for("search")
         packaging = search.packaging_from_scan(barcode)
         if packaging:
-            move = picking.move_lines.filtered(
-                lambda m: packaging in m.product_id.packaging_ids
+            move = fields.first(
+                picking.move_lines.filtered(
+                    lambda m: packaging in m.product_id.packaging_ids
+                    and m.quantity_done < m.product_uom_qty
```
mmequignon: use the force, use float_compare
```suggestion
                    and float_compare(m.quantity_done, m.product_uom_qty, rounding=m.product_uom.rounding) == -1
```

shopfloor_reception/services/reception.py
JuMiSanAr: As discussed, `child_of` itself uses `parent_path` so we can use the latter to simplify the check.

I've pushed a fixup commit to invert the location check as well.

jbaudoux (approved):

simahawk:
@mmequignon fine for you? @JuMiSanAr can you squash?

shopfloor_reception/services/reception.py
```diff
@@ -356,26 +365,44 @@ def _set_quantity__by_package(self, picking, selected_line, barcode):
...
                     )
-                else:
-                    # If the scanned package has a valid destination,
-                    # set both package and destination on the package,
-                    # and go back to the selection line screen
-                    selected_line.result_package_id = package
-                    selected_line.location_dest_id = pack_location
-                    return self._response_for_select_move(picking)
+                quantity = selected_line.qty_done
+                __, qty_check = selected_line._split_qty_to_be_done(
```
mmequignon, line 393: I'm kind of worried about that. If qty_check is "lesser", then qty_done has been altered on self, it created a new line, and we do nothing about this.

mmequignon (changes requested):

shopfloor_reception/services/reception.py
JuMiSanAr: If `qty_check` is "lesser", we proceed with the workflow and we post a new line, which is the expected behavior right?

mmequignon: @jbaudoux what do you think about this?

jbaudoux: Looks all good.
If qty done is 2 and todo is 3, then you split the line in 2. On the line with qty done = qty todo = 2, you then write the package & location. The other new line (qty done = 0 and qty todo = 1) will be available when you start over the process.

mmequignon (approved):

shopfloor_reception/services/reception.py
mmequignon: All good then :blush:

sebalix (approved):

PR #578, opened 2023-02-22
[14.0] shopfloor: cluster picking fix action done

simahawk:
Avoid calling '_action_done' on non assigned pickings.


sebalix (approved):

sebalix:
Note: the issue we got is mainly triggered by this override of `_action_done` in `mrp_subcontracting`:
https://github.com/odoo/odoo/blob/14.0/addons/mrp_subcontracting/models/stock_picking.py#L75
Transfers still `confirmed` will be processed by this loop and as there is no move lines the call to `min` will crash.

It probably needs to be fixed upstream (issue is still there in 16.0), but it's also a good thing to filter them out here in Shopfloor as well.

PR #581, opened 2023-02-24
[14.0][IMP] shopfloor_reception_mobile: Make states hookable

mt-software-de:


mt-software-de:
@simahawk could you please have a look

shopfloor_reception_mobile/static/src/scenario/reception_states.js
```diff
@@ -0,0 +1,295 @@
...
+
+/*
+    Define states for reception scenario.
+    @param $instance VueJS component instance
+*/
+export const reception_states = function ($instance) {
+    return {
+        init: {
+            enter: () => {
+                this.wait_call(this.odoo.call("start"));
```
simahawk, line 15: if you don't bind `this` to `$instance` you have to replace all this `this` w/ `$instance`.

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -454,295 +456,17 @@ const Reception = {
             }
             return picking.origin.includes(input);
         },
+        _get_states: function () {
+            return reception_states(this);
```
simahawk: can you try this? See comment below.
```suggestion
            return reception_states.bind(this)();
```

simahawk (changes requested):

shopfloor_reception_mobile/static/src/scenario/reception_states.js
mt-software-de: oh yes i forget to replace it

shopfloor_reception_mobile/static/src/scenario/reception.js
simahawk: Out of curiosity: have you tried this? If it works as I expect, is probably more clear.
Not a blocker though, as we've done the same for checkout_states.

mt-software-de: i tried it this should work

simahawk (approved):
LG, thanks

sebalix (approved):

PR #582, opened 2023-02-24
[14.0] shopfloor_mobile: improve list-item action handling

simahawk:
You can now decide if the action goes to the left or the right of the title.

ref: cos-3950

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/components/list.js
+++ b/shopfloor_mobile_base/static/wms/src/components/list.js
@@ -150,13 +150,28 @@ Vue.component("list", {
 
 Vue.component("list-item", {
     mixins: [ItemDetailMixin],
+    computed: {
+        show_title_action() {
+            return this.opts.on_title_action || this.opts.title_action_field;
+        },
+        title_action_position() {
+            return this.opts.title_action_position || "left";
+        },
+    },
+    methods: {
+        show_title_action_by_position(position) {
+            return this.show_title_action && position == this.title_action_position;
+        },
+    },
     template: `
     <div class="list-item">
         <v-list-item-title v-if="opts.show_title" :class="{'font-weight-bold mb-2': opts.loud_title}">
+            <list-item-title-action v-bind="$props" v-if="show_title_action_by_position('left')"/>
             <div class="item-counter" v-if="opts.showCounters">
                 <span>{{ index + 1 }} / {{ count }}</span>
             </div>
             <span v-text="_.result(record, opts.key_title)" />
+            <list-item-title-action v-bind="$props" v-if="show_title_action_by_position('right')"/>
         </v-list-item-title>
         <div class="details">
             <div v-for="(field, index) in options.fields" :class="'field-detail ' + field.path.replace('.', '-') + ' ' + (field.klass || '')">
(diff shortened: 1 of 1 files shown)


shopfloor_mobile_base/static/wms/src/components/list.js
```diff
@@ -184,3 +199,15 @@ Vue.component("list-item", {
...
+
+Vue.component("list-item-title-action", {
+    mixins: [ItemDetailMixin],
+    template: `
+<v-btn icon class="detail-action" :class="{'theme--dark': opts.theme_dark}" link
+        v-if=""
+        @click="opts.on_title_action ? opts.on_title_action(record): on_detail_action(record, opts.title_action_field)">
+    <btn-info-icon size="small" v-if="!opts.title_action_icon"/>
+    <v-icon v-if="opts.title_action_icon">{{ opts.title_action_icon }}</v-icon>
+</v-btn>
```
mmequignon: not sure I understand, why aren't you using the `show_title_action()` and `title_action_position()` methods?

shopfloor_mobile_base/static/wms/src/components/list.js
```diff
@@ -150,13 +150,28 @@ Vue.component("list", {
...
+    },
     template: `
     <div class="list-item">
         <v-list-item-title v-if="opts.show_title" :class="{'font-weight-bold mb-2': opts.loud_title}">
+            <list-item-title-action v-bind="$props" v-if="show_title_action_by_position('left')"/>
             <div class="item-counter" v-if="opts.showCounters">
                 <span>{{ index + 1 }} / {{ count }}</span>
             </div>
             <span v-text="_.result(record, opts.key_title)" />
+            <list-item-title-action v-bind="$props" v-if="show_title_action_by_position('right')"/>
```
JuMiSanAr, line 174: Are this detail action and the one that existed before (v-btn in this component) mutually exclusive? They do the same thing right?

If so, should we hide the v-btn if there's a title action?

shopfloor_mobile_base/static/wms/src/components/list.js
simahawk: Because they are not on this component. I render this component IF they say so. I forgot the empty v-if apparently

shopfloor_mobile_base/static/wms/src/components/list.js
simahawk: I don't get your point. I want to render this action on one side or the other.

JuMiSanAr: Yes, I was just thinking that the detail action you introduce in the title does the same thing as the action in the details part of this component:

```
    <v-btn icon class="detail-action"
            v-if="has_detail_action(record, field)"
            @click="on_detail_action(record, field, opts)">
        <btn-info-icon />
    </v-btn>
```

So maybe we could make them mutually exclusive to avoid making the frontend too crowded. But it's true that this all comes from the conditions in the props so there shouldn't be an issue.

simahawk: I see. That's another one and - in fact - I could reuse this component there too. I check.

simahawk: Ok, the scope and the behavior is different, I prefer to not touch it for now.

JuMiSanAr: Sounds good then!

JuMiSanAr (approved):

PR #584, opened 2023-02-24
[14.0][IMP] shopfloor: find_work

mt-software-de:
Before:
Only pickings were found which are printed = False
If the a split is happening because of different location in the move lines
Then the new picking will be assigned to the user and printed is set to true
https://github.com/OCA/wms/blob/c0c7c48a20263bb896a5fa9aa4cc65038de4d5c0/shopfloor/services/location_content_transfer.py#L291
If this user leaves the next screen (scan location) without clicking on cancel the picking is still assigned to the user
and not started.
If this user clicks now on Get work again, the picking is not recovered because of https://github.com/OCA/wms/blob/c0c7c48a20263bb896a5fa9aa4cc65038de4d5c0/shopfloor/services/location_content_transfer.py#L188

Because of the printed = True domain 
i will never find the picking again.

With this change all pickings would be found which are assigned to no or the current user. 
Also with the different order, the previous splitted picking would be the first one, which will be found.

### DIFF
--- a/shopfloor/services/location_content_transfer.py
+++ b/shopfloor/services/location_content_transfer.py
@@ -253,9 +253,9 @@ class LocationContentTransfer(Component):
             [
                 ("picking_type_id", "in", self.picking_types.ids),
                 ("state", "=", "assigned"),
-                ("printed", "=", False),
+                ("user_id", "in", (False, self.env.user.id)),
             ],
-            order="create_date",
+            order="user_id, priority desc, scheduled_date asc, id desc",
         )
 
         for next_picking in pickings:


mt-software-de:
@jbaudoux @simahawk 
this would be better then the change which was done in https://github.com/OCA/wms/pull/575

shopfloor/services/location_content_transfer.py
```diff
@@ -253,9 +253,11 @@ def _find_location_to_work_from(self):
             [
                 ("picking_type_id", "in", self.picking_types.ids),
                 ("state", "=", "assigned"),
-                ("printed", "=", False),
+                "|",
+                ("user_id", "=", False),
+                ("user_id", "=", self.env.user.id),
```
jbaudoux:
```suggestion
                ("user_id", "in", (False, self.env.uid)),
```

jbaudoux (approved):
Indeed.
This is consistent with what is done in the other scenarios.

shopfloor/services/location_content_transfer.py
```diff
@@ -253,9 +253,11 @@ def _find_location_to_work_from(self):
...
             ],
-            order="create_date",
+            order="user_id, create_date",
```
jbaudoux: As you touch this line, let's fix it also. The right sorting should be:
```suggestion
            order="user_id, priority desc, scheduled_date asc, id desc",
```

jbaudoux (approved):
Thanks

shopfloor/services/location_content_transfer.py
```diff
@@ -253,9 +253,9 @@ def _find_location_to_work_from(self):
             [
                 ("picking_type_id", "in", self.picking_types.ids),
                 ("state", "=", "assigned"),
-                ("printed", "=", False),
+                ("user_id", "in", (False, self.env.user.id)),
```
sebalix, line 256: I'm wondering if it's not a use case for te `=?` operator?
```suggestion
                ("user_id", "=?", self.env.user.id),
```

sebalix (approved):

shopfloor/services/location_content_transfer.py
jbaudoux: No, you cannot do that
`'=?' is a short-circuit that makes the term TRUE if right is None or False`

PR #596, opened 2023-03-06
[IMP] shopfloor_mobile_base: make item-detail-card clickable

mt-software-de:
Add possiblity to make the item-detail-card clickable
And to a box on the right of the item-detail-card which i also clickable

Shopfloor Reception: new config to make the move selectable by a user click

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/components/detail/detail_card.js
+++ b/shopfloor_mobile_base/static/wms/src/components/detail/detail_card.js
@@ -6,56 +6,83 @@
 
 import {ItemDetailMixin} from "/shopfloor_mobile_base/static/wms/src/components/detail/detail_mixin.js";
 
+const v_card_content = `
+    <v-card-title v-if="!opts.no_title">
+        <slot name="title">
+            <v-icon v-if="opts.title_icon" v-text="opts.title_icon" class="mr-2" />
+            <span v-text="_.result(record, opts.key_title)" />
+            <v-btn icon class="detail-action" :class="{'theme--dark': opts.theme_dark}" link
+                    v-if="opts.on_title_action || opts.title_action_field"
+                    @click="opts.on_title_action ? opts.on_title_action(record): on_detail_action(record, opts.title_action_field)">
+                <btn-info-icon v-if="!opts.title_action_icon"/>
+                <v-icon v-if="opts.title_action_icon">{{ opts.title_action_icon }}</v-icon>
+
+            </v-btn>
+        </slot>
+    </v-card-title>
+    <v-card-subtitle v-if="$slots.subtitle">
+        <slot name="subtitle"></slot>
+    </v-card-subtitle>
+    <slot name="details">
+        <!-- TODO: this loop is the same in list-item => make it a component -->
+        <v-card-text class="details" v-if="opts.fields.length">
+            <div v-for="(field, index) in opts.fields" :class="'field-detail ' + field.path.replace('.', '-') + ' ' + (field.klass || '')">
+                <div v-if="raw_value(record, field) || field.display_no_value">
+                    <span v-if="field.label" class="label">{{ field.label }}:</span>
+                    <component
+                        v-if="field.render_component"
+                        :is="field.render_component"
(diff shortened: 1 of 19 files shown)


jbaudoux (changes requested):
Can you extract first this generic feature of printing a product/product packaging/lot barcode. This should be available to perform from any scenario after clicking on the info button. And can you add the support to print the product packaging barcode?

Instead of making the reception list clickable, you should switch to use the other widget that is clickable. You can look at the checkout scenario, there the list is clickable. This should be done in a separate PR.

I don't like the printing from the reception product selection list as you don't know yet the quantity that is received and so how much to print. Instead of printing a label to select a line, you should select the line, set the quantity and then print the barcode.
Printing of the barcode could be handled as a generic feature on the product/package/lot if you know the supplier never provides you the barcode. It should be an exception in the reception process that the user requests printing the barcode, like unreadable barcode. Then you can click on info button and launch the printing. If you always have to print because you want to relabel everything, then you don't need a button, just always print.

cc @simahawk @JuMiSanAr @mmequignon

mt-software-de:
I removed the printing logic to a custom module, because after a discussion with @jbaudoux. 
There should be a automated printing (if configured) happening if a user selects the move by click.
But this will becoming an own PR.

shopfloor_mobile_base/static/wms/src/components/detail/detail_card.js
```diff
@@ -6,56 +6,83 @@
...
-                            </v-btn>
-                        </div>
-                    </div>
-                </v-card-text>
-            </slot>
-            <slot name="after_details"></slot>
-        </v-card>
+    template:
+        `
+    <div :class="wrapper_klass + (opts.action_box ? 'item-detail-card-box' : '')">
```
simahawk: hmm I'm not sure to understand this change.
All of these options could be handled through params.

shopfloor_reception/models/shopfloor_menu.py
```diff
@@ -17,13 +18,25 @@ class ShopfloorMenu(models.Model):
         default=False,
         help=AUTO_POST_LINE,
     )
-    auto_post_line_is_possible = fields.Boolean(
-        compute="_compute_auto_post_line_is_possible"
+    auto_post_line_is_possible = fields.Boolean(compute="_compute_scenario_options")
+
+    select_move_by_click = fields.Boolean(
```
simahawk, line 23: This is already possible by using manual-select widget instead of a simple list.
Please, have a look at `checkout`.

simahawk (changes requested):

shopfloor_reception/models/shopfloor_menu.py
mt-software-de: This would be needed the configure, if you can select a move without scanning the barcode.
I know that there is already the manual-select widget.

But in general. I wanted to add more possibilities to the item-detail-card, to make it easier to overwrite it by a other module without, overwriting the template. Because as it is right now, i think an overwrite of an existing scenario is not easy to implement or not nice.

Like you can see it here. 
https://github.com/OCA/wms/blob/c11fb21b1cd7ef5d125c13df82c248b4fc425ec2/shopfloor_checkout_package_measurement_mobile/static/src/scenario/checkout_package_measurement.js#L15

In our case we need a button on right sight of the item-detail-card at reception for only specific marked products. 

In fact the best would be to change, how/where the scenario template str is generated. I think it would be nice to move this out of javascript .

mt-software-de:
@simahawk how should we proceed here?
What we need is the possibility to make some shown move_lines clickable and some not clickable.
Depending on different values of the move_line. Also we need to show icon like object instead of the big checkbox like it is on the manual select. 

So we need either the ability to make a detail-card clickable or to disable the manual-select on certain cards.

Is it maybe better to use the manual-select for this PR and add this functions there?
cc @jbaudoux

mt-software-de:
This PR will be closed and the new approach is done in https://github.com/OCA/wms/pull/684 and https://github.com/OCA/wms/pull/683

PR #606, opened 2023-03-15
[14.0][FIX] shopfloor: Fix product/lot not found message

TDu:
Rephrasing the error message when a product or a lot has been found for a barcode. But there is no picking that matches it in the current context.

ref.: rau-138

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -412,13 +412,7 @@ class MessageAction(Component):
     def product_not_found_in_pickings(self):
         return {
             "message_type": "warning",
-            "body": _("No product found among current transfers."),
-        }
-
-    def product_not_found_in_picking(self):
-        return {
-            "message_type": "warning",
-            "body": _("Product not found in the current transfer."),
+            "body": _("No transfer found for this product."),
         }
 
     def x_not_found_or_already_in_dest_package(self, message_code):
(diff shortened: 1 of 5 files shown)


shopfloor/actions/message.py
```diff
@@ -481,16 +475,10 @@ def lot_multiple_packages_scan_package(self):
             "body": _("This lot is part of multiple packages, please scan a package."),
         }
 
-    def lot_not_found(self):
```
simahawk, line 497: removing methods w/o deprecation might break existing implementations

TDu: True.  I checked and it is not being used in this module or any merged related modules. I thought doing good by cleaning up.
What is the the proper way to do this, then ? Knowing that this module is still in it beta https://github.com/OCA/wms/blob/0c854c54458c932ea6ecdfa4d6e0f0bd5c5673ab/shopfloor/__manifest__.py#L10

simahawk: Yeah, we should change this to stable indeed :sweat_smile: 
However, in general, even if it's beta I tend to preserve backward compat when it's easy.
Now, this is a small thing and if you checked is not used anywhere... that's ok.
Let's move on :)

simahawk (approved):

jbaudoux (approved):

jbaudoux:
rebased

PR #607, opened 2023-03-20
shopfloor_mobile: improve qty picker calculation

JuMiSanAr:
When introducing a very large qty in the packaging-qty-picker, the screen would freeze as the calculations needed were very heavy.

With this change, we don't need to iterate many times to obtain the packaging quantities anymore.

ref: rau-140

### DIFF
--- a/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
+++ b/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
@@ -104,13 +104,9 @@ export var PackagingQtyPickerMixin = {
          */
         _qty_by_pkg: function (pkg_qty, qty) {
             const precision = this.unit_uom.rounding || 3;
-            let qty_per_pkg = 0;
-            // TODO: anything better to do like `float_compare`?
-            while (_.round(qty - pkg_qty, precision) >= 0.0) {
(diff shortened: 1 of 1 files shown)


shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -104,13 +104,9 @@ export var PackagingQtyPickerMixin = {
...
         _qty_by_pkg: function (pkg_qty, qty) {
             const precision = this.unit_uom.rounding || 3;
-            let qty_per_pkg = 0;
-            // TODO: anything better to do like `float_compare`?
-            while (_.round(qty - pkg_qty, precision) >= 0.0) {
-                qty -= pkg_qty;
-                qty_per_pkg += 1;
-            }
-            return [qty_per_pkg, qty];
+            const remainder = _.round(qty % pkg_qty, precision);
```
simahawk, line 107: remaining? reminder? which one? :)

JuMiSanAr: Remainder!

TDu (approved):

simahawk (approved):

PR #608, opened 2023-03-29
shopfloor_mobile: zone_picking always refresh list of lines

JuMiSanAr:
With this change, we ensure that the list of lines is always refreshed if two different users are working simultaneously on the same picking type.

ref: rau-95

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/zone_picking.js
+++ b/shopfloor_mobile/static/wms/src/scenario/zone_picking.js
@@ -601,9 +601,19 @@ const ZonePicking = {
                     },
                     on_select: (selected) => {
                         this.list_move_lines(selected.id);
+                        this.force_lines_refresh = true;
                     },
                 },
                 select_line: {
+                    enter: () => {
+                        if (!this.force_lines_refresh) {
+                            // Ensure that the list of lines is always refreshed
+                            // when landing on this screen.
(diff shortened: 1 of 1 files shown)


shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -601,9 +601,19 @@ const ZonePicking = {
                     },
                     on_select: (selected) => {
                         this.list_move_lines(selected.id);
+                        this.coming_from_select_picking_type = true;
```
```suggestion
                        this.force_lines_refresh = false;
```

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -601,9 +601,19 @@ const ZonePicking = {
...
                     },
                 },
                 select_line: {
+                    enter: () => {
+                        if (!this.coming_from_select_picking_type) {
```
```suggestion
                        if (this.force_line_refresh) {
```

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -796,6 +806,7 @@ const ZonePicking = {
                     },
                 },
             },
+            coming_from_select_picking_type: false,
```
```suggestion
            force_line_refresh: true
```

simahawk (commented):
WDYT?

jbaudoux (approved):

PR #609, opened 2023-03-29
[16.0][IMP] stock_available_to_promise_release: Remove unused fields

lmignon:
For some historical motivations, some informations from the partner are stored on stock.move(zip_code, city) and stock.picking (zip_code, city, state_id). We can imagines UC where these fields were used into the list views to filter pickings and/or moves to release according to one of these criteria. Nevertheless, this use of these fields is specific to some specific customers and is not justified by the way the addon works. Moreover, since these fields are defined as related on the res.partner table, any modification to one of these fields on your database where the module is installed will freeze it for a long time and most probably ends in a concurrent update exception since it will trigger an update on all the pickings and moves where this partner is referenced. Last but not least, if you install this addon on a large database with hundreds of thousands or millions of rows in the related tables, it will stuck your upgrade process for a very long time. If these fields are required for your customer it's safer to add them into a specific addon and put the necessary precautions to avoid any of the above issues

### DIFF
--- a/stock_available_to_promise_release/__init__.py
+++ b/stock_available_to_promise_release/__init__.py
@@ -1,2 +1,3 @@
 from . import models
 from . import wizards
+from .hooks import pre_init_hook
--- a/stock_available_to_promise_release/__manifest__.py
+++ b/stock_available_to_promise_release/__manifest__.py
@@ -24,4 +24,5 @@
     "license": "LGPL-3",
     "application": False,
     "development_status": "Beta",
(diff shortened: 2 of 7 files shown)


lmignon:
@rousseldenis @jbaudoux This one is ready to review and then merge :smirk:

jbaudoux (approved):

rousseldenis (approved):

lmignon:
@rousseldenis can you merge this one PLZ?

PR #614, opened 2023-04-13
[14.0] shopfloor: display location operations progress in delivery scenario

JuMiSanAr:
image

ref: rau-98

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -11,12 +11,36 @@ class DataAction(Component):
 
     @ensure_model("stock.location")
     def location(self, record, **kw):
-        return self._jsonify(
-            record.with_context(location=record.id), self._location_parser, **kw
-        )
+        parser = self._location_parser
+        data = self._jsonify(record.with_context(location=record.id), parser, **kw)
+        if "with_operation_progress" in kw:
+            operation_progress = self._get_location_operations_progress(record)
+            data.update({"operation_progress": operation_progress})
+        return data
 
     def locations(self, record, **kw):
         return self.location(record, multi=True)
+
+    def _get_location_operations_progress(self, location):
+        lines = self.env["stock.move.line"].search(
+            [
+                ("location_id", "=", location.id),
+                ("state", "in", ["partially_available", "assigned"]),
+                ("picking_id.state", "=", "assigned"),
+            ]
+        )
+        # operations_to_do = number of total operations that are pending for this location.
+        # operations_done = number of operations already done.
+        # A line with an assigned package counts as 1 operation.
+        operations_to_do = 0
+        operations_done = 0
+        for line in lines:
+            operations_done += line.qty_done if not line.package_id else 1
+            operations_to_do += line.product_uom_qty if not line.package_id else 1
+        return {
+            "done": operations_done,
+            "to_do": operations_to_do,
+        }
 
     @property
     def _location_parser(self):
--- a/shopfloor/actions/schema.py
+++ b/shopfloor/actions/schema.py
@@ -109,6 +109,14 @@ class ShopfloorSchemaAction(Component):
             "id": {"required": True, "type": "integer"},
             "name": {"type": "string", "nullable": False, "required": True},
             "barcode": {"type": "string", "nullable": True, "required": False},
+            "operation_progress": {
+                "type": "dict",
+                "required": False,
+                "schema": {
+                    "done": {"type": "float", "required": False},
+                    "to_do": {"type": "float", "required": False},
+                },
+            },
         }
 
     def packaging(self):
--- a/shopfloor_mobile/static/wms/src/scenario/delivery.js
+++ b/shopfloor_mobile/static/wms/src/scenario/delivery.js
@@ -29,6 +29,14 @@ const Delivery = {
                     :key="make_state_component_key(['picking-summary', 'detail-picking', state.data.picking.id])"
                     />
             </div>
+            <div v-if="state_is('deliver') && display_location_operations_progress()">
+                <item-detail-card
+                    :key="make_state_component_key(['deliver-operations-progress', state.data.sublocation.id])"
+                    :record="state.data.sublocation.operation_progress"
+                    :options="detail_card_location_operation_progress_options()"
+                    :card_color="detail_card_location_operation_progress_color()"
+                />
+            </div>
             <div v-if="state_is('manual_selection')">
                 <manual-select
                     class="with-progress-bar"
@@ -108,7 +116,7 @@ const Delivery = {
             if (this.current_location().id) {
                 return {title: "Working from location " + this.current_location().name};
             }
-            return "";
+            return {};
         },
         has_picking: function () {
             return !_.isEmpty(this.current_picking());
@@ -138,6 +146,40 @@ const Delivery = {
         },
         move_line_detail_fields: function () {
             return [{path: "package_src.name", klass: "loud"}];
+        },
+        display_location_operations_progress: function () {
+            const operation_progress = this._get_operation_progress_data();
+            if (operation_progress) {
+                return operation_progress.to_do > 0;
+            }
+        },
+        detail_card_location_operation_progress_options: function () {
+            return {
+                no_title: true,
+                fields: [
+                    {
+                        path: "to_do",
+                        klass: "loud",
+                        label: "Available",
+                    },
+                    {
+                        path: "done",
+                        klass: "loud",
+                        label: "Ongoing",
+                    },
+                ],
+            };
+        },
+        detail_card_location_operation_progress_color: function () {
+            const operation_progress = this._get_operation_progress_data();
+            if (operation_progress.done === operation_progress.to_do) {
+                return this.utils.colors.color_for("screen_step_done");
+            } else {
+                return this.utils.colors.color_for("screen_step_todo");
+            }
+        },
+        _get_operation_progress_data: function () {
+            return _.result(this.state, "data.sublocation.operation_progress", false);
         },
         manual_select_options: function () {
             return {
--- a/shopfloor/services/delivery.py
+++ b/shopfloor/services/delivery.py
@@ -52,7 +52,11 @@ class Delivery(Component):
                 "picking": self.data_detail.picking_detail(picking)
                 if picking
                 else None,
-                "sublocation": self.data.location(location) if location else None,
+                "sublocation": self.data.location(
+                    location, with_operation_progress=True
+                )
+                if location
+                else None,
             },
             message=message,
         )
--- a/shopfloor/tests/test_actions_data.py
+++ b/shopfloor/tests/test_actions_data.py
@@ -40,6 +40,22 @@ class ActionsDataCase(ActionsDataCaseBase):
         }
         self.assertDictEqual(data, expected)
 
+    def test_data_location_with_operation_progress(self):
+        location = self.stock_location
+        location.sudo().barcode = None
+        data = self.data.location(location, with_operation_progress=True)
+        self.assert_schema(self.schema.location(), data)
+        expected = {
+            "id": location.id,
+            "name": location.name,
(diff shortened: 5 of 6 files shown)


shopfloor/actions/data.py
```diff
@@ -11,13 +11,37 @@ class DataAction(Component):
...
+        )
+        # operations_to_do = number of total operations that are pending for this location.
+        # operations_done = number of operations already done.
+        # A line with an assigned package counts as 1 operation.
+        operations_to_do = 0
+        operations_done = 0
+        for line in location_pickings.move_line_ids:
+            operations_done += line.qty_done if not line.package_id else 1
+            operations_to_do += line.product_uom_qty if not line.package_id else 1
+        return (operations_done, operations_to_do)
```
simahawk: I'd return a dictionary instead

shopfloor/actions/data.py
```diff
@@ -11,13 +11,37 @@ class DataAction(Component):
 
     @ensure_model("stock.location")
     def location(self, record, **kw):
```
simahawk, line 13:
```suggestion
    def location(self, record, with_operations_progress_info=False, **kw):
```

shopfloor/actions/data.py
```diff
@@ -11,13 +11,37 @@ class DataAction(Component):
...
-        return self._jsonify(
-            record.with_context(location=record.id), self._location_parser, **kw
-        )
+        parser = self._location_parser
+        data = self._jsonify(record.with_context(location=record.id), parser, **kw)
+        if "with_operations_progress_info" in kw:
```
simahawk:
```suggestion
        if with_operations_progress_info:
```

shopfloor/actions/data.py
```diff
@@ -11,13 +11,37 @@ class DataAction(Component):
...
+            operations_done, operations_to_do = self._get_location_operations_progress(
+                record
+            )
+            data.update(
+                {
+                    "operations_done": operations_done,
+                    "operations_to_do": operations_to_do,
+                }
+            )
```
simahawk:
```suggestion
            progress_info = self._get_location_operations_progress(record)
            data.update(progress_info)
```

shopfloor/actions/schema.py
```diff
@@ -109,6 +109,8 @@ def location(self):
             "id": {"required": True, "type": "integer"},
             "name": {"type": "string", "nullable": False, "required": True},
             "barcode": {"type": "string", "nullable": True, "required": False},
+            "operations_done": {"type": "float", "required": False},
+            "operations_to_do": {"type": "float", "required": False},
```
simahawk: I'd wrap this into its own key with `done` and `todo` subkeys.
```suggestion
            "operation_info": {"type": "dict", "required": False, ...},
```

shopfloor_mobile/static/wms/src/scenario/delivery.js
```diff
@@ -13,6 +13,10 @@ const Delivery = {
         <Screen :screen_info="screen_info">
             <template v-slot:header>
                 <state-display-info :info="current_location_msg()" v-if="current_location()"/>
+                <state-display-info
```
simahawk: TBH I don't like the current design. It make info blocks confusing.
IMO we should have a specific block done w/ detail-card on top of the page.

Also, wondering: do we really need this inside location data? Can't we get this as a message in the response?

simahawk (changes requested):

shopfloor/actions/data.py
JuMiSanAr: I'm following the kw approach because this is what we did for the pickings data:

```    
def picking(self, record, **kw):
        parser = self._picking_parser
        # progress is a heavy computed field,
        # and it may reduce performance significatively
        # when dealing with a large number of pickings.
        # Thus, we make it optional.
        if "with_progress" in kw:
            parser.append("progress")
        return self._jsonify(record, parser, **kw)
```

We should try to keep it consistent, shouldn't we?

shopfloor_mobile/static/wms/src/scenario/delivery.js
JuMiSanAr: I've moved it to it's own detail card, and placed it below the searchbar (it didn't look good on top of the page):

image

I think it makes more sense to have this inside location data, in the end the messages we normally use them to display alerts / warnings / etc., which is not the case here.

shopfloor/actions/data.py
simahawk: Fair enough. In any case here we would just  "book" a specific key to be extracted from the generic kw.

shopfloor/actions/schema.py
```diff
@@ -109,6 +109,14 @@ def location(self):
             "id": {"required": True, "type": "integer"},
             "name": {"type": "string", "nullable": False, "required": True},
             "barcode": {"type": "string", "nullable": True, "required": False},
+            "progress_info": {
+                "type": "dict",
+                "required": False,
+                "schema": {
+                    "operations_done": {"type": "float", "required": False},
+                    "operations_to_do": {"type": "float", "required": False},
+                },
```
simahawk:
```suggestion
            "operation_progress": {
                "type": "dict",
                "required": False,
                "schema": {
                    "done": {"type": "float", "required": False},
                    "todo": {"type": "float", "required": False},
                },
```
not a blocker tho ;)

shopfloor/actions/data.py
```diff
@@ -11,13 +11,33 @@ class DataAction(Component):
 
     @ensure_model("stock.location")
     def location(self, record, **kw):
-        return self._jsonify(
-            record.with_context(location=record.id), self._location_parser, **kw
-        )
+        parser = self._location_parser
+        data = self._jsonify(record.with_context(location=record.id), parser, **kw)
+        if "with_operations_progress_info" in kw:
```
simahawk:
```suggestion
        if "with_operations_progress" in kw:
```

shopfloor_mobile/static/wms/src/scenario/delivery.js
```diff
@@ -139,6 +147,31 @@ const Delivery = {
...
+                "data.sublocation.progress_info",
+                false
+            );
+            if (progress_info) {
+                return (
+                    progress_info.operations_done && progress_info.operations_to_do > 0
+                );
+            }
+        },
+        detail_card_location_progress_info: function () {
```
simahawk:
```suggestion
        detail_card_location_progress_options: function () {
```

shopfloor_mobile/static/wms/src/scenario/delivery.js
```diff
@@ -29,6 +29,14 @@ const Delivery = {
                     :key="make_state_component_key(['picking-summary', 'detail-picking', state.data.picking.id])"
                     />
             </div>
+            <div v-if="state_is('deliver') && display_location_operations_progress()">
+                <item-detail-card
+                    :key="make_state_component_key(['deliver-operations-progress', state.data.sublocation.id])"
+                    :record="state.data.sublocation.progress_info"
+                    :options="detail_card_location_progress_info(state.data.sublocation.progress_info)"
+                    :card_color="utils.colors.color_for('screen_step_todo')"
```
simahawk: switch to "done" color if they are all done? (`todo == 0`)

simahawk:
can you update the screenshot in the description when done? :pray:

shopfloor/actions/data.py
```diff
@@ -11,13 +11,33 @@ class DataAction(Component):
...
+            operation_progress = self._get_location_operations_progress(record)
+            data.update({"operation_progress": operation_progress})
+        return data
 
     def locations(self, record, **kw):
         return self.location(record, multi=True)
 
+    def _get_location_operations_progress(self, location):
+        location_pickings = self.env["stock.picking"].search(
+            [("location_id", "=", location.id), ("state", "=", "assigned")]
```
jbaudoux: You need to collect the lines from the processed location. The location of the picking can be a parent of the processed location and cannot be used in the search.

stock.move.line search:
* location_id = location.id
* state in partially_available, assigned
* picking_id.state = assigned

shopfloor_mobile/static/wms/src/scenario/delivery.js
```diff
@@ -139,6 +147,37 @@ const Delivery = {
...
+        },
+        detail_card_location_operation_progress_options: function () {
+            return {
+                no_title: true,
+                fields: [
+                    {
+                        path: "done",
+                        klass: "loud",
+                        renderer: function (rec, field) {
+                            return `Operations done: ${rec.done} / ${rec.todo}`;
```
jbaudoux: Maybe?
```suggestion
                            return `Progress: ${rec.done} / ${rec.todo}`;
```

jbaudoux (changes requested):

jbaudoux (approved):

jbaudoux:

shopfloor/actions/data.py
```diff
@@ -11,13 +11,37 @@ class DataAction(Component):
...
+            ]
+        )
+        # operations_to_do = number of total operations that are pending for this location.
+        # operations_done = number of operations already done.
+        # A line with an assigned package counts as 1 operation.
+        operations_to_do = 0
+        operations_done = 0
+        for line in lines:
+            operations_done += line.qty_done if not line.package_id else 1
+            operations_to_do += line.product_uom_qty if not line.package_id else 1
```
simahawk, line 39: "nr of operations" means quantities?? :thinking:

shopfloor/actions/schema.py
```diff
@@ -109,6 +109,14 @@ def location(self):
             "id": {"required": True, "type": "integer"},
             "name": {"type": "string", "nullable": False, "required": True},
             "barcode": {"type": "string", "nullable": True, "required": False},
+            "operation_progress": {
+                "type": "dict",
+                "required": False,
+                "schema": {
+                    "done": {"type": "float", "required": False},
+                    "to_do": {"type": "float", "required": False},
```
simahawk, line 117:
```suggestion
                    "todo": {"type": "float", "required": False},
```

shopfloor_mobile/static/wms/src/scenario/delivery.js
```diff
@@ -29,6 +29,14 @@ const Delivery = {
                     :key="make_state_component_key(['picking-summary', 'detail-picking', state.data.picking.id])"
                     />
             </div>
+            <div v-if="state_is('deliver') && display_location_operations_progress()">
+                <item-detail-card
```
simahawk, line 33: can you add and option to `item-detail-card` to display fields inline? I think it's better for such cases as you save some screen space.

simahawk (changes requested):

shopfloor_mobile/static/wms/src/scenario/delivery.js
JuMiSanAr: In the beginning it was a one-liner, then it was decided we would keep it in two separate lines for clarity. What do you think?

simahawk: As long as they are readable (eg: proper separation w/ 2 columns of half width) I don't get why you would like burn more space...

JuMiSanAr: To me, it's much clearer like this (as an user I'd rather have one quantity above the other to easily see how much has been done out of the total). If the quantities get big, I think it helps the eye.
Specially in a screen which doesn't have many elements, like this one, where we don't need to save space.

jbaudoux: In most cases, ongoing will always be 0. I would display it only if != 0

JuMiSanAr: That's already the case (by default in this component, fields whose value is 0 are not displayed).
image

simahawk (approved):
Not yet convinced about few details but there's no strong reason to block ;)

PR #617, opened 2023-04-18
[14.0][FIX] stock_release_channel: _pickings_sort_key 

mt-software-de:
uses now as fallback the picking.scheduled_date field

The date_priority field of a stock.picking can be null, that's why with need a fallback field. The scheduled_date is a required field of a stock.move, which means it is also always computed on the a stock.picking

### DIFF
--- a/stock_release_channel/models/stock_release_channel.py
+++ b/stock_release_channel/models/stock_release_channel.py
@@ -649,7 +649,12 @@ class StockReleaseChannel(models.Model):
 
     @staticmethod
     def _pickings_sort_key(picking):
-        return (-int(picking.priority or 1), picking.date_priority, picking.id)
+        return (
+            -int(picking.priority or 1),
+            picking.scheduled_date,
+            picking.date_priority or picking.create_date,
+            picking.id,
+        )
 
     def _get_next_pickings(self):
         return getattr(self, "_get_next_pickings_{}".format(self.auto_release))()


mt-software-de:
@jbaudoux could you please review this one.

stock_release_channel/models/stock_release_channel.py
```diff
@@ -564,7 +564,11 @@ def get_action_picking_form(self):
 
     @staticmethod
     def _pickings_sort_key(picking):
-        return (-int(picking.priority or 1), picking.date_priority, picking.id)
+        return (
+            -int(picking.priority or 1),
+            picking.date_priority or picking.scheduled_date,
```
jbaudoux: I prefer this then
```suggestion
            picking.date_priority or picking.create_date,
```

mt-software-de: Why create_date instead of scheduled_date?

jbaudoux: Because both date_priority and create_date default to now() while scheduled_date is not
Nevertheless, date_priority should always have a value

mt-software-de: Yes they are both default to now, but also the scheduled_date. 
Because it gets computed from the date field of the moves. 
https://github.com/odoo/odoo/blob/e8834cb00b82e0bee7f6338aef940037a3e49471/addons/stock/models/stock_picking.py#LL291C8-L291C8
https://github.com/odoo/odoo/blob/e8834cb00b82e0bee7f6338aef940037a3e49471/addons/stock/models/stock_picking.py#L521
https://github.com/odoo/odoo/blob/e8834cb00b82e0bee7f6338aef940037a3e49471/addons/stock/models/stock_move.py#L39

Should we not process a picking/move which is created later, as an other picking/move, but with an earlier scheduled_date/date?

jbaudoux: It's sorting on pickings with scheduled_date in the past
Re-reading this, you are right to use the scheduled_date.
The date_priority is for the virtual reservation, not to sort for the next batch of release.
What about:
`(-int(picking.priority or 1), picking.scheduled_date, picking.date_priority or picking.create_date, picking.id)`

jbaudoux: cc @sebalix @lmignon

mt-software-de: Yes this is looking good, i change it like you mentioned.

jbaudoux (approved):

stock_release_channel/models/stock_release_channel.py
mt-software-de: ping @sebalix @lmignon

lmignon (approved):
All this makes sense.

PR #618, opened 2023-04-18
[14.0] shopfloor_checkout: ask to select child destination when marking as done

JuMiSanAr:
In the checkout scenario, when clicking mark as done, if the destination location has children locations we now show an extra screen asking for the destination location.
This location will serve as destination location for all the processed lines.

ref: rau-143

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -28,8 +28,8 @@ class DataAction(Component):
             data.update({"operation_progress": operation_progress})
         return data
 
-    def locations(self, record, **kw):
-        return self.location(record, multi=True)
+    def locations(self, records, **kw):
+        return [self.location(rec, **kw) for rec in records]
 
     @property
     def _location_parser(self):
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -71,6 +71,17 @@ class Checkout(Component):
             message=message,
         )
 
+    def _response_for_select_child_location(self, picking, message=None):
+        return self._response(
+            next_state="select_child_location",
+            data={
+                "picking": self._data_for_stock_picking(
+                    picking, done=True, with_lines=False, with_location=True
+                ),
+            },
+            message=message,
+        )
+
     def _response_for_select_document(self, message=None):
         data = {"restrict_scan_first": self.work.menu.scan_location_or_pack_first}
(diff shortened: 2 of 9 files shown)


shopfloor/actions/data.py
```diff
@@ -15,8 +15,8 @@ def location(self, record, **kw):
             record.with_context(location=record.id), self._location_parser, **kw
         )
 
-    def locations(self, record, **kw):
-        return self.location(record, multi=True)
+    def locations(self, records, **kw):
+        return [self.location(rec, **kw) for rec in records]
```
simahawk, line 32: this was a bug, isn't it? But I don't see its commit :wink:

shopfloor/services/checkout.py
```diff
@@ -1293,6 +1310,7 @@ def done(self, picking_id, confirmation=False):
         * summary: in case of error
         * select_document: after done, goes back to start
         * confirm_done: confirm a partial
+        * select_child_location: select a child location from the scanned location, if any
```
```suggestion
        * select_child_location_id: ID of a child location to use
```

shopfloor/services/checkout.py
```diff
@@ -1315,8 +1333,28 @@ def done(self, picking_id, confirmation=False):
...
+            for line in lines_done:
+                line.update({"location_dest_id": selected_child_location_id})
+        elif len(child_locations) > 0 and child_locations != dest_location:
+            return self._response_for_select_child_location(
+                picking,
+                child_locations,
+                message={
+                    "message_type": "warning",
+                    "body": _(
+                        "Select the destination location inside {}".format(
```
simahawk: Wrong interpolation of translated string. Also, it would be nice to move this msg to the message registry.

simahawk (changes requested):

jbaudoux (changes requested):
The selection should be like any other screen. You have to scan it.

Message in blue: Set destination location
Input barcode: Scan location
Box in yellow: picking destination location

image

TDu (approved):

jbaudoux (approved):
LG now

jbaudoux:

jbaudoux:
Tested, works well

jbaudoux:
@JuMiSanAr Can you rebase and fix merge conflict?

PR #619, opened 2023-04-19
[14.0][FIX] shopfloor_dangerous_goods_mobile: fix logo

JuMiSanAr:
The dangerous goods logo wasn't displayed correctly due to the fact that the content of the picking-select components changed and the CSS wasn't pointing to the new elements.

image
image

ref: cos-4038

### DIFF
--- a/shopfloor_dangerous_goods_mobile/static/wms/src/css/dangerous_goods.css
+++ b/shopfloor_dangerous_goods_mobile/static/wms/src/css/dangerous_goods.css
@@ -1,10 +1,16 @@
+.detail-picking-select .has-lq-products .record-name::after,
 .detail-picking-select .has-lq-products .v-list-item__title::after {
+    /* TODO: This PR fixes a bug that made the dangerous product logo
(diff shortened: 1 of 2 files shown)


shopfloor_dangerous_goods_mobile/static/wms/src/css/dangerous_goods.css
```diff
@@ -1,10 +1,12 @@
+.detail-picking-select .has-lq-products .record-name::after,
```
simahawk, line 1: hmm I fixed this... AFAIR `record-name` should not  be there anymore.
Can you check pls? And ping me if you don't have findings ;)

simahawk (approved):

shopfloor_dangerous_goods_mobile/static/wms/src/css/dangerous_goods.css
JuMiSanAr: As discussed, we still have `record-name` in the "picking-select-package-content". I added a TODO to remind us to update it in the future.

PR #620, opened 2023-04-19
[14.0] shopfloor: fix scan_destination not returning done qty

JuMiSanAr:
Endpoint "scan_location" was not returning any qty_done if certain conditions were met, which means that the qty-picker was being reset to 0.

ref: cos-4039

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -688,6 +688,7 @@ class ClusterPicking(Component):
             return self._response_for_scan_destination(
                 move_line,
                 message=self.msg_store.unable_to_pick_more(move_line.product_uom_qty),
+                qty_done=quantity,
             )
 
         search = self._actions_for("search")
@@ -716,6 +717,7 @@ class ClusterPicking(Component):
                         "The destination bin {} is not empty, please take another."
                     ).format(bin_package.name),
                 },
+                qty_done=quantity,
             )
         move_line.write({"qty_done": quantity, "result_package_id": bin_package.id})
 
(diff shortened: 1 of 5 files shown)


simahawk (approved):
seems legit

TDu (approved):

shopfloor/services/cluster_picking.py
```diff
@@ -992,13 +994,16 @@ def change_pack_lot(self, picking_batch_id, move_line_id, barcode):
...
                 return response
 
         package = search.package_from_scan(barcode)
         if package:
-            return change_package_lot.change_package(
+            response = change_package_lot.change_package(
                 move_line, package, response_ok_func, response_error_func
             )
+            response["data"]["scan_destination"].update({"qty_done": quantity})
+            return response
```
sebalix: Not sure it is correct, you expect the `response` to get the `scan_destination` state to set the `qty_done`, but the response could be returned by either `response_ok_func` or `response_error_func` here (respectively `self._response_for_scan_destination` and `self._response_for_change_pack_lot`).
Better to test first if `scan_destination` is present.

shopfloor/services/cluster_picking.py
```diff
@@ -992,13 +994,16 @@ def change_pack_lot(self, picking_batch_id, move_line_id, barcode):
                 move_line, lot, response_ok_func, response_error_func
             )
             if response:
+                response["data"]["scan_destination"].update({"qty_done": quantity})
```
sebalix: Same remark

shopfloor/services/cluster_picking.py
JuMiSanAr: Updated.

sebalix (approved):

PR #621, opened 2023-04-24
[14.0][IMP] shopfloor: zone picking handle complete mix package

TDu:
When a package has different product in it, it will be displayed with multiple move lines in the `select_line` step. And they will need to be move one after an other.

For such package, this change allows to move such package in one step.

ref.: rau-145

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -128,6 +128,10 @@ class DataAction(Component):
             "name",
             "shopfloor_weight:weight",
             ("package_storage_type_id:storage_type", ["id", "name"]),
+            (
+                "quant_ids:total_quantity",
+                lambda rec, fname: sum(rec.quant_ids.mapped("quantity")),
+            ),
         ]
 
     @property
--- a/shopfloor/actions/packaging.py
+++ b/shopfloor/actions/packaging.py
@@ -51,3 +51,13 @@ class PackagingAction(Component):
 
     def package_has_several_lots(self, package):
         return len(package.quant_ids.lot_id) > 1
+
+    def is_complete_mix_pack(self, package):
+        """Check if a package is mixed and completely reserved.
+
+        Will return true if the package has multiple distinct products and
+        all the package quantities are reserved.
+        """
+        return self.package_has_several_products(package) and all(
+            quant.quantity == quant.reserved_quantity for quant in package.quant_ids
+        )
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -161,6 +161,13 @@ class ZonePicking(Component):
     def _pick_pack_same_time(self):
         return self.work.menu.pick_pack_same_time
 
+    def _handle_complete_mix_pack(self, package):
+        packaging = self._actions_for("packaging")
+        return (
+            packaging.is_complete_mix_pack(package)
+            and not self.work.menu.no_prefill_qty
+        )
+
     def _response_for_start(self, message=None):
         zones = self.work.menu.picking_type_ids.mapped(
             "default_location_src_id.child_ids"
@@ -231,6 +238,9 @@ class ZonePicking(Component):
         data[
             "allow_alternative_destination_package"
         ] = self.work.menu.allow_alternative_destination_package
+        data["handle_complete_mix_pack"] = self._handle_complete_mix_pack(
+            move_line.package_id
+        )
         return self._response(
             next_state="set_line_destination", data=data, message=message
         )
@@ -349,11 +359,19 @@ class ZonePicking(Component):
             # to retrieve if location will be empty.
             # Maybe group lines by location and compute only once.
             move_line = self.env["stock.move.line"].browse(data_move_line["id"])
+            handle_complete_mix_pack = self._handle_complete_mix_pack(
+                move_line.package_id
+            )
+            data_move_line["handle_complete_mix_pack"] = handle_complete_mix_pack
             # `location_will_be_empty` flag states if, by processing this move line
             # and picking the product, the location will be emptied.
             data_move_line[
                 "location_will_be_empty"
-            ] = move_line.location_id.planned_qty_in_location_is_empty(move_line)
+            ] = move_line.location_id.planned_qty_in_location_is_empty(
+                move_line.package_id.move_line_ids
+                if handle_complete_mix_pack
+                else move_line
+            )
         return data
 
     def _data_for_location(self, location, zone_location=None, picking_type=None):
@@ -607,7 +625,13 @@ class ZonePicking(Component):
         move_lines = self._find_location_move_lines(
             locations=sublocation, package=package
         )
+        handle_complete_mix_pack = self._handle_complete_mix_pack(package)
         if move_lines:
+            if handle_complete_mix_pack:
+                response = self._response_for_set_line_destination(
+                    first(move_lines), qty_done=0
+                )
+                return response, message
             if packaging.package_has_several_products(package):
                 message = self.msg_store.several_products_in_package(package)
             if packaging.package_has_several_lots(package):
@@ -621,8 +645,8 @@ class ZonePicking(Component):
                     ),
                     message,
                 )
+
             move_line = first(move_lines)
-            # Fix me for a package prefill qty is zero ?
             qty_done = self._get_prefill_qty(move_line)
             response = self._response_for_set_line_destination(
                 move_line, qty_done=qty_done
@@ -864,7 +888,9 @@ class ZonePicking(Component):
             base_response=response, message=self.msg_store.barcode_not_found()
         )
 
-    def _set_destination_location(self, move_line, quantity, confirmation, location):
+    def _set_destination_location(
+        self, move_line, package, quantity, confirmation, location
+    ):
         location_changed = False
         response = None
 
@@ -903,21 +929,34 @@ class ZonePicking(Component):
             )
             return (location_changed, response)
         # destination location set to the scanned one
-        self._write_destination_on_lines(move_line, location)
+
         stock = self._actions_for("stock")
+        move_lines = move_line
+        if package:
+            quantity = None
+            # Handling all move lines from a complete mix pack
+            for _move_line in package.move_line_ids:
+                if _move_line.state not in ("assigned", "partially_available"):
+                    continue
+                _move_line.qty_done = move_line.product_uom_qty
+                move_lines |= _move_line
+        self._write_destination_on_lines(move_lines, location)
+
         try:
-            stock.mark_move_line_as_picked(move_line, quantity, check_user=True)
+            stock.mark_move_line_as_picked(move_lines, quantity, check_user=True)
         except ConcurentWorkOnTransfer as error:
+            values = {"qty_done": quantity} if quantity is not None else {}
             response = self._response_for_set_line_destination(
                 move_line,
                 message={
                     "message_type": "error",
                     "body": str(error),
                 },
-                qty_done=quantity,
+                **values,
             )
             return (location_changed, response)
-        stock.validate_moves(move_line.move_id)
+        stock.validate_moves(move_lines.move_id)
+
         location_changed = True
         # Zero check
         zero_check = self.picking_type.shopfloor_zero_check
@@ -1043,6 +1082,7 @@ class ZonePicking(Component):
         barcode,
         quantity,
         confirmation=False,
+        handle_complete_mix_pack=False,
     ):
         """Set a destination location (and done) or a destination package (in buffer)
 
@@ -1082,6 +1122,9 @@ class ZonePicking(Component):
         When the barcode is the product (or its packaging) or the lot on the line:
             * The done quantity is incremented by one or the packaging quantity.
 
+        The `handle_complete_mix_pack` option, when it is set to true. Will move all they
+        lines contained in the package of the move line passed in parameter.
+
         Transitions:
         * select_line: destination has been set, showing the next lines to pick
         * zero_check: if the option is active and if the quantity of product
@@ -1120,6 +1163,9 @@ class ZonePicking(Component):
             # only allow it if moving the full qty.
             location = search.location_from_scan(barcode)
             if location:
+                package = None
+                if handle_complete_mix_pack:
+                    package = move_line.package_id
                 if self._pick_pack_same_time():
                     (
                         good_for_packing,
@@ -1136,6 +1182,7 @@ class ZonePicking(Component):
                         )
                 pkg_moved, response = self._set_destination_location(
                     move_line,
+                    package,
                     quantity,
                     confirmation,
                     location,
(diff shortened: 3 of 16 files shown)


shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -116,20 +116,31 @@ const template_mobile = `
...
+
+        <item-detail-card
+            v-if="hide_qty_picker()"
+            :key="make_state_component_key(['detail-move-line-dest-pack', state.data.move_line.id])"
+            :record="state.data.move_line"
+            :options="{main: true, key_title: 'package_dest.name', title_action_field: {action_val_path: 'package_dest.name'}}"
+            :card_color="utils.colors.color_for(state_in(['set_line_destination']) ? 'screen_step_done': 'screen_step_todo')"
+            />
+
+
```
mt-software-de: What is the purpose of this item-detail-card?
With this item-detail-card the screen looks like this. 
image
without it
image
```suggestion
```

mt-software-de (changes requested):

shopfloor/services/zone_picking.py
```diff
@@ -803,10 +835,19 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
...
+            # Handling all move lines from a complete mix pack
+            for move_line in package.move_line_ids:
+                move_line.qty_done = move_line.product_uom_qty
+                self._write_destination_on_lines(move_line, location)
+            stock.mark_move_line_as_picked(package.move_line_ids)
+            stock.validate_moves(package.move_line_ids.move_id)
+        else:
+            self._write_destination_on_lines(move_line, location)
+            stock.mark_move_line_as_picked(move_line, quantity)
+            stock.validate_moves(move_line.move_id)
```
mt-software-de:
```suggestion
        move_lines = move_line
        if package:
            quantity = 0
            # Handling all move lines from a complete mix pack
            for move_line in package.move_line_ids:
                move_line.qty_done = move_line.product_uom_qty
                move_lines |= move_line
        self._write_destination_on_lines(move_lines, location)
        stock.mark_move_line_as_picked(move_lines, quantity)
        stock.validate_moves(move_lines.move_id)
```

This is would reduce the database calls.

mt-software-de (changes requested):

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
mt-software-de: ok the problem is something else :-)
```suggestion


        <item-detail-card
            v-if="hide_qty_picker()"
            :key="make_state_component_key(['detail-move-line-dest-pack', state.data.move_line.id])"
            :record="state.data.move_line"
            :options="{main: true, key_title: 'package_src.name', title_action_field: {action_val_path: 'package_src.name'}}"
            :card_color="utils.colors.color_for(state_in(['set_line_destination']) ? 'screen_step_done': 'screen_step_todo')"
            />


```

mt-software-de (changes requested):

shopfloor/services/zone_picking.py
TDu: Thanks, updated.

TDu:
@mt-software-de thanks for your review. I updated the code for both of your comments.

shopfloor/services/zone_picking.py
```diff
@@ -803,10 +835,18 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
...
             return (location_changed, response)
         # destination location set to the scanned one
-        self._write_destination_on_lines(move_line, location)
+
         stock = self._actions_for("stock")
-        stock.mark_move_line_as_picked(move_line, quantity)
-        stock.validate_moves(move_line.move_id)
+        move_lines = move_line
+        if package:
+            quantity = 0
```
mt-software-de:
```suggestion
            quantity = None
```
Sorry the quantity must be None instead of 0 otherwise noting will be set done

mt-software-de (changes requested):

shopfloor/services/zone_picking.py
```diff
@@ -803,10 +835,18 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
...
         stock = self._actions_for("stock")
-        stock.mark_move_line_as_picked(move_line, quantity)
-        stock.validate_moves(move_line.move_id)
+        move_lines = move_line
+        if package:
+            quantity = None
+            # Handling all move lines from a complete mix pack
+            for move_line in package.move_line_ids:
+                move_line.qty_done = move_line.product_uom_qty
+                move_lines |= move_line
```
mt-software-de: This will cause an error if the package was already used.
We should check the state or the picking of the move_lines
```suggestion
            for _move_line in package.move_line_ids:
                if _move_line.state != "assigned":
                    continue
                _move_line.qty_done = _move_line.product_uom_qty
                move_lines |= _move_line
```
or 
```suggestion
            for _move_line in package.move_line_ids:
                if _move_line.picking_id != move_line.picking_id:
                    continue
                _move_line.qty_done = _move_line.product_uom_qty
                move_lines |= _move_line
```

TDu: Updated

jbaudoux (changes requested):
The mobile version is not displaying correctly that a complete package is being moved. Only the desktop version is changed.

shopfloor/actions/packaging.py
```diff
@@ -51,3 +51,14 @@ def package_has_several_products(self, package):
...
+
+    def is_complete_mix_pack(self, package):
+        """Check if a package is mixed and completely reserved.
+
+        Will return true if the package has multiple distinct products and
+        all the package quantities are reserved.
+        """
+        return (
+            self.package_has_several_products(package)
+            and package.move_line_ids == package.reserved_move_line_ids
```
jbaudoux: There is something wrong here. While testing, partially reserved packages where considered and show as complete mix pack

jbaudoux (changes requested):

jbaudoux (changes requested):
I was not able to validate the move anymore. Odoo is complaining the package cannot be on 2 different locations

shopfloor/actions/packaging.py
TDu: I changed this implementation and added some tests

TDu:
> I was not able to validate the move anymore. Odoo is complaining the package cannot be on 2 different locations

I pretty sure I fixed that error after the rebase.

TDu:
> The mobile version is not displaying correctly that a complete package is being moved. Only the desktop version is changed.

We need to discuss about that one, because when I move a complete package I have this screen
image
And I can see the content of the package being moved
image

shopfloor/services/zone_picking.py
```diff
@@ -228,6 +235,11 @@ def _response_for_set_line_destination(
         data = self._data_for_move_line(move_line)
         data["move_line"].update(kw)
         data["confirmation_required"] = confirmation_required
+        self._actions_for("packaging")
+        data["handle_complete_mix_pack"] = (
+            self._handle_complete_mix_pack(move_line.package_id)
+            and not self.work.menu.no_prefill_qty
```
TDu: This line seems to be redundant with what is in the `_handle_complete_mix_pack` function

shopfloor/actions/packaging.py
```diff
@@ -51,3 +51,13 @@ def package_has_several_products(self, package):
...
         return len(package.quant_ids.lot_id) > 1
+
+    def is_complete_mix_pack(self, package):
+        """Check if a package is mixed and completely reserved.
+
+        Will return true if the package has multiple distinct products and
+        all the package quantities are reserved.
+        """
+        return self.package_has_several_products(package) and all(
+            [quant.quantity == quant.reserved_quantity for quant in package.quant_ids]
```
jbaudoux:
```suggestion
            quant.quantity == quant.reserved_quantity for quant in package.quant_ids
```

shopfloor/services/zone_picking.py
```diff
@@ -603,8 +623,17 @@ def _scan_source_package(
         move_lines = self._find_location_move_lines(
             locations=sublocation, package=package
         )
+
+        handle_complete_mix_pack = self._handle_complete_mix_pack(package)
+        handling_package = None
+        if handle_complete_mix_pack:
+            handling_package = package
+
```
jbaudoux: This should be in the below `if` statement. Handling this complete mix pack separately makes the code more readable. Then you don't have to change the code afterwards
```suggestion
        if move_lines:
            handle_complete_mix_pack = self._handle_complete_mix_pack(package)
            if handle_complete_mix_pack:
                move_line = first(move_lines)
                response = self._response_for_set_line_destination(
                    move_line, qty_done=0, package=package
                )
            return response, message
```

shopfloor/services/zone_picking.py
```diff
@@ -603,8 +623,17 @@ def _scan_source_package(
...
         if move_lines:
-            if packaging.package_has_several_products(package):
+            if (
+                packaging.package_has_several_products(package)
+                and not handle_complete_mix_pack
+            ):
```
jbaudoux: With above suggestion, you can drop this change

shopfloor/services/zone_picking.py
```diff
@@ -617,11 +646,12 @@ def _scan_source_package(
                     ),
                     message,
                 )
+
             move_line = first(move_lines)
             # Fix me for a package prefill qty is zero ?
             qty_done = self._get_prefill_qty(move_line)
             response = self._response_for_set_line_destination(
-                move_line, qty_done=qty_done
+                move_line, qty_done=qty_done, package=handling_package
```
jbaudoux: With above suggestion, you can drop this change

jbaudoux (changes requested):
Also in `change_package_lot.change_package` here https://github.com/OCA/wms/pull/621/files#diff-81fce66279454c4c0be2adf1f46eb4e629c63edfe409c4cf6bda95940686f9bcR676
it is lacking the new package in the ok func (response for set line destination) in case of complete mix pack

Btw, it is also lacking the qty_done there but not related to this PR

cc @mt-software-de

shopfloor/services/zone_picking.py
```diff
@@ -899,10 +931,21 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
...
+                    continue
+                _move_line.qty_done = move_line.product_uom_qty
+                move_lines |= _move_line
+        self._write_destination_on_lines(move_lines, location)
+
         try:
-            stock.mark_move_line_as_picked(move_line, quantity, check_user=True)
+            stock.mark_move_line_as_picked(move_lines, quantity, check_user=True)
         except ConcurentWorkOnTransfer as error:
             response = self._response_for_set_line_destination(
```
mt-software-de, line 949: Because of line 938 where you set the quantity = None
we should not pass qty_done as argument.
Because we can not set qty_done to None, also it is not needed because it is already done.  
```
kwargs = {}
if not package:
    kwargs["qty_done"] = quantity
response = self._response_for_set_line_destination(
    move_line,
    message={
        "message_type": "error",
        "body": str(error),
    },
    **kwargs
)
return (location_changed, response)```

**Review by mt-software-de — CHANGES_REQUESTED** [review]

**Comment by TDu** [comment]
Last fixup updates all the previous comments

File: shopfloor/services/zone_picking.py [hunk 12]
```diff
@@ -899,21 +927,33 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
...
                 move_line,
                 message={
                     "message_type": "error",
                     "body": str(error),
                 },
-                qty_done=quantity,
+                # qty_done=quantity,
```
mt-software-de: Now it never get passed to the method, it should only be passed to the method,
if we are not handling a package

TDu:
> Also in `change_package_lot.change_package` here https://github.com/OCA/wms/pull/621/files#diff-81fce66279454c4c0be2adf1f46eb4e629c63edfe409c4cf6bda95940686f9bcR676 it is lacking the new package in the ok func (response for set line destination) in case of complete mix pack

This pr has been refactored a few times. And I could not find a reason for passing that package to `_response_for_set_line_destination` , it actually does not make sense because this is an odoo record that was passed and not some json data.

So it is not needed and removed from previous implementation.

jbaudoux (approved):
Tested

TDu:
I squashed the commits and rebased.

shopfloor/services/zone_picking.py
```diff
@@ -899,21 +925,34 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
...
         # destination location set to the scanned one
-        self._write_destination_on_lines(move_line, location)
+
         stock = self._actions_for("stock")
+        move_lines = move_line
+        if package:
+            quantity = None
+            # Handling all move lines from a complete mix pack
+            for _move_line in package.move_line_ids:
+                if _move_line.state != "assigned":
```
```suggestion
                if _move_line.state not in ("assigned", "partially_available"):
```

jbaudoux (changes requested):
The move can be partially available

cc @mt-software-de

shopfloor/services/zone_picking.py
TDu: Fixed

jbaudoux (approved):

jbaudoux:
@mt-software-de Can you update your review?

shopfloor/actions/data.py
```diff
@@ -125,6 +125,10 @@ def _package_parser(self):
             "name",
             "shopfloor_weight:weight",
             ("package_storage_type_id:storage_type", ["id", "name"]),
+            (
+                "quant_ids:total_quantity",
+                lambda rec, fname: sum(rec.quant_ids.mapped("quantity")),
+            ),
```
mt-software-de, line 134: Due to the mass test, performance problems are gradually appearing. One performance problem is the data loading. Because of that, we should only load the data that is really needed, in the screen. With this change, total_quantity is always loaded for the package data, also in other scenarios where it is not needed.

mt-software-de: We could create a separate method, like _get_package_parser which is getting as argument the record to load and kwargs. And there we can add a kwarg like with_total_quantity.

mt-software-de (changes requested):

shopfloor/actions/data.py
jbaudoux: can we do this improvement in another PR?

jbaudoux:
@Tdu can you squash the last fixup commit?

shopfloor/actions/data.py
mt-software-de: yes we can do this. 
I think we should create a PR which is doing it for all scenarios then.

mt-software-de (approved):

TDu:
squashed and rebased.

TDu:
I had to fix a test after the rebase on `test_actions_data.py`

PR #622, opened 2023-04-25
[14.0] shopfloor_checkout: show full list of products after select_line

JuMiSanAr:
Before, in some cases, when scanning / selecting a line in the `select_line` screen, only the selected lines would be displayed in the `select_package` screen.

Now, all lines are displayed and the scanned lines appear as selected.

ref: rau-144

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -486,9 +486,9 @@ class Checkout(Component):
         return self._response_for_select_package(picking, lines)
 
     def _select_lines_from_product(
-        self, picking, selection_lines, product, prefill_qty=1, **kw
+        self, picking, selection_lines, product, prefill_qty=1, check_lot=True, **kw
     ):
-        if product.tracking in ("lot", "serial"):
+        if product.tracking in ("lot", "serial") and check_lot:
             return self._response_for_select_line(
                 picking, message=self.msg_store.scan_lot_on_product_tracked_by_lot()
             )
@@ -570,8 +570,14 @@ class Checkout(Component):
                 picking, selection_lines, packages, prefill_qty=prefill_qty, **kw
             )
 
-        self._select_lines(lines, prefill_qty=prefill_qty)
-        return self._response_for_select_package(picking, lines)
+        first_allowed_line = fields.first(lines)
+        return self._select_lines_from_product(
+            picking,
+            selection_lines,
+            first_allowed_line.product_id,
+            prefill_qty=prefill_qty,
+            check_lot=False,
+        )
 
     def _select_lines_from_serial(self, picking, selection_lines, lot, **kw):
         # Search for serial number is actually the same as searching for lot (as of v14...)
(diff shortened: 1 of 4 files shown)


JuMiSanAr:
@jbaudoux I've corrected this issue in the following cases:
- When scanning a lot.
- When scanning a package.
- When selecting a line manually.

I understand this behavior should be the same in all these cases, but if one of them should behave differently please let me know.

NOTE: Waiting for validation to add the tests.

jbaudoux:
@JuMiSanAr Tested and works properly. Can you update the unit tests?

jbaudoux (approved):

jbaudoux:
@simahawk Can you drop a review on this one?

shopfloor/services/checkout.py
```diff
@@ -478,15 +478,18 @@ def _select_lines_from_package(self, picking, selection_lines, package, **kw):
                     ),
                 },
             )
-        self._select_lines(lines)
+        related_lines = selection_lines.filtered(
+            lambda l: l.package_id != package and not l.shopfloor_checkout_done
+        )
+        lines = self._select_lines(lines, related_lines=related_lines)
```
jbaudoux: Just for this one, we should not show something out of the currently scanned package

jbaudoux (changes requested):

shopfloor/services/checkout.py
JuMiSanAr: Updated.

simahawk: Maybe I miss a piece of ctx but I don't understand this change. If you are asking for a specific line that's the only line that you should select, all the others shouldn't be shown on the select package screen IMO :thinking:

jbaudoux (approved):

shopfloor/services/checkout.py
jbaudoux: he dropped this change. When you scan a package, you see only the content of the package. Like it has always been

shopfloor_packing_info/tests/test_checkout_scan_line.py
```diff
@@ -26,6 +26,10 @@ def test_scan_line_package_ok_packing_info_filled_info(self):
         self._fill_stock_for_moves(move2, in_package=True)
         picking.action_assign()
         move_line = move1.move_line_ids
+        related_lines = picking.move_line_ids - move_line
         self._test_scan_line_ok(
-            move_line.package_id.name, move_line, packing_info=packing_info_text
+            move_line.package_id.name,
+            move_line,
+            related_lines,
+            packing_info=packing_info_text,
```
jbaudoux: @JuMiSanAr I think you should also revert this change as you reverted the change on the package scan.

You can also rebase and squash

shopfloor/services/checkout.py
```diff
@@ -628,8 +630,12 @@ def _select_line_move_line(self, picking, selection_lines, move_line):
             return self._select_lines_from_package(
                 picking, selection_lines, move_line.package_id
             )
-        self._select_lines(move_line)
-        return self._response_for_select_package(picking, move_line)
+
+        related_lines = selection_lines.filtered(
+            lambda l: not l.package_id and l.product_id != move_line.product_id
```
simahawk, line 645: @jbaudoux the change is still here or... I'm drunk :smile:  :thinking:

JuMiSanAr: But we don't make it here when we're scanning a package, this is part of `_select_line_move_line`. In fact, this will take us to the `select_package` screen, right?

jbaudoux: see the condition just above, if the line is in a package, it jumps to _select_lines_from_package

So this change here is correct

simahawk: Ok, but why do you want to show lines that have nothing to deal w/ the current line?
When you select a specific line that has no package you should not list any other line IMO.
But again... I might miss the ctx or the final goal.

jbaudoux: How can you pack the products without seeing the lines? After selecting the first line, I need to see the other lines otherwise I cannot select more than one line to put in a pack

PR #623, opened 2023-05-01
[14.0][IMP] shopfloor: zone_picking scan location or pack first

TDu:
With the option scan_location_or_pack_first on.
When scanning a location, only the lines without a source package are offered to be processed.
If none are found, ask for scanning a package.

ref.: rau-147

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -378,6 +378,12 @@ class MessageAction(Component):
             "body": _("Please scan the location first."),
         }
 
+    def scan_the_package(self):
+        return {
+            "message_type": "warning",
+            "body": _("Please scan the package."),
+        }
+
     def product_multiple_packages_scan_package(self):
         return {
             "message_type": "warning",
@@ -635,6 +641,12 @@ class MessageAction(Component):
         return {
             "message_type": "error",
             "body": _("Location {} empty").format(location.name),
+        }
+
+    def location_empty_scan_package(self, location):
+        return {
+            "message_type": "warning",
+            "body": _("Location empty. Try scanning a package"),
         }
 
     def location_not_found(self):
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -200,11 +200,12 @@ class ZonePicking(Component):
         confirmation_required=False,
         product=False,
         sublocation=False,
+        package=False,
     ):
         if confirmation_required and not message:
             message = self.msg_store.need_confirmation()
         data = self._data_for_move_lines(
-            move_lines, product=product, sublocation=sublocation
+            move_lines, product=product, sublocation=sublocation, package=package
         )
         data["confirmation_required"] = confirmation_required
         data["scan_location_or_pack_first"] = self.work.menu.scan_location_or_pack_first
@@ -325,6 +326,7 @@ class ZonePicking(Component):
         picking_type=None,
         product=None,
         sublocation=None,
+        package=None,
     ):
         zone_location = zone_location or self.zone_location
         picking_type = picking_type or self.picking_type
@@ -337,6 +339,8 @@ class ZonePicking(Component):
             data["product"] = self.data.product(product)
         if sublocation and sublocation != zone_location:
             data["sublocation"] = self.data.location(sublocation)
+        if package:
+            data["package"] = self.data.package(package)
         for data_move_line in data["move_lines"]:
             # TODO: this could be expensive, think about a better way
             # to retrieve if location will be empty.
@@ -401,6 +405,7 @@ class ZonePicking(Component):
         product=None,
         lot=None,
         match_user=False,
+        enforce_empty_package=False,
     ):
         """Find lines that potentially need work in given locations."""
         return self.search_move_line.search_move_lines_by_location(
@@ -411,6 +416,7 @@ class ZonePicking(Component):
             product=product,
             lot=lot,
             match_user=match_user,
+            enforce_empty_package=enforce_empty_package,
         )
 
     def _find_buffer_move_lines_domain(self, dest_package=None):
@@ -487,16 +493,23 @@ class ZonePicking(Component):
         """
         return self._list_move_lines(self.zone_location)
 
-    def _list_move_lines(self, location, product=False, sublocation=False):
+    def _list_move_lines(
+        self, location, product=False, sublocation=False, package=False
+    ):
         move_lines = self._find_location_move_lines(
-            sublocation or location, product=product, match_user=True
+            sublocation or location, product=product, package=package, match_user=True
         )
         return self._response_for_select_line(
-            move_lines, product=product, sublocation=sublocation
+            move_lines, product=product, sublocation=sublocation, package=package
         )
 
     def _scan_source_location(
-        self, barcode, confirmation=False, product_id=False, sublocation=False
+        self,
+        barcode,
+        confirmation=False,
+        product_id=False,
+        sublocation=False,
+        package=False,
     ):
         """Search a location and find available lines into it."""
         response = None
@@ -511,17 +524,25 @@ class ZonePicking(Component):
             message = self.msg_store.location_not_allowed()
             return response, message
 
-        product, lot, package = self._find_product_in_location(location, product_id)
-        if len(product) > 1:
-            message = self.msg_store.several_products_in_location(location)
-        elif len(lot) > 1:
-            message = self.msg_store.several_lots_in_location(location)
-        elif len(package) > 1:
+        if package and package.location_id != location:
+            # Do not search based on a package from a previous location
+            package = False
+        product, lot, packages = self._find_product_in_location(
+            location, product_id, package
+        )
+        if len(packages) > 1:
             message = self.msg_store.several_packs_in_location(location)
+        elif len(packages) == 1 and self.work.menu.scan_location_or_pack_first:
+            message = self.msg_store.scan_the_package()
+        elif len(product) > 1 and not message:
+            message = self.msg_store.several_products_in_location(location)
+        elif len(lot) > 1 and not message:
+            message = self.msg_store.several_lots_in_location(location)
         if message:
-            response = self._list_move_lines(location, sublocation=location)
+            response = self._list_move_lines(
+                location, sublocation=location, package=package
+            )
             return response, message
-
         move_lines = self._find_location_move_lines(
             location,
             product=product,
@@ -535,16 +556,17 @@ class ZonePicking(Component):
                 move_line, qty_done=self._get_prefill_qty(move_line)
             )
         else:
-            # if no move line, narrow the list of move lines on the scanned location
-            response = self._list_move_lines(location)
-            message = self.msg_store.location_empty(location)
+            response = self._list_move_lines(self.zone_location)
+            message = self.msg_store.wrong_record(location)
         return response, message
 
-    def _find_product_in_location(self, location, product_id):
+    def _find_product_in_location(self, location, product_id, package=False):
         """Find the prooducts in stock in given location move line in the location."""
         domain = [("location_id", "=", location.id)]
         if product_id:
             domain.append(("product_id", "=", product_id))
+        if package:
+            domain.append(("package_id", "=", package.id))
         quants = self.env["stock.quant"].search(domain)
         product = quants.product_id
         lot = quants.lot_id
@@ -552,7 +574,12 @@ class ZonePicking(Component):
         return product, lot, package
 
     def _scan_source_package(
-        self, barcode, confirmation=False, product_id=False, sublocation=False
+        self,
+        barcode,
+        confirmation=False,
+        product_id=False,
+        sublocation=False,
+        package=False,
     ):
         """Search a package and find available lines for it.
 
@@ -567,6 +594,12 @@ class ZonePicking(Component):
         package = search.package_from_scan(barcode)
         if not package:
             return response, message
+        if not package.location_id.is_sublocation_of(self.zone_location):
+            # Package is not in an allowed location
+            response = self._list_move_lines(self.zone_location)
+            message = self.msg_store.location_not_allowed()
+            return response, message
+
         move_lines = self._find_location_move_lines(
             locations=sublocation, package=package
         )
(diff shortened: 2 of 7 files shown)


shopfloor/services/zone_picking.py
```diff
@@ -535,15 +544,22 @@ def _scan_source_location(
...
+            else:
+                message = self.msg_store.location_empty_scan_package(location)
         return response, message
 
     def _find_product_in_location(self, location, product_id):
         """Find the prooducts in stock in given location move line in the location."""
         domain = [("location_id", "=", location.id)]
         if product_id:
             domain.append(("product_id", "=", product_id))
+        if self.work.menu.scan_location_or_pack_first:
```
sebalix: I would add a comment for our future selves:
```suggestion
        # As we scanned a location, we want to list products that are not part of packages
        if self.work.menu.scan_location_or_pack_first:
```

shopfloor/services/zone_picking.py
```diff
@@ -535,15 +544,22 @@ def _scan_source_location(
...
             # if no move line, narrow the list of move lines on the scanned location
-            response = self._list_move_lines(location)
-            message = self.msg_store.location_empty(location)
+            response = self._list_move_lines(
+                location, no_package=scan_location_or_pack_first
+            )
+            if not scan_location_or_pack_first:
+                message = self.msg_store.location_empty(location)
+            else:
+                message = self.msg_store.location_empty_scan_package(location)
```
sebalix: Generally when we have a `else` block I prefer to invert the checks to avoid the negation here + a comment
```suggestion
            # Location is empty: invite the user to scan a package if the option is enabled
            if scan_location_or_pack_first:
                message = self.msg_store.location_empty_scan_package(location)
            else:
                message = self.msg_store.location_empty(location)
```

shopfloor/services/zone_picking.py
TDu: Added

shopfloor/services/zone_picking.py
TDu: Updated

shopfloor/actions/message.py
```diff
@@ -611,6 +611,12 @@ def location_empty(self, location):
             "body": _("Location {} empty").format(location.name),
         }
 
+    def location_empty_scan_package(self, location):
+        return {
+            "message_type": "warning",
+            "body": _("Try scanning a package"),
```
```suggestion
            "body": _("Location empty. Try scanning a package"),
```

simahawk (approved):
LG

shopfloor/actions/message.py
TDu: Changed

sebalix (approved):
I restarted the runboat which was failing, we'll see...

TDu:
WIP do not merge.

shopfloor/services/zone_picking.py
```diff
@@ -510,17 +524,20 @@ def _scan_source_location(
...
-        product, lot, package = self._find_product_in_location(location, product_id)
+        product, lot, packages = self._find_product_in_location(
+            location, product_id, package
+        )
         if len(product) > 1:
             message = self.msg_store.several_products_in_location(location)
         elif len(lot) > 1:
             message = self.msg_store.several_lots_in_location(location)
-        elif len(package) > 1:
+        elif len(packages) > 1:
```
jbaudoux: I would do the tests in this order: packages, product, lot to raise the most useful message

shopfloor/services/zone_picking.py
```diff
@@ -510,17 +524,20 @@ def _scan_source_location(
             message = self.msg_store.location_not_allowed()
             return response, message
 
-        product, lot, package = self._find_product_in_location(location, product_id)
+        product, lot, packages = self._find_product_in_location(
```
jbaudoux, line 530: If we change location, we have to remove the package from the context
```suggestion
        if package and package.location_id != location:
          package = False
        product, lot, packages = self._find_product_in_location(
```

shopfloor/services/zone_picking.py
```diff
@@ -535,23 +552,34 @@ def _scan_source_location(
...
         else:
             # if no move line, narrow the list of move lines on the scanned location
-            response = self._list_move_lines(location)
-            message = self.msg_store.location_empty(location)
+            response = self._list_move_lines(location, package=package)
+            # Location is empty: invite the user to scan a package if the option is enabled
+            if scan_location_or_pack_first:
+                message = self.msg_store.location_empty_scan_package(location)
+            else:
+                message = self.msg_store.location_empty(location)
```
jbaudoux: The wrong location has been scanned. We should list again all possibilities of the zone and raise an error

```suggestion
            response = self._list_move_lines(self.zone_location)
            message = self.msg_store.wrong_record(location)
```

shopfloor/services/zone_picking.py
```diff
@@ -535,23 +552,34 @@ def _scan_source_location(
...
         return product, lot, package
 
     def _scan_source_package(
-        self, barcode, confirmation=False, product_id=False, sublocation=False
+        self,
+        barcode,
+        confirmation=False,
+        product_id=False,
+        sublocation=False,
+        package=False,
```
jbaudoux, line 582: This method is lacking a first check like there is in _scan_source_location
After `if not package`

```
        if not package.location_id.is_sublocation_of(self.zone_location):
            response = self._list_move_lines(self.zone_location)
            message = self.msg_store.location_not_allowed()
            return response, message
```

and below in the `Check if the package selected can be a substitute on a move line`, the sublocation check can be removed as done in any case at the beginning with this new addition

shopfloor/services/zone_picking.py
```diff
@@ -577,7 +605,9 @@ def _scan_source_package(
             if message:
                 return (
                     self._list_move_lines(
-                        self.zone_location, sublocation=package.location_id or False
+                        self.zone_location,
+                        sublocation=sublocation,
+                        package=package,
```
jbaudoux, line 616: The sublocation optional argument should be dropped. It is useless
```suggestion
                        sublocation,
                        package=package,
```

jbaudoux (changes requested):
partial review

jbaudoux (changes requested):
partial review

shopfloor/services/zone_picking.py
TDu: Done, I also added a check on the `elif` that the message was not yet assigned by the previous `if`clauses.

shopfloor/services/zone_picking.py
TDu: Fixed

shopfloor/services/zone_picking.py
TDu: Done

shopfloor/services/zone_picking.py
TDu: IMO, it is still needed because _list_move_lines will call _response_for_select_line which needs to know if the sublocation has to be passed in the response to the front-end.

shopfloor/services/zone_picking.py
TDu: Fixed

shopfloor/services/zone_picking.py
```diff
@@ -566,6 +591,12 @@ def _scan_source_package(
...
         if not package:
             return response, message
+        if not package.location_id.is_sublocation_of(self.zone_location):
+            # Package is not in an allowed location
+            response = self._list_move_lines(self.zone_location)
+            message = self.msg_store.location_not_allowed()
+            return response, message
+
         move_lines = self._find_location_move_lines(
             locations=sublocation, package=package
```
jbaudoux, line 604: The package has it's own location. We don't need the sublocation anymore in this call.
```suggestion
            package=package
```

shopfloor/services/zone_picking.py
jbaudoux: What I mean is that I find it cleaner to only have one location argument in _list_move_lines and give the right location in the call. Instead of
```
def fn(location, sublocation):
  location = sublocation or location
  ...

fn(zone_location, sublocation)
```
I prefer
```
def fn(location):
   ...

fn(sublocation or zone_location)
```
Here we want to list the move lines on the package.location_id. The zone_location serves nothing in this call. So it would be:
```
self._list_move_lines(
    package.location_id,
    package=package)
```

But that's a detail, it changes nothing to the behavior and result

shopfloor/services/zone_picking.py
```diff
@@ -664,13 +703,20 @@ def _scan_source_product(
             )
         else:
             response = self._list_move_lines(
-                sublocation or self.zone_location, sublocation=sublocation
+                sublocation or self.zone_location,
+                sublocation=sublocation,
```
jbaudoux, line 721: regarding my previous remark, this would be sufficient
```suggestion
                sublocation or self.zone_location,
```

shopfloor/services/zone_picking.py
```diff
@@ -646,15 +682,18 @@ def _scan_source_product(
         if not product:
             return response, message
         move_lines = self._find_location_move_lines(
-            locations=sublocation, product=product
+            locations=sublocation,
+            product=product,
+            package=package,
+            enforce_empty_package=self.work.menu.scan_location_or_pack_first,
         )
```
jbaudoux, line 692: For raising the right error message, we still need to know if there are lines with a package in case the location has been scanned but not the package.
```
            locations=sublocation, product=product
            locations=sublocation,
            product=product,
            package=package,
        )
if not package and self.work.menu.scan_location_or_pack_first:
   move_lines_with_package_ids = []
   move_lines_without_package_ids = []
   for move_line in move_lines:
      if move_line.package_id:
          move_lines_with_package_ids.append(move_line.id)
      else:
          move_lines_without_package_ids.append(move_line.id)
    move_lines = move_lines.browse(move_lines_without_package_ids)
```
And then at the end you can check move_lines_with_package_ids for raising the right error message

shopfloor/services/zone_picking.py
```diff
@@ -664,13 +703,20 @@ def _scan_source_product(
             )
         else:
             response = self._list_move_lines(
-                sublocation or self.zone_location, sublocation=sublocation
+                sublocation or self.zone_location,
+                sublocation=sublocation,
+                package=package,
             )
             message = self.msg_store.product_not_found()
```
```suggestion
            if move_lines_with_package_ids:
                 message = <This product is part of a package, please scan the package first.>
            else:
                message = self.msg_store.product_not_found()
```

shopfloor/services/zone_picking.py
```diff
@@ -682,7 +728,9 @@ def _scan_source_lot(
         if not lots:
             return response, message
         for lot in lots:
-            move_lines = self._find_location_move_lines(locations=sublocation, lot=lot)
+            move_lines = self._find_location_move_lines(
+                locations=sublocation, lot=lot, package=package
+            )
```
jbaudoux, line 752: Similar to product
```
move_lines_without_package_ids = []
for lot in lots:
  ...
  if not package and self.work.menu.scan_location_or_pack_first:
   move_lines_with_package_ids = []
   for move_line in move_lines:
      if move_line.package_id:
          move_lines_with_package_ids.append(move_line.id)
      else:
          move_lines_without_package_ids.append(move_line.id)
    move_lines = move_lines.browse(move_lines_without_package_ids)
```

shopfloor/services/zone_picking.py
```diff
@@ -695,12 +743,19 @@ def _scan_source_lot(
                     move_line, qty_done=qty_done
                 )
             return response, message
-        response = self._list_move_lines(sublocation or self.zone_location)
+        response = self._list_move_lines(
+            sublocation or self.zone_location, package=package
+        )
         message = self.msg_store.lot_not_found()
```
````
            if move_lines_with_package_ids:
                 message = <This lot is part of a package, please scan the package first.>
            else:
                message = self.msg_store.lot_not_found()
````

**Review by jbaudoux — CHANGES_REQUESTED** [review]
scan product / lot also reviewed now

File: shopfloor/services/zone_picking.py (hunk 11 above)
Comment (TDu) [reply shopfloor/services/zone_picking.py:692]: But before the end the default implementation of this scenario will process the first available move line (line 698..699) the end message you mention will never be displayed, no ?

File: shopfloor/services/zone_picking.py (hunk 13 above)
Comment (TDu) [reply shopfloor/services/zone_picking.py:752]: And I have the similar answer...

File: shopfloor/services/zone_picking.py (hunk 11 above)
Comment (jbaudoux) [reply shopfloor/services/zone_picking.py:692]: it will be if move_lines (i.e. without package) is empty, then you reach the else

Comment (TDu) [reply shopfloor/services/zone_picking.py:692]: Ok, but another issue with this change which prevents it to work. We have added the `enforce_empty_package` in the `find_location_move_lines` to use with the option scan location or package first.
Which will add the package into the search domain even if it is empty. So we will have no lines without package returned.

Comment (jbaudoux) [reply shopfloor/services/zone_picking.py:692]: that's why I dropped enforce_empty_package in the call

**Review by jbaudoux — CHANGES_REQUESTED** [review]
I retried scanning a location containing only a package, when I scan the location, it is not asking to scan the package.
cf rau-147

**Comment by TDu** [comment]
> I retried scanning a location containing only a package, when I scan the location, it is not asking to scan the package. cf rau-147

Seems to be an edge case, when there is only one package in the source location. Fixed with the last fixup.

File: shopfloor/services/zone_picking.py [hunk 15]
```diff
@@ -510,17 +523,25 @@ def _scan_source_location(
...
+        if package and package.location_id != location:
+            # Do not search based on a package from a previous location
+            package = False
+        product, lot, packages = self._find_product_in_location(
+            location, product_id, package
+        )
+        if len(packages) > 1 or (
+            len(packages) >= 1 and self.work.menu.scan_location_or_pack_first
+        ):
+            message = self.msg_store.several_packs_in_location(location)
```
jbaudoux: When `len(packages) == 1`, you cannot say that there are several packs in location
:astonished:

jbaudoux (changes requested):
:scream:

shopfloor/services/zone_picking.py
TDu: Fixed

jbaudoux (approved):
LG now

PR #625, opened 2023-05-03
[14.0][IMP]  shopfloor: checkout auto post lines

JuMiSanAr:
In the checkout scenario, we want to give the option to split lines to a new picking after package selection.

ref: rau-150


shopfloor/tests/test_checkout_auto_post.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
```
simahawk:
```suggestion
# Copyright 2022 Camptocamp SA (http://www.camptocamp.com)
```

shopfloor/tests/test_checkout_auto_post.py
```diff
@@ -0,0 +1,78 @@
...
+        self.service.dispatch(
+            "scan_package_action",
+            params={
+                "picking_id": picking.id,
+                "selected_line_ids": [selected_move_line_a.id, selected_move_line_b.id],
+                "barcode": packaging.barcode,
+            },
+        )
+
+        # Check that two new lines for products s a and b are in a split picking,
```
simahawk:
```suggestion
        # Check that two new lines for products a and b are in a split picking,
```

simahawk (approved):
LG

TDu (approved):

sebalix (approved):

shopfloor/services/checkout.py
```diff
@@ -844,6 +844,10 @@ def _put_lines_in_allowed_package(self, picking, selected_lines, package):
             lines_to_pack.write(
                 {"result_package_id": package.id, "shopfloor_checkout_done": True}
             )
+            if self.work.menu.auto_post_line:
+                # If option auto_post_line is active in the shopfloor menu,
+                # create a split order with this line.
```
sebalix:
```suggestion
                # create a split order with these packed lines.
```

sebalix (approved):

shopfloor/services/checkout.py
```diff
@@ -844,6 +844,10 @@ def _put_lines_in_allowed_package(self, picking, selected_lines, package):
             lines_to_pack.write(
                 {"result_package_id": package.id, "shopfloor_checkout_done": True}
             )
+            if self.work.menu.auto_post_line:
```
jbaudoux: I don't like that this is occurring in this method. We need to post when we reach the end of the process.
We need to keep compatibility with shopfloor_checkout_package_measurement that hooks this method to collect the package measurements. The posting should occur after collecting the measurement. You'll have to add this there also

jbaudoux (changes requested):

shopfloor/services/checkout.py
JuMiSanAr: I've added a fixup to make sure auto_post is done afterwards, let me know if it's ok!

shopfloor/services/checkout.py
```diff
@@ -860,7 +859,16 @@ def _post_put_lines_in_package(self, lines_packaged):
     def _create_and_assign_new_packaging(self, picking, selected_lines, packaging=None):
         actions = self._actions_for("packaging")
         package = actions.create_package_from_packaging(packaging=packaging)
-        return self._put_lines_in_allowed_package(picking, selected_lines, package)
+        return self._pack_lines(picking, selected_lines, package)
+
+    def _pack_lines(self, picking, selected_lines, package):
+        lines_to_pack = selected_lines.filtered(self._filter_lines_to_pack)
+        response = self._put_lines_in_allowed_package(picking, lines_to_pack, package)
```
jbaudoux, line 866: * move check for `message = _("No line to pack found")` out of `_put_lines_in_allowed_package` and put it before this line.
* move `message = self.msg_store.goods_packed_in(package)` out of `_put_lines_in_allowed_package` and put it at the end of this method
* if `_put_lines_in_allowed_package` returns a response, return it immediately without posting (case coming from measurement) as we need to collect the measurement before posting the move.
* add `_auto_post_lines` in measurement module after the measurement have been successfully collected

jbaudoux (changes requested):
Already better, still need a bit more refactoring

jbaudoux (approved):
Looks good now, thanks.
You can squash the fixup commits

PR #630, opened 2023-05-09
[14.0] shopfloor: cluster picking add option first scan location or pack

TDu:
ref.: rau-148

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -533,14 +533,6 @@ class ClusterPicking(Component):
 
     def _scan_line_by_package(self, picking, move_line, package, batch, sublocation):
         """Package scanned, just work with it."""
-        packaging = self._actions_for("packaging")
-        message = None
-        if packaging.package_has_several_products(package):
-            message = self.msg_store.several_products_in_package(package)
-        elif packaging.package_has_several_lots(package):
-            message = self.msg_store.several_lots_in_package(package)
-        if message:
-            return self._pick_next_line(batch, message=message)
         quantity = self._get_prefill_qty(move_line)
         return self._response_for_scan_destination(move_line, qty_done=quantity)
 
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -93,10 +93,14 @@ class ClusterPicking(Component):
         }
         return self._response(next_state="manual_selection", data=data, message=message)
 
-    def _response_for_start_line(self, move_line, message=None, popup=None):
+    def _response_for_start_line(
+        self, move_line, message=None, popup=None, sublocation=None
+    ):
+        kw = {"sublocation": self.data.location(sublocation)} if sublocation else {}
+        data = self._data_move_line(move_line, **kw)
         return self._response(
             next_state="start_line",
-            data=self._data_move_line(move_line),
+            data=data,
             message=message,
             popup=popup,
         )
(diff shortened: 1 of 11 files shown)


shopfloor/services/cluster_picking.py
```diff
@@ -416,7 +417,7 @@ def unassign(self, picking_batch_id):
             batch.write({"state": "draft", "user_id": False})
         return self._response_for_start()
 
-    def scan_line(self, picking_batch_id, move_line_id, barcode):
+    def scan_line(self, picking_batch_id, move_line_id, barcode, sublocation_id=None):
```
jbaudoux, line 424:
```suggestion
    def scan_line(self, picking_batch_id, move_line_id, barcode, package=None):
```

shopfloor/services/cluster_picking.py
```diff
@@ -456,17 +463,21 @@ def scan_line(self, picking_batch_id, move_line_id, barcode):
...
             return self._scan_line_by_package(picking, move_line, package, batch)
 
-        product = search.product_from_scan(barcode)
-        if product and move_line.product_id == product:
-            return self._scan_line_by_product(picking, move_line, product)
+        product = self.env["product.product"]
+        if not self.work.menu.scan_location_or_pack_first or sublocation:
+            product = search.product_from_scan(barcode)
+            if product and move_line.product_id == product:
+                return self._scan_line_by_product(picking, move_line, product)
```
jbaudoux:
```suggestion
        product = search.product_from_scan(barcode)
        if product and move_line.product_id == product:
            if self.work.menu.scan_location_or_pack_first:
              if move_line.package_id == package or False:
                 return self._scan_line_by_product(picking, move_line, product)
            else:
               return self._scan_line_by_product(picking, move_line, product)
```

jbaudoux (changes requested):
You need to give the package in the context

shopfloor/services/cluster_picking.py
TDu: After discussion we do not need to pass the package between the frontend and backend.
But I still believe we need to do this for the sublocation, example :

> The user scan a location containing multiple products. He will first scan the location but the current implementation will ask him to scan the product. And when the scan line or packaging first option is enabled the backend needs to remember that a location was scanned previously

.

shopfloor/services/cluster_picking.py
```diff
@@ -488,7 +507,31 @@ def _get_prefill_qty(self, move_line, qty=0):
...
+            message = self.msg_store.scan_the_location_first()
+        if message:
+            return self._response_for_start_line(
+                move_line,
+                message=message,
+                sublocation=location_scanned or sublocation or None,
+            )
+        return None
+
+    def _scan_line_by_package(self, picking, move_line, package, batch, sublocation):
```
jbaudoux, line 534: I would drop this section so that we always jump to next screen when the right package is scanned. The next screen will take care of scanning the product/lot

```
        if packaging.package_has_several_products(package):
            message = self.msg_store.several_products_in_package(package)
        elif packaging.package_has_several_lots(package):
            message = self.msg_store.several_lots_in_package(package)
        if message:
            return self._pick_next_line(batch, message=message)
```

TDu: Removed

jbaudoux (approved):

simahawk (approved):

simahawk:
@TDu can you rebase?

PR #631, opened 2023-05-10
[14.0] shopfloor zone_picking: block move selection

JuMiSanAr:
Until now, in the mobile version of the zone_picking scenario, we were allowed to select manually a move by clicking on it. This is not the case anymore.

image

ref: rau-153

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/zone_picking.js
+++ b/shopfloor_mobile/static/wms/src/scenario/zone_picking.js
@@ -56,11 +56,13 @@ const template_mobile = `
         </div>
 
         <div v-if="state_is('select_line')">
-            <manual-select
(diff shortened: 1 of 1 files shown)


jbaudoux (changes requested):
We should not loose the condensed display on mobile. Can"t switch manual-select to item-detail-card ?

JuMiSanAr:
> We should not loose the condensed display on mobile. Can"t switch manual-select to item-detail-card ?

Yes. I've attached a screenshot of how that looks like to this PR, let me know if that's ok.

jbaudoux (approved):
Thanks. Looks good now :)

mt-software-de (approved):
LGTM

PR #520, opened 2022-12-13
[14.0][ADD] shopfloor_single_product_transfer_frontend

JuMiSanAr:
Note: This PR includes a hook that's needed to make sure that the endpoints are synced if the shopfloor application was installed prior to adding this scenario.

Depends on:

- https://github.com/OCA/wms/pull/483
- https://github.com/OCA/wms/pull/513

ref: rau-48

PR #635, opened 2023-05-15
[14.0] shopfloor_single_product_transfer: allow to scan packages

JuMiSanAr:
Now the user can scan packages on the first screen (before, only locations were allowed).

ref: rau-146

- [x] Add tests for scan_location endpoint.

Depends on:
- https://github.com/OCA/wms/pull/513
- https://github.com/OCA/wms/pull/520


shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -102,6 +108,34 @@ def _scan_location__check_stock(self, location):
             message = self.msg_store.location_empty(location)
             return self._response_for_select_location(message=message)
 
+    def _scan_location__check_package(self, location):
+        """Check if the location has lines without an assigned package."""
+        lines_without_package = self.env["stock.move.line"].search(
```
jbaudoux: checking existing move lines should be only if allow_create_moves is False
In _scan_location__check_stock, I would also add a check that if all the quants in the location are part of a package, you have to scan a package

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -102,6 +108,34 @@ def _scan_location__check_stock(self, location):
...
+            [
+                ("location_id", "=", location.id),
+                ("package_id", "=", False),
+                ("picking_id.picking_type_id", "in", self.picking_types.ids),
+            ]
```
jbaudoux: add `"(state", "in", ("assigned", "partially_available"))`

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -102,6 +108,34 @@ def _scan_location__check_stock(self, location):
...
+            return self._response_for_select_location(message=message)
+
+    def _scan_package__check_location(self, package):
+        """Check if this package corresponds to any of the allowed locations."""
+        locations = self.picking_types.default_location_src_id
+        child_locations = self.env["stock.location"].search(
+            [("id", "child_of", locations.ids)]
+        )
+        allowed_locations = locations | child_locations
+        if package.location_id not in allowed_locations:
```
jbaudoux: Use instead `if not self.is_src_location_valid(package.location_id)`

Do the same change in `_scan_location__check_location`

jbaudoux (changes requested):

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -106,22 +106,21 @@ def _scan_location__check_stock(self, location):
...
-            return self._response_for_select_location(message=message)
+            return self._response_for_select_location_or_package(message=message)
 
     def _scan_location__check_package(self, location):
         """Check if the location has lines without an assigned package."""
         lines_without_package = self.env["stock.move.line"].search(
             [
                 ("location_id", "=", location.id),
                 ("package_id", "=", False),
-                ("picking_id.picking_type_id", "in", self.picking_types.ids),
```
jbaudoux: I would have kept this `("picking_id.picking_type_id", "in", self.picking_types.ids)`

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -325,29 +354,45 @@ def _use_handlers(self, handlers, *args, **kwargs):
             if response:
                 return response
 
+    def _add_location_package_lot_domain(
+        self, domain, location=None, package=None, lot=None
+    ):
+        if location:
+            domain = AND([domain, [("location_id", "=", location.id)]])
+        if package:
+            domain = AND([domain, [("package_id", "=", package.id)]])
```
jbaudoux: If no package, do not match a line with package
`domain = AND([domain, [("package_id", "=", package.id if package else False)]])`

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -356,11 +401,14 @@ def _create_move_from_location(
...
             "product_uom_qty": quantity,
-            "location_id": location.id,
             "location_dest_id": picking_type.default_location_dest_id.id,
             "origin": self.work.menu.name,
             "picking_type_id": picking_type.id,
         }
+        if location:
+            move_vals["location_id"] = location.id
+        if package:
+            move_vals["package_id"] = package.id
```
jbaudoux: This field does not exist. You need to leverage forced_package_id in the action_assign().
I would create a package_level and set it on the move after action_confirm. This way the move has a picking_id you can use on the package level creation.

jbaudoux (changes requested):

JuMiSanAr:
Closed as this work is already included in https://github.com/OCA/wms/pull/640

PR #638, opened 2023-05-19
[14.0 ]  shopfloor_single_product_transfer: remove dest package

JuMiSanAr:
The result package must always be removed from the moved line as this scenario is moving products.

Depends on:
- https://github.com/OCA/wms/pull/513
- https://github.com/OCA/wms/pull/520

ref: rau-149


JuMiSanAr:
This is already done in https://github.com/OCA/wms/pull/640

PR #643, opened 2023-05-29
[14.0][IMP] stock_warehouse_flow: Split a move

mt-software-de:
This PR allows to specify a qty on a flow, it reduces the possible flows.
Flows to apply limited now also with flow.qty <= move.product_qty

Also it introduces the possibility to split moves before a flow is getting applied. 
Therefor a new selection field is added, split_method. 
A selection value of it defines the way the move is getting split. 
Within the module stock_warehouse_flow the split_method "simple" is added. 
It will split a move by a multiple qty of the configured qty of the move. 

As example:
Flow.qty = 2
Move.product_qty = 7
Then the split will do the following
Move.product_qty = 6
New move is created with a product_qty of 1

This PR also adds a new Module called stock_warehouse_flow_packaging_type, which integrates https://github.com/OCA/product-attribute/tree/14.0/product_packaging_type
This is adding a new field called packaging_type_ids. The field is limiting the possible flows.
It limits the flows, flow.packaging_type_ids = False or flow.packaging_type_ids in move.product_id.packaging_ids.packaging_type_id

Also this modules adds a second split_method called "packaging_type"
The moves are getting the same way split as you would use simple, but the used qty will come from the packaging.qty


mt-software-de:
FYI: @jbaudoux

stock_warehouse_flow/models/stock_move.py
```diff
@@ -10,12 +11,16 @@ class StockMove(models.Model):
...
+        moves_to_confirm = self.browse([])
         for move in self:
             if not move._apply_flow_on_action_confirm():
+                moves_to_confirm |= move
                 continue
-            flow = self.env["stock.warehouse.flow"]._search_for_move(move)
-            flow._apply_on_move(move)
-        return super()._action_confirm(merge=merge, merge_into=merge_into)
+            moves_to_confirm |= FLOW._search_and_apply_for_move(move)
+        return super(StockMove, moves_to_confirm)._action_confirm(
```
jbaudoux: don't concatenate recordset, this is slow
append ids and call browse on the ids before the call to super

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -385,6 +363,68 @@ def _get_rule_from_delivery_route(self, html_exc=False):
...
+            move.product_qty - split_qty, move.product_uom, round=False
+        )
+        move.product_uom_qty = new_product_qty
+        return split_move
+
+    def _compute_split_qty_by_packaging(self, move, packaging):
+        """Returns the qty to split for move and packaging"""
+        rounding = move.product_id.uom_id.rounding
+        if (
+            packaging.packaging_type_id not in self.packaging_type_ids
```
jbaudoux: I would move this test `if packaging.packaging_type_id not in self.packaging_type_ids`

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -385,6 +363,68 @@ def _get_rule_from_delivery_route(self, html_exc=False):
...
+
+    def split_move(self, move):
+        """Try to split the move by the first matching packaging"""
+        self.ensure_one()
+        if not self.split_by_packaging_qty or not self.packaging_type_ids:
+            return
+
+        # Move gets split by first matching packaging
+        for packaging in move.product_id.packaging_ids:
+            split_move = self._split_move_by_packaging(move, packaging)
```
jbaudoux: here
```
if packaging.packaging_type_id not in self.packaging_type_ids:
  continue
```

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -385,6 +363,68 @@ def _get_rule_from_delivery_route(self, html_exc=False):
...
+            if split_move:
+                return split_move
```
jbaudoux: Maybe the split move could be split again with another packaging. I would continue looping

jbaudoux (approved):
LG

jbaudoux:
ping @sebalix for review

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -176,8 +181,29 @@ def _compute_warning(self):
             except UserError as exc:
                 flow.warning = str(exc)
 
+    def _is_equal(self, flow):
```
jbaudoux: You can implement == operator if you also add
```
and self.warehouse_id == flow.warehouse_id
and self.from_picking_type_id == flow.from_picking_type_id
```

```suggestion
    def __eq__(self, flow):
```

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -136,6 +139,8 @@ class StockWarehouseFlow(models.Model):
             "global overview of the flow configuration."
         ),
     )
+    qty = fields.Float("Qty", default=0)
```
jbaudoux: add help

stock_warehouse_flow_packaging_type/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class StockWarehouseFlow(models.Model):
+    _inherit = "stock.warehouse.flow"
+
+    packaging_type_ids = fields.Many2many(
+        "product.packaging.type", string="Packaging types"
+    )
+    split_method = fields.Selection(
+        selection_add=[("packaging_type", "By Packaging Type")]
```
jbaudoux: The model changed to packaging level in v16 https://github.com/OCA/product-attribute/pull/1215, so I would make the key agnostic of this
```suggestion
        selection_add=[("packaging", "By Product Packaging")]
```

stock_warehouse_flow_packaging_type/models/stock_warehouse_flow.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    def _is_equal(self, flow):
+        res = super()._is_equal(flow)
```
jbaudoux:
```suggestion
    def __eq__(self, flow):
        res = super().__eq__(flow)
```

stock_warehouse_flow_packaging_type/README.rst
```diff
@@ -0,0 +1,81 @@
+===================================
+Stock Warehouse Flow Packaging Type
```
jbaudoux: Maybe rename to stock_warehouse_flow_product_packaging ?
As product_packaging_type is now product_packaging_level in v16 https://github.com/OCA/product-attribute/pull/1215

jbaudoux (changes requested):
Minor remarks

stock_warehouse_flow/models/stock_warehouse_flow.py
mt-software-de: I renamed it to `_are_apply_conditions_equal`
Because a wanted to keep the odoo defaults behavior https://github.com/odoo/odoo/blob/36459d26f1adb92f92d52ce05329e8ad3e95dd91/odoo/models.py#L5636

jbaudoux (approved):

mt-software-de:
> ping @sebalix for review

Ping @sebalix could you maybe Review this PR as well?

stock_warehouse_flow/__manifest__.py
```diff
@@ -3,10 +3,10 @@
 {
     "name": "Stock Warehouse Flow",
     "summary": "Configure routing flow for stock moves",
-    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "author": "Camptocamp, BCIM, MT Software, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/wms",
     "category": "Warehouse Management",
-    "version": "14.0.1.0.3",
+    "version": "14.0.1.0.4",
```
sebalix: No need to increase the version number manually if there is no migration script, it's up to the maintainer to bump it through the `ocabot` command (and with the proposed change it'll be more a `14.0.1.1.0` version at the end).

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -136,6 +139,17 @@ class StockWarehouseFlow(models.Model):
             "global overview of the flow configuration."
         ),
     )
+    qty = fields.Float(
+        "Qty",
+        default=0,
+        help="If a qty is set the flow can be applied on moves "
+        "where the move's qty >= the qty set on the flow\n",
+    )
```
sebalix, line 147: Should we not implement instead a One2many with a data model that express the `qty` in a given `uom`? Then the move has to match this UoM first to retrieve the expected qty.

Having a simple `qty` field could lead to weird results if we stock units, kg, liters in our stock. A given flow could not apply the same on products expressed in units or kg.

cc @jbaudoux

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -136,6 +139,17 @@ class StockWarehouseFlow(models.Model):
...
+    split_method = fields.Selection(
+        [("simple", "Simple")],
+        "Split method",
+        help="Simple => move will be split by the qty of the flow or a multiple of the it",
```
sebalix: Typo: multiple of it?

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -325,42 +356,6 @@ def action_generate_route(self):
...
-        for flow in flows:
-            if flow._is_valid_for_move(move):
-                return flow
-        if flows:
-            raise UserError(
-                _(
-                    "No routing flow available for the move {move} in transfer {picking}."
-                ).format(move=move.display_name, picking=move.picking_id.name)
-            )
-        return self.browse()
```
sebalix: Why moving all these methods? It generates a lot of diff and it's difficult to understand what has been changed

stock_warehouse_flow/models/stock_warehouse_flow.py
jbaudoux: As the purpose is to have a very basic split, I would just add a m2o to a uom and check that the stock move also matches the uom.
For more complex cases, it is covered by stock_warehouse_flow_product_packaging

sebalix: Yes, we need at least the UoM if we want to keep things simple.

mt-software-de: Yes this Sounds great i will do this

stock_warehouse_flow/__manifest__.py
mt-software-de: ok thanks for this information, i didn't know that

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -334,32 +379,71 @@ def _search_for_move_domain(self, move):
...
         domain = safe_eval(self.move_domain or "[]")
         if not domain:
             return move
         return move.filtered_domain(domain)
 
+    def _is_qty_valid_for_move(self, move):
+        if not self.qty:
+            return move
+        qty = self.uom_id._compute_quantity(self.qty, move.product_id.uom_id)
+        if qty <= move.product_qty:
```
jbaudoux, line 406: missing test on uom

jbaudoux (changes requested):

stock_warehouse_flow/models/stock_warehouse_flow.py
mt-software-de: This should only be called after _search_for_move there is a Domain for uom

mt-software-de: But you are right it also makes sense to test it here

stock_warehouse_flow/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Stock Warehouse Flow",
     "summary": "Configure routing flow for stock moves",
-    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "author": "Camptocamp, BCIM, MT Software, Odoo Community Association (OCA)",
```
jbaudoux: You can add you as a contributor but not as author

jbaudoux (changes requested):

stock_warehouse_flow/README.rst
```diff
@@ -106,12 +106,14 @@ Authors
 
 * Camptocamp
 * BCIM
+* MT Software
```
jbaudoux: same here

sebalix (approved):

stock_warehouse_flow/views/stock_warehouse_flow.xml
```diff
@@ -42,11 +42,14 @@
...
               </div>
               <field name="carrier_ids" widget="many2many_tags" />
+              <field name="qty" />
+              <field name="uom_id" attrs="{'required': [('qty', '>', 0)]}" />
               <field
                                 name="move_domain"
                                 widget="domain"
                                 options="{'model': 'stock.move', 'in_dialog': true}"
                             />
+              <field name="split_method" />
```
jbaudoux:
```suggestion
              <field
                                name="move_domain"
                                widget="domain"
                                options="{'model': 'stock.move', 'in_dialog': true}"
                            />
              <field name="split_method" string="Max quantity multiple" />
              <label
                                for="qty"
                                attrs="{'invisible': [('split_method', '!=', 'simple')]}"
                            />
              <div name="qty" class="o_row">
                  <field
                                    name="qty"
                                    attrs="{'required': [('split_method', '=', 'simple')], 'invisible': [('split_method', '!=', 'simple')]}"
                                />
                  <field
                                    name="uom_id"
                                    options="{'no_open': True, 'no_create': True}"
                                    groups="uom.group_uom"
                                    attrs="{'required': [('split_method', '=', 'simple')], 'invisible': [('split_method', '!=', 'simple')]}"
                                />
              </div>
```

stock_warehouse_flow_product_packaging/views/stock_warehouse_flow.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="name">stock.warehouse.flow.form</field>
+        <field name="model">stock.warehouse.flow</field>
+        <field
+            name="inherit_id"
+            ref="stock_warehouse_flow.stock_warehouse_flow_view_form"
+        />
+        <field name="arch" type="xml">
+            <field name="qty" position="after">
+                <field name="packaging_type_ids" widget="many2many_tags" />
+            </field>
```
jbaudoux:
```suggestion
            <xpath expr="//div[@name='qty']" position="after">
                <field
                    name="packaging_type_ids"
                    widget="many2many_tags"
                    attrs="{'required': [('split_method', '=', 'packaging')], 'invisible': [('split_method', '!=', 'packaging')]}"
                />
            </xpath>
```

jbaudoux (changes requested):
Configuration is clearer with those changes

mt-software-de:
There is still something missing, the module stock_warehouse_flow_release has also be updated

jbaudoux (approved):

TDu (approved):

PR #645, opened 2023-06-02
[14.0][ADD] shopfloor location content transfer scan dest package

TDu:
On the location content transfer scenario allow to scan a destination
package.
The package can already be in the destination location, goods will
be move into it.
The package can be empty it will be used to move the goods to the
location the user scan next.

Also add `shopfloor_location_content_transfer_force_package` a new
module extending the location content transfer scenario.
It allows to force the user to use a package to move goods in some
location.
It relies on the module `stock_location_package_restriction`.

ref.: rau-152


shopfloor/services/location_content_transfer.py
```diff
@@ -479,31 +544,45 @@ def set_destination_all(self, location_id, barcode, confirmation=False):
...
-        scanned_location = self._actions_for("search").location_from_scan(barcode)
-        if not scanned_location:
-            return self._response_for_scan_destination_all(
-                pickings, message=self.msg_store.barcode_not_found()
-            )
 
-        if not self.is_dest_location_valid(move_lines.move_id, scanned_location):
+        empty_package = self._set_dest_get_empty_pack(package_id)
+        message = None
+        scan_location, scan_package, message = self._set_dest_handle_barcode(barcode)
```
jbaudoux: What about using 
```
         handlers_by_type = {
             "package": self._set_destination_all__by_package,
             "location": self._set_destination_all__by_location,
         }
         self._actions_for("search").find(barcode, types=handlers_by_type.keys())
```
I find it hard to understand in which situation we are afterwards.

jbaudoux: In case of confirmation, do you check scanned package is the same than package_id. Otherwise, confirmation should be considered as False

shopfloor/services/location_content_transfer.py
```diff
@@ -448,23 +457,79 @@ def _find_transfer_move_lines(self, location):
...
+        """
+        message = None
+        if scanned_package and self._is_package_empty(scanned_package):
+            # New empty pack scanned, ask for a destination location to use.
+            empty_package = scanned_package
+            message = self.msg_store.package_selected_is_empty(scanned_package)
+        elif empty_package:
+            if not self._is_package_empty(empty_package):
+                message = self.package_not_empty_anymore(empty_package)
+                empty_package = self.env["stock.quant.pacakge"]
```
jbaudoux: Split in 2 different handler. One is when scanning a package, one is when scanning a location.

If the package is not empty but already on the scanned location, it is not an issue.

jbaudoux (commented):
I find the code hard to follow in which situation we are and what we try to achieve

shopfloor/services/location_content_transfer.py
TDu: The confirmation is done on the destination location (the destination scanned or the location of the package scanned).
And there is no check if the destination being used when asking for confirmation is the same than the first one scanned.

shopfloor/services/location_content_transfer.py
TDu: I thought it would be good to inform the user.
Next step for the user if he wants to use this package is to scan it again.

shopfloor/services/location_content_transfer.py
TDu: I have refactored the part that handles the barcode.

TDu:
I rebased

jbaudoux (approved):

PR #646, opened 2023-06-02
[14.0][IMP] shopfloor: change zone picking option restriction

TDu:
Decouple the scenario options

* Unload package at destination
* Collect multiple moves on a same destination package

So they can be selected independently.

ref.: rau-159


TDu:
I don't know if there was good reason to have this restriction. I was just told to remove it :see_no_evil:

jbaudoux (approved):

PR #647, opened 2023-06-08
[14.0][ADD] shopfloor_reception_packaging_dimension

TDu:
~~This will still need to be rebased after #559 has been merged.~~

ref.: rau-156


shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -469,7 +469,7 @@ const Reception = {
             return picking.origin.includes(input);
         },
         _get_states: function () {
-            return reception_states.bind(this)();
```
simahawk: this shouldn't be changed

shopfloor_reception_mobile/static/src/scenario/reception_states.js
```diff
@@ -6,13 +6,13 @@
 
 /*
     Define states for reception scenario.
-    @param this VueJS component instance
+    @param $instance VueJS component instance
 */
-export const reception_states = function () {
+export const reception_states = function ($instance) {
```
simahawk: don't change this, we found the option to make it work w/ bind

shopfloor_reception_mobile/static/src/scenario/reception.js
TDu: Fixed

shopfloor_reception_mobile/static/src/scenario/reception_states.js
TDu: Fixed

TDu:
rebased

TDu:
The last 2 commits refactor the modules to allow extending the fields to update on a product.packaging.

shopfloor_reception_packaging_dimension/tests/test_set_package_dimension.py
```diff
@@ -0,0 +1,160 @@
...
+    #             "barcode": self.product_a.barcode,
+    #         },
+    #     )
+    #     self.data.picking(self.picking)
+    #     selected_move_line = self.picking.move_line_ids.filtered(
+    #         lambda l: l.product_id == self.product_a
+    #     )
+    #     self._assert_response_set_dimension(
+    #         response, self.picking, selected_move_line, self.product_a_packaging
+    #     )
```
jbaudoux: needs fixing

shopfloor_reception/services/reception.py
```diff
@@ -272,7 +272,8 @@ def _scan_line__assign_user(self, picking, line, qty_done):
         self._assign_user_to_line(line)
         line.qty_done += qty_done
         if product.tracking not in ("lot", "serial") or (line.lot_id or line.lot_name):
-            return self._response_for_set_quantity(picking, line)
+            # return self._response_for_set_quantity(picking, line)
```
jbaudoux: to drop

shopfloor_reception_packaging_dimension/tests/test_set_package_dimension.py
TDu: Fixed and enabled.

TDu:
Squashed and rebased

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,241 @@
...
+                },
+                "qty": {
+                    "coerce": to_float,
+                    "required": False,
+                    "type": "float",
+                    "nullable": True,
+                },
+                "barcode": {"type": "string", "required": False, "nullable": True},
+                "cancel": {"type": "boolean"},
+            }
```
mt-software-de:
```suggestion


class ShopfloorReceptionValidator(Component):
    _inherit = "shopfloor.reception.validator"

    def set_packaging_dimension(self):
        return {
            "picking_id": {"coerce": to_int, "required": True, "type": "integer"},
            "selected_line_id": {
                "coerce": to_int,
                "required": True,
                "type": "integer",
            },  
            "packaging_id": {"coerce": to_int, "required": True, "type": "integer"},
            "height": {
                "coerce": to_float,
                "required": False,
                "type": "float",
                "nullable": True,
            },  
            "length": {
                "coerce": to_float,
                "required": False,
                "type": "float",
                "nullable": True,
            },
            "width": {
                "coerce": to_float,
                "required": False,
                "type": "float",
                "nullable": True,
            },  
            "weight": {
                "coerce": to_float,
                "required": False,
                "type": "float",
                "nullable": True,
            },  
            "shipping_weight": {
                "coerce": to_float,
                "required": False,
                "type": "float",
                "nullable": True,
            },  
            "qty": {
                "coerce": to_float,
                "required": False,
                "type": "float",
                "nullable": True,
            },  
            "barcode": {"type": "string", "required": False, "nullable": True},
            "cancel": {"type": "boolean"},
        } 
```

mt-software-de: There are to much leading spaces

mt-software-de (changes requested):

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,241 @@
...
+            ("width", "=", 0),
+            ("width", "=", False),
+            ("height", "=", 0),
+            ("height", "=", False),
+            ("max_weight", "=", 0),
+            ("max_weight", "=", False),
+            ("qty", "=", 0),
+            ("qty", "=", False),
+            ("barcode", "=", False),
+        ]
```
mt-software-de:
```suggestion
        return expression.OR([
            [("packaging_length", "=", 0)],
            [("packaging_length", "=", False)],
            [("width", "=", 0)],
            [("width", "=", False)],
            [("height", "=", 0)],
            [("height", "=", False)],
            [("max_weight", "=", 0)],
            [("max_weight", "=", False)],
            [("qty", "=", 0)],
            [("qty", "=", False)],
            [("barcode", "=", False)],
        ])
```
I think this is a little bit easier to read

mt-software-de (changes requested):

shopfloor_reception_packaging_dimension/services/reception.py
mt-software-de: Therefor you need to import
`from odoo.osv import expression`

mt-software-de (commented):
Wouldn't it be also good, 
to have some button or a logic to create a packaging if none exists?
cc @jbaudoux

jbaudoux:
> Wouldn't it be also good, to have some button or a logic to create a packaging if none exists? cc @jbaudoux

Not in this PR

shopfloor_reception_packaging_dimension/services/reception.py
TDu: True, updated.

jbaudoux (approved):
LG

jbaudoux:
@TDu Can you squash and rebase?

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,234 @@
...
+        )
+
+    def _get_next_packaging_to_set_dimension(self, product, previous_packaging=None):
+        """Return for a product the next packaging needing dimension to be set."""
+        next_packaging_id = previous_packaging.id + 1 if previous_packaging else 0
+        domain_dimension = self._get_domain_packaging_needs_dimension()
+        domain_packaging_id = [
+            ("product_id", "=", product.id),
+            ("id", ">=", next_packaging_id),
+        ]
```
mt-software-de, line 50:
```suggestion
        previous_packaging_id = previous_packaging.id if previous_packaging else 0
        domain_dimension = self._get_domain_packaging_needs_dimension()
        domain_packaging_id = [
            ("product_id", "=", product.id),
            ("id", ">", previous_packaging_id),
        ]
```

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,234 @@
...
+        """Update dimension on the packaging."""
+        dimension_to_update = ["height", "length", "width", "weight", "barcode", "qty"]
+        for dimension in dimension_to_update:
+            value = self.dimension_to_update.get(dimension, None)
+            if value is not None:
+                if dimension == "weight":
+                    dimension = "max_weight"
+                elif dimension == "length":
+                    dimension = "packaging_length"
+                packaging[dimension] = value
```
mt-software-de: Please do not repeat the fields. 

```suggestion
    def _get_packaging_fields_to_update(self):
        return ["height", "length", "width", "weight", "barcode", "qty"]
    
    def _check_dimension_to_update(self):
        """Return True if there is any dimension that needs to be updated on the packaging."""
        dimension_to_check = self._get_packaging_fields_to_update()
        return any(
            [self.dimension_to_update.get(dim, False) for dim in dimension_to_check]
        )

    def _update_packaging_dimension(self, packaging):
        """Update dimension on the packaging."""
        dimension_to_update = self._get_packaging_fields_to_update()
        for dimension in dimension_to_update:
            value = self.dimension_to_update.get(dimension, None)
            if value is not None:
                if dimension == "weight":
                    dimension = "max_weight"
                elif dimension == "length":
                    dimension = "packaging_length"
                packaging[dimension] = value
```

mt-software-de (changes requested):
Some minor remarks

shopfloor_reception_packaging_dimension/services/reception.py
TDu: Updated

TDu:
Squashed and rebased.

shopfloor/actions/message.py
```diff
@@ -445,6 +445,12 @@ def packaging_not_found_in_picking(self):
             "body": _("Packaging not found in the current transfer."),
         }
 
+    def packaging_dimension_updated(self, packaging):
+        return {
+            "message_type": "success",
+            "body": _("Packaging {} dimension updated.".format(packaging.name)),
```
simahawk: this formatting is wrong
```suggestion
            "body": _("Packaging %(name)% dimension updated.", name=packaging.name),
```

shopfloor_reception/services/reception.py
```diff
@@ -770,6 +770,10 @@ def _align_product_uom_qties(self, move):
             for line in lines:
                 line.product_uom_qty = line.qty_done + remaining_todo
 
+    def _before_response_for_set_quantity(self, picking, line):
```
simahawk: the name of this method seems wrong... Is not "before" is actually replacing the direct call to the other method :thinking:

shopfloor_reception_packaging_dimension/hooks.py
```diff
@@ -0,0 +1,28 @@
...
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__file__)
+
+
+def post_init_hook(cr, registry):
+    _logger.info("Add set packaging dimension option on reception scenario")
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    scenario = env.ref("shopfloor_reception.scenario_reception")
+    options = json.loads(scenario.options_edit)
```
simahawk:
```suggestion
    options = scenario.options
```

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,237 @@
...
+class Reception(Component):
+    _inherit = "shopfloor.reception"
+
+    @property
+    def dimension_to_update(self):
+        if not hasattr(self.work, "dimension_to_update"):
+            self.work.dimension_to_update = {}
+        return self.work.dimension_to_update
+
+    def _before_response_for_set_quantity(self, picking, line):
```
simahawk: I don't understand why you needed a new method to override instead of overriding `_response_for_set_qty` :thinking:

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,237 @@
...
+        ]
+        domain = expression.AND([domain_packaging_id, domain_dimension])
+        return self.env["product.packaging"].search(domain, order="id", limit=1)
+
+    def _response_for_set_packaging_dimension(
+        self, picking, line, packaging=None, message=None
+    ):
+        if not packaging:
+            packaging = self._get_next_packaging_to_set_dimension(line.product_id)
+        if not packaging:
```
simahawk: do we really need this check and call for something that should happen before?

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+class Reception(Component):
+    _inherit = "shopfloor.reception"
+
+    @property
+    def dimension_to_update(self):
+        if not hasattr(self.work, "dimension_to_update"):
+            self.work.dimension_to_update = {}
+        return self.work.dimension_to_update
```
simahawk:
```suggestion
    def __init__(self, work_ctx):
        super().__init__(work_ctx)
        if not hasattr(self.work, "dimension_to_update"):
            self.work.dimension_to_update = {}
```

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,237 @@
...
+            data={
+                "picking": self.data.picking(picking),
+                "selected_move_line": self.data.move_line(line),
+                "packaging": self.data_detail.packaging_detail(packaging),
+            },
+            message=message,
+        )
+
+    def _store_dimension_to_update(self, dimensions, local_table):
+        """Store values to update in the context manager of the component.
```
simahawk: Is not a "context manager"
```suggestion
        """Store values to update in the work context.
```
What is local_table?

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,237 @@
...
+        """Update dimension on the packaging."""
+        dimension_to_update = self._get_packaging_fields_to_update()
+        for dimension in dimension_to_update:
+            value = self.dimension_to_update.get(dimension, None)
+            if value is not None:
+                if dimension == "weight":
+                    dimension = "max_weight"
+                elif dimension == "length":
+                    dimension = "packaging_length"
+                packaging[dimension] = value
```
simahawk:
```suggestion
        fields_conv_map = {"weight": "max_weight", "length": "packaging_length"}
        for dimension in dimension_to_update:
            value = self.dimension_to_update.get(dimension, None)
            if value is not None:
                dimension = fields_conv_map.get(dimension, dimension)
                packaging[dimension] = value
```

shopfloor_reception_packaging_dimension_mobile/static/src/scenario/reception_packaging_dimension.js
```diff
@@ -0,0 +1,144 @@
...
+    </div>
+</div>
+
+` +
+    template.substring(pos);
+
+// Extend the reception scenario with :
+//   - the new patched template
+//   - the js code for the new state
+const ReceptionPackageDimension = process_registry.extend("reception", {
```
simahawk, line 104: Can you try this?

```
Vue.componen("reception", {
     "methods": {
          [....]
      }
})
```

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,237 @@
...
+
+        If the user cancel the dimension update we still propose the next
+        possible packgaging.
+
+        Transitions:
+            - set_packaging_dimension: if more packaging needs dimension
+            - set_quantity: otherwise
+        """
+        self._store_dimension_to_update(
+            ["height", "length", "width", "weight", "barcode", "qty"], locals()
```
simahawk: I would not pass `locals`. Can't we wrap dimension in its own dict? IMO we should have `measures` dict containing all of them.

simahawk (changes requested):

shopfloor_reception/services/reception.py
TDu: It actually inserts a new front end state before the set quantity state if needed. Names can always be improved...

shopfloor_reception_packaging_dimension/services/reception.py
TDu: It seemed easier at the time. Looking at it now, because `_response_for_set_quantity` is called in many places in the scenario code and only in a few places we need to inject the packaging dimension check to return a specific response.

shopfloor_reception_packaging_dimension/services/reception.py
TDu: Indeed the packaging parameter should not be optional, and the ifs can be removed.

shopfloor_reception_packaging_dimension/services/reception.py
TDu: The local symbol table returned by https://docs.python.org/3/library/functions.html#locals

shopfloor_reception_packaging_dimension/services/reception.py
TDu: Updated

shopfloor_reception_packaging_dimension/services/reception.py
TDu: Changed.
@mt-software-de this @jbaudoux  this change impacts the related extending module, I pushed a fixup on that branch as well.

shopfloor_reception_packaging_dimension/hooks.py
TDu: Updated

shopfloor/actions/message.py
TDu: Fixed

shopfloor_reception_packaging_dimension/services/reception.py
TDu: I am happy to keep it as it is.

shopfloor_reception/services/reception.py
mt-software-de: why don't you call it then maybe _response_after_scan_line__assign_user ?

shopfloor_reception_packaging_dimension/services/reception.py
mt-software-de: Why don't you use **kwargs?
Then the store_dimension_to_update is not needed
`def set_packaging_dimension(self, picking_id, selected_line_id, packaging_id, cancel=False, **kwargs):`

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,233 @@
...
+            [
+                self.work.dimension_to_update.get(dim, False)
+                for dim in dimension_to_check
+            ]
+        )
+
+    def _update_packaging_dimension(self, packaging):
+        """Update dimension on the packaging."""
+        dimension_to_update = self._get_packaging_fields_to_update()
+        fields_conv_map = {"weight": "max_weight", "length": "packaging_length"}
```
mt-software-de: Could you maybe move this also to an own method, this would make it easier to hook.

mt-software-de (changes requested):

shopfloor_reception/services/reception.py
TDu: well, it is also called after `set_lot_confirm_action` so I don't think it it really fits.

shopfloor_reception_packaging_dimension/services/reception.py
TDu: Done

shopfloor_reception_packaging_dimension/services/reception.py
TDu: You are right, it seems the solution was over complex. Now no need to use the work context of the component or the `locals`.
I refactored.

shopfloor_reception_packaging_dimension_mobile/static/src/scenario/reception_packaging_dimension.js
TDu: 2 reasons for this not to be working.
First one is in how we declare the Vue component for each scenario. At the moment those components are registered as `Anonymous Component`. But it can be fixed by adding the `name` key in the scenario component.

The 2nd one is at the time this file is being interpreted by the browser, the Vue component is not created yet, it is created later at the app creation https://github.com/OCA/wms/blob/eca1a4208830e96565016b7e204b307e10eb9539/shopfloor_mobile_base/static/wms/src/main.js#L65

mt-software-de (approved):
<strike>LGTM - except the naming of _before_response_for_set_quantity</strike>

After testing there must be a change for the js overwrite of the reception states

shopfloor_reception_packaging_dimension_mobile/static/src/scenario/reception_packaging_dimension.js
simahawk: Yeah, you are right. I had an eureka moment some days ago in the shower (and forgot to reply here): the main issue is that scenario components are registered as local components on the root app. That's it.
We should refactor this https://github.com/OCA/wms/blob/eca1a4208830e96565016b7e204b307e10eb9539/shopfloor_mobile_base/static/wms/src/main.js#L73 (I mean, drop it, and register such components in the std way).
Go ahead like this for now. Maybe add a TODO then we'll see.

shopfloor_reception_packaging_dimension_mobile/static/src/scenario/reception_packaging_dimension.js
```diff
@@ -0,0 +1,144 @@
...
+    template: new_template,
+    "methods.get_packaging_measurements": function () {
+        return ["length", "width", "height", "max_weight", "qty", "barcode"];
+    },
+    "methods._get_states": function () {
+        let states = reception_states.bind(this)();
```
mt-software-de:
```suggestion
        let states = _get_states.bind(this)();
```
Otherwise if an other module tries to overwrite the states, it will not work, because always the one from the reception_states are used + the new state which is defined here.

shopfloor_reception_packaging_dimension_mobile/static/src/scenario/reception_packaging_dimension.js
```diff
@@ -0,0 +1,144 @@
...
+ * Copyright 2023 Camptocamp SA (http://www.camptocamp.com)
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+ */
+
+import {process_registry} from "/shopfloor_mobile_base/static/wms/src/services/process_registry.js";
+
+import {reception_states} from "/shopfloor_reception_mobile/static/src/scenario/reception_states.js";
+
+// Get the original template of the reception scenario
+const reception_scenario = process_registry.get("reception");
```
mt-software-de:
```suggestion
const reception_scenario = process_registry.get("reception");
const _get_states = reception_scenario.component.methods._get_states;
// Get the original template of the reception scenario
```

mt-software-de (changes requested):

shopfloor_reception_packaging_dimension/hooks.py
simahawk: it's not :rofl:

shopfloor_reception_packaging_dimension/hooks.py
```diff
@@ -0,0 +1,28 @@
...
+    options.update({"set_packaging_dimension": True})
+    scenario.options_edit = json.dumps(options)
```
simahawk, line 22: shouldn't you use changes from #700 to enable the endpoint?

shopfloor_reception/services/reception.py
simahawk: TBH I'd prefer something more explicit. The main issue here is that we are trying to change the behavior of something that does not take into account the possibility to handle dimensions.
We could assume - that being a WMS app - we can loosely support such extension.

What if we simply offer an explicit hook here (and everywhere else needed)?
Eg: 
```
       if product.tracking not in ("lot", "serial") or (line.lot_id or line.lot_name):
            if self._requires_packaging_dimension(...):
                return self._response_for_set_packaging_dimension(...)
            return self._response_for_set_quantity(picking, line)

[....]

def _response_for_set_packaging_dimension(self, .....):
    raise NotImplementedError()
```

WDYT?

shopfloor_reception_packaging_dimension/services/reception.py
mmequignon: `_response_for_set_quantity` could be called from elsewhere, it's nice to be able to use this hook or not.
I would check if there's packaging available to set dimension where it's relevant, and return `_response_for_set_packaging_dimension` when we find some.

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,194 @@
...
+                "required": False,
+                "type": "float",
+                "nullable": True,
+            },
+            "weight": {
+                "coerce": to_float,
+                "required": False,
+                "type": "float",
+                "nullable": True,
+            },
```
mt-software-de: We have to check the request for max_weight, because this will be send from javascript.
```suggestion
            "max_weight": {
                "coerce": to_float,
                "required": False,
                "type": "float",
                "nullable": True,
            },
```

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -0,0 +1,194 @@
...
+        return self._response_for_set_packaging_dimension(
+            picking, selected_line, next_packaging, message=message
+        )
+
+    def _check_dimension_to_update(self, dimensions):
+        """Return True if there is any dimension that needs to be updated on the packaging."""
+        return any([value is not None for key, value in dimensions.items()])
+
+    def _get_dimension_fields_conversion_map(self):
+        return {"weight": "max_weight", "length": "packaging_length"}
```
mt-software-de: mapping for weight is not needed, because we will get max_weight from js
```suggestion
        return {"length": "packaging_length"}
```

mt-software-de (changes requested):
Javascript tries to send max_weight but python is looking for weight.

shopfloor_reception_packaging_dimension/services/reception.py
mmequignon: max_weight is another field, IIRC.
If I'm not mistaking, we should rather fix the frontend.

mt-software-de: Yes there is weight defined in OCA/product-attribute/product_packaging_dimension
But this is not a dependency of this module. 
In odoo standard there is only max_weight

mt-software-de: also the packaging parser its getting the data from max_weight and putting into in js also in max_weight.
So i should be max_weight here

shopfloor_reception_packaging_dimension/tests/test_set_package_dimension.py
```diff
@@ -0,0 +1,161 @@
...
+            ),
+        )
+        response = self.service.dispatch(
+            "set_packaging_dimension",
+            params={
+                "picking_id": self.picking.id,
+                "selected_line_id": line.id,
+                "packaging_id": self.product_c_packaging_2.id,
+                "height": 200,
+                "weight": 1000,
```
mt-software-de: shopfloor_reception_packaging_dimension_mobile is never sending weight.
it sends max_weight
```suggestion
                "max_weight": 1000,
```

mt-software-de (changes requested):

shopfloor_reception_packaging_dimension/services/reception.py
TDu: This part of the code has been removed.

shopfloor_reception/services/reception.py
TDu: I am not sure why this is better. It adds more coupling between module when not needed.

What is being done here is inserting a new backend state in a scenario before an existing one. When that new state has achieved its purpose (multiple conversation/interaction between the frontend and backend) the scenario will proceed as initially intended.

Now I still agree that the naming is not good. Because the insertion should not be done before generating the response for `set_quantity` (which is called for the first and subsequent access to the state) but before entering the state `set_quantity`.

So in that light I refactored the code renaming the function to `_before_state__set_quantity`.
It does not add coupling and can be used by other modules.

shopfloor_reception_packaging_dimension/services/reception.py
TDu: Thank you, updated

shopfloor_reception_packaging_dimension/services/reception.py
TDu: Fixed

shopfloor_reception_packaging_dimension/tests/test_set_package_dimension.py
TDu: Fixed

shopfloor_reception_packaging_dimension_mobile/static/src/scenario/reception_packaging_dimension.js
TDu: changed

shopfloor_reception_packaging_dimension_mobile/static/src/scenario/reception_packaging_dimension.js
TDu: changed

TDu:
I have answered or/and fixed all comments.
Rebased as well.

shopfloor_reception_packaging_dimension/models/shopfloor_menu.py
```diff
@@ -0,0 +1,24 @@
...
+        help="If for the product being processed, its related packaging "
+        "dimension are not set, ask to fill them up.",
+    )
+
+    @api.depends("scenario_id")
+    def _compute_set_packaging_dimension_is_possible(self):
+        for menu in self:
+            menu.set_packaging_dimension_is_possible = menu.scenario_id.has_option(
+                "set_packaging_dimension"
+            )
```
mmequignon, line 24: I wouldn't add options.
If you don't want this to be active, you uninstall the module.
Or is there some other need I'm missing?

TDu: With the option, it allows to have multiple menus using the reception scenario and a different configuration.
Also easier to enable/disable this functionality, so why not have it

jbaudoux:
@simahawk Are we good now with this one? Can you update your review?
@mt-software-de Your comments have been tackled, can you update your review?

mt-software-de (approved):
LGTM

jbaudoux:
@Tdu Can you squash ?

simahawk (approved):

PR #649, opened 2023-06-08
[14.0] shopfloor_delivery_shipment: some improvements to load trucks

sebalix:
Some changes are required on the front-end part, will be done later (maybe in another PR).

Depends on:
- https://github.com/OCA/stock-logistics-transport/pull/96

Ref. rau-154

### DIFF
--- a/shopfloor_delivery_shipment/services/delivery_shipment.py
+++ b/shopfloor_delivery_shipment/services/delivery_shipment.py
@@ -420,7 +420,13 @@ class DeliveryShipment(Component):
         if picking:
             data.update(
                 picking=self.data.picking(picking),
-                content=self._data_for_content_to_load(shipment_advice, picking),
+                content=self._data_for_content_to_load_from_picking(
+                    shipment_advice, picking
+                ),
+            )
+        else:
+            data.update(
+                content=self._data_for_content_to_load_from_pickings(shipment_advice),
             )
         return self._response(next_state="scan_document", data=data, message=message)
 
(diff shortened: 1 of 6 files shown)


shopfloor_delivery_shipment/services/delivery_shipment.py
```diff
@@ -459,8 +465,43 @@ def _response_for_validate(self, shipment_advice, message=None):
...
+        loaded_lines = self.env["stock.move.line"].search(
+            [
+                ("picking_id", "in", all_pickings.ids),
+                ("id", "not in", all_lines_to_load.ids),
+                ("shipment_advice_id", "!=", False),
+            ]
+        )
+        pickings_partially_loaded = loaded_lines.picking_id
+        domain += [("picking_id", "in", pickings_partially_loaded.ids)]
+        move_lines = self.env["stock.move.line"].search(domain)
```
jbaudoux: Looking at `_find_move_lines_domain`, I think this should only apply if there is no planned moves on the shipment advice
```
if not shipment_advice.planned_move_ids:
   ...
```

jbaudoux (changes requested):
I hope it loads quickly even when all_lines_to_load is more than a 1000
before it's reduced to a few that are in partially loaded picking

jbaudoux:
It would be good to move this feature out of shopfloor and put it on the shipment advice.
I would even have put an action to list those moves lines grouped by package from a shipment advice. That's a useful control report from a shipment advice without having to open shopfloor.

Nice to have, non blocking, #ReduceCodeInShopfloor

TDu (approved):

shopfloor_delivery_shipment/services/delivery_shipment.py
sebalix: Done. I also moved and adapted the `_find_move_lines_domain` to `shipment_advice` module.

sebalix:
@jbaudoux this PR now depends on https://github.com/OCA/stock-logistics-transport/pull/96

jbaudoux (approved):

PR #650, opened 2023-06-09
[14.0] stock_picking_consolidation_priority: prioritize moves

JuMiSanAr:
Until now, the pickings were being marked as "Urgent" when consolidating priority.

Now, it's the moves that are marked as such, and the picking will only be considered urgent if all moves in it are as well.

ref: rau-162

### DIFF
--- a/stock_picking_consolidation_priority/models/stock_move.py
+++ b/stock_picking_consolidation_priority/models/stock_move.py
@@ -121,4 +121,19 @@ class StockMove(models.Model):
                 [[("id", "in", move_ids)], self._consolidate_priority_domain()]
             )
         )
-        moves.picking_id.write(self._consolidate_priority_values())
+        for move in moves:
+            # Flag the move as urgent.
+            move.write(self._consolidate_priority_values())
+        for picking in moves.picking_id:
+            # Flag the picking as urgent if it only contains urgent moves.
+            picking_moves = picking.move_lines.filtered(
+                lambda m: m.state not in ("cancel", "done")
+            )
+            if (
+                all(
+                    m.priority == self._consolidate_priority_value
+                    for m in picking_moves
+                )
+                and picking.priority is not self._consolidate_priority_value
+            ):
+                picking.write(self._consolidate_priority_values())


stock_picking_consolidation_priority/models/stock_move.py
```diff
@@ -121,4 +121,17 @@ def _consolidate_priority(self):
...
+            # Flag the picking as urgent if it only contains urgent moves.
+            picking = move.picking_id
+            picking_moves = picking.move_lines
+            urgent_moves = picking.move_lines.filtered(
+                lambda m: m.priority == self._consolidate_priority_value
+            )
+            if (
+                picking_moves == urgent_moves
+                and picking.priority is not self._consolidate_priority_value
+            ):
```
jbaudoux:
```suggestion
            if (
                all(m.priority == self._consolidate_priority_value for m in picking_moves)
                and picking.priority is not self._consolidate_priority_value
            ):
```

jbaudoux (approved):

stock_picking_consolidation_priority/models/stock_move.py
```diff
@@ -121,4 +121,17 @@ def _consolidate_priority(self):
...
+        for move in moves:
+            # Flag the move as urgent.
+            move.write(self._consolidate_priority_values())
+            # Flag the picking as urgent if it only contains urgent moves.
+            picking = move.picking_id
+            picking_moves = picking.move_lines
+            if (
+                all(
+                    m.priority == self._consolidate_priority_value
+                    for m in picking_moves
```
jbaudoux, line 135: forgot to exclude the cancel / done moves

jbaudoux (changes requested):

stock_picking_consolidation_priority/models/stock_move.py
JuMiSanAr: Isn't that already done in `_consolidate_priority_domain`?

jbaudoux (approved):
Tested & working

stock_picking_consolidation_priority/models/stock_move.py
```diff
@@ -121,4 +121,19 @@ def _consolidate_priority(self):
...
+                lambda m: m.state not in ("cancel", "done")
+            )
+            if (
+                all(
+                    m.priority == self._consolidate_priority_value
+                    for m in picking_moves
+                )
+                and picking.priority is not self._consolidate_priority_value
+            ):
+                picking.write(self._consolidate_priority_values())
```
mmequignon: This looks overkill to me.
Can't we factorize the calls to _consolidate_priority_values?
If there's multiple moves, each with the same picking_id, we might end up to prioritize it multiple times.
Is that ok?
Even though I don't have the full picture, I don't like this, and I think we can do better.

mmequignon (commented):
Won't block this, but I don't like this fix.

stock_picking_consolidation_priority/models/stock_move.py
jbaudoux: Indeed. Something like:
```
        for move in moves:
            # Flag the move as urgent.
            move.write(self._consolidate_priority_values())
        for picking in moves.picking_id:
            # Flag the picking as urgent if it only contains urgent moves.
```

JuMiSanAr: We really wouldn't prioritize the same picking multiple times due to the `picking.priority is not self._consolidate_priority_value` condition, we would simply skip that picking.
I did it this way to avoid having to implement a second loop on pickings, and instead reuse the first one that we're doing on the moves.

If looping twice is clearer for you, I'll change it.

jbaudoux: On each move, you loop on each move of the move's picking with filtered. So it is O(n²).
It's better to change :)

JuMiSanAr: Don't we still have to apply that filtered anyway to exclude cancelled and done moves?

jbaudoux: yes you do, but only once per picking, not once per move

JuMiSanAr: I don't see much of a difference in that case, we end up having to do loops everywhere anyway, but ok, updated. Thanks!

stock_picking_consolidation_priority/models/stock_move.py
```diff
@@ -121,4 +121,19 @@ def _consolidate_priority(self):
                 [[("id", "in", move_ids)], self._consolidate_priority_domain()]
             )
         )
-        moves.picking_id.write(self._consolidate_priority_values())
+        for move in moves:
+            # Flag the move as urgent.
+            move.write(self._consolidate_priority_values())
+        for picking in moves.picking_id:
```
mmequignon, line 127: Maybe we could remove duplicated pickings here. We might have multiple moves with the same picking.

stock_picking_consolidation_priority/models/stock_move.py
```diff
@@ -121,4 +121,19 @@ def _consolidate_priority(self):
...
+            picking_moves = picking.move_lines.filtered(
+                lambda m: m.state not in ("cancel", "done")
+            )
+            if (
+                all(
+                    m.priority == self._consolidate_priority_value
+                    for m in picking_moves
+                )
+                and picking.priority is not self._consolidate_priority_value
+            ):
```
mmequignon, line 138: I'm pretty sure we can do all that in a single SQL query.
```sql
SELECT sp.id, ALL(sm.priority = {consolidate_value}) as has_consolidate_value, 
FROM stock_move sm
JOIN stock_picking sp ON sm.picking_id = sp.id
WHERE id in {move.picking_id}
AND sp.priority != {consolidate_value}
GROUP BY sm.picking_id
HAVING has_consolidate_value;
```
haven't tested though

stock_picking_consolidation_priority/models/stock_move.py
mmequignon: This can be done with a distinct in SQL.
Check the next comment.

jbaudoux: you cannot have duplicates, it's a records set not a records list

stock_picking_consolidation_priority/models/stock_move.py
jbaudoux: no reason to do in sql

mmequignon: yeah, why should be do things all at once, while we could use the ORM?

PR #651, opened 2023-06-11
[16.0][IMP] - stock_release_channel: add partners as selection criteria

sbejaoui:


rousseldenis (approved):
Code review

stock_release_channel/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+
+    stock_release_channel_id = fields.Many2one(
```
lmignon: @sbejaoui A partner could be in many release channels. Some can be asleep or closed or ...

lmignon (changes requested):
A partner could be in many release channels. Some can be asleep or closed or ...

stock_release_channel/models/stock_release_channel.py
```diff
@@ -472,6 +472,10 @@ def assign_release_channel(self, picking):
         ):
             return
         message = ""
+        partner_release_channel = picking.partner_id._get_partner_release_channel()
```
lmignon: IMO we should add a m2m also on the release channel and display this field on the channel form. Moreover, we must add something in place to instruct the user that when a partner is assigned to a channel, others criteria are no more evaluated. I've the feeling this rule functionality should be implemented into a dedicated addon since it break the normal flow selection flow.

lmignon: ping @jbaudoux

jbaudoux: We must still validate the criteria like the picking type, the shipping method, the domain.... You can only skip the toursolver resolution.

stock_release_channel/readme/USAGE.rst
```diff
@@ -26,3 +26,7 @@ A asleep channel can be waken up by using the "Wake up" button. When the "Wake u
 button is used, in addition to the state change, the system looks for pending
 transfers requiring a release and try to assign them to a channel in the
 "Open" or "Locked" state.
+
+You can choose a dedicated release channel for this partner. All of their
+transfers will be assigned to the selected channel, regardless of the automatic
+assigning process.
```
jbaudoux: Add that as an additional explicit criteria but don't skip the other defined criteria

jbaudoux (changes requested):

stock_release_channel/models/res_partner.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+
+    stock_release_channel_ids = fields.Many2many(
```
lmignon, line 11: @sbejaoui To be sure that the same table is used for each side of the many2many relation I would prefer to explicitely reference the relation, column1 and column2 paramters (id ResPartner and StockReleaseChannel)
```suggestion
    stock_release_channel_ids = fields.Many2many(
        comodel_name="stock.release.channel",
        relation="res_partner_stock_release_channel_rel",
        column1="partner_id",
        column2="channel_id",
        string="Release Channels"
    )
```

stock_release_channel/models/stock_release_channel.py
```diff
@@ -202,6 +202,7 @@ class StockReleaseChannel(models.Model):
         help="Technical field to check if the "
         "action 'Release Next Batch' is allowed.",
     )
+    partner_ids = fields.Many2many(comodel_name="res.partner", string="Partners")
```
lmignon:
```suggestion
partner_ids = fields.Many2many(
        comodel_name="stock.release.channel",
        relation="res_partner_stock_release_channel_rel",
        column1="channel_id",
        column2="partner_id",
        string="Partners"
    )
```

lmignon (changes requested):

lmignon (approved):
LGTM (Code review)

A little comment, I'm not sure that the many2many_tags is the most appropriate widget.... (at least if the list becomes large)

stock_release_channel/views/res_partner.xml
```diff
@@ -0,0 +1,27 @@
...
+            <xpath expr="//page[@name='sales_purchases']/group" position="inside">
+                <group
+                    name="container_row_stock"
+                    groups="stock.group_stock_user"
+                    priority="6"
+                >
+                    <group string="Release Channels" name="release_channel" colspan="2">
+                        <field
+                            name="stock_release_channel_ids"
+                            widget="many2many_tags"
```
jbaudoux, line 19:
```suggestion
                            widget="many2many_tags"
                            options="{'no_create': True}"

```

stock_release_channel/views/stock_release_channel_views.xml
```diff
@@ -89,6 +89,11 @@
                         </page>
                         <page string="Selection criteria" name="criteria">
                             <group name="criteria">
+                                <field
+                                    name="partner_ids"
+                                    options="{'no_create': True}"
+                                    widget="many2many_tags"
+                                />
```
jbaudoux, line 95: fields order should be:

1. warehouse + option no_create
2. picking_type_ids + option no_create but without many2many_tags widget (the picking type names are usually long and prefixed by the warehouse and you can't read it in a many2many_tags
3. rule_domain
4. partner_ids + option no_create but without many2many_tags

jbaudoux (approved):

PR #655, opened 2023-06-13
[14.0] shopfloor_single_product_transfer: improvements for auto_move_create

JuMiSanAr:
If the `auto_move_create` option is enabled:

- We cancel the transfer when clicking on the "cancel" button.
- We don't split the move when scanning a location / package in "set_quantity", we simply mark it as done.

ref: rau-161


shopfloor_single_product_transfer/tests/test_set_quantity.py
```diff
@@ -559,6 +559,32 @@ def test_action_cancel(self):
...
+        move_line.qty_done = 10.0
+        response = self.service.dispatch(
+            "set_quantity__action_cancel", params={"selected_line_id": move_line.id}
+        )
+        data = {}
+        self.assert_response(
+            response, next_state="select_location_or_package", data=data
+        )
+        # Ensure the picking is cancelled if allow_move_create is enabled
+        self.assertTrue(move_line.picking_id.state == "cancel")
```
mmequignon, line 587: we have to check with another user.
I suspect that:
user uses SF, but the backend creates the picking as some special user (admin?).
If so, `picking.create_uid != picking.shopfloor_user_id`.
To ensure that (or the opposite) I would create a SF user for this testcase.

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -599,7 +599,12 @@ def _set_quantity__post_move(self, move_line, location, confirmation=False):
         # TODO qty_done = 0: transfer_no_qty_done
         # TODO qty done < product_qty: transfer_confirm_done
         self._write_destination_on_lines(move_line, location)
-        self._post_move(move_line)
+        if self.is_allow_move_create():
+            self._post_move(move_line)
+        else:
+            # If allow_move_create is enabled,
+            # we create a backorder.
+            self._split_move(move_line)
```
mmequignon: aren't you doing the opposite?

shopfloor_single_product_transfer/tests/test_set_quantity.py
JuMiSanAr: The picking is created with the user that is using the application, as far as I know:

image

shopfloor_single_product_transfer/services/single_product_transfer.py
JuMiSanAr: Fixed the comment, thanks!

shopfloor_single_product_transfer/tests/test_set_quantity.py
mmequignon: this is user_id, not create_uid.

JuMiSanAr: When testing it, `create_uid` gives me this user as well:

image

mmequignon: all good then :)
BTW, you can check this from the UI, through `view_metadata`

JuMiSanAr: Ah cool, thanks!

jbaudoux (approved):

PR #660, opened 2023-06-21
[16.0][FIX] stock_available_to_promise_release: Fix unrelease on backorder …

lmignon:
…moves

When we check if we can release a move, we must only check the remeaining qties on origin moves that are not canceled or done and not part of a printed picking.

Before this change, we checked the 'printed' flag on origin moves's pickings. This check wil not work for backorder moves since in suck a case, the origin moves will be at same time in a 'printed' picking for the qty already processed and delivered and into a picking to do for the remaining qty.

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -85,21 +85,25 @@ class StockMove(models.Model):
 
     def _is_unrelease_allowed_on_origin_moves(self, origin_moves):
         """We check that the origin moves are in a state that allows the unrelease
-        of the current move. At this stage, a move can't be unreleased if
-          * a picking is already printed. (The work on the picking is planed and
-            we don't want to change it)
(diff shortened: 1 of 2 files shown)


lmignon:
ping @jbaudoux @rousseldenis @TDu

jbaudoux:
If you partially process the picking, somehow you started working on the order and you should not be able to unrelease.
With your change, it can leads to issues when you do grouped picking (picking move serving multiple deliveries). Example:
move1 on locX for prodA for SO1 qty 3 is done
move2 on locY for prodA for SO1 & SO2 & SO3 with qty 6 (2 for SO1, 2 for SO2, 2 for SO3).
You should not be able to unrelease SO1 while unprocessed move2 qty 6 > SO1 qty 5

jbaudoux (changes requested):
Needs clarification

rousseldenis:
@lmignon

PR #661, opened 2023-06-21
[14.0] shopfloor zone_picking: block scanning of destination packages

JuMiSanAr:
When scanning a source package in the zone picking scenario, we then have to scan either a destination location or a destination package.
However, scanning a destination package should be prevented if needed.

This PR adds an option to not allow scanning destination packages, prompting the user to scan a location instead.

ref: rau-167

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -883,10 +883,11 @@ class MessageAction(Component):
             "body": _("No line to pack found."),
         }
 
-    def package_not_allowed_scan_location(self):
+    def package_transfer_not_allowed_scan_location(self):
         return {
             "message_type": "warning",
             "body": _(
-                "Scanning a package is not allowed, please scan a location instead."
+                "Transferring to a different package is not allowed, "
+                "please scan a location instead."
             ),
         }
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -882,3 +882,11 @@ class MessageAction(Component):
             "message_type": "warning",
             "body": _("No line to pack found."),
         }
+
+    def package_not_allowed_scan_location(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "Scanning a package is not allowed, please scan a location instead."
+            ),
+        }
--- a/shopfloor/models/shopfloor_menu.py
+++ b/shopfloor/models/shopfloor_menu.py
@@ -46,7 +46,7 @@ A new move will be added as a return of the delivery,
 decreasing the delivered quantity of the related SO line.
 """
 
-ALLOW_SCAN_DESTINATION_PACKAGE_HELP = """
+ALLOW_ALTERNATIVE_DESTINATION_PACKAGE_HELP = """
 When moving a whole package, the user normally scans
 a destination location.
 If enabled, they will also be allowed
@@ -218,13 +218,13 @@ class ShopfloorMenu(models.Model):
     auto_post_line_is_possible = fields.Boolean(
         compute="_compute_auto_post_line_is_possible"
     )
-    allow_scan_destination_package = fields.Boolean(
-        string="Allow to scan a destination package",
+    allow_alternative_destination_package = fields.Boolean(
+        string="Allow to change the destination package",
         default=False,
-        help=ALLOW_SCAN_DESTINATION_PACKAGE_HELP,
+        help=ALLOW_ALTERNATIVE_DESTINATION_PACKAGE_HELP,
     )
-    allow_scan_destination_package_is_possible = fields.Boolean(
-        compute="_compute_allow_scan_destination_package_is_possible"
+    allow_alternative_destination_package_is_possible = fields.Boolean(
+        compute="_compute_allow_alternative_destination_package_is_possible"
     )
 
     @api.onchange("unload_package_at_destination")
@@ -449,8 +449,9 @@ class ShopfloorMenu(models.Model):
         for menu in self:
             menu.allow_return_is_possible = menu.scenario_id.has_option("allow_return")
 
-    def _compute_allow_scan_destination_package_is_possible(self):
+    @api.depends("scenario_id")
+    def _compute_allow_alternative_destination_package_is_possible(self):
         for menu in self:
-            menu.allow_scan_destination_package_is_possible = (
-                menu.scenario_id.has_option("allow_scan_destination_package")
+            menu.allow_alternative_destination_package_is_possible = (
+                menu.scenario_id.has_option("allow_alternative_destination_package")
             )
--- a/shopfloor/models/shopfloor_menu.py
+++ b/shopfloor/models/shopfloor_menu.py
@@ -46,6 +46,13 @@ A new move will be added as a return of the delivery,
 decreasing the delivered quantity of the related SO line.
 """
 
+ALLOW_SCAN_DESTINATION_PACKAGE_HELP = """
+When moving a whole package, the user normally scans
+a destination location.
+If enabled, they will also be allowed
+to scan a destination package.
+"""
+
 
 class ShopfloorMenu(models.Model):
     _inherit = "shopfloor.menu"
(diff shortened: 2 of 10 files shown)


shopfloor/models/shopfloor_menu.py
```diff
@@ -186,15 +193,22 @@ class ShopfloorMenu(models.Model):
...
     auto_post_line = fields.Boolean(
         string="Automatically post line",
         default=False,
-        help=AUTO_POST_LINE,
+        help=AUTO_POST_LINE_HELP,
     )
     auto_post_line_is_possible = fields.Boolean(
         compute="_compute_auto_post_line_is_possible"
     )
+    allow_scan_destination_package = fields.Boolean(
```
jbaudoux:
```suggestion
    allow_alternative_destination_package = fields.Boolean(
```

shopfloor/models/shopfloor_menu.py
```diff
@@ -186,15 +193,22 @@ class ShopfloorMenu(models.Model):
...
+        string="Allow to scan a destination package",
```
jbaudoux:
```suggestion
        string="Allow to transfer to another package",
```

shopfloor/services/zone_picking.py
```diff
@@ -1086,7 +1089,14 @@ def set_destination(
...
         search = self._actions_for("search")
-        accept_only_package = not self._move_line_full_qty(move_line, quantity)
+        # Only allow scanning a destination package if this option is enabled.
+        allow_package = self.work.menu.allow_scan_destination_package
+        # If scanning a destination package is allowed,
+        # and the scanned qty isn't equal to the qty todo,
+        # we force the user to scan a package.
+        accept_only_package = allow_package and not self._move_line_full_qty(
+            move_line, quantity
+        )
```
jbaudoux: don't change this.
If the qty is partial, we have to scan a package

shopfloor/services/zone_picking.py
```diff
@@ -1135,6 +1145,11 @@ def set_destination(
 
         # When the barcode is a package
         package = search.package_from_scan(barcode)
+        if package and not allow_package:
```
```suggestion
        if package:
           if not self.work.menu.allow_alternative_destination_package and move_line.result_package_id and move_line.result_package_id != package:
```

jbaudoux: We could also add a check in case of partial that we are effectively putting on a different package
```
if accept_only_package and move_line.package_id == package:
  message = ...dest_package_not_valid(package)
```

shopfloor/actions/message.py
```diff
@@ -807,3 +807,11 @@ def no_line_to_pack(self):
             "message_type": "warning",
             "body": _("No line to pack found."),
         }
+
+    def package_not_allowed_scan_location(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "Scanning a package is not allowed, please scan a location instead."
```
```suggestion
                "Transferring to a different package is not allowed, please scan a location instead."
```

jbaudoux (changes requested):
We don't want to prevent putting on a package. We want to forbid transferring to a different package.
So if you take partial content out of a source package, you still have to put that on a different package. However, if you take the full source package, then you're not allowed to modify the package.

shopfloor/services/zone_picking.py
```diff
@@ -1136,6 +1141,24 @@ def set_destination(
         # When the barcode is a package
         package = search.package_from_scan(barcode)
         if package:
+
+            if move_line.result_package_id == package:
+                # Check we're not scanning the same dest package.
```
```suggestion
            if not moving_full_quantity and move_line.package_id == package:
                # Check we're not using the source package as transfer package
```

shopfloor/services/zone_picking.py
```diff
@@ -1136,6 +1141,24 @@ def set_destination(
...
+                message = self.msg_store.dest_package_not_valid(package)
+                return self._response_for_set_line_destination(
+                    move_line, message=message, qty_done=quantity
+                )
+
+            allow_alternative_package = self._check_allow_alternative_package(
+                move_line, package
+            )
+            if moving_full_quantity and not allow_alternative_package:
```
```suggestion
            if not allow_alternative_package and move_line.result_package_id and move_line.result_package_id != package:
```

jbaudoux (changes requested):
Some tests are wrong

simahawk:
@JuMiSanAr what about JE's comments?

JuMiSanAr:
> @JuMiSanAr what about JE's comments?

I already implemented them.

jbaudoux (approved):
LG

shopfloor/models/shopfloor_menu.py
```diff
@@ -186,15 +193,22 @@ class ShopfloorMenu(models.Model):
...
         default=False,
-        help=AUTO_POST_LINE,
+        help=AUTO_POST_LINE_HELP,
     )
     auto_post_line_is_possible = fields.Boolean(
         compute="_compute_auto_post_line_is_possible"
     )
+    allow_alternative_destination_package = fields.Boolean(
+        string="Allow to transfer a destination package",
+        default=False,
```
jbaudoux, line 223: You need to add a migration script to enable this on existing menu using this scenario as you are adding a restriction that was not there before

shopfloor/models/shopfloor_menu.py
```diff
@@ -186,15 +193,22 @@ class ShopfloorMenu(models.Model):
...
         string="Automatically post line",
         default=False,
-        help=AUTO_POST_LINE,
+        help=AUTO_POST_LINE_HELP,
     )
     auto_post_line_is_possible = fields.Boolean(
         compute="_compute_auto_post_line_is_possible"
     )
+    allow_alternative_destination_package = fields.Boolean(
+        string="Allow to transfer a destination package",
```
jbaudoux:
```suggestion
        string="Allow to change the destination package",
```

jbaudoux (changes requested):

shopfloor/models/shopfloor_menu.py
JuMiSanAr: Added.

jbaudoux (approved):

jbaudoux (changes requested):
Don't we miss the new configuration on the zone picking scenario data file?

JuMiSanAr:
> Don't we miss the new configuration on the zone picking scenario data file?

It's false by default right?

jbaudoux:
> > Don't we miss the new configuration on the zone picking scenario data file?
> 
> It's false by default right?

It is missing in this file https://github.com/OCA/wms/blob/14.0/shopfloor/data/shopfloor_scenario_data.xml#L18
It is true in this file to enable the option, and yes the config is by default on False on the scenario

JuMiSanAr:
> > > Don't we miss the new configuration on the zone picking scenario data file?
> > 
> > 
> > It's false by default right?
> 
> It is missing in this file https://github.com/OCA/wms/blob/14.0/shopfloor/data/shopfloor_scenario_data.xml#L18 It is true in this file to enable the option, and yes the config is by default on False on the scenario

Ah yes, added.

shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -658,21 +658,50 @@ const ZonePicking = {
...
-                        if (this.state.data.move_line.quantity != qty) {
-                            this.state.display_info.scan_placeholder = this.state.display_info.scan_placeholder_partial;
+
+                        // Display different placeholder messages (package / location / both).
+                        const is_scan_package_allowed = this.state.data
+                            .allow_alternative_destination_package;
+                        const full_qty = this.state.data.move_line.quantity === qty;
+
+                        const display_info = this.state.display_info;
+                        if (!is_scan_package_allowed) {
```
jbaudoux, line 691: While testing I had always "Scan location"

jbaudoux: Tested again and this is working fine

jbaudoux (approved):
I increase the version number as this version was already used.
I made the new option major

PR #663, opened 2023-06-26
[14.0] shopfloor_single_product_transfer: show message if line already done

JuMiSanAr:
If a line is already done, we need to display a message in case the user is trying to update the qty_done through the qty picker.

ref: rau-173

### DIFF
--- a/shopfloor_single_product_transfer/services/single_product_transfer.py
+++ b/shopfloor_single_product_transfer/services/single_product_transfer.py
@@ -217,18 +217,7 @@ class ShopfloorSingleProductTransfer(Component):
     def _scan_product__select_move_line(
         self, product, location=None, package=None, lot=None, packaging=None
     ):
-        domain = self._scan_product__select_move_line_domain(
-            product, location=location, package=package, lot=lot
-        )
-        query = self.env["stock.move.line"]._search(domain, limit=1)
-        order_elems = [
-            "stock_move_line__picking_id.user_id",
-            "stock_move_line__picking_id.priority DESC",
-            "stock_move_line__picking_id.scheduled_date ASC",
-            "id DESC",
-        ]
-        query.order = ",".join(order_elems)
-        move_line = self.env["stock.move.line"].browse(query)
+        move_line = self._select_move_line_from_product(product, location, package, lot)
         if move_line:
             stock = self._actions_for("stock")
             if self.work.menu.no_prefill_qty:
@@ -243,6 +232,28 @@ class ShopfloorSingleProductTransfer(Component):
             else:
                 stock.mark_move_line_as_picked(move_line)
             return self._response_for_set_quantity(move_line)
+
+    def _select_move_line_from_product(self, product, location, package, lot):
+        domain = self._scan_product__select_move_line_domain(
+            product, location=location, package=package, lot=lot
+        )
+        # We add a default order by "id" to avoid the _search method
+        # setting up its own order, which will result in an error.
+        query = self.env["stock.move.line"]._search(domain, order="id", limit=1)
+        # After we retrieve the query, we update the order ourselves.
+        order_elems = [
+            "stock_move_line__picking_id.user_id",
+            "stock_move_line__picking_id.priority DESC",
+            "stock_move_line__picking_id.scheduled_date ASC",
+            "id DESC",
+        ]
+        query.order = ",".join(order_elems)
+        query_str, query_params = query.select()
+        query_str += " FOR UPDATE"
+        self.env.cr.execute(query_str, query_params)
+        ml_ids = [row[0] for row in self.env.cr.fetchall()]
+        move_line = self.env["stock.move.line"].browse(ml_ids)
+        return move_line
 
     def _scan_product__check_create_move_line(
         self, product, location=None, package=None, lot=None, packaging=None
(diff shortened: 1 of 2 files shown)


shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -796,6 +796,9 @@ def set_quantity(self, selected_line_id, barcode, quantity, confirmation=False):
         if not move_line.exists():
             # TODO Should probably return to scan_product or scan_location?
             return self._response_for_set_quantity(move_line)
+        if move_line.state == "done":
```
jbaudoux, line 812: We need to be sure the line is not changed to done during the whole transaction
```suggestion
        self._lock_lines(move_line)
        if move_line.state == "done":
```

jbaudoux (changes requested):
Reviewing the code, 2 users can still select the same line ending up updating something the other would have already processed.
To prevent this, the query in `_scan_product__select_move_line` requires a `FOR UPDATE`.

JuMiSanAr:
> Reviewing the code, 2 users can still select the same line ending up updating something the other would have already processed. To prevent this, the query in `_scan_product__select_move_line` requires a `FOR UPDATE`.

Not sure how to implement this. From what I see, FOR UPDATE should go together with a SELECT, but the query in this method doesn't contain any. Here's an example:

```
        ('"stock_move_line" LEFT JOIN "stock_picking" AS "stock_move_line__picking_id" ON ("stock_move_line"."picking_id" = '
         '"stock_move_line__picking_id"."id") LEFT JOIN "stock_quant_package" AS "stock_move_line__result_package_id" ON '
         '("stock_move_line"."result_package_id" = "stock_move_line__result_package_id"."id")',
         '((((((("stock_move_line"."product_id" = %s) AND ("stock_move_line"."state" in (%s,%s))) AND (("stock_move_line__picking_id"."user_id" in '
         '(%s)) OR "stock_move_line__picking_id"."user_id" IS NULL)) AND ("stock_move_line__picking_id"."picking_type_id" in (%s))) AND '
         '("stock_move_line"."location_id" = %s)) AND ("stock_move_line"."lot_id" = %s)) AND "stock_move_line"."package_id" IS NULL ) AND '
         '("stock_move_line"."company_id" IS NULL  OR ("stock_move_line"."company_id" in (%s)))',
         [47, 'assigned', 'partially_available', 10, 17, 8, 5, 1])
```

jbaudoux:
> > Reviewing the code, 2 users can still select the same line ending up updating something the other would have already processed. To prevent this, the query in `_scan_product__select_move_line` requires a `FOR UPDATE`.
> 
> Not sure how to implement this. From what I see, FOR UPDATE should go together with a SELECT, but the query in this method doesn't contain any. Here's an example: 

replace `move_line = self.env["stock.move.line"].browse(query)` by
```
         query_str, query_params = query.select()
         query_str += " FOR UPDATE"
         self._cr.execute(query_str, query_params)
         ml_ids = [row[0] for row in self._cr.fetchall()]
         move_line = self.env["stock.move.line"].browse(ml_ids)
```

mmequignon:
> > > Reviewing the code, 2 users can still select the same line ending up updating something the other would have already processed. To prevent this, the query in `_scan_product__select_move_line` requires a `FOR UPDATE`.
> > 
> > 
> > Not sure how to implement this. From what I see, FOR UPDATE should go together with a SELECT, but the query in this method doesn't contain any. Here's an example:
> 
> replace `move_line = self.env["stock.move.line"].browse(query)` by
> 
> ```
>          query_str, query_params = query.select()
>          query_str += " FOR UPDATE"
>          self._cr.execute(query_str, query_params)
>          ml_ids = [row[0] for row in self._cr.fetchall()]
>          move_line = self.env["stock.move.line"].browse(ml_ids)
> ```

bah, are you sure about that?
As I see it, once a line is selected, it should be locked.
Once the work is done, it should be unlocked.
No ?

jbaudoux:
> bah, are you sure about that? As I see it, once a line is selected, it should be locked. Once the work is done, it should be unlocked. No ?

But if 2 transactions are modifying the same line, it will raise a concurrent update error which will likely occur if 2 users are scanning the same product. Adding a waiting lock will prevent this

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -244,6 +233,25 @@ def _scan_product__select_move_line(
                 stock.mark_move_line_as_picked(move_line)
             return self._response_for_set_quantity(move_line)
 
+    def _select_move_line_from_product(self, product, location, package, lot):
+        domain = self._scan_product__select_move_line_domain(
+            product, location=location, package=package, lot=lot
+        )
+        query = self.env["stock.move.line"]._search(domain, limit=1)
```
jbaudoux: as order by is overwritten afterwards, don't apply default order by
```suggestion
        query = self.env["stock.move.line"]._search(domain, order="id", limit=1)
```
Default order by is on "result_package_id" which is a non required field so it will cause a useless nullable join if you don't remove it from the search. Nullable joins are not allowed in FOR UPDATE

jbaudoux (changes requested):

jbaudoux:
@JuMiSanAr Can you squash the fixup commit?

jbaudoux (approved):

TDu (approved):

PR #664, opened 2023-06-27
[14.0][IMP] shopfloor_delivery_shipment: filter by location

TDu:
ref.: rau-175


jbaudoux:
@TDu Can you rebase ?

TDu:
@jbaudoux It has been rebased.

jbaudoux (changes requested):
There is an inconsistency between _scan_package, _scan_lot and _scan_product.
They should all take the same context picking & location as argument.
`_find_move_lines_from_product` should then search based on this context.
`_response_for_scan_document` should also keep that picking & location context.

TDu:
> There is an inconsistency between _scan_package, _scan_lot and _scan_product. They should all take the same context picking & location as argument. `_find_move_lines_from_product` should then search based on this context. `_response_for_scan_document` should also keep that picking & location context.

The inconsistency was already there because with this scenario you can only scan a product if you have scanned a picking first.
At the moment the filtering is either on a location or a picking but not both.
Do we want to change so that when scanning a picking after having scanned a location the filtering will be on both of them ?

jbaudoux:
> The inconsistency was already there because with this scenario you can only scan a product if you have scanned a picking first. At the moment the filtering is either on a location or a picking but not both. Do we want to change so that when scanning a picking after having scanned a location the filtering will be on both of them ?

No, we don't want to change that. The context is either on picking or location.

But you scan a product/lot/package, it should respect the context

TDu:
@jbaudoux Updated, the filtering is either on the location or picking and the back button will reset the filtering applied.

jbaudoux:
@TDu `_find_move_lines_from_*` are still not respecting the location/picking context. Only `_find_move_lines_from_product` is respecting the picking context but not the location.
When you scan a location, you cannot match stuffs that are not on this location. Same for picking.

TDu:
> @TDu `_find_move_lines_from_*` are still not respecting the location/picking context. Only `_find_move_lines_from_product` is respecting the picking context but not the location. When you scan a location, you cannot match stuffs that are not on this location. Same for picking.

Ok. So before when scanning a location or a picking it would only filter the lines that were displayed to the user. But now it will also restrict what can be scanned.
That is, if the user is filtering on a location and then scan a package that is NOT in that location BUT still valid to put in the shipment he will face an error "Packgage ... can not be loaded in the shipment..." which would not be raised without the location scanned.

shopfloor_delivery_shipment/services/delivery_shipment.py
```diff
@@ -746,8 +769,15 @@ def _find_move_lines_from_product(
         """
         domain = self._find_move_lines_domain(shipment_advice)
         domain.extend(
-            [("product_id", "=", product.id), ("picking_id", "=", picking.id)]
+            [
+                ("product_id", "=", product.id),
+                ("picking_id", "=", picking.id),
```
jbaudoux, line 794: if picking

jbaudoux:
> Ok. So before when scanning a location or a picking it would only filter the lines that were displayed to the user. But now it will also restrict what can be scanned. That is, if the user is filtering on a location and then scan a package that is NOT in that location BUT still valid to put in the shipment he will face an error "Packgage ... can not be loaded in the shipment..." which would not be raised without the location scanned.

You can still raise a proper error message like "Package xxx not found in location" or "Package xxx not found in picking". Same for product & lot. I'm sure we should have similar message in other scenarios.

shopfloor_delivery_shipment_mobile/static/wms/src/scenario/delivery_shipment.js
```diff
@@ -151,6 +159,18 @@ const DeliveryShipment = {
             }
             return data.picking;
         },
+        filter_location: function (from_state = "") {
```
simahawk, line 163: AFAIU from your discussions, the location - and likely the shipment advice - is an information that we carry on across the whole flow until it gets changed specifically. If that's the case, I'd say that a better approach would be to store the selected record  in a specific vue.js field and then send it via header (as we do for menu item in all scenario or for picking type + zone + lines order on zone picking).
This way, such records become part of a "global context" info and you don't have to carry them on everywhere.

simahawk (commented):
Overall in line w/ JE remarks

shopfloor_delivery_shipment/services/delivery_shipment.py
TDu: The scenario does not allow to search lines for a product if a transfer has not been scanned before. So not needed. https://github.com/OCA/wms/pull/664/files#diff-20525928439cdd42c96888983cc73a09f37ce4d30a8870ab7aad714b2d78c90cR301

shopfloor_delivery_shipment_mobile/static/wms/src/scenario/delivery_shipment.js
TDu: Using the http header to store the menu item makes sense and is a good approach. But for data specific to endpoints (here not all endpoint have a location or shipment advice as parameter) I am not sure it is a real help and IMO removes some clarity.
Although I am (I hope) just at the end of this change and I would be happy not to refactor at this stage.

TDu:
The last 2 commits improve the messages displayed to the user as discussed in previous conversation (above)
I also had to rebase.

jbaudoux (approved):
tested, looks good

Can you squash commits?

TDu:
@jbaudoux Thanks for your reviews. Squashed.

jbaudoux:
@simahawk Good for you?

simahawk:
@jbaudoux pls do not trigger a merge for too many PRs in a row otherwise we get tons of useless build (there's no queue for such commands).

PR #666, opened 2023-07-03
[14.0] shopfloor_reception: Allow to parse carrier return barcodes

mmequignon:
based on https://github.com/OCA/delivery-carrier/pull/668

ref: RAU-164


shopfloor/__manifest__.py
```diff
@@ -41,6 +41,8 @@
         # and to validate packaging/carrier in checkout scenario
         # This must be an optional dep
         "delivery",
+        # TODO: Used to parse return barcodes coming from carriers
+        "delivery_carrier_return_barcode_pattern",
```
simahawk: This should stay only in sf_reception

shopfloor/actions/search.py
```diff
@@ -169,11 +169,12 @@ def generic_packaging_from_scan(self, barcode):
         )
 
     def origin_move_from_scan(self, barcode, extra_domain=None):
+        origins = self.env["delivery.carrier"]._get_origin_from_barcode(barcode)
```
simahawk:
```suggestion
    def origin_move_from_scan(self, barcode, extra_domain=None, origins=None):
        origins = origins or [barcode]
```
And move `self.env["delivery.carrier"]._get_origin_from_barcode(barcode)` to the override in sf_reception.

simahawk (changes requested):

shopfloor/actions/search.py
mmequignon: I don't like this "adding args because we expect smth that's implemented in another module" thing.
I think we should either we move this whole "origin_move_from_scan" method in sf_reception or keep it like this.
But we might need this find method in another scenario in the future, and we will move it here anyway.

simahawk: Then delegate it to an hookable method :man_shrugging: 
Whatever you like most but don't leave it here :smile:

simahawk:
@mmequignon ping :)

mmequignon:
> @mmequignon ping :)

done

shopfloor/actions/search.py
sebalix: We can't call `self.env["delivery.carrier"]._get_origin_from_barcode(barcode)` from `shopfloor` anyway unless to add a dependency on `delivery_carrier_return_barcode` in `shopfloor`

shopfloor/actions/search.py
```diff
@@ -168,12 +168,18 @@ def generic_packaging_from_scan(self, barcode):
             [("barcode", "=", barcode), ("product_id", "=", False)], limit=1
         )
 
+    def _get_origin_from_barcode(self, barcode):
+        """Returns a list of valid origins for a given barcode."""
+        # Hookable method
+        return [barcode]
+
     def origin_move_from_scan(self, barcode, extra_domain=None):
+        origins = self.env["delivery.carrier"]._get_origin_from_barcode(barcode)
```
sebalix: You probably wanted to write this:
```suggestion
        origins = self._get_origin_from_barcode(barcode)
```

shopfloor_reception/__manifest__.py
```diff
@@ -9,7 +9,11 @@
     "maintainers": ["mmequignon", "JuMiSanAr"],
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["shopfloor"],
+    "depends": [
+        "shopfloor",
+        # TODO: Used to parse return barcodes coming from carriers
```
sebalix: Why TODO?

shopfloor/actions/search.py
mmequignon: yeah, added a hook, forgot to change the call.
This was the case before I addressed this comment. shopfloor was depending on `delivery_carrier_return_barcode`

shopfloor/actions/search.py
```diff
@@ -168,12 +168,18 @@ def generic_packaging_from_scan(self, barcode):
             [("barcode", "=", barcode), ("product_id", "=", False)], limit=1
         )
 
+    def _get_origin_from_barcode(self, barcode):
+        """Returns a list of valid origins for a given barcode."""
+        # Hookable method
+        return [barcode]
+
     def origin_move_from_scan(self, barcode, extra_domain=None):
```
sebalix, line 171: I'm not sure to understand the goal of this method, a docstring is welcome (I know it was written before this PR).
From what I see, in `shopfloor_reception` we are overloading the result of this method, are we not breaking something here for scenarios using this method outside of a return/reception context?

shopfloor_reception/__manifest__.py
mmequignon: Because the manifest is written like that for shopfloor, and I moved this to `shopfloor_reception` to address @simahawk 's comment.

simahawk: but there's nothing todo, no?

shopfloor_reception/actions/search.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo.addons.component.core import Component
+
+
+class SearchAction(Component):
+
+    _inherit = "shopfloor.search.action"
+
+    def _get_origin_from_barcode(self, barcode):
+        carriers = self.env["delivery.carrier"].search([])
```
simahawk: this should be probably controlled via ctx key or work context var to enable it only for the reception scenario

simahawk (changes requested):

shopfloor_reception/actions/search.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class SearchAction(Component):
+
+    _inherit = "shopfloor.search.action"
+
+    def _get_origin_from_barcode(self, barcode):
+        if self.work.menu.scenario_id.key == "reception":
+            carriers = self.env["delivery.carrier"].search([])
+            return carriers._get_origin_from_barcode(barcode)
```
mt-software-de: There should be a super call

```suggestion
        if self.work.menu.scenario_id.key == "reception":
            carriers = self.env["delivery.carrier"].search([])
            return carriers._get_origin_from_barcode(barcode)
        return super()._get_origin_from_barcode(barcode)
```

also you have to adapt shopfloor_base/services/service.py https://github.com/OCA/wms/blob/fd917e896659271d1dfa36493510a1b6b0791dd0/shopfloor_base/services/service.py#L39

To something similar like this, otherwise there will be no "self.work.menu" in actions search
```
    def _actions_for(self, usage, **kw):
        propagate_kwargs = kw.get("propagate_kwargs", []) 
        if "menu" not in propagate_kwargs:
            propagate_kwargs.append("menu")
            kw["propagate_kwargs"] = propagate_kwargs
        return get_actions_for(self, usage, **kw)
```

mt-software-de (changes requested):

shopfloor_reception/actions/search.py
mt-software-de: It should be like this.
If there is nothing found from the carrier, return the super result

```suggestion
        if self.work.menu.scenario_id.key == "reception":
            carriers = self.env["delivery.carrier"].search([])
            res = carriers._get_origin_from_barcode(barcode)
            if res:
                return res
        return super()._get_origin_from_barcode(barcode)
```

mmequignon: true, I'll fix that

jbaudoux: @mmequignon Do you fix it? Or @TDu ?

mt-software-de: What is the state here?

TDu:
I pushed force to remove the commit with non merged related PR, that now has been merged.

shopfloor_reception/actions/search.py
mt-software-de: Thx. @TDu 
We also need to include 
https://github.com/OCA/wms/pull/822
as test-requirement otherwise this will not work because work has no attribute menu

TDu: > There should be a super call
> 
> also you have to adapt shopfloor_base/services/service.py
> 
> https://github.com/OCA/wms/blob/fd917e896659271d1dfa36493510a1b6b0791dd0/shopfloor_base/services/service.py#L39
> 
> To something similar like this, otherwise there will be no "self.work.menu" in actions search
> 
> ```
>     def _actions_for(self, usage, **kw):
>         propagate_kwargs = kw.get("propagate_kwargs", []) 
>         if "menu" not in propagate_kwargs:
>             propagate_kwargs.append("menu")
>             kw["propagate_kwargs"] = propagate_kwargs
>         return get_actions_for(self, usage, **kw)
> ```

I think the actions should be kept independent from the shopfloor service implementation.
Instead of relying on a shopfloor menu, probably better to add a context key.

mt-software-de: But we have to otherwise, this will not work. It is only working with tests because they are not acting like, if it is called from the app. 
The problem is here https://github.com/OCA/wms/blob/d246abf384c343c69fb2ded4473164f7a43a4039/shopfloor_reception/tests/common.py#L101. Here the service is initialized with the menu from the start and https://github.com/OCA/wms/blob/d246abf384c343c69fb2ded4473164f7a43a4039/shopfloor_base/tests/common.py#L73 passes the menu arg also to https://github.com/OCA/wms/blob/d246abf384c343c69fb2ded4473164f7a43a4039/shopfloor_base/tests/common.py#L63 which initializes the data with _actions_for with the menu as kwarg. Thats why it is working in the tests but not within the app. 
@simahawk @jbaudoux @TDu please check this PR, it ensures that the menu is set on the data property.
If you have a look at https://github.com/OCA/wms/blob/d246abf384c343c69fb2ded4473164f7a43a4039/shopfloor_base/services/service.py#L202 which is called within the dispatch of service you can see that the menu is set after the init.

TDu: But you don't think it is feasible with a context key ?

The real question, I raised,  is do we want to have the functions define in the actions tightly coupled with the shopfloor implementation, which I don't ?

mt-software-de: I think a context will also not Work. Of course moving the search into the service we would Not have to Deal with this problem at all. 👍🏼

shopfloor/actions/search.py
simahawk: ping

shopfloor_reception/actions/search.py
simahawk: I agree w/ @TDu and @mt-software-de it shouldn't stay there.
If there's a difference in tests we have to fix that too.
IMO that method is specific to services and should stay in the base service mixin.

sebalix: Basically in the long run we want to extract all helper of Shopfloor in dedicated Odoo modules (that could provide these features in the Odoo backend), and Shopfloor will use these ones instead to perform some actions (extract_and_action_done, etc). It's not easy, but that's what we target.

TDu: Indeed the context does not work, probably because it is a component. But I updated the code with an optional parameter (see the last fixup commit)

mt-software-de: ok then we should just move the method origin_move_from_scan to service mixin and the overwrite in the reception scenario.

mt-software-de: @simahawk  and @TDu  i updated https://github.com/OCA/wms/pull/822
To not have a menu in the action components while testing

mt-software-de: @TDu #822 is now merged can you do a rebase? Then the tests should fail

TDu: I just did a rebase.
Hopefully the test will still be ok, because of my last fixup from last week

shopfloor_reception/actions/search.py
```diff
@@ -0,0 +1,17 @@
...
+
+    _inherit = "shopfloor.search.action"
+
+    def _get_origin_from_barcode(self, barcode, **kwargs):
+        if kwargs.get("reception_return"):
+            carriers = self.env["delivery.carrier"].search([])
+            res = carriers._get_origin_from_barcode(barcode)
+            if res:
+                return res
+        return super()._get_origin_from_barcode(barcode)
```
mt-software-de: Could we just move this into the reception scenario, it is only used there. 
This would then simplify the code. 
Also it is not needed to search all carriers. If we merge https://github.com/OCA/delivery-carrier/pull/767
```suggestion
    def _get_origin_from_barcode(self, barcode, **kwargs):
        if kwargs.get("reception_return"):
            res = self.env["delivery.carrier"]._get_origin_from_barcode(barcode)
            if res:
                return res
        return super()._get_origin_from_barcode(barcode)
```

mt-software-de (changes requested):

shopfloor_reception/actions/search.py
TDu: I agree. Updated

mt-software-de (commented):
LGTM. Could some reopen this one please. 
@jbaudoux ?

mt-software-de (approved):

shopfloor_reception/README.rst
```diff
@@ -7,7 +7,7 @@ Shopfloor Reception
    !! This file is generated by oca-gen-addon-readme !!
    !! changes will be overwritten.                   !!
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-   !! source digest: sha256:339c718c4454459205067e05b3ff5c6dac381126cd1aaaf9f7503be87b410a62
+   !! source digest: sha256:f0987430ffb4dffad60516f1fbbb17e2ce2ce37d437b54c7b6354f15c06c8b31
```
jbaudoux, line 10: Can you drop changes to auto-generated files?
@TDu Can you solve this?

shopfloor_reception/static/description/index.html
```diff
@@ -367,7 +367,7 @@ <h1 class="title">Shopfloor Reception</h1>
 !! This file is generated by oca-gen-addon-readme !!
 !! changes will be overwritten.                   !!
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-!! source digest: sha256:339c718c4454459205067e05b3ff5c6dac381126cd1aaaf9f7503be87b410a62
```
simahawk, line 370: just rollback this change and always use this env key:

```export SKIP=oca-gen-addon-readme``` 

Source https://github.com/OCA/wms/blob/16.0/.github/workflows/pre-commit.yml#L38

shopfloor/actions/search.py
simahawk: any answer to this?

mt-software-de:
> LGTM. Could some reopen this one please. 
> @jbaudoux ?

Again? @jbaudoux 😄

mt-software-de:
@mmequignon whats the state here?

Could you remove the changes from `README.rst` and `index.html` then this should be ready to merge.

PR #667, opened 2023-07-03
[FIX] - don't copy release channel when duplicating a picking

sbejaoui:


rousseldenis (approved):

rousseldenis:
Seems legit.

/ocabot merge minor

jbaudoux:
What about backorders ?

PR #668, opened 2023-07-10
[14.0] shopfloor_reception: Fixes

mmequignon:
see atomic comments for details

ref: RAU-172


TDu (approved):

shopfloor/actions/data.py
```diff
@@ -93,13 +93,19 @@ def package(self, record, picking=None, with_packaging=False, **kw):
...
+                [
+                    ("picking_id.picking_type_id", "=", picking.picking_type_id.id),
+                    ("result_package_id", "=", record.id),
+                    ("state", "in", ["partially_available", "assigned"]),
+                ]
             )
-            data.update({"move_line_count": len(lines)})
+            qty += move_line_count
+            # TODO does this name really makes sense?
+            data.update({"move_line_count": qty})
```
sebalix, line 108: As we are now mixing a number of quants + a number of move lines not `done`, this deserves a comment explaining why we are doing this.
I'm still trying to understand why the previous implementation was giving wrong results, it's not obvious.

jbaudoux: Because each processed qty is extracted in a new picking. So picking.move_line_ids.filtered(state="done") returns nothing. Moreover, there could also be other receptions on the same destination package.
The best result is given by counting what is already on the destination package.

jbaudoux (approved):

JuMiSanAr (approved):

PR #669, opened 2023-07-10
[14.0] Add  shopfloor_reception_refund_return

mmequignon:
This module, atm, just sets `to_refund` on return moves created by shopfloor.


shopfloor_account/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Glue module between `shopfloor` and `stock_account`
```
TDu: Maybe explain why ?

TDu (approved):

shopfloor_reception/migrations/14.0.2.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+except (ImportError, IOError) as err:
+    _logger.debug(err)
+
+
+def move_fields_to_new_module(cr):
+    # is_shopfloor_created has been wrongly put in shopfloor_reception.
+    # this script moves it in shopfloor.
+    openupgrade.update_module_moved_fields(
+        cr, "stock.picking", "is_shopfloor_created", "shopfloor_reception", "shopfloor"
+    )
```
sebalix: <s>Does it work well if `shopfloor` is upgraded first (which will create the `stock_picking.is_shopfloor_created` column in DB), then comes the upgrade of `shopfloor_reception` with this script?</s>
EDIT: it's OK, I said nothing

shopfloor_account/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "name": "Shopfloor Account",
+    "summary": "Glue module between shopfloor and stock_account",
+    "version": "14.0.1.0.0",
+    "category": "Inventory, Accounting",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "maintainers": ["mmequignon"],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": False,
```
sebalix:
```suggestion
    "auto_install": True,
```

sebalix (commented):
`auto_install` flag to fix

shopfloor_account/__manifest__.py
jbaudoux: I would keep auto install False as this is not a glue module but an optional module

jbaudoux (changes requested):
This is not a glue module.
This is an optional module that enables the refund of all the created reception returns

Can you rename the module to `shopfloor_reception_refund_returns` ?

simahawk:
Final "s" should be removed from the name :wink:

jbaudoux (approved):
@TDu @sebalix I changed the module name, rewrote commit messages. Ok for you?

jbaudoux:
> Final "s" should be removed from the name wink

removed

shopfloor_reception/migrations/14.0.2.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from openupgradelib import openupgrade
+except (ImportError, IOError) as err:
+    _logger.debug(err)
```
jbaudoux: Why `try?. If it cannot be imported, the migration will fail. Why not adding it in the repo requirements.txt ?

jbaudoux (changes requested):

shopfloor_reception/migrations/14.0.2.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def move_fields_to_new_module(cr):
+    # is_shopfloor_created has been wrongly put in shopfloor_reception.
+    # this script moves it in shopfloor.
+    openupgrade.update_module_moved_fields(
+        cr, "stock.picking", "is_shopfloor_created", "shopfloor_reception", "shopfloor"
```
jbaudoux: It raises an error:
```
shopfloor_reception/migrations/14.0.2.1.0/pre-migration.py", line 17, in move_fields_to_new_module
    openupgrade.update_module_moved_fields(
  File "python3.8/site-packages/openupgradelib/openupgrade.py", line 2676, in update_module_moved_fields
    do_raise("moved_fields %s must be a tuple or list!" % moved_fields)
```

shopfloor_reception/migrations/14.0.2.1.0/pre-migration.py
TDu: Agree. Removed the try and added the dependency.

TDu:
Rebased and added the dependency on openupgradelib.
But also IMO the shopfloor module should have its version bumped (a new field is added).
So shouldn't the script be actually in the shopfloor module ?

shopfloor_reception/__manifest__.py
```diff
@@ -1,7 +1,7 @@
 {
     "name": "Shopfloor Reception",
     "summary": "Reception scenario for shopfloor",
-    "version": "14.0.2.3.0",
+    "version": "14.0.2.4.0",
```
jbaudoux, line 4: the migration script must have the same version

TDu: changed

jbaudoux (approved):

simahawk:
@mmequignon pls rebase

TDu:
@simahawk  Can we reopen this one

simahawk:
I cannot. It says "the branch was force-pushed or recreated". You'll have to open a new one.

TDu:
I have opened https://github.com/OCA/wms/pull/669 to follow up

PR #670, opened 2023-07-10
[14.0] shopfloor single_pack_transfer: display weight in package info card

JuMiSanAr:
image

ref: rau-182

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/single_pack_transfer.js
+++ b/shopfloor_mobile/static/wms/src/scenario/single_pack_transfer.js
@@ -76,6 +76,12 @@ const SinglePackTransfer = {
                     :card_color="utils.colors.color_for('screen_step_done')"
                 >
                     <template v-slot:after_details>
+                        <v-card-subtitle>
+                            <span class="font-weight-bold">Weight:</span>
+                            <span>
+                                {{ _get_pack_weight() }}
+                            </span>
+                        </v-card-subtitle>
                         <v-card-text class="details pt-0">
                             <div v-for="product in state.data.products" class="field-detail">
                                 {{ product.display_name}}
@@ -108,6 +114,17 @@ const SinglePackTransfer = {
                 },
             },
         };
+    },
+    methods: {
+        _get_pack_weight: function () {
+            let weight = this.state.data.weight;
+            let uom = this.state.data.weight_uom;
+            if (!weight) {
+                weight = this.state.data.estimated_weight_kg;
+                uom = "kg";
+            }
+            return weight.toFixed(3) + " " + uom;
+        },
     },
 };
 process_registry.add("single_pack_transfer", SinglePackTransfer);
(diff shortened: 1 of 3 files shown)


TDu (approved):

shopfloor_mobile/static/wms/src/scenario/single_pack_transfer.js
```diff
@@ -76,6 +76,11 @@ const SinglePackTransfer = {
                     :card_color="utils.colors.color_for('screen_step_done')"
                 >
                     <template v-slot:after_details>
+                        <v-card-subtitle>
+                            <span>
+                                {{ state.data.picking.weight.toFixed(2)}} {{state.data.weight_uom}}
```
jbaudoux: The picking weight is an estimation based on product data. And there could be more than just the pack in the picking.

I would display the package `pack_weight` or, if empty, the `estimated_pack_weight_kg` converted to the package weight_uom_id

Maybe add the `estimated_pack_weight` on stock-logistics-workflow/stock_quant_package_dimension

jbaudoux (changes requested):

shopfloor_mobile/static/wms/src/scenario/single_pack_transfer.js
```diff
@@ -76,6 +76,11 @@ const SinglePackTransfer = {
                     :card_color="utils.colors.color_for('screen_step_done')"
                 >
                     <template v-slot:after_details>
+                        <v-card-subtitle>
```
jbaudoux, line 79: Maybe add a title like:
Weight: 14.25 kg

shopfloor_mobile/static/wms/src/scenario/single_pack_transfer.js
jbaudoux: Weight is usually displayed with 3 decimals

JuMiSanAr: I see the exact opposite was explicitly made some time ago, to make sure `estimated_pack_weight` was always in kg:

image

Do we wanna change that back?

I've modified the scenario to use the uom if there's a `pack_weight`, and otherwise use `estimated_pack_weight_kg` and display it as kg. What do you think?

jbaudoux: Sounds great. Indeed if there is no pack_weight, there is likely no uom, so we better display estimated_pack_weight_kg in kg

jbaudoux (approved):

jbaudoux:
@JuMiSanAr Can you squash the commits for merging?

shopfloor_mobile/static/wms/src/scenario/single_pack_transfer.js
```diff
@@ -109,6 +115,17 @@ const SinglePackTransfer = {
             },
         };
     },
+    methods: {
+        _get_pack_weight: function () {
```
simahawk, line 119: I see no reason for making it a "private" method but no harm :)

simahawk (approved):

PR #672, opened 2023-07-11
[14.0][IMP] shopfloor_reception: Allow users to work concurrently

mmequignon:
ping @jbaudoux

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -748,9 +748,43 @@ class Reception(Component):
             message=message,
         )
 
+    def _align_product_uom_qties(self, move):
+        # This method aligns product uom qties on move lines.
+        # In the shopfloor context, we might have multiple users working at
+        # the same time on the same move. This is done by creating one move line
+        # per user, with shopfloor_user_id = user.
+        # This method ensures that the product_uom_qty reflects what remains to
+        # be done, so we can display coherent numbers on the UI.
+
+        # for a given line, product_uom_qty is computed as this:
+        # remaining_todo = move.product_uom_qty - move.quantity_done
+        # line.product_uom_qty = line.qty_done + remaining_todo
+
+        # However, if the overall qty_done is > to the move's product_uom_qty,
+        # then we're not updating the line's product_uom_qty.
+
+        # TODO, do we need to check move's state?
+        # If move is already done, do not update lines qties
+        # if move.state in ("done", "cancel"):
+        #     return
+
+        qty_todo = move.product_uom_qty
+        qty_done = sum(move.move_line_ids.mapped("qty_done"))
+        rounding = move.product_id.uom_id.rounding
+        compare = float_compare(qty_done, qty_todo, precision_rounding=rounding)
+        if compare < 1:  # If qty done <= qty todo, align qty todo on move lines
+            remaining_todo = qty_todo - qty_done
+            # if we didn't bypass reservation update, the quant reservation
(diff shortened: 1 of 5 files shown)


shopfloor_reception/tests/test_set_quantity.py
```diff
@@ -381,3 +381,137 @@ def test_scan_new_package(self):
...
+        self.assertEqual(selected_move_line.qty_done, 3.0)
+        self.assertEqual(selected_move_line.product_uom_qty, 3.0)
+
+        self.assertEqual(new_line.qty_done, 7.0)
+        self.assertEqual(new_line.product_uom_qty, 7.0)
+
+        # And everything's fine on the move
+        move = move_lines.move_id
+        self.assertEqual(move.product_uom_qty, move.quantity_done)
+        self.assertEqual(move.product_uom_qty, 10.0)
```
mmequignon, line 547: need one more test:
What happens if move quantity_done == move.product_uom_qty, and some other user starts working on this?

jbaudoux: A new move line with product_uom_qty=0 will be created here https://github.com/OCA/wms/blob/380408f11c56caded17e753306cbc1eea303eb61/shopfloor_reception/services/reception.py#L264
Seems already fine as user will see in red that the quantity_done > 0

shopfloor_reception/services/reception.py
```diff
@@ -748,9 +748,21 @@ def _response_for_set_lot(self, picking, line, message=None):
...
 
+    def _align_product_uom_qties(self, move):
+        qty_todo = move.product_uom_qty
+        qty_done = sum(move.move_line_ids.mapped("qty_done"))
+        rounding = move.product_id.uom_id.rounding
+        compare = float_compare(qty_done, qty_todo, precision_rounding=rounding)
+        if compare < 1:  # If qty done <= qty todo, align qty todo on move lines
+            remaining_todo = qty_todo - qty_done
+            for line in move.move_line_ids:
+                line.product_uom_qty = line.qty_done + remaining_todo
```
TDu, line 771: Does this work if there is multiple move lines on the move ?

mmequignon: yes, that's the point.
Look at the test I added: 2 users working on the same move.
Each one gets its own move line and is working in //
If we have a move with product_uom_qty = 10, each move line has qty_done = 2.
Then quantity_done on the move is 4. We're missing 6 units.
Since both move lines has qty_done 2, then the max qty should be updated to 2 + 6 == 8.
Is it clearer?

shopfloor_reception/tests/test_set_quantity.py
mmequignon: I mean, are we sure product_uom_qty will be 0?

shopfloor_reception/services/reception.py
TDu: But then on the move qty to do is 6, but the total on move lines is 12 ?! 
Maybe this reasoning makes sense on the shopfloor app level (only for this scenario ?) because each move line is a presentation of the state of the move to the user (I am not sure of that). But at Odoo level it does not make sense (or for me anyway). And if I am partially right there should be a docstring on this function.
(I don't know the detail implementation of this scenario)

TDu (approved):

jbaudoux (approved):
Tested

JuMiSanAr (approved):

PR #673, opened 2023-07-11
[IMP] - stock_available_to_promise_release: Improve unrelease check message

sbejaoui:
It's hard to debug why a move can't be unreleased. This PR adds more details about the picking and its origins.


rousseldenis (approved):
Code review.

Even if a little test should be great

lmignon (approved):

PR #674, opened 2023-07-13
[14.0][DO NOT MERGE] shopfloor: search product by barcode only

mmequignon:


simahawk (changes requested):
It should be made configurable

PR #678, opened 2023-07-17
[14.0][FIX] shopfloor reception: fix select destination package with location

TDu:


shopfloor_reception/services/reception.py
```diff
@@ -623,22 +630,22 @@ def _set_quantity__by_package(self, picking, selected_line, package):
...
-        # screen
+        line.result_package_id = package
+        return self._response_for_select_move(picking)
+
+    def _set_quantity__by_package(self, picking, selected_line, package):
+        if package.location_id:
+            response = self._set_package_on_move_line(picking, selected_line, package)
+            return response
+        # Package with no location, move to location selection screen
         selected_line.result_package_id = package
```
```suggestion
    def _set_quantity__by_package(self, picking, selected_line, package):
        response = self._set_package_on_move_line(picking, selected_line, package)
        if response:
            return response
```

shopfloor_reception/services/reception.py
```diff
@@ -1301,11 +1308,8 @@ def select_dest_package(
                     selected_line,
                     message=self.msg_store.package_not_empty(package),
                 )
-            selected_line.result_package_id = package
-            response = self._post_line(selected_line)
-            if response:
-                return response
-            return self._response_for_select_move(picking)
+            response = self._set_package_on_move_line(picking, selected_line, package)
+            return response
```
```suggestion
            if response:
                return response
            return self._response_for_select_move(picking)
```

jbaudoux (changes requested):
Code looks better structured to me with those last changes

TDu:
@jbaudoux I refactored following your comment and some more...

jbaudoux (approved):
Thanks, looks good now

shopfloor_reception/services/reception.py
```diff
@@ -599,47 +599,53 @@ def _set_quantity__by_packaging(self, picking, selected_line, packaging):
...
-                        selected_line.product_uom_qty
-                    ),
-                )
-            # If the scanned package has a valid destination,
-            # set both package and destination on the package.
-            selected_line.result_package_id = package
-            selected_line.location_dest_id = pack_location
+        line.result_package_id = package
+        line.location_dest_id = pack_location
+        response = self._post_line(line)
```
jbaudoux: Shouldn't this call to post line be outside this method?

TDu: It already is, so I removed it.

shopfloor_reception/services/reception.py
```diff
@@ -599,47 +599,50 @@ def _set_quantity__by_packaging(self, picking, selected_line, packaging):
...
+        line.location_dest_id = pack_location
 
-            response = self._post_line(selected_line)
-            if response:
-                return response
+    def _set_quantity__by_package(self, picking, selected_line, package):
+        response = self._set_package_on_move_line(picking, selected_line, package)
+        if response:
+            return response
+        if package.location_id:
```
jbaudoux, line 637: I still think you lost the call to `_post_line` when scanning a package that is already on a location
```suggestion
        if package.location_id:
            response = self._post_line(selected_line)
            if response:
                return response
```

shopfloor_reception/services/reception.py
```diff
@@ -1301,8 +1304,7 @@ def select_dest_package(
                     selected_line,
                     message=self.msg_store.package_not_empty(package),
                 )
-            selected_line.result_package_id = package
-            response = self._post_line(selected_line)
+            response = self._set_package_on_move_line(picking, selected_line, package)
```
jbaudoux, line 1348: And here also
```suggestion
            response = self._set_package_on_move_line(picking, selected_line, package)
            if response:
                return response
            response = self._post_line(selected_line)
```

jbaudoux (changes requested):
call to post line looks now broken

shopfloor_reception/services/reception.py
jbaudoux: good but you removed it also outside the method

TDu:
> call to post line looks now broken

Unfortunately it seems that this `_post_line` is not well covered by the unit test :thinking:

jbaudoux (approved):

jbaudoux:
@TDu Can you rebase and squash?

PR #642, opened 2023-05-25
[16.0][MIG] - shopfloor_base

sbejaoui:

PR #679, opened 2023-07-17
[16.0][MIG] - shopfloor

sbejaoui:
needs:

- #642 
- https://github.com/OCA/stock-logistics-workflow/pull/1330
- https://github.com/OCA/product-attribute/pull/1215
- https://github.com/OCA/stock-logistics-workflow/pull/1469


jbaudoux:
@sbejaoui Why you dropped the migration scripts ?

sbejaoui:
> @sbejaoui Why you dropped the migration scripts ?

Based on the migration [procedure](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0), migration scripts should be about the new version, and the scripts for the old versions should be removed (as it's supposed to have been executed already).

jbaudoux:
> > @sbejaoui Why you dropped the migration scripts ?
> 
> Based on the migration [procedure](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0), migration scripts should be about the new version, and the scripts for the old versions should be removed (as it's supposed to have been executed already).

Ah thanks for the info :)

shopfloor/services/cluster_picking.py
```diff
@@ -1122,17 +1122,19 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=False):
...
+    def _unload_set_picking_to_done(self, picking, picking_lines):
+        if picking.state == "done":
+            return
+        # We set the picking to done only when the last line is
+        # unloaded to avoid backorders.
+        if all(line.shopfloor_unloaded for line in picking.move_line_ids):
+            picking._action_done()
         if self.work.menu.unload_package_at_destination:
-            lines.result_package_id = False
+            picking_lines.result_package_id = False
```
lmignon: @sbejaoui It's difficult to know what's the semantic behind this new method '_unload_set_picking_to_done'. In your change, you still iterate multi times n the lines.....



```python
from collections import namedtuple

        lines.write({"shopfloor_unloaded": True, "location_dest_id": location.id})
        lines.package_level_id.location_dest_id = location
        LineIdsByPickingValue =  namedtuple("LINESBYP", ["ids", "all_unloaded_status"])
        bypicking = {} 
        for line in lines:
            if line.picking_id.state == "done":
                continue
            value = lines_by_picking_value[line.picking_id]
            if value:
                value = LineIdsByPickingValue(ids = value.ids + line.ids, all_unloaded_status = value.all_unloaded_status and line.shopfloor_unloaded)
            else:
                value = LineIdsByPickingValue(ids=line.ids, all_unloaded_status=line.shopfloor_unloaded)
            lines_by_picking_value[line.picking_id] = value

        for picking, line_values in lines_by_picking_value.items():
            if line_values.all_unloaded_status:
                picking._action_done()
           if self.work.menu.unload_package_at_destination:
               self.env["stock.move.line"].browse(line_values.ids).write({"result_package_id" : False})
     
```

```

Comment (sbejaoui) [reply shopfloor/services/cluster_picking.py]: This refactoring has the purpose of separating the picking validation from the line location update. The idea behind it is to enable the addition of custom behavior during picking validation. Your proposition won't help, as we are still handling all picking in the same method

File: shopfloor/services/cluster_picking.py [hunk 2]
```diff
@@ -1122,17 +1122,19 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=False):
...
-        for line in lines:
-            # We set the picking to done only when the last line is
-            # unloaded to avoid backorders.
-            picking = line.picking_id
-            if picking.state == "done":
-                continue
-            picking_lines = picking.mapped("move_line_ids")
-            if all(line.shopfloor_unloaded for line in picking_lines):
-                picking._action_done()
+        for picking in lines.batch_id.picking_ids:
```
sbejaoui, line 1125: @lmignon ,I replaced the loop on the lines with a loop on the pickings. I don't understand how it iterates multiple times on the lines

shopfloor/services/location_content_transfer.py
```diff
@@ -764,9 +764,10 @@ def set_destination_line(
         self._lock_lines(move_line)
 
         move_line.qty_done = quantity
-        remaining_move_line = move_line._split_partial_quantity()
+        # If a move line reference a move linked to others move lines (different
+        # locations), we have to split the move to process only the lines related to
+        # the current location.
         move_line._extract_in_split_order({"user_id": self.env.uid})
```
jbaudoux: How is that possible as when you scan the location at the start, you isolate the move lines from that location in a split order? So at this stage inside the picking we should only have move lines from that location.

In set_destination_all and set_destination_package, I think such is not done. We should have the same behavior in all 3 cases.

I would have dropped that line also unless you can reproduce the use case.

cc @TDu @sebalix

lmignon: It's required by a test (https://github.com/OCA/wms/pull/679/files#diff-d468159958b3bfcb32afaa09c6a77a69e1811aa92d2abe6d4022f1be66a9ad19R637) but reading the code, It seems that the test is not right since the move_line to process is not selected by using the `scan_location` method and therefore the initial split is not done... I'll fix the test and remove the line.

lmignon: @jbaudoux It's fixed into the last commit https://github.com/OCA/wms/pull/679/commits/dbab97dca13c8402b45275840363f8b54ec7ce06

lmignon:
@rousseldenis @jbaudoux This one is finally :green_circle: Can we proceed to the merge so I can move forward with others PRs and thereafter forward port  the last changes from 14.0

lmignon:
An important element for the future. Shopfloor should be split into several smaller modules (1 per scenario). The current approach is not healthy for maintenance.

shopfloor/actions/inventory.py
```diff
@@ -51,8 +51,10 @@ def _get_existing_quant(self, location, product, package=None, lot=None, limit=1
             domain.append(("lot_id", "=", False))
         return self.inventory_model.search(domain, limit=limit)
 
-    def _create_draft_inventory(self, location, product, lot=None):
-        quants = self._get_existing_quant(location, product, lot=lot, limit=None)
+    def _create_draft_inventory(self, location, product, package=None, lot=None):
```
rousseldenis, line 54: This should be completed with owner_id field too.

rousseldenis (approved):

shopfloor/actions/inventory.py
lmignon: maybe after the forward port from  v14

lmignon:
ping @simahawk Can we merge this.... It works in production for months (cluster + location_content) and we want to finalize all the pending PRs on shopfloor  for v16 and forward port the last changes from v14 to have both branches aligned.

rousseldenis:
Let's merge this one. It will allow to do enhancements in separate PR's.

/ocabot merge nobump

shopfloor/services/location_content_transfer.py
```diff
@@ -450,7 +450,7 @@ def _find_transfer_move_lines(self, location):
     # hook used in module shopfloor_checkout_sync
     def _write_destination_on_lines(self, lines, location):
         lines.location_dest_id = location
-        lines.package_level_id.location_dest_id = location
+        lines.package_level_id.picking_id.location_dest_id = location
```
jbaudoux: I think this change is not correct and should be reverted

cc @rousseldenis @lmignon @mmequignon

shopfloor/services/location_content_transfer.py
```diff
@@ -0,0 +1,1194 @@
...
+        """Find move lines currently being moved by the user"""
+        lines = self.env["stock.move.line"].search(
+            self._find_transfer_move_lines_domain(location)
+        )
+        return lines
+
+    # hook used in module shopfloor_checkout_sync
+    def _write_destination_on_lines(self, lines, location):
+        lines.location_dest_id = location
+        lines.package_level_id.picking_id.location_dest_id = location
```
jbaudoux, line 453: This line needs fixing. Each package taken from the source location can have their own destination

PR #681, opened 2023-07-18
[14.] shopfloor checkout: always update qty done after selecting line

JuMiSanAr:
Make sure the qty of the line is set to 1 after selecting / scanning a line if the no_prefill_qty option is enabled

ref: rau-184

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -464,7 +464,9 @@ class Checkout(Component):
             picking, message=self.msg_store.barcode_not_found()
         )
 
-    def _select_lines_from_package(self, picking, selection_lines, package, **kw):
+    def _select_lines_from_package(
+        self, picking, selection_lines, package, prefill_qty=0, **kw
+    ):
         lines = selection_lines.filtered(
             lambda l: l.package_id == package and not l.shopfloor_checkout_done
         )
@@ -478,7 +480,7 @@ class Checkout(Component):
                     ),
                 },
             )
-        self._select_lines(lines)
+        self._select_lines(lines, prefill_qty=prefill_qty)
         if self.work.menu.no_prefill_qty:
             lines = picking.move_line_ids
         return self._response_for_select_package(picking, lines)
@@ -514,7 +516,9 @@ class Checkout(Component):
         elif packages:
             # Select all the lines of the package when we scan a product in a
             # package and we have only one.
-            return self._select_lines_from_package(picking, selection_lines, packages)
+            return self._select_lines_from_package(
+                picking, selection_lines, packages, prefill_qty=prefill_qty
+            )
         else:
             # There is no package on selected lines, so also select all other lines
             # not in a package. But only the quantity on first selected lines
@@ -533,7 +537,9 @@ class Checkout(Component):
             picking, selection_lines, packaging.product_id, prefill_qty=packaging.qty
         )
 
-    def _select_lines_from_lot(self, picking, selection_lines, lot, **kw):
+    def _select_lines_from_lot(
+        self, picking, selection_lines, lot, prefill_qty=1, **kw
+    ):
         lines = selection_lines.filtered(lambda l: l.lot_id == lot)
         if not lines:
             return self._response_for_select_line(
@@ -561,10 +567,10 @@ class Checkout(Component):
             # Select all the lines of the package when we scan a lot in a
             # package and we have only one.
             return self._select_lines_from_package(
-                picking, selection_lines, packages, **kw
+                picking, selection_lines, packages, prefill_qty=prefill_qty, **kw
             )
 
-        self._select_lines(lines, prefill_qty=1)
+        self._select_lines(lines, prefill_qty=prefill_qty)
         return self._response_for_select_package(picking, lines)
 
     def _select_lines_from_serial(self, picking, selection_lines, lot, **kw):


simahawk:
I guess there was no test for that?

JuMiSanAr:
> 

Added.

JuMiSanAr:
> I guess there was no test for that?

Added.

simahawk (approved):

jbaudoux (changes requested):
Can you elaborate the use case? Which screen needs fixing?
Currently when you select the line in the second screen, it sets the quantity done = todo
It doesn't look right to always set the qty done to 1.

JuMiSanAr:
> Can you elaborate the use case? Which screen needs fixing? Currently when you select the line in the second screen, it sets the quantity done = todo It doesn't look right to always set the qty done to 1.

This is to solve the issue where we scan a product / lot / packaging / package and we don't have the scanned lines selected with their qty increased.

I've updated the case where we scan a delivery packaging, it increases by the packaging qty now.

simahawk (changes requested):
Looking again at this: seems wrong.
`prefill_qty` should be always be controlled by the flag.
Please check what has been done in other scenario (zone picking, cluster picking, etc) where there's `_get_prefill_qty` to centralize it.

JuMiSanAr:
> Looking again at this: seems wrong. `prefill_qty` should be always be controlled by the flag. Please check what has been done in other scenario (zone picking, cluster picking, etc) where there's `_get_prefill_qty` to centralize it.

It's basically the same implementation as in the other scenarios, where you can find things like:
`quantity = self._get_prefill_qty(move_line, packaging.qty)`
and
`quantity = self._get_prefill_qty(move_line, 1.0)`

Basically, if the option is enabled, we want to select a line and increase `qty_done` by 1 or by the packaging qty.

In checkout, this is done inside the `_select_lines` method:
```
         if self.work.menu.no_prefill_qty and i == 0:
              # For prefill quantity we only want to increment one line
              line.qty_done += prefill_qty
          elif not self.work.menu.no_prefill_qty:
              line.qty_done = line.product_uom_qty
```

shopfloor/services/checkout.py
```diff
@@ -478,7 +478,7 @@ def _select_lines_from_package(self, picking, selection_lines, package, **kw):
                     ),
                 },
             )
-        self._select_lines(lines)
+        self._select_lines(lines, prefill_qty=1)
```
jbaudoux: @JuMiSanAr But this new argument is not used ??

JuMiSanAr: What do you mean? It's used to increase qty_done inside this method:

```
            if self.work.menu.no_prefill_qty and i == 0:
                # For prefill quantity we only want to increment one line
                line.qty_done += prefill_qty
```

jbaudoux: wrong reading from me

jbaudoux (approved):
Tested

shopfloor/services/checkout.py
jbaudoux: When scanning a source package, we don't increase the quantity
```suggestion
        self._select_lines(lines, prefill_qty=0)
```

jbaudoux (changes requested):
We increase by one the qty when scanning a product or lot.
We increase by the product packaging qty when scanning a product packaging.
We don't increase when scanning a location or package

shopfloor/services/checkout.py
```diff
@@ -600,7 +600,7 @@ def _select_lines_from_delivery_packaging(
             )
         if confirm_pack_all:
             # Select all lines and pack them all w/o passing for select_package state
-            self._select_lines(selection_lines)
+            self._select_lines(selection_lines, prefill_qty=packaging.qty)
```
jbaudoux: This can't be right. A delivery package is scanned to put lines in a new package.

I have the impression this feature is not properly made @simahawk 

* If no lines are selected (i.e. none have a done qty) and no_qty_prefill, then ask the user to select lines => missing
* If no lines are selected (i.e. none have a done qty) and not no_qty_prefill, then confirm_put_all_goods_in_delivery_package => how selection_lines is build? Then on confirmation, select_lines and put in pack.
* If lines are selected (i.e. have a done qty), then don't ask confirmation, don't change quantities, just _create_and_assign_new_packaging

shopfloor/services/checkout.py
```diff
@@ -628,7 +628,7 @@ def _select_line_move_line(self, picking, selection_lines, move_line):
             return self._select_lines_from_package(
                 picking, selection_lines, move_line.package_id
             )
-        self._select_lines(move_line)
+        self._select_lines(move_line, prefill_qty=1)
```
jbaudoux: No, nothing has been scanned, so no quantity increase

jbaudoux (changes requested):
All fixes here are wrong.

The only bug regarding the PR title is when `self._select_lines_from_package` is called from `_select_lines_from_product` or `_select_lines_from_lot` where prefill_qty should be set to 1.
This is when the scanned product/lot is part of a source package.

shopfloor/services/checkout.py
jbaudoux: This should be fixed in another PR (on bscos)

JuMiSanAr:
@jbaudoux updated.

jbaudoux (approved):

PR #682, opened 2023-07-19
[14.0][IMP] stock_warehouse_flow: assign_picking only if needed

mt-software-de:
Do not call assign_picking on apply
if it is getting called from action_confirm

cc @jbaudoux


mt-software-de:
@sebalix and @jbaudoux could also please review this one?

jbaudoux (commented):
I guess we lack a unit test for checking that after the flow is applied the delivery can be inside a different release channel

stock_warehouse_flow/models/stock_move.py
```diff
@@ -14,7 +15,7 @@ def _action_confirm(self, merge=True, merge_into=False):
             if not move._apply_flow_on_action_confirm():
                 continue
             flow = self.env["stock.warehouse.flow"]._search_for_move(move)
-            flow._apply_on_move(move)
+            flow._apply_on_move(move, assign_picking=False)
```
jbaudoux: A comment why you do this would be helpful for the future

mt-software-de:
> I guess we lack a unit test for checking that after the flow is applied the delivery can be inside a different release channel
yes but this should only occur if stock_release_channel and stock_warehouse_flow would be installed

jbaudoux:
> > I guess we lack a unit test for checking that after the flow is applied the delivery can be inside a different release channel

> yes but this should only occur if stock_release_channel and stock_warehouse_flow would be installed

Then we need a glue module that auto install. This module can contain just that test.

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -439,7 +439,7 @@ def apply_on_move(self, move):
             # Try to apply the rest of the flows to the split move
             for split_move in split_moves:
                 move_ids += (flows - flow).apply_on_move(split_move).ids
-            flow._apply_on_move(move)
+            flow._apply_on_move(move, assign_picking)
```
RodrigoBM, line 442:
```suggestion
            flow._apply_on_move(move, assign_picking=assign_picking)
```

RodrigoBM (changes requested):

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -421,14 +421,14 @@ def _search_for_move(self, move):
...
 
     @api.model
-    def _search_and_apply_for_move(self, move):
+    def _search_and_apply_for_move(self, move, assign_picking=True):
         move.ensure_one()
         flows = self._search_for_move(move)
         if not flows:
             return move
-        return flows.apply_on_move(move)
+        return flows.apply_on_move(move, assign_picking)
```
mt-software-de, line 429: Thx for your Review.
I am fine with calling the Method without the Keyword.

jbaudoux (approved):

PR #683, opened 2023-07-19
[14.0][IMP] shopfloor_mobile_base: Display manual-select list-items separted

mt-software-de:
This allows to display the list-items of the manual-select component as separate items. 
Set list_items_separated = true
Then it will display the items like this
image
instead of
image

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/components/manual_select.js
+++ b/shopfloor_mobile_base/static/wms/src/components/manual_select.js
@@ -1,6 +1,7 @@
 /**
  * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
  * @author Simone Orsi <simahawk@gmail.com>
+ * Copyright 2022 Michael Tietz (MT Software) <mtietz@mt-software.de>
  * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
  */
 import event_hub from "../services/event_hub.js";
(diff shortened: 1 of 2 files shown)


mt-software-de:
cc @jbaudoux @simahawk

simahawk (changes requested):
Thanks for you contrib, few discussion items:

* I'm not sure we need a specific param for this... `card_klass` should be enough
* can we avoid defining our own shadow rules? It could be possible to get them from vuetifyjs
* we should support the same in list.js IMO
*

mt-software-de:
> * vuetifyjs

mt-software-de:
> Thanks for you contrib, few discussion items:
> 
>     * I'm not sure we need a specific param for this... `card_klass` should be enough

I would provide a specific parameter as that makes it easier to know what the component is capable of . 

> 
>     * can we avoid defining our own shadow rules? It could be possible to get them from vuetifyjs

i have found .v-stepper this is adding the same shadow so we can use this.

> 
>     * we should support the same in list.js IMO
Yes this would be good.

simahawk (approved):

jbaudoux (commented):
We can improve how the list is displayed but I don't think we should have different display in different scenario.
I liked the fact that each element is inside the same box. Maybe just a line separator is enough. This is displayed on a mobile screen and we should not loose vertical space.
Just my opinion...

mt-software-de:
> We can improve how the list is displayed but I don't think we should have different display in different scenario. I liked the fact that each element is inside the same box. Maybe just a line separator is enough. This is displayed on a mobile screen and we should not loose vertical space. Just my opinion...

Yes we are loosing some space with this option, but this is not fixed. 
I wanted to achieve a lookalike of the item-detail-card.

mt-software-de:
How should we go on with that PR? Should we merge it?
Or close it? Maybe a better solution would be to introduce different themes for the shopfloor app?

PR #695, opened 2023-07-28
[16.0][ADD] stock_release_channel_plan

jbaudoux:
Make a wake-up plan of release channels

image

cc @lmignon @sbejaoui @rousseldenis @sebalix @TDu


stock_release_channel_plan/wizards/launch_plan.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.model
+    def _action_launch(self, channels):
+        channels.filtered("is_action_unlock_allowed").action_unlock()
+
+        channels_to_wakeup = channels.filtered("is_action_wake_up_allowed")
+        channels_to_wakeup.action_wake_up()
+
+    def action_launch(self):
+        self.ensure_one()
+        channels_to_launch = self.preparation_plan_id.get_channels_to_launch()
```
lmignon: @jbaudoux If you ask to a plan the channels to launch. It's not possible to extend the wizard with specific additional paramters used to filter the channels to launch from the plan as it's done today into Alcyon code

stock_release_channel_plan/wizards/launch_plan.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.model
+    def _action_launch(self, channels):
+        channels.filtered("is_action_unlock_allowed").action_unlock()
+
+        channels_to_wakeup = channels.filtered("is_action_wake_up_allowed")
+        channels_to_wakeup.action_wake_up()
+
+    def action_launch(self):
+        self.ensure_one()
+        channels_to_launch = self.preparation_plan_id._get_channels_to_launch()
```
lmignon, line 34: @jbaudoux In our case, the channels to launch could be filtered according to additional criteria on the wizard. The `_get_channels_to_launch` method should therefore be provided by the wizard

lmignon (changes requested):

stock_release_channel_plan/wizards/launch_plan.py
jbaudoux: This was done on purpose as the plan should define the rules and not the wizard. That's the purpose of having a plan. Otherwise you have half a plan and the wizard is providing the missing other half. This is not ideal.
I keep it like that for now.

TDu (approved):
LG

simahawk (changes requested):
~LGTM but the missing handling of the wizard method~

stock_release_channel_plan/wizards/launch_plan.py
simahawk, line 34: To address @lmignon concern...

```suggestion
        channels_to_launch = self._get_channels_to_launch()
        [...]
    def _get_channels_to_launch(self):
        return self.preparation_plan_id._get_channels_to_launch()
```

simahawk (approved):

stock_release_channel_plan/wizards/launch_plan.py
```diff
@@ -0,0 +1,40 @@
...
+        self._action_launch(channels_to_launch)
```
simahawk, line 35: This must be defined and do nothing or raise NotImplementedError. Not a blocker but still...

jbaudoux: This is defined here https://github.com/OCA/wms/pull/695/files#diff-5db6170d4ad5dd149e55b3294fb038ecbc3abb6e7bcfac7ec64e9a4566968114R26

stock_release_channel_plan/wizards/launch_plan.py
jbaudoux: No, it's the plan providing the channels to launch, not the wizard. No need to create a second hook on the wizard

simahawk:
@lmignon good for you?

lmignon:
The problem is that these changes are not yet tested on our side. JE extracted this code from our project but didn't end the PR to use it back. We've to validate that we can still covers our needs with the change he made into the implementation.

simahawk:
Good to know, thanks for your feedback :) Have nice weekend!

cyrilmanuel (approved):
not an expert but LGTM

grindtildeath:
https://github.com/jbaudoux/wms/pull/1

jbaudoux:
rebased

jbaudoux:
@grindtildeath Can you check why `stock_release_channel_propagate_channel_picking` is now failing ?

grindtildeath:
@jbaudoux The test is failing because of this SQL query: https://github.com/OCA/wms/blob/16.0/stock_release_channel_process_end_time/models/stock_picking.py#L40-L56

Basically the query is returning pickings having a scheduled date that is before _any_ channel that has a process_end_date defined.

As we are adding new channels as demo data, these channels will have a process_end_date defined to today when the module is installed, and as tests are frozen in time in 2023, the pickings will always be scheduled before a channel. I'm not sure what is the right way to fix this as it took me already quite a while to figure out what's wrong and why 2023/01/28 was before 2024/02/27 :dizzy_face: 

EDIT: please check https://github.com/jbaudoux/wms/pull/3

stock_release_channel_plan/views/res_partner.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record model="ir.ui.view" id="res_partner_form_view">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="stock_release_channel.res_partner_form_view" />
+        <field name="arch" type="xml">
+            <field name="stock_release_channel_ids" position="after">
+                <field name="stock_release_channel_delivery_weekday_ids" />
```
santostelmo:
```suggestion
                <field name="stock_release_channel_delivery_weekday_ids"  widget="many2many_tags" />
```

cyrilmanuel (approved):

cyrilmanuel (approved):

jbaudoux:
rebased

jbaudoux:
I fixed test in stock_release_channel_propagate_channel_picking. Now it's again green :)
@lmignon Can we move on and merge this one ?

cyrilmanuel:
@lmignon , little ping, is it good for you ?

cyrilmanuel (approved):
LGTM

lmignon:
> I fixed test in stock_release_channel_propagate_channel_picking. Now it's again green :) @lmignon Can we move on and merge this one ?

@jbaudoux Yes, you can.

PR #700, opened 2023-08-01
[14.0] shopfloor_*: fix install/uninstall

simahawk:
Due to the way sf.app endpoints are registered
if at install/uninstall you don't take care of refreshing routes
you'll end up w/ non working services or stale entries in the route table.

See atomic commits for details.

ref: rau-160

### DIFF
--- a/shopfloor_base/models/shopfloor_app.py
+++ b/shopfloor_base/models/shopfloor_app.py
@@ -2,6 +2,8 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
+import logging
+
 from odoo import api, fields, models, tools
 from odoo.tools import DotDict
 
@@ -9,6 +11,8 @@ from odoo.addons.base_rest.tools import _inspect_methods
 from odoo.addons.component.core import _component_databases
 
 from ..utils import APP_VERSION, RUNNING_ENV
+
+_logger = logging.getLogger(__file__)
 
 
 class ShopfloorApp(models.Model):
@@ -281,6 +285,13 @@ class ShopfloorApp(models.Model):
         return comp_registry and comp_registry.ready
 
     def _get_services(self):
+        forced_services = self.env.context.get("sf_service_components")
+        if forced_services:
+            _logger.debug(
+                "_get_services forced services: %s",
+                ", ".join([x._usage for x in forced_services]),
+            )
+            return forced_services
         if not self._is_component_registry_ready():
             # No service is available before the registry has been loaded.
             # This is a very special case, when the odoo registry is being
--- a/shopfloor_base/utils.py
+++ b/shopfloor_base/utils.py
@@ -1,11 +1,17 @@
 # Copyright 2021 Camptocamp SA (http://www.camptocamp.com)
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 # @author Simone Orsi <simahawk@gmail.com>
+import logging
 import os
 from functools import wraps
 
 from odoo.modules.module import load_information_from_description_file
 from odoo.tools.config import config as odoo_config
+
+from odoo.addons.base_rest.controllers.main import _PseudoCollection
+from odoo.addons.component.core import WorkContext, _component_databases
+
+_logger = logging.getLogger(__name__)
 
 
 def ensure_model(model_name):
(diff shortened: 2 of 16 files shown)


simahawk:
Build failing for an issue present in 14.0 too.

shopfloor_base/utils.py
```diff
@@ -65,3 +71,66 @@ def _get_running_env():
...
+    work = WorkContext(model_name=rest_service._name, collection=collection)
+    return [comp(work) for comp in component_classes]
+
+
+def purge_endpoints(env, service_usage):
+    """Remove stale services' endpoints routes.
+
+    When scenario are removed (eg: module uninstalled)
+    their routes must be cleaned up.
+    You can use this func to easily clean them app for all apps.
```
TDu:
```suggestion
    You can use this function to easily clean them for all apps.
```

shopfloor_base/models/shopfloor_app.py
```diff
@@ -281,6 +285,13 @@ def _is_component_registry_ready(self):
         return comp_registry and comp_registry.ready
 
     def _get_services(self):
+        forced_services = self.env.context.get("sf_service_components")
+        if forced_services:
+            _logger.debug(
+                "_get_services forced services: %s",
+                ", ".join([x._usage for x in forced_services]),
+            )
+            return self.env.context.get("sf_service_components")
```
TDu:
```suggestion
            return forced_services
```

shopfloor_base/utils.py
```diff
@@ -65,3 +71,66 @@ def _get_running_env():
...
+    Motivation: registered routes for a given app are taken
+    from the endpoint_route table via `route_group`.
+    As the table is populated dynamically with registered components
+    when a component is removed there's no direct update to the table.
+    Hence, is the responsibility of the dev to take care of the cleanup.
+    """
+    apps = env["shopfloor.app"].search([])
+    for app in apps:
+        route = app.api_url_for_service(service_usage)
+        env.cr.execute("DELETE FROM endpoint_route WHERE route like %s", (f"{route}%",))
```
TDu, line 137:
```suggestion
        env.cr.execute("DELETE FROM endpoint_route WHERE route like %s", (f"{route}/%",))
```
I think the trailing slash is not included, this would be safer.

Also this function will not work for Odoo modules that extend an existing shopfloor scenario with new route.

simahawk: > I think the trailing slash is not included, this would be safer.

Why? We must clean them ALL when they belong to the service.

> Also this function will not work for Odoo modules that extend an existing shopfloor scenario with new route.

What do you mean? A new route will still be under the same base route of the scenario.

TDu: What if one day there is two services, the 2nd one starting with the name of the first one ('reception', `reception_something' ?

TDu: For the 2nd point I was thinking that this implementation can not be used for some module like in https://github.com/OCA/wms/pull/647 for the module `shopfloor_reception_packaging_dimension` but that is not blocking I guess just a warning

simahawk: Good catches, you are right. Please check my last commit: 
1. use "/" for main routes
2. accept "endpoint" if you want to affect only specific endpoints

Untested: could you give it a try for #647?

TDu: Tested and it works, with a little gotcha, when importing a service that extends an existing component `_usage`, `_name` are not set. More info in #647 when I push the fix

shopfloor_base/models/shopfloor_app.py
```diff
@@ -281,6 +285,13 @@ def _is_component_registry_ready(self):
         return comp_registry and comp_registry.ready
 
     def _get_services(self):
+        forced_services = self.env.context.get("sf_service_components")
+        if forced_services:
+            _logger.debug(
+                "_get_services forced services: %s",
+                ", ".join([x._usage for x in forced_services]),
+            )
+            return forced_services
```
mmequignon, line 294: shouldn't we add those to the regular result?

mmequignon (approved):
Not sure to understand when this is run.
I thought we also had issues when updating those modules.

But I don't know much about this. LG

shopfloor_base/utils.py
```diff
@@ -115,3 +115,22 @@ def load_components_without_registry(env, *component_classes):
         return []
     work = WorkContext(model_name=rest_service._name, collection=collection)
     return [comp(work) for comp in component_classes]
+
+
+def purge_endpoints(env, service_usage):
```
sebalix: So once merged, we should write uninstall hook using this helper for all Shopfloor modules?
EDIT: Ok this is what you did in the next commit https://github.com/OCA/wms/pull/700/commits/ab093b0223faaea5eed6269c113adc590f53957f !

sebalix (approved):

shopfloor_base/models/shopfloor_app.py
simahawk: There's no result when you use this (that's why we use it :wink: ).

simahawk:
> Not sure to understand when this is run. I thought we also had issues when updating those modules.
> 
> But I don't know much about this. LG

You must use it (as I did for `shopfloor_*` commit) when you add new services or new endpoints or modify them (eg: POST becomes GET).
@TDu is also using it in #647 .

TDu (approved):

PR #705, opened 2023-08-03
[14.0][FIX] stock_warehouse_flow: fix RollbackException on write

Ricardoalso:
In order to avoid UserError (_('You still have some product in locations %s') coming from stock module https://github.com/odoo/odoo/blob/076baeacbbdf92ae1b500e6716fd95d32acc4a53/addons/stock/models/stock_location.py#L114

Plus, fixes the exception handling to get real exceptions that were badly catched.

### DIFF
--- a/stock_warehouse_flow/models/stock_warehouse_flow.py
+++ b/stock_warehouse_flow/models/stock_warehouse_flow.py
@@ -10,6 +10,10 @@ from odoo.tools.safe_eval import safe_eval
 logger = logging.getLogger(__name__)
 
 
+class ForceRollback(Exception):
+    pass
+
+
 class StockWarehouseFlow(models.Model):
     _name = "stock.warehouse.flow"
     _description = "Stock Warehouse Routing Flow"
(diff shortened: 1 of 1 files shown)


stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -10,6 +10,10 @@
 logger = logging.getLogger(__name__)
 
 
+class RollbackException(Exception):
```
yvaucher: What about renaming it:

```suggestion
class ForceRollback(Exception):
```

While it can seems unconventional, I think `raise ForceRollback` is self explanatory.

yvaucher (approved):
Approving the patch and I confirm it was tested.

The except is too broad and any constraint that might occur on warehouse would result in a failure where we don't know the original cause.

Just adding a suggestion on the name of the exception.

(About the trick with the savepoint we might be able to trigger the savepoint differently or not use it at all but this is not a refactor just a patch.)

jbaudoux (approved):
LGTM

RodrigoBM (approved):

stock_warehouse_flow/models/stock_warehouse_flow.py
Ricardoalso: Indeed, I made the change

PR #708, opened 2023-08-04
[14.0] shopfloor_mobile: display only qty reserved for picking in picking-detail

JuMiSanAr:
The picking-detail component used to display the information of the picking was taking the "qty reserved" from the total amount of reserved qty for that product, regardless of the picking.

Now, we only display the reserved amount for the specific move.

ref: cos-4149


sebalix (approved):

jbaudoux (approved):
LGTM

PR #710, opened 2023-08-07
[14.0][IMP] shopfloor_single_product_transfer improve confirmation

TDu:


jbaudoux (approved):

simahawk (approved):

shopfloor_single_product_transfer/services/single_product_transfer.py
```diff
@@ -578,34 +578,35 @@ def _valid_dest_location_for_menu(self):
...
+                # Confirmation is valid, return no response
+                pass
+            else:
                 # Ask for confirmation
                 orig_location = move_line.location_dest_id
                 message = self.msg_store.confirm_location_changed(
                     orig_location, location
                 )
-                asking_confirmation = True
+                confirmation = location.barcode
```
jbaudoux: You have to compare with the scanned barcode (you need to add it in the method arguments)
You don't know how the location was matched. And this is also called when a package is scanned

jbaudoux (changes requested):

shopfloor_single_product_transfer/services/single_product_transfer.py
TDu: New edge case because of the improvement on the confirmation.
Indeed it can also be called when a package is scanned (the package is not empty so using the location).
But this will lead to an incoherent situation. The user scanned a package in a location that need confirmation so the user is being asked to confirm the location change by scanning the location again ? But the system will not remember it should be put in a package.

TDu: And do we not have a bug on https://github.com/OCA/wms/pull/710/files#diff-f871450f795713d4ab1061b7117262fdee975730cb8f85f5963d06d339d7e39dL723
Because if the user is being asked to confirm a location in the response. The `package_id` on the move.line should not yet be changed....

jbaudoux: The user scan a barcode and we ask to confirm by scanning the same barcode. You first put the scanned barcode as the value to confirm. You then need to compare the newly scanned barcode with the barcode that must be confirmed.
We don't care if the barcode is a package name, a location barcode, or whatever complex gs1 sscc label.
We just want the user to scan twice the same barcode.

jbaudoux: > And do we not have a bug on https://github.com/OCA/wms/pull/710/files#diff-f871450f795713d4ab1061b7117262fdee975730cb8f85f5963d06d339d7e39dL723 Because if the user is being asked to confirm a location in the response. The `package_id` on the move.line should not yet be changed....

No there is no such bug. The user scan a package that is on another location. We ask to confirm this location change. You set scanned barcode in `confirmation`. The user can confirm this change by scanning again the package.

TDu: But the destination package is set on the move.line before the user has confirmed the action !

mmequignon: No. `_set_quantity__check_location` is called before, right here
https://github.com/OCA/wms/pull/710/files#diff-f871450f795713d4ab1061b7117262fdee975730cb8f85f5963d06d339d7e39dR720-R723

mmequignon: yes, that is true.
I think we have to check that ask confirmation is not set in the response here (and there's no error) here.
This _set_quantity__check_location is part of the handlers.
https://github.com/OCA/wms/blob/0108725c0ca9b5332727e9144d17ead1006ff04e/shopfloor_single_product_transfer/services/single_product_transfer.py#L724
Regarding the barcode check, it can be done in the endpoint here https://github.com/OCA/wms/blob/0108725c0ca9b5332727e9144d17ead1006ff04e/shopfloor_single_product_transfer/services/single_product_transfer.py#L835
confirmation is true if barcode == confirmation, so you don't have to check this in the other methods.

shopfloor_single_product_transfer_mobile/static/src/scenario/single_product_transfer.js
```diff
@@ -239,9 +239,8 @@ const SingleProductTransfer = {
                         this.state.data.quantity = qty;
                     },
                     on_scan: (scanned) => {
-                        const confirmation = this.state.data.asking_confirmation
-                            ? true
-                            : false;
+                        const confirmation =
+                            this.state.data.asking_confirmation || null;
```
mt-software-de: shouldn't this be like
```suggestion
                            this.state.data.asking_confirmation || "";
```
like you doing it here? https://github.com/OCA/wms/pull/716

mt-software-de (changes requested):

shopfloor_single_product_transfer_mobile/static/src/scenario/single_product_transfer.js
TDu: changed, thanks for your review

jbaudoux (approved):

jbaudoux:
@TDu Can you squash and rebase?
@mt-software-de Can you update your review? Your remark has been fixed

mt-software-de (approved):
LGTM

TDu:
@jbaudoux squashed and rebased

PR #711, opened 2023-08-07
[14.0]  shopfloor_reception: take into account backorders qty done if auto_post is enabled

JuMiSanAr:
For reception, in the select_move screen, we were displaying the qty done for each move in the picking.
However, this would always be 0 if auto posting was enabled.

Now, it will display the qty done for the move (if it hasn't been posted elsewhere) + the qty done for that product in all related backorders.

ref: rau-120

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -245,8 +245,8 @@ class DataAction(Component):
                     ("product_id", "=", record.product_id.id),
                 ]
             )
-            qty_done = sum(related_moves.mapped("quantity_done"))
-            data.update({"backorders_quantity_done": qty_done})
+            backorders_qty_done = sum(related_moves.mapped("quantity_done"))
+            data["quantity_done"] += backorders_qty_done
         return data
 
     def moves(self, records, **kw):
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -236,7 +236,18 @@ class DataAction(Component):
     def move(self, record, **kw):
         record = record.with_context(location=record.location_id.id)
         parser = self._move_parser
-        return self._jsonify(record, parser)
+        data = self._jsonify(record, parser)
+        if kw.get("auto_post_enabled"):
+            # Return the quantities done for the related backorders.
+            related_moves = self.env["stock.move"].search(
+                [
+                    ("picking_id.backorder_id", "=", record.picking_id.id),
+                    ("product_id", "=", record.product_id.id),
+                ]
+            )
+            qty_done = sum(related_moves.mapped("quantity_done"))
+            data.update({"backorders_quantity_done": qty_done})
+        return data
 
     def moves(self, records, **kw):
         return [self.move(rec, **kw) for rec in records]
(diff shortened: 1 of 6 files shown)


shopfloor/actions/data.py
```diff
@@ -229,7 +229,20 @@ def _move_line_parser(self):
...
+            # Return the quantities done for the related backorders.
+            related_moves = self.env["stock.move"].search(
+                [
+                    ("picking_id.backorder_id", "=", record.picking_id.id),
+                    ("product_id", "=", record.product_id.id),
+                ]
+            )
+            qty_done = 0.0
+            for move in related_moves:
+                qty_done += move.quantity_done
```
mmequignon:
```python
qty_done = sum(related_moves.mapped("quantity_done"))
```

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -367,6 +367,9 @@ const Reception = {
                             path: "quantity_done",
                             label: "Qty done",
                             display_no_value: true,
+                            renderer: (rec, field) => {
+                                return rec.quantity_done + rec.backorders_quantity_done;
+                            },
```
mmequignon: this shouldn't be the frontend responsibility to compute that.

shopfloor/actions/data.py
mmequignon: see the other comment, and then, the front just displays the quantity done
```python
qty_done = data.get("qty_done", 0.0)
qty_done += sum(related_moves.mapped("quantity_done"))
data["qty_done"] = qty_done
```

shopfloor_reception_mobile/static/src/scenario/reception.js
JuMiSanAr: Hmmmm I think it's important to make the distinction between the qty done in the picking and the qty done from the backorders, isn't it?
Because what if auto_post is not enabled, we start working on a picking, and then we enable it and keep working on it, we may not have accurate qties done.

Also, qty_done can be used anywhere in the scenario, we will probably want to keep it untouched as it is now.

To me the simplest is to give the frontend both pieces of information and let it decide what to do with it.

TDu (approved):

jbaudoux:
@mmequignon Can you update your review?

shopfloor/actions/data.py
```diff
@@ -229,7 +229,18 @@ def _move_line_parser(self):
     def move(self, record, **kw):
         record = record.with_context(location=record.location_id.id)
         parser = self._move_parser
-        return self._jsonify(record, parser)
+        data = self._jsonify(record, parser)
+        if "auto_post_enabled" in kw:
```
jbaudoux:
```suggestion
        if kw.get("auto_post_enabled"):
```

mmequignon (changes requested):
backorders_quantity_done can be empty in the response if auto_post_enabled is false in the scenario config.
You cannot use this field like this in the frontend.
Also, I still think this shouldn't be the frontend task to compute the qty done.

JuMiSanAr:
> backorders_quantity_done can be empty in the response if auto_post_enabled is false in the scenario config. You cannot use this field like this in the frontend. Also, I still think this shouldn't be the frontend task to compute the qty done.

I've updated it so that it falls back to 0 if nothing is returned.

Regarding moving to the backend, I understand but I still think it's ok to give both pieces of information back to the frontend and then let it decide what to do with them, as each quantity has its own meaning.

jbaudoux (changes requested):
I agree with @mmequignon It is up to the backend to provide the list of moves & qty to display depending on the scope (incoming picking or shipment advice)

PR #712, opened 2023-08-07
[14.0] shopfloor: checkout scan change lot (rebased)

sebalix:
Same PR than https://github.com/OCA/wms/pull/569/ but rebased on top of 14.0 to resolve conflicts (we still need the previous PR to not bring regressions as it is used as pending-merge).

Comments from @jbaudoux have been addressed too.

cc @simahawk

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -41,7 +41,7 @@ class Checkout(Component):
     _description = __doc__
 
     def _response_for_select_line(
-        self, picking, message=None, need_confirm_pack_all=False, need_confirm_lot=False
+        self, picking, message=None, need_confirm_pack_all="", need_confirm_lot=None
     ):
         if all(line.shopfloor_checkout_done for line in picking.move_line_ids):
             return self._response_for_summary(picking, message=message)
@@ -56,7 +56,7 @@ class Checkout(Component):
         )
 
     def _data_for_select_line(
-        self, picking, need_confirm_pack_all=False, need_confirm_lot=False
+        self, picking, need_confirm_pack_all="", need_confirm_lot=None
     ):
         return {
             "picking": self._data_for_stock_picking(picking),
@@ -428,7 +428,7 @@ class Checkout(Component):
             {"qty_done": 0, "shopfloor_user_id": False}
         )
 
-    def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=False):
+    def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=None):
         """Scan move lines of the stock picking
 
         It allows to select move lines of the stock picking for the next
@@ -1582,7 +1582,7 @@ class ShopfloorCheckoutValidator(Component):
                 "required": False,
             },
             "confirm_lot": {
-                "type": "boolean",
+                "type": "integer",
                 "nullable": True,
                 "required": False,
             },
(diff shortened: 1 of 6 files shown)


simahawk:
@sebalix are you going to address JE's comments too?

sebalix:
@simahawk @jbaudoux comments addressed, excepting the check on the length of pickings in `_select_document_from_package` as it was breaking tests, indeed the relevant response is generated by `_select_picking` afterwards, I preferred to not change too much things for now and risk to break something else.

shopfloor/services/checkout.py
```diff
@@ -556,15 +559,58 @@ def _select_lines_from_packaging(self, picking, selection_lines, packaging, **kw
...
+                )
+                # If there's at least one product matching we are good to go.
+                # In any case, only the 1st line matching will be affected.
+                if lines_same_product:
+                    return self._response_for_select_line(
+                        picking,
+                        message=self.msg_store.lot_different_change(),
+                        need_confirm_lot=lot.name,
+                    )
+                    # TODO: add a msg saying the lot has been changed
```
jbaudoux: to drop or misplaced TODO

shopfloor/services/checkout.py
```diff
@@ -556,15 +559,58 @@ def _select_lines_from_packaging(self, picking, selection_lines, packaging, **kw
...
+                        selection_lines, lot
+                    )
+                )
+                # If there's at least one product matching we are good to go.
+                # In any case, only the 1st line matching will be affected.
+                if lines_same_product:
+                    return self._response_for_select_line(
+                        picking,
+                        message=self.msg_store.lot_different_change(),
+                        need_confirm_lot=lot.name,
```
jbaudoux: This is a bit unsafe as you can have many lots with the same name as the name is only unique per product.
A string with "model,id" would be more secure.

jbaudoux: Or just lot.id as we enter this method when a lot is scanned

shopfloor/services/checkout.py
```diff
@@ -557,15 +569,58 @@ def _select_lines_from_packaging(self, picking, selection_lines, packaging, **kw
...
+            message = change_package_lot.change_lot(
+                line, lot, response_ok_func, response_error_func
+            )
+            if message["message_type"] == "error":
+                return self._response_for_select_line(picking, message=message)
+            else:
+                lines = line
+                # Some lines have been recreated, refresh the recordset
+                # to avoid CacheMiss error
+                selection_lines = self._lines_to_pack(picking)
```
TDu, line 625:
```suggestion
                selection_lines = self._lines_to_pack(picking)
                selection_lines = self._picking_lines_by_lot(picking, selection_lines, lot)
```
So only the lines with the lot that changed are displayed in the next state ?

simahawk:
@TDu @jbaudoux @sebalix what's left to be done here beside rebasing n squashing?

shopfloor/services/checkout.py
TDu: This is not relevant anymore.
A change related to this #785 will any one cause changes here

TDu:
> @TDu @jbaudoux @sebalix what's left to be done here beside rebasing n squashing?

For me nothing but it is closely related to #785  and will need to be fixed on relation

TDu (approved):

jbaudoux:
@TDu @sebalix Can this be rebased & squashed ? I understand there was nothing left to be done

TDu:
@jbaudoux

PR #715, opened 2023-08-08
[14.0][FIX] shopfloor action validate_moves

jbaudoux:
When a package with 2 moves in 2 different pickings has to be replenished, then the 2 pickings must be validated together otherwise you get the error "You cannot move the same package content more than once in the same transfer or split the same package into two location".

cc @mt-software-de

### DIFF
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -185,7 +185,7 @@ class StockAction(Component):
             if self._check_backorder(picking, moves_todo):
                 existing_backorder_ids += picking.backorder_ids.ids
                 pickings_to_validate_ids.append(picking.id)
-            elif moves_todo != picking.move_lines:
+            else:
                 new_picking = moves_todo._extract_in_split_order()
                 if new_picking.state != "assigned":
                     raise UserError(_("Internal Error. Split order is not available"))
@@ -203,15 +203,25 @@ class StockAction(Component):
 
         We want to create a normal backorder if:
 
-            - the moves are equal to all available moves of the current picking
-              but there are still unavailable moves to process
-            - the moves are not linked to unprocessed ancestor moves
+            - the moves are equal to all reserved moves of the current picking
+            - the unavailable moves are not linked to unprocessed ancestor moves
         """
-        assigned_moves = picking.move_lines.filtered(lambda m: m.state == "assigned")
-        has_ancestors = bool(
-            moves.move_orig_ids.filtered(lambda m: m.state not in ("cancel", "done"))
+        available_moves = picking.move_lines.filtered(
+            lambda m: m.state in ("partially_available", "assigned")
         )
-        return moves == assigned_moves and not has_ancestors
+        if moves != available_moves:
+            return False
+        unavailable_moves = picking.move_lines.filtered(
+            lambda m: m.state in ("waiting", "confirmed", "partially_available")
+        )
+        has_ancestors = bool(
+            unavailable_moves.move_orig_ids.filtered(
+                lambda m: m.state not in ("cancel", "done")
+            )
+        )
+        if has_ancestors:
+            return False
+        return True
 
     def put_package_level_in_move(self, package_level):
         """Ensure to put the package level in its own move.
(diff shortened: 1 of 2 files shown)


shopfloor/actions/stock.py
```diff
@@ -177,16 +177,23 @@ def validate_moves(self, moves):
             the transfer is validated as usual, creating a backorder.
         """
         moves.split_unavailable_qty()
+        existing_backorder_ids = []
+        pickings_to_validate = moves.picking_id.browse()
```
simahawk: why collecting records here and ids above? ids in all the cases maybe?

shopfloor/actions/stock.py
```diff
@@ -177,16 +177,23 @@ def validate_moves(self, moves):
...
-                new_backorders = picking.backorder_ids - existing_backorders
-                if new_backorders:
-                    new_backorders.write({"user_id": False})
-            else:
-                moves_todo.extract_and_action_done()
+                existing_backorder_ids += picking.backorder_ids.ids
+                pickings_to_validate |= picking
+            elif moves_todo != picking.move_lines:
+                new_picking = moves_todo._extract_in_split_order()
+                assert new_picking.state == "assigned"
```
simahawk: as this can happen on an end user call, I would raise a UserError

shopfloor/actions/stock.py
jbaudoux: You could have many backorders but you usually process only one picking (in rare cases a few pickings). But let's use ids always and browse at the end :)

shopfloor/actions/stock.py
jbaudoux: changed

mt-software-de:
~~LGTM except the missing test~~

shopfloor/actions/stock.py
```diff
@@ -177,16 +178,25 @@ def validate_moves(self, moves):
             the transfer is validated as usual, creating a backorder.
```
sebalix, line 178: I would complete this docstring for the use case you are fixing here

shopfloor/actions/stock.py
```diff
@@ -177,16 +178,25 @@ def validate_moves(self, moves):
...
+                if new_picking.state != "assigned":
+                    raise UserError(_("Internal Error. Split order is not available"))
+                pickings_to_validate_ids.append(new_picking.id)
+        if pickings_to_validate_ids:
+            pickings_to_validate = moves.picking_id.browse(pickings_to_validate_ids)
+            pickings_to_validate._action_done()
+            existing_backorders = moves.picking_id.browse(existing_backorder_ids)
+            new_backorders = pickings_to_validate.backorder_ids - existing_backorders
+            if new_backorders:
+                new_backorders.write({"user_id": False})
```
mt-software-de: This is causing an issue now for moves of one picking, where the moves are the same as the assigned moves of the picking when they have "unprocessed ancestor moves".

By adding the check "elif moves_todo != picking.move_lines"
The moves will be never set to done.
Before they were set to done, because the were passed to "extract_and_action_done()". https://github.com/OCA/wms/blob/1059fec1fb507e4cbbfd022b6790dfe4284b5446/shopfloor/models/stock_move.py#L92
And there is the check https://github.com/OCA/wms/blob/1059fec1fb507e4cbbfd022b6790dfe4284b5446/shopfloor/models/stock_move.py#L106 that no split will be created, but the picking will be set do done

mt-software-de: I think we should redo the "_check_backorder" method/rename/reverse it to _check_extract.
I would make it much easier to read and also the docstring fits not to what it is doing. 
```
    
    def _check_for_extract(self, picking, moves):
        """Check if we have to extract the moves to a new picking

        We want to extract them to a new picking if:

           - there are still unavailable moves to process and
             and if they linked to unprocessed ancestor moves
        """
        has_ancestors = bool(
            moves.move_orig_ids.filtered(lambda m: m.state not in ("cancel", "done"))
        )
        if not has_ancestors:
            return False
        assigned_moves = picking.move_lines.filtered(lambda m: m.state == "assigned")
        if assigned_moves == picking.move_lines:
            return False
        return True
```

mt-software-de (changes requested):

shopfloor/actions/stock.py
jbaudoux: @mt-software-de Can you check this fix in the fixup commit https://github.com/OCA/wms/pull/715/commits/94d296fdb40247144508f052f0081f7e71186709 ?

jbaudoux: ping @mt-software-de

PR #719, opened 2023-08-10
[14.0][IMP] shopfloor: do not compute move_line_count by default for package data

mt-software-de:
If you want to get the data for a package it always computes the move_line_count.
This is very bad for example for the zone picking scenario where can have a long list of moves with packages.
Where this isn't used at all.

The only scenario which is using the move_line_count from the package data is the checkout.
So i think it is convenient to not compute it by default only if it is really needed.

~~The unittest are not updated for this fix yet.~~
cc @jbaudoux

### DIFF
--- a/shopfloor/tests/test_actions_data_detail.py
+++ b/shopfloor/tests/test_actions_data_detail.py
@@ -80,7 +80,6 @@ class TestActionsDataDetailCase(ActionsDataDetailCaseBase):
                 "name": package.location_id.display_name,
             },
             "name": package.name,
-            "move_line_count": 2,
             "packaging": self.data_detail.packaging(package.packaging_id),
             "weight": 20.0,
             "pickings": self.data_detail.pickings(pickings),
@@ -90,6 +89,11 @@ class TestActionsDataDetailCase(ActionsDataDetailCaseBase):
                 "name": self.storage_type_pallet.name,
             },
         }
+        self.assertDictEqual(data, expected)
+        data = self.data_detail.package_detail(
+            package, picking=self.picking, with_package_move_line_count=True
+        )
+        expected.update({"move_line_count": 2})
         self.assertDictEqual(data, expected)
 
     def test_data_picking(self):
@@ -185,14 +189,12 @@ class TestActionsDataDetailCase(ActionsDataDetailCaseBase):
             "package_src": {
                 "id": move_line.package_id.id,
                 "name": move_line.package_id.name,
-                "move_line_count": 1,
                 "weight": 20.0,
                 "storage_type": None,
             },
             "package_dest": {
                 "id": result_package.id,
                 "name": result_package.name,
-                "move_line_count": 1,
                 "weight": 6.0,
                 "storage_type": None,
             },
@@ -201,6 +203,10 @@ class TestActionsDataDetailCase(ActionsDataDetailCaseBase):
             "priority": "1",
             "progress": 30.0,
         }
+        self.assertDictEqual(data, expected)
+        data = self.data_detail.move_line(move_line, with_package_move_line_count=True)
+        expected["package_src"]["move_line_count"] = 1
+        expected["package_dest"]["move_line_count"] = 1
         self.assertDictEqual(data, expected)
 
     def test_data_move_line_lot(self):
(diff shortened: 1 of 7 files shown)


mt-software-de:
@jbaudoux and @TDu the unittests are now also green, could you do the review?

shopfloor/tests/test_actions_data_detail.py
```diff
@@ -91,6 +90,10 @@ def test_data_package(self):
             },
         }
         self.assertDictEqual(data, expected)
+        data = self.data_detail.package_detail(
+            package, picking=self.picking, with_package_move_line_count=True
+        )
+        expected.update({"move_line_count": 2})
```
TDu, line 96: I don't understand this change with no assertion added after ?

shopfloor/tests/test_actions_data_detail.py
```diff
@@ -202,6 +203,18 @@ def test_data_move_line_package(self):
             "progress": 30.0,
         }
         self.assertDictEqual(data, expected)
+        data = self.data_detail.move_line(move_line, with_package_move_line_count=True)
+        package_src = expected["package_src"]
+        package_src.update({"move_line_count": 1})
+        package_dest = expected["package_dest"]
+        package_dest.update({"move_line_count": 1})
+        expected.update(
```
TDu: The tests in this file have all the same workflow: generate some data, check it matches a schema and validate with some static content.
Here the data is generated and then modified, which does not make sense to me.

shopfloor/tests/test_actions_data_detail.py
mt-software-de: oh yes this was missing. Thx

shopfloor/tests/test_actions_data_detail.py
mt-software-de: the call self.data_detail.move_line(move_line, with_package_move_line_count=True)
creates the same data it only adds "move_line_count" on package_src and package_dest.
That's why i don't need to define the expected data again.

jbaudoux (approved):

shopfloor/tests/test_actions_data_detail.py
```diff
@@ -202,6 +204,18 @@ def test_data_move_line_package(self):
...
+            {
+                "package_src": package_src,
+                "package_dest": package_dest,
+            }
+        )
```
simahawk: It makes sense because is already tested above that the pkgs are correct. Yet, this seems too verbose.
What about this?
```suggestion
        expected["package_src"]["move_line_count"] = 1
        expected["package_dest"]["move_line_count"] = 1
```
That's all you need IMO. Same in the other tests.

shopfloor/tests/test_actions_data_detail.py
simahawk: See https://github.com/OCA/wms/pull/719#discussion_r1296758759

simahawk:
@mt-software-de can you pls shorten the commit msg and include the description on the next push?

```
[IMP] shopfloor: compute package move_line_count when needed

If you want to get the data for a package it always computes the move_line_count.
This is very bad for perf on scenario - like zone picking - where you have a long list of moves with packages
w/o needing such information.

The only scenario which is using the move_line_count from the package data is the checkout.
So i think it is convenient to not compute it by default only if it is really needed.
```

Note that if you write a commit like this the description of the PR (for 1 commit PRs) is free ;)

shopfloor/tests/test_actions_data_detail.py
mt-software-de: yes was already change not yet commited :-)

mt-software-de:
> @mt-software-de can you pls shorten the commit msg and include the description on the next push?
> 
> ```
> [IMP] shopfloor: compute package move_line_count when needed
> 
> If you want to get the data for a package it always computes the move_line_count.
> This is very bad for perf on scenario - like zone picking - where you have a long list of moves with packages
> w/o needing such information.
> 
> The only scenario which is using the move_line_count from the package data is the checkout.
> So i think it is convenient to not compute it by default only if it is really needed.
> ```
> 
> Note that if you write a commit like this the description of the PR (for 1 commit PRs) is free ;)

Done. Thx

simahawk (approved):

PR #720, opened 2023-08-10
[14.0][IMP] shopfloor: zone picking is package not valid

mt-software-de:
- a package is now also not valid if it already contains move lines from an other picking type

cc @jbaudoux 

Maybe we should consider to check the unload_package_at_destination configuration, 
because it only would make problems if there is a difference in the shopfloor menu configuration of unload_package_at_destination

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -940,25 +940,34 @@ class ZonePicking(Component):
             move_line.product_uom_qty - qty, precision_rounding=rounding
         )
 
-    def _set_destination_package(self, move_line, quantity, package):
-        package_changed = False
-        response = None
+    def _is_package_not_valid(self, package):
+        message = False
         # A valid package is:
         # * an empty package
         # * not used as destination for another move line
+        # * not contains move lines with different operation type
         if not self._is_package_empty(package):
+            message = self.msg_store.package_not_empty(package)
+        elif package.planned_move_line_ids:
+            if not self.work.menu.multiple_move_single_pack:
+                message = self.msg_store.package_already_used(package)
+            else:
+                for line in package.planned_move_line_ids:
+                    if line.picking_id.picking_type_id.id in self.picking_types.ids:
+                        continue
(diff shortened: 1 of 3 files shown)


shopfloor/services/zone_picking.py
```diff
@@ -940,25 +940,33 @@ def _move_line_full_qty(self, move_line, qty):
...
         # * an empty package
         # * not used as destination for another move line
+        # * not contains move lines with different operation type
         if not self._is_package_empty(package):
+            message = self.msg_store.package_not_empty(package)
+        elif package.planned_move_line_ids:
+            if not self.work.menu.multiple_move_single_pack:
+                message = self.msg_store.package_already_used(package)
+            elif any(
+                line.picking_id.picking_type_id != move_line.picking_id.picking_type_id
```
mt-software-de: maybe it is better to do it like that
```suggestion
                line.picking_id.picking_type_id.id not in self.picking_types.ids
```

shopfloor/services/zone_picking.py
```diff
@@ -940,25 +940,33 @@ def _move_line_full_qty(self, move_line, qty):
...
+            message = self.msg_store.package_not_empty(package)
+        elif package.planned_move_line_ids:
+            if not self.work.menu.multiple_move_single_pack:
+                message = self.msg_store.package_already_used(package)
+            elif any(
+                line.picking_id.picking_type_id.id not in self.picking_types.ids
+                for line in package.planned_move_line_ids
+            ):
+                message = self.msg_store.package_different_picking_type(
+                    package, line.picking_id.picking_type_id
```
jbaudoux: `line` should not be defined as it is outside generator scope

jbaudoux (changes requested):
It's missing a test for the new addition.

Besides that, this is a nice improvement

mt-software-de:
> It's missing a test for the new addition.
> 
> Besides that, this is a nice improvement

Tests are added.

jbaudoux (approved):

simahawk (approved):
LG

PR #721, opened 2023-08-14
[14.0][PERF] stock_available_to_promise_release priority

jbaudoux:
Prevent odoo to try to compute if a state change must be tracked when the priority is changed. As many moves can be released at the same time this useless computation is not neglectable.

cc @mt-software-de

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -514,7 +514,9 @@ class StockMove(models.Model):
         super()._assign_picking_post_process(new)
         priorities = self.mapped("move_dest_ids.picking_id.priority")
         if priorities:
-            self.picking_id.write({"priority": max(priorities)})
+            self.picking_id.with_context(tracking_disable=True).write(
+                {"priority": max(priorities)}
+            )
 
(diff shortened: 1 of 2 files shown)


stock_available_to_promise_release/models/stock_move.py
```diff
@@ -514,7 +514,9 @@ def _assign_picking_post_process(self, new=False):
         super()._assign_picking_post_process(new)
         priorities = self.mapped("move_dest_ids.picking_id.priority")
         if priorities:
-            self.picking_id.write({"priority": max(priorities)})
+            self.picking_id.with_context(tracking_disable=True).write(
```
mt-software-de, line 517: Maybe you can add a comment?

mt-software-de (changes requested):

TDu (approved):

jbaudoux:
@TDu Besides the perf improvement, there is something we need to adapt in the code implementation you did in https://github.com/OCA/wms/commit/1d4e50facc92307124ca769dd53925baa6983b14
We cannot write the priority on the picking as this will override the priority of all moves.
Something similar to this fix https://github.com/OCA/wms/commit/046e10e33b1ab63bef06aa8c74accb74b24b06af

PR #722, opened 2023-08-14
[14.0][PERF] dynamic routing

jbaudoux:
As the call to _check_entire_pack can be very slow, bypass it's call when pre-computing the dynamic routing pull rules as this is not needed for the computation and will be rollbacked.

Also in case of source relocation, ensure to call _check_entire_pack on the recordset of moves and not on each move one by one. This is because _check_entire_pack triggers the check on the picking of the move and so you perform the same check for each move of the picking. When you reserve from source packages you have many moves inside a picking, the check is quite long and this change makes a important performance improvement.

When the dynamic routing does not change the source location, do not write the same value on the move as this triggers many useless additional recomputations.

cc @mt-software-de


simahawk (commented):
Changes looks good.
I think it's time to upgrade these modules to dev status = production/stable as they are used in prod since years w/o major issues. WDYT?

jbaudoux:
> Changes looks good. I think it's time to upgrade these modules to dev status = production/stable as they are used in prod since years w/o major issues. WDYT?

`stock_helper` first needs to be changed to stable and it is in another repo. So I won't do the change in this PR

jbaudoux:
There was a feature for the re-routing of waiting moves in the chain (even when the pick location destination is not changed) that was not covered by a test. To be sure the performance changes are not breaking any feature, I have added a test for this.

This feature is deprecated in favor of the `stock_warehouse_flow` module but as this was initially offered by the dynamic routing you can still rely on the `stock_dynamic_routing` config for this.

simahawk:
> > Changes looks good. I think it's time to upgrade these modules to dev status = production/stable as they are used in prod since years w/o major issues. WDYT?
> 
> `stock_helper` first needs to be changed to stable and it is in another repo. So I won't do the change in this PR

You are no the maintainer :wink:  Can you take care of this on another PR? :pray:

simahawk (approved):

mt-software-de (approved):
LGTM

stock_move_source_relocate/models/stock_move.py
```diff
@@ -18,22 +22,33 @@ def _action_assign(self):
         unconfirmed_moves = unconfirmed_moves.filtered(
             lambda m: m.state in ["confirmed", "partially_available"]
         )
-        unconfirmed_moves._apply_source_relocate()
+        if unconfirmed_moves:
+            unconfirmed_moves._apply_source_relocate()
```
sebalix, line 26: Does it change something to check the recordset here?

Should we instead do this check in the `_apply_source_relocate` method, if `self` is empty return.

stock_move_source_relocate/models/stock_move.py
```diff
@@ -63,3 +78,7 @@ def _apply_source_relocate_rule(self, relocation, reserved_availability, roundin
         new_move.location_id = relocation.relocate_location_id
         self._action_assign()
         return new_move
+
+    def _after_apply_source_relocate_rule(self):
+        # Hook for stock_dynamic_routing
```
sebalix, line 83: Not sure we need to mention this module in `stock_move_source_relocate`? It's the job of the glue module (or anything else) to use this hook.

sebalix (approved):
LG, minor comments/questions

stock_move_source_relocate/models/stock_move.py
jbaudoux: No the method should not be called if not necessary.
Otherwise, all methods in odoo would start with `if not self: return` which is ugly

sebalix: Well, here except the debug log the method will do nothing. And as the Odoo API tends to be "functional-programming" in some way it's OK to call methods on an object without thinking if it's empty or not.

PR #725, opened 2023-08-15
[14.0][IMP] shopfloor checkout scan delivery package

TDu:
ref: cos-4147


shopfloor/services/checkout.py
```diff
@@ -41,7 +41,7 @@ class Checkout(Component):
     _description = __doc__
 
     def _response_for_select_line(
-        self, picking, message=None, need_confirm_pack_all=False
+        self, picking, message=None, need_confirm_pack_all=""
```
simahawk, line 44: IMO this should be still None when empty

simahawk (approved):
Minor remark

shopfloor/services/checkout.py
```diff
@@ -620,16 +623,27 @@ def _select_lines_from_delivery_packaging(
                     packaging, carrier
                 ),
             )
+        message = None
+        need_confirm_pack_all = ""
+        lines_to_pack = selection_lines.filtered(self._filter_lines_to_pack)
+        if not lines_to_pack:
+            if self.work.menu.no_prefill_qty:
+                message = self.msg_store.no_lines_to_process_set_quantities()
+            elif not confirm_pack_all == packaging.barcode:
```
jbaudoux:
```suggestion
            elif confirm_pack_all != packaging.barcode:
```

shopfloor/services/checkout.py
```diff
@@ -620,16 +623,27 @@ def _select_lines_from_delivery_packaging(
...
+                )
+            if message:
+                return self._response_for_select_line(
+                    picking,
+                    message=message,
+                    need_confirm_pack_all=need_confirm_pack_all,
+                )
         if confirm_pack_all:
-            # Select all lines and pack them all w/o passing for select_package state
             self._select_lines(selection_lines)
```
jbaudoux: Should be the `else` condition in the above `if not lines_to_pack:` block

shopfloor/services/checkout.py
```diff
@@ -620,16 +623,27 @@ def _select_lines_from_delivery_packaging(
                     packaging, carrier
                 ),
             )
+        message = None
+        need_confirm_pack_all = ""
+        lines_to_pack = selection_lines.filtered(self._filter_lines_to_pack)
```
jbaudoux:
```suggestion
        has_lines_to_pack = any(self._filter_lines_to_pack(ml) for ml in selection_lines)
```

shopfloor/services/checkout.py
TDu: That does not work, on a valid confirmation to pack everything there will still be nothing to pack !
I refined the if case to be more explicit.

JuMiSanAr (approved):

jbaudoux (approved):

jbaudoux:
@TDu Can you squash the fixups in order to merge?

TDu:
@jbaudoux squashed and rebased

PR #726, opened 2023-08-16
[14.0][FIX] shopfloor_reception: Fix qty todo computation

mmequignon:
We had a concurrent update introduced here https://github.com/OCA/wms/pull/672

Updating product uom qties were causing issues.
This is basically fixed by modifying the product_qty in responses instead of updating the move lines qties.

### DIFF
--- a/shopfloor_reception/data/shopfloor_scenario_data.xml
+++ b/shopfloor_reception/data/shopfloor_scenario_data.xml
@@ -9,7 +9,8 @@
       <field name="options_edit">
 {
       "auto_post_line": true,
-      "allow_return": true
+      "allow_return": true,
+      "scan_location_or_pack_first": true
 }
       </field>
   </record>
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1301,6 +1301,17 @@ class Reception(Component):
         new_move._action_confirm(merge=False)
         new_move._recompute_state()
         new_move._action_assign()
+        # Set back the quantity to do on one of the lines
+        line = fields.first(
+            move.move_line_ids.filtered(
+                lambda line: line.state not in ("cancel", "done")
+            )
+        )
+        if line:
+            move_quantity = move.product_uom._compute_quantity(
+                move.product_uom_qty, line[0].product_uom_id
+            )
+            line.product_uom_qty = move_quantity
         move._recompute_state()
         new_move.extract_and_action_done()
 
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -257,12 +257,7 @@ class Reception(Component):
             )
         )
         if not line:
-            qty_todo_remaining = max(
-                0,
-                move.product_uom_qty
-                - sum(move.move_line_ids.mapped("product_uom_qty")),
-            )
-            values = move._prepare_move_line_vals(quantity=qty_todo_remaining)
+            values = move._prepare_move_line_vals()
             line = self.env["stock.move.line"].create(values)
         return self._scan_line__assign_user(picking, line, qty_done)
 
@@ -1249,6 +1244,7 @@ class Reception(Component):
         return self._response_for_set_destination(picking, selected_line)
 
     def _post_line(self, selected_line):
+        selected_line.product_uom_qty = selected_line.qty_done
         if (
             selected_line.picking_id.is_shopfloor_created
             and self.work.menu.allow_return
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1242,15 +1242,26 @@ class Reception(Component):
         )
 
     def _auto_post_line(self, selected_line):
-        new_move = selected_line.move_id.split_other_move_lines(
-            selected_line, intersection=True
-        )
-        if new_move:
-            # A new move is created in case of partial quantity
-            new_move.extract_and_action_done()
-            return
-        # In case of full quantity, post the initial move
-        selected_line.move_id.extract_and_action_done()
+        # If user only processed 1/5 and is the only one working on the move,
+        # then selected_line is the only one related to this move.
+        # In such case, we must ensure there's another move line with the remaining
+        # quantity to do, so selected_line is extracted in a new move as expected.
+        if selected_line.product_uom_qty:
+            selected_line.product_uom_qty = 0
+        move = selected_line.move_id
+        if selected_line.qty_done == move.product_uom_qty:
+            # In case of full quantity, post the initial move
+            selected_line.move_id.extract_and_action_done()
+        split_move_vals = move._split(selected_line.qty_done)
+        new_move = move.create(split_move_vals)
+        new_move.move_line_ids = selected_line
+        new_move._action_confirm(merge=False)
+        new_move._recompute_state()
+        new_move._action_assign()
+        lock = self._actions_for("lock")
+        lock.for_update(move)
+        move._recompute_state()
+        new_move.extract_and_action_done()
 
     def set_destination(
         self, picking_id, selected_line_id, location_name, confirmation=False
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1164,6 +1164,22 @@ class Reception(Component):
             )
         return self._response_for_set_quantity(picking, selected_line)
 
+    def set_quantity__cancel_action(self, picking_id, selected_line_id):
+        picking = self.env["stock.picking"].browse(picking_id)
+        selected_line = self.env["stock.move.line"].browse(selected_line_id)
+        message = self._check_picking_status(picking)
+        if message:
+            return self._response_for_set_quantity(
+                picking, selected_line, message=message
+            )
+        if selected_line.exists():
+            if selected_line.product_uom_qty:
+                stock = self._actions_for("stock")
+                stock.unmark_move_line_as_picked(selected_line)
+            else:
+                selected_line.unlink()
+        return self._response_for_select_move(picking)
+
     def _set_quantity__process__set_qty_and_split(self, picking, line, quantity):
         move = line.move_id
         sum(move.move_line_ids.mapped("qty_done"))
@@ -1442,6 +1458,16 @@ class ShopfloorReceptionValidator(Component):
             "confirmation": {"type": "string", "nullable": True},
         }
 
+    def set_quantity__cancel_action(self):
+        return {
+            "picking_id": {"coerce": to_int, "required": True, "type": "integer"},
+            "selected_line_id": {
+                "coerce": to_int,
+                "type": "integer",
+                "required": True,
+            },
+        }
+
     def process_with_existing_pack(self):
         return {
             "picking_id": {"coerce": to_int, "required": True, "type": "integer"},
@@ -1573,6 +1599,9 @@ class ShopfloorReceptionValidatorResponse(Component):
     def _set_quantity_next_states(self):
         return {"set_quantity", "select_move", "set_destination"}
 
+    def _set_quantity__cancel_action_next_states(self):
+        return {"set_quantity", "select_move"}
+
     def _set_destination_next_states(self):
         return {"set_destination", "select_move"}
 
@@ -1654,6 +1683,16 @@ class ShopfloorReceptionValidatorResponse(Component):
         }
 
     @property
+    def _schema_set_quantity__cancel_action(self):
+        return {
+            "selected_move_line": {
+                "type": "list",
+                "schema": {"type": "dict", "schema": self.schemas.move_line()},
+            },
+            "picking": {"type": "dict", "schema": self.schemas.picking()},
+        }
+
+    @property
     def _schema_set_destination(self):
         return {
             "selected_move_line": {
@@ -1729,6 +1768,11 @@ class ShopfloorReceptionValidatorResponse(Component):
 
     def set_quantity(self):
         return self._response_schema(next_states=self._set_quantity_next_states())
+
+    def set_quantity__cancel_action(self):
+        return self._response_schema(
+            next_states=self._set_quantity__cancel_action_next_states()
+        )
 
     def process_with_existing_pack(self):
         return self._response_schema(
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1283,10 +1283,12 @@ class Reception(Component):
             lambda line: line.state not in ("cancel", "done")
             and line.product_uom_qty > 0
         )
+        move = selected_line.move_id
+        lock = self._actions_for("lock")
+        lock.for_update(move)
         if lines_with_qty_todo:
             lines_with_qty_todo.product_uom_qty = 0
 
-        move = selected_line.move_id
         move_quantity = move.product_uom._compute_quantity(
             move.product_uom_qty, selected_line.product_uom_id
         )
@@ -1299,8 +1301,6 @@ class Reception(Component):
         new_move._action_confirm(merge=False)
         new_move._recompute_state()
         new_move._action_assign()
-        lock = self._actions_for("lock")
-        lock.for_update(move)
         move._recompute_state()
         new_move.extract_and_action_done()
 
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -137,12 +137,10 @@ class Reception(Component):
         return self._response_for_select_move(picking)
 
     def _response_for_select_move(self, picking, message=None):
-        self._assign_user_to_picking(picking)
         data = {"picking": self._data_for_stock_picking(picking, with_lines=True)}
         return self._response(next_state="select_move", data=data, message=message)
 
     def _response_for_confirm_done(self, picking, message=None):
-        self._assign_user_to_picking(picking)
         data = {"picking": self._data_for_stock_picking(picking, with_lines=True)}
         return self._response(next_state="confirm_done", data=data, message=message)
 
@@ -161,7 +159,6 @@ class Reception(Component):
     def _select_document_from_move_lines(self, move_lines, msg_func):
         pickings = move_lines.move_id.picking_id
         if len(pickings) == 1:
-            self._assign_user_to_picking(pickings)
             if (
                 move_lines.product_id.tracking not in ("lot", "serial")
                 or move_lines.lot_id
@@ -263,7 +260,6 @@ class Reception(Component):
 
     def _scan_line__assign_user(self, picking, line, qty_done):
         product = line.product_id
-        self._assign_user_to_picking(picking)
         self._assign_user_to_line(line)
         line.qty_done += qty_done
         if product.tracking not in ("lot", "serial") or (line.lot_id or line.lot_name):
@@ -328,7 +324,6 @@ class Reception(Component):
                 < today_end
             )
             if len(picking_filter_result_due_today) == 1:
-                self._assign_user_to_picking(picking_filter_result_due_today)
                 return self._select_picking(picking_filter_result_due_today)
             if len(picking_filter_result) > 1:
                 return self._response_for_select_document(
@@ -681,9 +676,6 @@ class Reception(Component):
             response = handler(*args, **kwargs)
             if response:
                 return response
-
-    def _assign_user_to_picking(self, picking):
-        picking.user_id = self.env.user
 
     def _assign_user_to_line(self, line):
         line.shopfloor_user_id = self.env.user
@@ -1404,7 +1396,6 @@ class Reception(Component):
                 return response
             return self._response_for_select_move(picking)
         message = self.msg_store.create_new_pack_ask_confirmation(barcode)
-        self._assign_user_to_picking(picking)
         return self._response_for_confirm_new_package(
             picking, selected_line, new_package_name=barcode, message=message
         )
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -256,12 +256,12 @@ class Reception(Component):
                 )
             )
         )
-        if line:
-            # The line quantity to do needs to correspond to
-            # the remaining quantity to do of its move.
-            line.product_uom_qty = move.product_uom_qty - move.quantity_done
-        else:
-            qty_todo_remaining = max(0, move.product_uom_qty - move.quantity_done)
+        if not line:
+            qty_todo_remaining = max(
+                0,
+                move.product_uom_qty
+                - sum(move.move_line_ids.mapped("product_uom_qty")),
+            )
             values = move._prepare_move_line_vals(quantity=qty_todo_remaining)
             line = self.env["stock.move.line"].create(values)
         return self._scan_line__assign_user(picking, line, qty_done)
@@ -751,7 +751,7 @@ class Reception(Component):
             message=message,
         )
 
-    def _align_product_uom_qties(self, move):
+    def _align_display_product_uom_qty(self, line, response):
         # This method aligns product uom qties on move lines.
         # In the shopfloor context, we might have multiple users working at
         # the same time on the same move. This is done by creating one move line
@@ -770,19 +770,18 @@ class Reception(Component):
         # If move is already done, do not update lines qties
         # if move.state in ("done", "cancel"):
         #     return
-
+        move = line.move_id
         qty_todo = move.product_uom_qty
         qty_done = sum(move.move_line_ids.mapped("qty_done"))
         rounding = move.product_id.uom_id.rounding
         compare = float_compare(qty_done, qty_todo, precision_rounding=rounding)
-        if compare < 1:  # If qty done <= qty todo, align qty todo on move lines
+        if compare < 1:  # If qty done < qty todo, align qty todo in the response
             remaining_todo = qty_todo - qty_done
-            # if we didn't bypass reservation update, the quant reservation
-            # would be reduced as much as the deduced quantity, which is wrong
-            # as we only moved the quantity to a new move line
-            lines = move.move_line_ids.with_context(bypass_reservation_update=True)
-            for line in lines:
-                line.product_uom_qty = line.qty_done + remaining_todo
+            line_todo = line.qty_done + remaining_todo
+            response["data"]["set_quantity"]["selected_move_line"][0][
+                "quantity"
+            ] = line_todo
+        return response
 
     def _before_state__set_quantity(self, picking, line, message=None):
         # Used by inherting module  see shopfloor_reception_packaging_dimension
@@ -791,8 +790,7 @@ class Reception(Component):
     def _response_for_set_quantity(
         self, picking, line, message=None, asking_confirmation=None
     ):
-        self._align_product_uom_qties(line.move_id)
-        return self._response(
+        response = self._response(
             next_state="set_quantity",
             data={
                 "selected_move_line": self._data_for_move_lines(line),
@@ -801,6 +799,7 @@ class Reception(Component):
             },
             message=message,
         )
+        return self._align_display_product_uom_qty(line, response)
 
     def _response_for_set_destination(self, picking, line, message=None):
         return self._response(
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -758,9 +758,6 @@ class Reception(Component):
         # remaining_todo = move.product_uom_qty - move.quantity_done
         # line.product_uom_qty = line.qty_done + remaining_todo
 
-        # However, if the overall qty_done is > to the move's product_uom_qty,
-        # then we're not updating the line's product_uom_qty.
-
         # TODO, do we need to check move's state?
         # If move is already done, do not update lines qties
         # if move.state in ("done", "cancel"):
@@ -769,7 +766,7 @@ class Reception(Component):
         qty_todo = move.product_uom_qty
         other_lines_qty_done = 0.0
         move_uom = move.product_uom
-        for move_line in (move.move_line_ids - line):
+        for move_line in move.move_line_ids - line:
             # Use move's uom
             line_uom = move_line.product_uom_id
             other_lines_qty_done += line_uom._compute_quantity(
@@ -780,9 +777,10 @@ class Reception(Component):
         line_todo = line.product_uom_id._compute_quantity(
             remaining_todo, move_uom, round=False
         )
-        response["data"]["set_quantity"]["selected_move_line"][0][
-            "quantity"
-        ] = line_todo
+        # If more has been done keep the quantity to zero
+        response["data"]["set_quantity"]["selected_move_line"][0]["quantity"] = max(
+            line_todo, 0
+        )
         return response
 
     def _before_state__set_quantity(self, picking, line, message=None):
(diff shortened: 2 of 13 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -766,25 +766,23 @@ def _align_product_uom_qties(self, move):
         # If move is already done, do not update lines qties
         # if move.state in ("done", "cancel"):
         #     return
-
+        move = line.move_id
         qty_todo = move.product_uom_qty
         qty_done = sum(move.move_line_ids.mapped("qty_done"))
```
```suggestion
        qty_done = sum(
                    self.env['uom.uom'].browse(product_uom_id)._compute_quantity(qty_done, line.product_uom_id, round=False)
                     for product_uom_id, qty_done in move.move_line_ids)
                   )
```

shopfloor_reception/services/reception.py
```diff
@@ -766,25 +766,23 @@ def _align_product_uom_qties(self, move):
         # If move is already done, do not update lines qties
         # if move.state in ("done", "cancel"):
         #     return
-
+        move = line.move_id
         qty_todo = move.product_uom_qty
```
jbaudoux, line 774: Compute all in line uom
```suggestion
        qty_todo = self.env['uom.uom'].browse(move.product_uom_id)._compute_quantity(move.product_uom_qty, move.product_uom_id, round=False)
```

jbaudoux (commented):
What about the uom?

shopfloor_reception/services/reception.py
```diff
@@ -1235,6 +1243,11 @@ def _post_shopfloor_created_line(self, selected_line):
         )
 
     def _auto_post_line(self, selected_line):
+        # If user only processed 1/5 and is the only one working on the move,
+        # then selected_line is the only one related to this move.
+        # In such case, we must ensure there's another move line with the remaining
+        # quantity to do, so selected_line is extracted in a new move as expected.
+        selected_line._split_partial_quantity()
```
mt-software-de:
```suggestion
        new_move_line = selected_line._split_partial_quantity()
```
We should call unmark_move_line_as_picked and set the location_dest_id back to stock.move's location_dest_id
if a split was happening 
```
            stock.unmark_move_line_as_picked(new_move_line)
            new_move_line.location_dest_id = new_move_line.move_id.location_dest_id
```

mt-software-de (changes requested):

shopfloor_reception/services/reception.py
mt-software-de: https://github.com/camptocamp/wms/pull/118

mmequignon: Hello @mt-software-de , I updated the branch, can you please rebase? I'm checking this.
Thanks!

shopfloor_reception/models/shopfloor_menu.py
```diff
@@ -0,0 +1,28 @@
...
+
+ALLOW_SELECT_DOCUMENT_BY_PRODUCT_HELP = """
+If enabled, users will be able to select transfers by product.
+"""
+
+
+class ShopfloorMenu(models.Model):
+    _inherit = "shopfloor.menu"
+
+    allow_select_document_by_product = fields.Boolean(
```
jbaudoux: This is the already existing option `scan_location_or_pack_first` that allows to disable scanning product/lot to open a transfer. There is TODO to rename the field as it is misleading

shopfloor_reception/models/shopfloor_menu.py
```diff
@@ -0,0 +1,28 @@
...
+        string="Allow select document by product",
+        default=False,
+        help=ALLOW_SELECT_DOCUMENT_BY_PRODUCT_HELP,
+    )
+    allow_select_document_by_product_is_possible = fields.Boolean(
+        compute="_compute_pick_pack_same_time_is_possible"
+    )
+
+    @api.depends("scenario_id")
+    def _compute_pick_pack_same_time_is_possible(self):
```
jbaudoux: wrong method name

shopfloor_reception/services/reception.py
```diff
@@ -839,7 +842,12 @@ def start(self):
     def _scan_document__get_handlers_by_type(self):
         return {
             "picking": self._scan_document__by_picking,
-            "product": self._scan_document__by_product,
+            # only add the handler if allow_select_document_by_product is enabled
+            "product": (
+                self._scan_document__by_product
+                if self.work.menu.allow_select_document_by_product
```
```suggestion
                if not scan_location_or_pack_first
```

shopfloor_reception/services/reception.py
```diff
@@ -1216,6 +1240,7 @@ def process_without_pack(self, picking_id, selected_line_id, quantity):
         return self._response_for_set_destination(picking, selected_line)
 
     def _post_line(self, selected_line):
+        selected_line.product_uom_qty = selected_line.qty_done
```
jbaudoux, line 1253: I would not do this

shopfloor_reception/services/reception.py
```diff
@@ -1237,6 +1262,11 @@ def _post_shopfloor_created_line(self, selected_line):
         )
 
     def _auto_post_line(self, selected_line):
+        # If user only processed 1/5 and is the only one working on the move,
+        # then selected_line is the only one related to this move.
+        # In such case, we must ensure there's another move line with the remaining
+        # quantity to do, so selected_line is extracted in a new move as expected.
+        selected_line._split_partial_quantity()
```
jbaudoux: This block requires a `if product_uom_qty:`

shopfloor_reception/services/reception.py
```diff
@@ -256,13 +256,8 @@ def _scan_line__find_or_create_line(self, picking, move, qty_done=1):
                 )
             )
         )
-        if line:
-            # The line quantity to do needs to correspond to
-            # the remaining quantity to do of its move.
-            line.product_uom_qty = move.product_uom_qty - move.quantity_done
```
jbaudoux, line 262: The posting is way easier if we always set the product_uom_qty to 0
```
if line:
    line.product_uom_qty = 0
```

shopfloor_reception/services/reception.py
jbaudoux: This is correct but you also need to recompute the putaway.
It is easier if you always set product_uom_qty to 0, then there is no move line backorder (i.e. nothing to split).
I would do:
```
if product_uom_qty:
    selected_line.product_uom_qty = 0
```
and drop the `_split_partial_quantity`

shopfloor_reception/services/reception.py
```diff
@@ -1237,6 +1262,11 @@ def _post_shopfloor_created_line(self, selected_line):
         )
 
     def _auto_post_line(self, selected_line):
+        # If user only processed 1/5 and is the only one working on the move,
+        # then selected_line is the only one related to this move.
+        # In such case, we must ensure there's another move line with the remaining
+        # quantity to do, so selected_line is extracted in a new move as expected.
+        selected_line._split_partial_quantity()
         new_move = selected_line.move_id.split_other_move_lines(
```
jbaudoux: If some other move lines have a product_uom_qty, you need to modify them so that the new_move qty done is not smaller than the remaining move lines product_uom_qty. To prevent this complexity, it is easier if all move lines have a product_uom_qty of 0.
```suggestion
        selected_line.move_id.move_line_ids.filtered("product_uom_qty").product_uom_qty = 0
        new_move = selected_line.move_id.split_other_move_lines(
```

jbaudoux (changes requested):

shopfloor_reception/services/reception.py
mt-software-de: > Hello @mt-software-de , I updated the branch, can you please rebase? I'm checking this. Thanks!

rebase was done

shopfloor_reception/services/reception.py
mmequignon: when would you do this then?
We agreed that we should set product_uom_qty to 0 at creation, and align it to qty_done at posting.
Am I missing smth?

jbaudoux: I don't think you need to align at posting. If you do it in odoo interface, it will be 0 at creation and remain like that at posting

mmequignon: oh, ok, thanks

shopfloor_reception/services/reception.py
mmequignon: We said that if the line wasn't created by shopfloor, we might have product_uom_qty != 0, and we want to keep it like that..
I'm handling this here.
In the cancel action, I unlink move lines with product_uom_qty = 0, and reset the others
https://github.com/OCA/wms/pull/726/files/dd90a411d6c5f84ca0b8c6924297c553da3bd870#diff-57b252eb869558e97f9dbe4859b1f4cafc7db1e67ccdaecc92d2fc38d46a71bdR1167-R1171

mmequignon:
@jbaudoux Just checked, and we need the product uom qty to be set on the move line when we split the move.
In split_other_move lines, we use this product_uom_qty to determine the qty to be done on the new move.
Before we post the move though, we can use 0 as a product uom qty, that's ok, I guess.

shopfloor_reception/services/reception.py
mmequignon: We decided to revert the change where I introduced the "align_qties" method, and to modify the product_uom_qty in the response instead, in order to not alter a move line the user is not working on.
I'm, we should set the qty in `_scan_line__find_or_create_line` every time, whether we create a new line or find one.

jbaudoux: Drop the call to split_other_move_lines and do sth like:
```
if selected_line.qty_done < selected_line.move_id.qty_todo:
            split_move_vals = self._split(selected_line.qty_done)
            split_move = self.create(split_move_vals)
            split_move.move_line_ids = selected_line
            split_move._action_confirm(merge=False)
            split_move._recompute_state()
            split_move._action_assign()
            selected_line.move_id._recompute_state()
```

mmequignon: @jbaudoux can you make another pass please?

shopfloor_reception/services/reception.py
```diff
@@ -768,25 +763,32 @@ def _align_product_uom_qties(self, move):
...
+        move_uom = move.product_uom
+        for move_line in move.move_line_ids:
+            # Use move's uom
+            qty_done += move_uom._compute_quantity(
+                move_line.qty_done, move_line.product_uom_id, round=False
+            )
         rounding = move.product_id.uom_id.rounding
         compare = float_compare(qty_done, qty_todo, precision_rounding=rounding)
-        if compare < 1:  # If qty done <= qty todo, align qty todo on move lines
+        if compare < 1:  # If qty done < qty todo, align qty todo in the response
```
jbaudoux: I think we should always compute it. If qty done >= qty todo, you should not return 0 as todo. So I would drop this `if`.

Maybe you could skip `line` in above `qty_done` computation; That would be a `other_lines_qty_done`.
Then below, you don't need to re-add `line.qty_done` to the `remaining_todo`

shopfloor_reception/services/reception.py
```diff
@@ -1237,15 +1262,24 @@ def _post_shopfloor_created_line(self, selected_line):
...
-        # In case of full quantity, post the initial move
-        selected_line.move_id.extract_and_action_done()
+        # If user only processed 1/5 and is the only one working on the move,
+        # then selected_line is the only one related to this move.
+        # In such case, we must ensure there's another move line with the remaining
+        # quantity to do, so selected_line is extracted in a new move as expected.
+        if selected_line.product_uom_qty:
+            selected_line.product_uom_qty = 0
+        move = selected_line.move_id
+        if selected_line.qty_done == move.product_uom_qty:
```
jbaudoux: small remark: line and move line could have different uom

shopfloor_reception/services/reception.py
```diff
@@ -1237,15 +1262,24 @@ def _post_shopfloor_created_line(self, selected_line):
...
+            # In case of full quantity, post the initial move
+            selected_line.move_id.extract_and_action_done()
```
jbaudoux: I think we are then done here and can return
```suggestion
            return selected_line.move_id.extract_and_action_done()
```

shopfloor_reception/services/reception.py
```diff
@@ -1237,15 +1262,24 @@ def _post_shopfloor_created_line(self, selected_line):
...
+        split_move_vals = move._split(selected_line.qty_done)
```
jbaudoux, line 1298: small remark: line uom vs move uom

shopfloor_reception/services/reception.py
```diff
@@ -1237,15 +1262,24 @@ def _post_shopfloor_created_line(self, selected_line):
...
-        )
-        if new_move:
-            # A new move is created in case of partial quantity
-            new_move.extract_and_action_done()
-            return
-        # In case of full quantity, post the initial move
-        selected_line.move_id.extract_and_action_done()
+        # If user only processed 1/5 and is the only one working on the move,
+        # then selected_line is the only one related to this move.
+        # In such case, we must ensure there's another move line with the remaining
```
```suggestion
        # In such case, we must ensure there's another move with the remaining
```

shopfloor_reception/services/reception.py
```diff
@@ -1237,15 +1262,24 @@ def _post_shopfloor_created_line(self, selected_line):
...
+        # quantity to do, so selected_line is extracted in a new move as expected.
+        if selected_line.product_uom_qty:
+            selected_line.product_uom_qty = 0
```
jbaudoux: Safer to do that on any line having a product_uom_qty. Especially that before your change in this PR, the lines had a computed product_uom_qty.
```suggestion
        lines_with_qty_todo = selected_line.move_id.move_line_ids.filtered("product_uom_qty")
        if lines_with_qty_todo:
          lines_with_qty_todo.product_uom_qty = 0
```

If you have a line with 1/5 and another user creates a line with 1/0 then he cannot post his line unless you change the other initial line

jbaudoux (changes requested):
Looks promising. I still have a few remarks

shopfloor_reception/services/reception.py
TDu: If the `if` case is removed, when the users are processing more than the total quantity, how should the computation work ?
Because if the users end up scanning much more that required, we will see negative values in the to do quantity on the front end.

shopfloor_reception/services/reception.py
```diff
@@ -768,25 +763,32 @@ def _align_product_uom_qties(self, move):
...
+        move = line.move_id
         qty_todo = move.product_uom_qty
-        qty_done = sum(move.move_line_ids.mapped("qty_done"))
+        qty_done = 0.0
+        move_uom = move.product_uom
+        for move_line in move.move_line_ids:
+            # Use move's uom
+            qty_done += move_uom._compute_quantity(
+                move_line.qty_done, move_line.product_uom_id, round=False
+            )
```
TDu: Should we not use the move_line_uom
```suggestion
            line_uom = move_line.product_uom_id
            qty_done += line_uom._compute_quantity(
                move_line.qty_done, move_uom.product_uom_id, round=False
            )
```

TDu: Fixed

shopfloor_reception/services/reception.py
TDu: done

shopfloor_reception/services/reception.py
TDu: Fixed

TDu:
@jbaudoux @mt-software-de  Could I get a review update on this one, thanks

shopfloor_reception/data/shopfloor_scenario_data.xml
```diff
@@ -9,7 +9,8 @@
       <field name="options_edit">
 {
       "auto_post_line": true,
-      "allow_return": true
+      "allow_return": true,
+      "scan_location_or_pack_first": true
```
mt-software-de, line 13: Shouldn't scan_location_or_pack_first set by a migration script to True, on a existing reception scenario?

shopfloor_reception/services/reception.py
```diff
@@ -256,13 +256,8 @@ def _scan_line__find_or_create_line(self, picking, move, qty_done=1):
...
         )
-        if line:
-            # The line quantity to do needs to correspond to
-            # the remaining quantity to do of its move.
-            line.product_uom_qty = move.product_uom_qty - move.quantity_done
-        else:
-            qty_todo_remaining = max(0, move.product_uom_qty - move.quantity_done)
-            values = move._prepare_move_line_vals(quantity=qty_todo_remaining)
+        if not line:
+            values = move._prepare_move_line_vals()
```
mt-software-de: Shouldn't there not a kind of lock?
If users work in parrallel, they will get serialize errors.
Because in the method _scan_line__assign_user the picking.user_id will be set and also of course the shopfloor_user_id of the stock.move.line. 
Also i think we should remove the self._assign_user_to_picking from _scan_line__assign_user

If a line was found, shouldn't we change the product_uom_qty to 0. Otherwise _auto_post_line can maybe make some trouble if we work in parallel.

shopfloor_reception/services/reception.py
```diff
@@ -1237,15 +1258,34 @@ def _post_shopfloor_created_line(self, selected_line):
...
+            "product_uom_qty"
         )
-        if new_move:
-            # A new move is created in case of partial quantity
-            new_move.extract_and_action_done()
-            return
-        # In case of full quantity, post the initial move
-        selected_line.move_id.extract_and_action_done()
+        if lines_with_qty_todo:
+            lines_with_qty_todo.product_uom_qty = 0
```
mt-software-de, line 1290: We should ensure that there is only one line with product_uom_qty on _scan_line__find_or_create_line we should ensure that the product_uom_qty is 0. Otherwise we can find maybe more then one line?

shopfloor_reception/services/reception.py
```diff
@@ -1237,15 +1258,34 @@ def _post_shopfloor_created_line(self, selected_line):
...
+            # In case of full quantity, post the initial move
+            return selected_line.move_id.extract_and_action_done()
+        split_move_vals = move._split(selected_line.qty_done)
+        new_move = move.create(split_move_vals)
+        new_move.move_line_ids = selected_line
+        new_move._action_confirm(merge=False)
+        new_move._recompute_state()
+        new_move._action_assign()
+        move._recompute_state()
+        new_move.extract_and_action_done()
```
mt-software-de: Also here if we work in parallel we have to add a lock.
Otherwise the recompute_state will run into an error.

mt-software-de (changes requested):
I think we need some locks. Otherwise working in parallel will not work.

shopfloor_reception/services/reception.py
mt-software-de: The user should be unassigned from the picking or never set the user on the picking?

shopfloor_reception/data/shopfloor_scenario_data.xml
TDu: Not if we want to keep the same behavior for existing implementation.

shopfloor_reception/services/reception.py
TDu: With this scenario there can be multiple shopfloor users on the same move. Each one will have its own `stock.move.line` assigned. The `product_uom_qty` are kept to zero on all `stock.move.line` until they are being posted. This is to avoid concurrency issues, when users are scanning more than what is required by the `stock.move`.
So it is actually the goal to update all lines with a positive value on `product_uom_qty` back to zero.
All the done line that we would not want to update have been extracted on done.

shopfloor_reception/services/reception.py
TDu: > The user should be unassigned from the picking or never set the user on the picking?

The picking are never selected based on the ` user_id`

TDu: > Also here if we work in parallel we have to add a lock. Otherwise the recompute_state will run into an error.

What kind of error due to concurrent update, example ?

shopfloor_reception/services/reception.py
TDu: I think removing `self._assign_user_to_picking `   from _scan_line__assign_user is a good idea because the user assigned to picking is not used in this scenario.
@jbaudoux  ?

About reseting `product_uom_qty` to zero it is probably already done enough in this scenario (Before line is posted, all are reset)

shopfloor_reception/services/reception.py
TDu: I also improved the filter above to exclude on the state as well, just in case and to be more explicit.

TDu:
Rebased on 14.0

shopfloor_reception/services/reception.py
mt-software-de: I mean if a user starts and there is already an unassigned move.line where product_uom_qty is already set. I think it would be safe to set it to 0, then it would be the same as creating a new move.line

shopfloor_reception/services/reception.py
mt-software-de: What i mean. In _post_line you are setting the product_uom_qty
This line will be found bei auto_post_line normally it should only find the selected_line,
because other lines from other users should also have set  a qty of 0. 
But in _scan_line__find_or_create_line there could be a existing unassigned move.line returned with a qty set.

Also i think we can remove lines_with_qty_todo and always use the selected line.

shopfloor_reception/services/reception.py
mt-software-de: > > The user should be unassigned from the picking or never set the user on the picking?
> 
> The picking are never selected based on the ` user_id`

I mean if keep the assign of the user on the picking, we should remove it on a split case.

mt-software-de: > > Also here if we work in parallel we have to add a lock. Otherwise the recompute_state will run into an error.
> 
> What kind of error due to concurrent update, example ?

I mean a serialize access error. Because on a split case we always creating a new move which will be set to done. The initial move stays. This means if the users work in parallel and processing the same move which own a move.line one by one. The recompute_state will be always triggered for the same move. And this can run it a serialize access error.

shopfloor_reception/services/reception.py
mt-software-de: If two users start a the same time with the same product this will cause an serialize access error, because the will both get maybe the same existing move.line if there is one.
And if we are not remove the assign_user_to_picking this will cause also an serialize access error.

shopfloor_reception/services/reception.py
TDu: I added a commit to remove this assignation of the shopfloor user to the picking.

mmequignon: it's done, I resolve the conversation (to make it easier for JE to ignore such discussion)

mt-software-de: But we still need a lock

shopfloor_reception/services/reception.py
```diff
@@ -1241,15 +1270,35 @@ def _post_shopfloor_created_line(self, selected_line):
...
+        if selected_line.qty_done == move_quantity:
+            # In case of full quantity, post the initial move
+            return selected_line.move_id.extract_and_action_done()
+        split_move_vals = move._split(selected_line.qty_done)
+        new_move = move.create(split_move_vals)
+        new_move.move_line_ids = selected_line
+        new_move._action_confirm(merge=False)
+        new_move._recompute_state()
+        new_move._action_assign()
+        move._recompute_state()
```
mt-software-de, line 1300: There should also be a lock.

mt-software-de (changes requested):

shopfloor_reception/services/reception.py
mt-software-de, line 1315: There should also be a lock. 
```suggestion
        lock = self._actions_for("lock")
        lock.for_update(move)
        move._recompute_state()
```

mt-software-de (changes requested):

shopfloor_reception/services/reception.py
TDu: This function has been improved by the commit [shopfloor_reception: improve move line assignation to user](https://github.com/OCA/wms/pull/726/commits/d56d148c0feddef38fe6045852f1e29acb98b2f0)
First to lock a row being assigned to the user.
Second the first implementation missed some sorting and the first line being choosen could have been one without shopfloor_user assigned although a line with the current user assigned does exist.

shopfloor_reception/services/reception.py
TDu: Fixed

TDu:
@mt-software-de PR updated could you give another review, thank you

shopfloor_reception/services/reception.py
```diff
@@ -1241,15 +1270,37 @@ def _post_shopfloor_created_line(self, selected_line):
...
+            # In case of full quantity, post the initial move
+            return selected_line.move_id.extract_and_action_done()
+        split_move_vals = move._split(selected_line.qty_done)
+        new_move = move.create(split_move_vals)
+        new_move.move_line_ids = selected_line
+        new_move._action_confirm(merge=False)
+        new_move._recompute_state()
+        new_move._action_assign()
+        lock = self._actions_for("lock")
+        lock.for_update(move)
```
jbaudoux: Should be before the call to _split that modifies the qty to do on the move

jbaudoux: should even be before `if lines_with_qty_todo:`

jbaudoux:
@Tdu Can you squash fixups?

shopfloor_reception/services/reception.py
TDu: I moved it.

shopfloor_reception/services/reception.py
```diff
@@ -248,27 +245,37 @@ def _select_document_from_lot(self, lot):
...
+                unassigned_lines |= move_line
+        if not line and unassigned_lines:
+            lock = self._actions_for("lock")
+            for move_line in unassigned_lines:
+                if lock.for_update(move_line, skip_locked=True):
+                    line = move_line
+                    break
+        if not line:
+            values = move._prepare_move_line_vals()
             line = self.env["stock.move.line"].create(values)
```
mt-software-de, line 274: I think we should ensure that the product_uom_qty is always 0. 

```suggestion
        if not line:
            values = move._prepare_move_line_vals()
            line = self.env["stock.move.line"].create(values)
        else:
            line.product_uom_qty = 0
```

shopfloor_reception/services/reception.py
mt-software-de: Is it really needed to check for other stock.move.lines with a product_uom_qty > 0?
Because i think every scanned line should have a product_uom_qty of 0, until _post_line is called.

jbaudoux (approved):

jbaudoux:
@mt-software-de I think everything as been addressed now. I have not re-tested. Can you update your review?

mt-software-de (changes requested):

mt-software-de:
> @mt-software-de I think everything as been addressed now. I have not re-tested. Can you update your review?
Not everything. Could you have a look at my last review?

shopfloor_reception/services/reception.py
mmequignon: yes, because nobody prevents anybody to modify this from the UI.
Hence, we might have issues if `sum(move.move_line_ids.mapped("product_uom_qty")) > move.product_uom_qty`, when recomputing the state of the move.

mmequignon: We can set product_uom_qty = 0 everywhere in the code, but what really matters, here, is that those are aligned when recomputing the state of the move.
Otherwise the move state moves from assigned to partially available, and is not visible in shopfloor.

shopfloor_reception/services/reception.py
jbaudoux: We do this on the post for backward compatibility with previous implementation (lines that would already be there with a qty != 0). Also if you have only one line pre-generated by odoo, you have a qty != 0 and this is not causing any issue.
When posting, it is required we care other lines (if any) are well set to 0 otherwise you cannot post.

mmequignon: https://github.com/OCA/wms/pull/726#discussion_r1394116955

mmequignon:
> > @mt-software-de I think everything as been addressed now. I have not re-tested. Can you update your review?
> > Not everything. Could you have a look at my last review?

Thanks for your insights! :pray: 

However, this PR was meant to revert a specific change, and I believe we went too far.
Now that this looks (IMO) good, we should focus on merging this, and open a new PR later if needed.
This will help to have kind of atomic PRs, and also make the merging faster.
3 months for what was initially supposed to be a "simple" revert change is way too long.
It doesn't help maintaining those pending merges.

mt-software-de:
> > > @mt-software-de I think everything as been addressed now. I have not re-tested. Can you update your review?
> > > Not everything. Could you have a look at my last review?
> 
> Thanks for your insights! 🙏
> 
> However, this PR was meant to revert a specific change, and I believe we went too far. Now that this looks (IMO) good, we should focus on merging this, and open a new PR later if needed. This will help to have kind of atomic PRs, and also make the merging faster. 3 months for what was initially supposed to be a "simple" revert change is way too long. It doesn't help maintaining those pending merges.

I understand you, but i have unanswered questions to code which was introduced within this PR, maybe i am missing something. But i think this should be answered?

shopfloor_reception/services/reception.py
mt-software-de: Thx for the explanation. I tested this quick. 
If i'm working with two users. Both are scanning both getting an own move.line. 
The first one creates a new package and sets the the destination and enters the _post_line method.
Within the _auto_post_line the split is happening a new picking is created in state done.
And the move._recompute_state() sets the state to "confirmed". 
So the move state is not anymore 'assigned' as you said. 
What is clear if i am looking at the code. 
https://github.com/odoo/odoo/blob/f572335756ef60402236cf5e4e11c0c843c3879e/addons/stock/models/stock_move.py#L1653
https://github.com/odoo/odoo/blob/f572335756ef60402236cf5e4e11c0c843c3879e/addons/stock/models/stock_move.py#L1659C25-L1659C26
It will only be set to 'assigned' if the move.reserved_availability ( sum of move.lines product_uom_qty) is the same as the move's product_uom_qty, and this isn't the case.

TDu: I could reproduce the issue you mentioned.
My last commit fixes that.

TDu:
Rebased

TDu:
Rebased

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -153,6 +153,7 @@ const Reception = {
                         <v-row align="center">
                             <v-col class="text-center" cols="12">
                                 <btn-back/>
+                                <cancel-button/>
```
mt-software-de, line 158:
```suggestion
                        </v-col>
                        <v-col class="text-center" cols="12">
                                <cancel-button/>
```

mt-software-de: A own v-col would be great i think. 
image

With suggestion, it would look like this. 
image

mt-software-de (changes requested):

shopfloor_reception_mobile/static/src/scenario/reception.js
jbaudoux: Do we need both back & cancel ?

TDu: I don't know, but
the Cancel button will call the back end to cancel the done quantity on the current line (line will be removed).
the Back button goes back to the previous screen without calling the backend.

TDu: I have done the change to have the button spaced out.

mt-software-de (approved):
LGTM now :-). Thanks

shopfloor_reception_mobile/static/src/scenario/reception.js
jbaudoux: In my opinion, it should be either cancel or back. So cancel should replace the back button. Maybe @mmequignon can clarify the intention here?

PR #732, opened 2023-08-25
[14.0][ADD] shopfloor_location_package_restriction

TDu:
~~Needs : https://github.com/OCA/stock-logistics-warehouse/pull/1797~~

This is not a good solution of the problem, needing 1 shopfloor module to nicely handle the errors/exception of another module.
A generic solution for this problem can be seen in https://github.com/OCA/wms/pull/768

ref.: rau-200


shopfloor/services/zone_picking.py
```diff
@@ -869,10 +869,11 @@ def _set_destination_location(self, move_line, quantity, confirmation, location)
         # if `confirmation is True
         # Ask confirmation to the user if the scanned location is not in the
         # expected ones but is valid (in picking type's default destination)
-        if not self.is_dest_location_valid(move_line.move_id, location):
+        message = self._validate_destination(location, move_line.move_id)
+        if message:
```
mmequignon: looks like a handler to me :wink:

TDu: Everything looks like a nail when you have a hammer.

shopfloor/services/zone_picking.py
```diff
@@ -1451,17 +1452,18 @@ def set_destination_all(self, barcode, confirmation=False):
...
             if len(location_dest) != 1:
-                error = self.msg_store.lines_different_dest_location()
+                message = self.msg_store.lines_different_dest_location()
             # check if the scanned location is allowed
             moves = buffer_lines.mapped("move_id")
-            if not self.is_dest_location_valid(moves, location):
-                error = self.msg_store.location_not_allowed()
-            if error:
-                return self._set_destination_all_response(buffer_lines, message=error)
+            if not message:
```
jbaudoux, line 1545: `else:` would be easier to read

shopfloor/actions/message.py
```diff
@@ -64,6 +64,15 @@ def location_requires_package(self):
...
 
+    def location_has_restrictions(self):
+        return {
+            "message_type": "warning",
+            "body": _(
+                "This location has restrictions. It requires a package and may have "
+                "some limit on package quantity."
+            ),
+        }
+
```
jbaudoux: The message is provided by `_check_location_package_restriction`. You should not build a new one

shopfloor/services/single_pack_transfer.py
```diff
@@ -205,10 +205,21 @@ def _create_package_level(self, package):
...
+        moves = package_level.move_line_ids.move_id
+        if not location:
+            return self._response_for_scan_location(
+                package_level, message=self.msg_store.no_location_found()
+            )
+        if not self.is_dest_location_valid(moves, location):
+            return self._response_for_scan_location(
+                package_level, message=self.msg_store.dest_location_not_allowed()
+            )
+        return None
```
jbaudoux: Can't we move this to service level instead of repeating it in each service?

shopfloor/services/zone_picking.py
```diff
@@ -1451,17 +1452,18 @@ def set_destination_all(self, barcode, confirmation=False):
         message = None
         buffer_lines = self._find_buffer_move_lines()
         if location:
-            error = None
+            # error = None
```
jbaudoux: line to remove

jbaudoux (changes requested):
Instead of doing many glue modules for catching validation errors can't we catch any validation error, rollback the transaction and return the error ?
Here you solved the ValidationError from location_package_restriction, but there will be the same issue with [location_product_restriction](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_location_product_restriction)
cc @simahawk @lmignon

shopfloor/services/zone_picking.py
mmequignon: aha

mmequignon:
> Instead of doing many glue modules for catching validation errors can't we catch any validation error, rollback the transaction and return the error ? Here you solved the ValidationError from location_package_restriction, but there will be the same issue with [location_product_restriction](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_location_product_restriction) cc @simahawk @lmignon

This is basically what I did here (except that I'm not looking for exceptions).
https://github.com/OCA/wms/blob/14.0/shopfloor_single_product_transfer/services/single_product_transfer.py#L18
Then you decorate your methods with this and you're good.
Not very flexible atm, but it can be improved for more generic needs.
For instance, we could pass a list of blocking message types and a list exceptions, which would trigger a rollback

```python
@with_savepoint(blocking_types=["error"], blocking_exceptions=[UserError, ValidationError])
def my_endpoints(...)
```

jbaudoux:
> > Instead of doing many glue modules for catching validation errors can't we catch any validation error, rollback the transaction and return the error ? Here you solved the ValidationError from location_package_restriction, but there will be the same issue with [location_product_restriction](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_location_product_restriction) cc @simahawk @lmignon
> 
> This is basically what I did here (except that I'm not looking for exceptions). https://github.com/OCA/wms/blob/14.0/shopfloor_single_product_transfer/services/single_product_transfer.py#L18 Then you decorate your methods with this and you're good. Not very flexible atm, but it can be improved for more generic needs. For instance, we could pass a list of blocking message types and a list exceptions, which would trigger a rollback
> 
> ```python
> @with_savepoint(blocking_types=["error"], blocking_exceptions=[UserError, ValidationError])
> def my_endpoints(...)
> ```

How do you build properly the response to pass all the right arguments ?

mmequignon:
> > > Instead of doing many glue modules for catching validation errors can't we catch any validation error, rollback the transaction and return the error ? Here you solved the ValidationError from location_package_restriction, but there will be the same issue with [location_product_restriction](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_location_product_restriction) cc @simahawk @lmignon
> > 
> > 
> > This is basically what I did here (except that I'm not looking for exceptions). https://github.com/OCA/wms/blob/14.0/shopfloor_single_product_transfer/services/single_product_transfer.py#L18 Then you decorate your methods with this and you're good. Not very flexible atm, but it can be improved for more generic needs. For instance, we could pass a list of blocking message types and a list exceptions, which would trigger a rollback
> > ```python
> > @with_savepoint(blocking_types=["error"], blocking_exceptions=[UserError, ValidationError])
> > def my_endpoints(...)
> > ```
> 
> How do you build properly the response to pass all the right arguments ?

I'm not sure to understand your question.
Let's have a call about that.

shopfloor/services/zone_picking.py
TDu: You mean the next `if` clause ? But no because the first `if` could return a message ?

shopfloor/actions/message.py
TDu: I fixed that

shopfloor/services/zone_picking.py
TDu: Removed

shopfloor/actions/message.py
TDu: And I removed the message

shopfloor/services/single_pack_transfer.py
TDu: I did.

TDu:
> Instead of doing many glue modules for catching validation errors can't we catch any validation error, rollback the transaction and return the error ? Here you solved the ValidationError from location_package_restriction, but there will be the same issue with [location_product_restriction](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_location_product_restriction) cc @simahawk @lmignon

I have opened this PR https://github.com/OCA/wms/pull/768 that implements a way to handle any ValidationError when handling a response.

It could replace this one

mmequignon:
> > Instead of doing many glue modules for catching validation errors can't we catch any validation error, rollback the transaction and return the error ? Here you solved the ValidationError from location_package_restriction, but there will be the same issue with [location_product_restriction](https://github.com/OCA/stock-logistics-warehouse/tree/16.0/stock_location_product_restriction) cc @simahawk @lmignon
> 
> I have opened this PR #768 that implements a way to handle any ValidationError when handling a response.
> 
> It could replace this one

We're still missing what was the requirement in this case:
Such as, rollback in case of error, or error management.

PR #734, opened 2023-08-29
[14.0][FIX] shopfloor: fix bug in checkout if there are two lines with the same article

JuMiSanAr:
If you had the same article from on two different locations in the `select_package` screen, you were only able to increment the qty done of the first one when scanning a barcode.

Now, you will:
- Increment the first line that hasn't been fully processed yet (`qty_done < product_uom_qty`).
- If all lines have been processed, increment the first one found.

ref: rau-216

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -970,6 +970,18 @@ class Checkout(Component):
             ),
         )
 
+    def _find_line_to_increment(self, product_lines):
+        """Find which line should have its qty incremented.
+
+        Return the first line for the scanned product
+        which still has some qty todo.
+        If none are found, return the first line for that product.
+        """
+        return next(
+            (line for line in product_lines if line.qty_done < line.product_uom_qty),
+            fields.first(product_lines),
+        )
+
     def _scan_package_action_from_product(
         self, picking, selected_lines, product, packaging=None, **kw
     ):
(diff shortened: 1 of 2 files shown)


shopfloor/services/checkout.py
```diff
@@ -970,6 +970,16 @@ def _scan_package_find(self, picking, barcode, search_types=None):
             ),
         )
 
+    def _find_line_to_increment(self, product_lines):
+        # Return the first line for the scanned product
```
TDu: This comment could be a docstring (nicer)

TDu (approved):

shopfloor/services/checkout.py
```diff
@@ -970,6 +970,16 @@ def _scan_package_find(self, picking, barcode, search_types=None):
...
+        # which still has some qty todo.
+        # If none are found, return the first line for that product.
+        line = next(
+            (line for line in product_lines if line.qty_done < line.product_uom_qty),
+            None,
+        )
+        return line or fields.first(product_lines)
```
jbaudoux: You could have put the default value in next
```suggestion
        return next(
            (line for line in product_lines if line.qty_done < line.product_uom_qty),
            fields.first(product_lines),
        )
```

jbaudoux (approved):

PR #738, opened 2023-09-04
[16.0][FIX] stock_release_channel: restriction to partner is applied only if partner is assigned to a release channel

sbejaoui:


lmignon (approved):
LGTM (Code review only)

jbaudoux (changes requested):
With your change, if a partner is not assigned to any release channel then you can match a release channel that is restricted to a specific list of partner. This does not look a good idea to me.

sbejaoui:
> With your change, if a partner is not assigned to any release channel then you can match a release channel that is restricted to a specific list of partner. This does not look a good idea to me.

Indeed, you are right; it was a misunderstanding on my part.

I close this PR.

PR #741, opened 2023-09-08
[FIX] stock_storage_type: fix es translation

nilshamerlinck:
`pre-commit` pylint check is failing because of incorrect translations done through weblate


nilshamerlinck:
Hi @sbidoul
- the issue here had a similar root cause as https://github.com/OCA/l10n-belgium/pull/200 (`{period}` mistakenly translated to `{período}` on weblate)
- if I understand correctly, enabling [`python-brace-format`](https://docs.weblate.org/en/weblate-4.17/user/checks.html#python-brace-format) check in weblate could help to prevent it

sbidoul:
> if I understand correctly, enabling [python-brace-format](https://docs.weblate.org/en/weblate-4.17/user/checks.html#python-brace-format) check in weblate could help to prevent it

Thanks for the tip, I'll look into that.

/ocabot merge patch

PR #742, opened 2023-09-09
[14.0][FIX] sf_single_product_transfer: fix backorder responsible

jbaudoux:
Remove user from backorder

cc @mmequignon

### DIFF
--- a/shopfloor_single_product_transfer/services/single_product_transfer.py
+++ b/shopfloor_single_product_transfer/services/single_product_transfer.py
@@ -666,13 +666,15 @@ class ShopfloorSingleProductTransfer(Component):
         # backorder, which should not be the case.
         # See if there's a way to identify the moves
         # generated through this mechanism and avoid creating them.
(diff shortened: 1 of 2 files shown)


mmequignon:
Why not setting "shopfloor_user_id" to False when copying the move line, in _split_partial_quantity ?
https://github.com/OCA/wms/blob/14.0/shopfloor/models/stock_move_line.py#L48

Also, when looking for move lines, why are we looking for move lines having their picking unassigned?
https://github.com/OCA/wms/blob/14.0/shopfloor_single_product_transfer/services/single_product_transfer.py#L230-L235
Why aren't we looking for move lines with  `shopfloor_user_id in (False, self.env.uid)` ?

jbaudoux:
Rebased

jbaudoux:
Can you review this one @Tdu @mt-software-de ?

mt-software-de (approved):
LGTM

PR #747, opened 2023-09-12
[14.0][FIX] shopfloor: fix scan anything for packaging

TDu:
When scanning a packaging barcode, the backend crashes because it tries to generate the data for a packaging using the function designed for products.

I guess it is by design that when scanning a packaging barcode, it is its content (the product) that is returned to the frontend.

ref.: rau-232


jbaudoux (approved):

simahawk (commented):
can you add a test please?

TDu:
> can you add a test please?

done

simahawk (approved):

PR #748, opened 2023-09-14
[14.0][IMP] shopfloor: imp cancel line/package in location content transfer

TDu:
ref.: rau-229

PR #749, opened 2023-09-14
[14.0][IMP] shopfloor: imp cancel on single pack transfer

TDu:
Before the package level was only set as not done, when calling the cancel route.

Now if the package level has been created by the
current shopfloor user, its related transfer is also canceled.

ref.: rau-229

### DIFF
--- a/shopfloor/services/single_pack_transfer.py
+++ b/shopfloor/services/single_pack_transfer.py
@@ -283,6 +283,12 @@ class SinglePackTransfer(Component):
             return self._response_for_start(message=self.msg_store.already_done())
 
         package_level.is_done = False
+        if package_level.picking_id.create_uid == self.env.user:
+            # Cancel the transfer when it has been created by the shopfloor user
+            moves._do_unreserve()
+            moves._recompute_state()
+            moves.picking_id.action_cancel()
+
         return self._response_for_start(
             message=self.msg_store.confirm_canceled_scan_next_pack()
         )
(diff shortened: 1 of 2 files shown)


shopfloor/services/single_pack_transfer.py
```diff
@@ -283,6 +283,12 @@ def cancel(self, package_level_id):
             return self._response_for_start(message=self.msg_store.already_done())
 
         package_level.is_done = False
+        if package_level.picking_id.create_uid == self.env.user:
```
jbaudoux:
```suggestion
        if self.is_allow_move_create() and package_level.picking_id.create_uid == self.env.user:
```

shopfloor/services/single_pack_transfer.py
```diff
@@ -283,6 +283,12 @@ def cancel(self, package_level_id):
...
+            # Cancel the transfer when it has been created by the shopfloor user
+            moves._do_unreserve()
+            moves._recompute_state()
+            moves.picking_id.action_cancel()
```
jbaudoux: You can drop the 2 first lines, action_cancel is doing it

shopfloor/services/single_pack_transfer.py
jbaudoux: and where is the else mark picking as unpicked like you just did on the other scenario in https://github.com/OCA/wms/pull/748 ?

jbaudoux (changes requested):

jbaudoux (approved):
LGTM

mt-software-de (approved):
LG

PR #750, opened 2023-09-15
[16.0] [ADD] stock_release_channel_shipment_lead_time

anothingguy:
depends on:
- https://github.com/OCA/wms/pull/550
- https://github.com/OCA/stock-logistics-transport/pull/88

PR #751, opened 2023-09-18
[16.0][ADD] stock_release_channel_partner_public_holidays

nguyenminhchien:
Depends on:
* https://github.com/OCA/wms/pull/750
* https://github.com/OCA/wms/pull/550
* https://github.com/OCA/stock-logistics-transport/pull/106
* https://github.com/OCA/wms/pull/760


cyrilmanuel (approved):
not an expert but lgtm

stock_release_channel_partner_public_holidays/models/stock_release_channel.py
```diff
@@ -0,0 +1,54 @@
...
+        res = False
+        shipment_date = self.shipment_date
+        if not shipment_date:
+            return res
+        holidays_filter = [("year", "=", shipment_date.year)]
+        if partner:
+            if partner.country_id:
+                holidays_filter.append("|")
+                holidays_filter.append(("country_id", "=", False))
+                holidays_filter.append(("country_id", "=", partner.country_id.id))
```
jbaudoux: FYI, you can write this as `holidays_filter.append(("country_id", "in", (False, partner.country_id.id)))`

stock_release_channel_partner_public_holidays/models/stock_release_channel.py
```diff
@@ -0,0 +1,54 @@
...
+            states_filter += [
+                "|",
+                ("state_ids", "=", False),
+                ("state_ids", "=", partner.state_id.id),
+            ]
+        else:
+            states_filter.append(("state_ids", "=", False))
+        states_filter.append(("date", "=", shipment_date))
+        hhplo = self.env["hr.holidays.public.line"]
+        holidays_line = hhplo.search(states_filter, limit=1)
```
jbaudoux: Can't we combine the 2 searches ?
```
domain = [
  ("year_id.year", "=", shipment_date.year), <= not sure we even need this as we already filter on the date
  ("year_id.country_id", ...),
  ("state_ids", ...),
  ("date", "=", shipment_date),
]
holidays_line = hhplo.search(domain, limit=1)

This should result in a single SQL request (maybe auto_join on year_id is needed)

File: stock_release_channel_partner_public_holidays/models/stock_release_channel.py (hunk 1 above)
Comment (jbaudoux) [reply stock_release_channel_partner_public_holidays/models/stock_release_channel.py]: This should even work if there is no country on the partner as it will resolve as `(False, False)`, allowing to drop the if/else

**Review by jbaudoux — CHANGES_REQUESTED** [review]
Code can be simplified

File: stock_release_channel_partner_public_holidays/models/stock_release_channel.py (hunk 1 above)
Comment (nguyenminhchien) [reply stock_release_channel_partner_public_holidays/models/stock_release_channel.py]: `("country_id", "in", (False, False))` is not good, it's against the indexes search.

Comment (jbaudoux) [reply stock_release_channel_partner_public_holidays/models/stock_release_channel.py]: https://dev.to/yugabyte/postgresql-do-different-equality-predicates-make-a-difference-1c7

Comment (nguyenminhchien) [reply stock_release_channel_partner_public_holidays/models/stock_release_channel.py]: there is no different between `("country_id", "in", [False])` and `("country_id", "=", [False])`, because DMS is smart enough to convert "in" into "=". But `("country_id", "in", (False, False))` is not.

**Review by jbaudoux — APPROVED** [review]
Thanks

File: stock_release_channel_partner_public_holidays/models/stock_release_channel.py [hunk 3]
```diff
@@ -0,0 +1,39 @@
...
+            if channel._is_shipment_date_a_public_holiday(partner):
+                channels -= channel
+        return channels
+
+    def _is_shipment_date_a_public_holiday(self, partner):
+        """
+        Returns True if shipment_date is a public holiday
+        :return: bool
+        """
+        self.ensure_one()
```
simahawk, line 25: there's no need to call ensure_one for a private method that is used only in this loop

stock_release_channel_partner_public_holidays/models/stock_release_channel.py
```diff
@@ -0,0 +1,39 @@
...
+            return res
+        domain = [
+            ("year_id.country_id", "in", (False, partner.country_id.id)),
+            "|",
+            ("state_ids", "=", False),
+            ("state_ids", "=", partner.state_id.id),
+            ("date", "=", shipment_date),
+        ]
+        hhplo = self.env["hr.holidays.public.line"]
+        holidays_line = hhplo.search(domain, limit=1, order="id")
```
simahawk, line 38: should use search_count instead

stock_release_channel_partner_public_holidays/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Nguyen Minh Chien <chien@trobz.com>
```
simahawk, line 1: JE missing

simahawk (approved):
Overall LG, some minor glitches found but can be fixed in another round.
Rebased and dropped test req.

PR #767, opened 2023-10-16
[14.0][FIX] shopfloor_mobile_base: fix searchbar focus

JuMiSanAr:
The searchbar wasn't always ready for scanned inputs unless the user manually clicked on it.
Using `focus()` is not enough, but combining it with `click()` did work in our tests.

ref: cos-4175  & cos-4242

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/zone_picking.js
+++ b/shopfloor_mobile/static/wms/src/scenario/zone_picking.js
@@ -7,6 +7,8 @@
 import {ScenarioBaseMixin} from "/shopfloor_mobile_base/static/wms/src/scenario/mixins.js";
 import {process_registry} from "/shopfloor_mobile_base/static/wms/src/services/process_registry.js";
 
+// TODO: consider replacing the dynamic "autofocus" in the searchbar by an event.
+// At the moment, we need autofocus to be disabled if there's a user popup.
 const template_mobile = `
     <Screen :screen_info="screen_info">
         <template v-slot:header>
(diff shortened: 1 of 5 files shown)


shopfloor_mobile/static/wms/src/scenario/zone_picking.js
```diff
@@ -16,6 +16,7 @@ const template_mobile = `
             v-if="state.on_scan"
             v-on:found="on_scan"
             :input_placeholder="search_input_placeholder"
+            :autofocus="!screen_info.user_popup"
```
TDu, line 21: Should it not be added to all the scenarios ?

simahawk: Good point.

simahawk: BTW any chance to link avoid having to specify this for all scenario in the future? Eg: event or ref?

JuMiSanAr: I only implemented it in the scenarios that use this popup (to avoid repeating it in all scenarios).

Regarding the event, I didn't want to touch the base `searchbar` component and keep its logic as is. I prefer just passing this dynamic autofocus, as we don't really gain anything by using events right?

simahawk: We do in fact. Consumers of the searchbar won't have to know about internal implementations.
Anyway, it's ok. Leave it like this for now but add a TODO if you don't mind pls.

simahawk (approved):

PR #768, opened 2023-10-17
[14.0][IMP] shopfloor: add decorator to gracefully handle exceptions on response

TDu:
This PR adds a decorator that can be used on any endpoint to catch some exceptions and return to the user a clean error message and not a 500 error message.

This is to solve the problem discussed on https://github.com/OCA/wms/pull/732

### DIFF
--- a/shopfloor_base/utils.py
+++ b/shopfloor_base/utils.py
@@ -43,15 +43,18 @@ def catch_errors(response_error):
 
     def _catch_errors(fn):
         @wraps(fn)
-        def wrapper(*args, **kwargs):
+        def wrapper(self, *args, **kwargs):
+            savepoint = self._actions_for("savepoint").new()
             try:
-                res = fn(*args, **kwargs)
+                res = fn(self, *args, **kwargs)
             except ValidationError as ex:
+                savepoint.rollback()
                 message = {
                     "message_type": "error",
                     "body": ex.name,
                 }
-                return response_error(*args, **kwargs, message=message)
+                return response_error(self, *args, **kwargs, message=message)
+            savepoint.release()
             return res
 
         return wrapper
--- a/shopfloor_base/utils.py
+++ b/shopfloor_base/utils.py
@@ -5,6 +5,7 @@ import logging
 import os
 from functools import wraps
 
+from odoo.exceptions import ValidationError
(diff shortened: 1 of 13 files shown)


shopfloor_base/utils.py
```diff
@@ -33,6 +34,31 @@ def wrapped(*args, **kwargs):
...
+                message = {
+                    "message_type": "error",
+                    "body": ex.name,
+                }
+                return response_error(*args, **kwargs, message=message)
+            return res
+
+        return wrapper
+
+    return _catch_errors
```
mmequignon: it would be nice to have a way to rollback the transaction in case of an error:
```suggestion
    def _catch_errors(fn):
        @wraps(fn)
        def wrapper(self, *args, **kwargs, rollback=False):
            savepoint = self._actions_for("savepoint").new()
            try:
                res = fn(self, *args, **kwargs)
            except ValidationError as ex:
                if rollback:
                    savepoint.rollback()
                else:
                    savepoint.release()
                message = {
                    "message_type": "error",
                    "body": ex.name,
                }
                return response_error(*args, **kwargs, message=message)
            savepoint.release()
            return res

        return wrapper

    return _catch_errors
```
or smth like that.
In this implementation, I'm also checking that response isn't an error.
https://github.com/OCA/wms/blob/14.0/shopfloor_single_product_transfer/services/single_product_transfer.py#L20

TDu: Thank you, yes it is needed, added a fixup

shopfloor_base/utils.py
```diff
@@ -33,6 +34,37 @@ def wrapped(*args, **kwargs):
     return _ensure_model
 
 
+def catch_errors(response_error, exceptions_to_catch=None):
```
mmequignon, line 37:
```suggestion
def with_savepoint(response_error, exceptions_to_catch=None):
```
I would prefer this name, not blocking

mmequignon (approved):

jbaudoux:
@bguillot As discussed on OCA days, here is the PR proposal for catching errors in shopfloor

jbaudoux:
@bguillot Can you drop a review as you were the first to suggest such improvement?

PR #770, opened 2023-10-24
[14.0] shopfloor: add option to order pickings by priority

JuMiSanAr:
This PR adds an option to return the priority number of a picking to the frontend.
In the checkout scenario, we display this priority if available.

It also fixes a bug related to a missing key in shopfloor_menu.xml

ref: cos-4215

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -360,7 +360,7 @@ class Checkout(Component):
         ]
 
     def _order_for_list_stock_picking(self):
-        return "scheduled_date asc, id asc"
+        return "priority desc, scheduled_date asc, id asc"
 
     def list_stock_picking(self):
         """List stock.picking records available
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -13,7 +13,7 @@ class DataAction(Component):
     def location(self, record, **kw):
         parser = self._location_parser
         data = self._jsonify(record.with_context(location=record.id), parser, **kw)
-        if "with_operation_progress" in kw:
+        if kw.get("with_operation_progress"):
             lines_blacklist = (
                 kw.get("progress_lines_blacklist")
                 or self.env["stock.move.line"].browse()
@@ -46,7 +46,7 @@ class DataAction(Component):
         # and it may reduce performance significatively
         # when dealing with a large number of pickings.
         # Thus, we make it optional.
-        if "with_progress" in kw:
+        if kw.get("with_progress"):
             parser.append("progress")
         return parser
 
@@ -72,6 +72,7 @@ class DataAction(Component):
             "bulk_line_count",
             "total_weight:weight",
             "scheduled_date",
+            "priority",
         ]
 
     @ensure_model("stock.quant.package")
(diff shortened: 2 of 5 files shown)


simahawk (approved):
LG

shopfloor/models/shopfloor_menu.py
```diff
@@ -53,6 +53,12 @@
 to scan a destination package.
 """
 
+ALLOW_ORDER_PICKINGS_BY_PRIORITY_HELP = """
+When listing all pickings in the scenario, display the priority number
```
sebalix:
```suggestion
When listing all transfers in the scenario, display the priority number
```

shopfloor/models/shopfloor_menu.py
```diff
@@ -226,6 +232,14 @@ class ShopfloorMenu(models.Model):
     allow_alternative_destination_package_is_possible = fields.Boolean(
         compute="_compute_allow_alternative_destination_package_is_possible"
     )
+    order_pickings_by_priority = fields.Boolean(
+        string="Order pickings by priority",
```
sebalix:
```suggestion
        string="Order transfers by priority",
```

sebalix (approved):

simahawk:
@JuMiSanAr can you rebase?

JuMiSanAr:
> @JuMiSanAr can you rebase?

When I try to push new content to this PR I get:
```
To github.com:camptocamp/wms.git
 ! [rejected]          14.0--shopfloor--priority -> 14.0--shopfloor--priority (stale info)
error: failed to push some refs to 'github.com:camptocamp/wms.git'
```

sebalix:
@JuMiSanAr I already rebased manually this branch, you'll have to get the history first then re-rebase :)

>  @sebalix sebalix force-pushed the 14.0--shopfloor--priority branch from 8494871 to 0fcc5c4

jbaudoux (changes requested):
Why do we need an option to allow sorting picking by priority?
In the zone picking, you can sort by date or priority. There is a button to choose the sorting.
Why don't you do the same in checkout? (I guess this concerns manual selection).
Let the user choose on shopfloor. I don't see the benefit of such option.

An option to set the default sorting (date or priority) would be a better approach, wouldn't be?

jbaudoux:
> Why do we need an option to allow sorting picking by priority? In the zone picking, you can sort by date or priority. There is a button to choose the sorting. Why don't you do the same in checkout? (I guess this concerns manual selection). Let the user choose on shopfloor. I don't see the benefit of such option.
> 
> An option to set the default sorting (date or priority) would be a better approach, wouldn't be?

Coming back to my comment, in the zone picking you sort either by "priority, date" or "location". So in the checkout, it should be sorted by priority, date and there is no need for an option.

shopfloor/services/checkout.py
```diff
@@ -360,7 +363,10 @@ def _domain_for_list_stock_picking(self):
         ]
 
     def _order_for_list_stock_picking(self):
-        return "scheduled_date asc, id asc"
```
jbaudoux, line 363: Just change this line to `priority desc, scheduled_date asc, id asc" and drop all the rest that looks overkill

jbaudoux (changes requested):

shopfloor/services/checkout.py
JuMiSanAr: Do I also add the button to choose how to sort them, like in zone_picking? Or just leave it ordered by priority first all the time?

JuMiSanAr: I just pushed the updated version, without the button.

jbaudoux (approved):

jbaudoux:
@JuMiSanAr There is a test to fix with the priority

JuMiSanAr:
Closed and replaced by https://github.com/OCA/wms/pull/782

PR #775, opened 2023-11-03
[14.0] shopfloor: add checkout option ask leaf location dest

TDu:
Following this change on the checkout scenario:

- https://github.com/OCA/wms/pull/618

A check has been added to force the user to scan a child location if the destination location is not a leaf in the locations tree.

To allow for flexibility this change adds the options to disable this feature.


shopfloor/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     "name": "Shopfloor",
     "summary": "manage warehouse operations with barcode scanners",
-    "version": "14.0.4.2.4",
+    "version": "14.0.5.0.0",
```
sebalix: Maybe this version is enough
```suggestion
    "version": "14.0.4.3.0",
```

sebalix (approved):

shopfloor/services/checkout.py
```diff
@@ -1405,13 +1405,14 @@ def done(self, picking_id, confirmation=False):
                 )
         lines_done = self._lines_checkout_done(picking)
         dest_location = picking.location_dest_id
```
jbaudoux, line 1407: Should be the move destination location

shopfloor/services/checkout.py
```diff
@@ -1405,13 +1405,14 @@ def done(self, picking_id, confirmation=False):
...
-            [("id", "child_of", dest_location.id), ("usage", "!=", "view")]
-        )
-        if len(child_locations) > 0 and child_locations != dest_location:
-            return self._response_for_select_child_location(
-                picking,
+        if self.work.menu.ask_for_leaf_destination_location:
+            child_locations = self.env["stock.location"].search(
+                [("id", "child_of", dest_location.id), ("usage", "!=", "view")]
             )
+            if len(child_locations) > 0 and child_locations != dest_location:
```
jbaudoux, line 1412: Can we simply test if `location_dest_id.usage == "view"` then we always need to choose a destination location?

jbaudoux (commented):
Why do you need an option for this?

jbaudoux (changes requested):
Discussed with @sebalix . Looks like proposed changes should solve the issue without the need of a new option.

TDu:
> Discussed with @sebalix . Looks like proposed changes should solve the issue without the need of a new option.

If Odoo allows to store products in stock locations that are not a leaf of the location structure, the shopfloor app should do the same.
This is why the options is needed. IMO

shopfloor/services/checkout.py
sebalix: There is something fishy there:

1. as said, we should take the destination location of the move
2. but what move? we get all lines done of `picking`, so we could get several moves as well, each having a different destination location (e.g. a sub-location of the picking destination location thanks to a put-away)
3. so we should ask the user to scan a relevant leaf location for each move that targets a non-leaf location?

Right now this code is asking the user to scan a leaf location even if it's probably not necessary IMO :thinking:

shopfloor/services/checkout.py
sebalix: Odoo std allows to put internal locations inside internal ones, so checking only usage == `view` is not enough.
In such case we would need such option to decide what behavior we want for each Shopfloor menu.

Now I'm not sure about the current implementation of this feature, see my comment above.

shopfloor/services/checkout.py
jbaudoux: We should always scan a destination if the moves target different destinations

shopfloor/services/checkout.py
jbaudoux: Following sebalix remark that we are processing multiple moves, use this condition:
`len(lines_done.location_dest_id) != 1 or lines_done.location_dest_id.usage == "view"`

jbaudoux:
> If Odoo allows to store products in stock locations that are not a leaf of the location structure, the shopfloor app should do the same. This is why the options is needed. IMO

It's just that the option is currently providing a feature that nobody needs. So, let's not introduce more complexity. Adapting the `if` condition will solve the issue.

TDu:
Following on the comments here, I opened another PR #780 as the fix changes completely.

PR #780, opened 2023-11-14
[14.0][FIX] shopfloor: checkout fix asking destination location

TDu:
Supersedes #775

There could be multiple moves with different destination The destination location could be of type `view`

For both cases, require the user to scan a new location.


sebalix (approved):
LG

jbaudoux (approved):

PR #782, opened 2023-11-17
[14.0]  shopfloor_mobile: display pickings by priority in checkout

JuMiSanAr:
Replaces https://github.com/OCA/wms/pull/770

Change the order of the pickings in the checkout scenario so that they're displayed by priority.
Also, display the priority as star icons.

Ref: cos-4264.

### DIFF
--- a/shopfloor_mobile/static/wms/src/scenario/checkout.js
+++ b/shopfloor_mobile/static/wms/src/scenario/checkout.js
@@ -293,7 +293,20 @@ const Checkout = {
                         {path: "origin"},
                         {path: "carrier.name", label: "Carrier"},
                         {path: "move_line_count", label: "Lines"},
-                        {path: "priority", label: "Priority"},
+                        {
+                            path: "priority",
+                            render_component: "priority-widget",
+                            render_options: function (record) {
+                                const priority = parseInt(record.priority);
+                                // We need to pass the label to the component as an option instead of using "display_no_value"
+                                // because pickings with no priority will still have a string value of "0"
+                                // and the label would always be displayed.
+                                return {
+                                    priority,
+                                    label: priority ? "Priority: " : null,
+                                };
+                            },
+                        },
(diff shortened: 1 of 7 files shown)


shopfloor_mobile/static/wms/src/scenario/checkout.js
```diff
@@ -293,7 +293,17 @@ const Checkout = {
                         {path: "origin"},
                         {path: "carrier.name", label: "Carrier"},
                         {path: "move_line_count", label: "Lines"},
-                        {path: "priority", label: "Priority"},
+                        {
+                            path: "priority",
+                            render_component: "priority-widget",
+                            render_options: function (record) {
+                                const priority = parseInt(record.priority);
+                                // We need to pass the label to the component as an option instead of using "display_no_value"
```
simahawk, line 301: can't we prevent that instead? Empty stars or no star at all?

JuMiSanAr: I tried and I always got the "Priority:" label if there was no assigned priority, followed by no stars.
Displaying empty stars could work, but then it wouldn't be consistent with the way in which we display stars in the other scenario.

jbaudoux:
image

jbaudoux:
image

TDu (approved):

jbaudoux (approved):

shopfloor_mobile/static/wms/src/scenario/checkout.js
jbaudoux: Instead of adding a line, you could also have displayed the stars in-front of the title (picking name) like it is shown on odoo backend

JuMiSanAr: I tried that option but I thought it looked better this way. I think it's ok as is.

PR #784, opened 2023-11-17
[16.0][FIX] stock_storage_type: v16 migration - fix quants for product put-away

jbaudoux:
In case of product put-away computation, the quants extracted from the moves in the context are for all the moves of the transfer. So we need to filter on the product for which we compute the put-away.

cc @rousseldenis @sebalix


rousseldenis (approved):
Great!

PR #785, opened 2023-11-30
[14.0][FIX] shopfloor checkout scanning a lot and quantity done

TDu:


shopfloor/services/checkout.py
```diff
@@ -564,35 +564,37 @@ def _select_lines_from_lot(
                     "body": _("Lot is not in the current transfer."),
                 },
             )
-
-        # When lots are as units outside of packages, we can select them for
-        # packing, but if they are in a package, we want the user to scan the packages.
-        # If the product is only in one package though, scanning the lot selects
-        # the package.
```
sebalix, line 630: I'm not sure to understand in what this comment was wrong?

shopfloor/services/checkout.py
```diff
@@ -564,35 +564,37 @@ def _select_lines_from_lot(
...
                 picking, message=self.msg_store.lot_multiple_packages_scan_package()
             )
         elif packages:
-            # Select all the lines of the package when we scan a lot in a
-            # package and we have only one.
-            return self._select_lines_from_package(
-                picking, selection_lines, packages, prefill_qty=prefill_qty, **kw
+            # The lot is only in one package
+            # Related lines are the one in the same package
+            related_lines = selection_lines.filtered(
```
sebalix, line 644: I don't know what means "related lines" here compared to "line_to_update" or "lines" below. Can you give a definition in a comment?

sebalix (commented):
LG, maybe some clarification needed?

shopfloor/services/checkout.py
TDu: It was not wrong, I thought it was clearer to have the explanation on each `if` cases

shopfloor/services/checkout.py
TDu: I have added a comment above with the declaration of the variable

shopfloor/services/checkout.py
sebalix: Both is good IMO, to get the big picture before entering in the "if/else" dance, and each block gets its own comment

shopfloor/services/checkout.py
sebalix: So if the related lines are the lines belonging to the same package than the lot line, we want to scan a package right? I don't know where this is happening.

shopfloor/services/checkout.py
TDu: But it is the same comment split so each comment is exactly were it is meaningful.

simahawk:
@sebalix @jbaudoux ping

shopfloor/services/checkout.py
sebalix: It was to get the big picture before reading all the code details. But OK

sebalix (approved):

PR #788, opened 2023-12-01
[IMP] stock_storage_type: improve location_is_empty compute performance

sbejaoui:
cc/ @sbidoul

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -7,9 +7,12 @@ from psycopg2 import sql
 
 from odoo import api, fields, models
 from odoo.fields import Command
-from odoo.tools import float_compare
+from odoo.tools import float_compare, index_exists
 
 _logger = logging.getLogger(__name__)
+OUT_MOVE_LINE_DOMAIN = [
+    ("state", "in", ("waiting", "confirmed", "partially_available", "assigned"))
+]
 
 
 class StockLocation(models.Model):
@@ -87,9 +90,7 @@ class StockLocation(models.Model):
     out_move_line_ids = fields.One2many(
         "stock.move.line",
         "location_id",
-        domain=[
-            ("state", "in", ("waiting", "confirmed", "partially_available", "assigned"))
-        ],
+        domain=OUT_MOVE_LINE_DOMAIN,
         help="technical field: the pending outgoing "
         "stock.move.lines in the location",
     )
@@ -140,6 +141,19 @@ class StockLocation(models.Model):
     only_empty = fields.Boolean(
         compute="_compute_only_empty", store=True, recursive=True
     )
+
+    def init(self):  # pylint: disable=missing-return
+        super().init()
+        if not index_exists(self._cr, "stock_move_line_location_state_index"):
+            self._cr.execute(
+                """
+                CREATE INDEX stock_move_line_location_state_index
+                ON
+                    stock_move_line (location_id, state)
+                WHERE
+                    (state IS NULL OR state NOT IN ('cancel', 'done'))
+                """
+            )
 
     @api.depends(
         "usage",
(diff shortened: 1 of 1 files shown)


stock_storage_type/models/stock_location.py
```diff
@@ -314,9 +327,12 @@ def _compute_location_is_empty(self):
             # to enforce concurrent transaction safety: 2 moves taking
             # quantities in a location have to be executed sequentially
             # or the location could remain "not empty"
+            location_domain = [("location_id", "=", rec.id)]
+            domain = OUT_MOVE_LINE_DOMAIN + location_domain
+            out_move_lines = self.env["stock.move.line"].search(domain, order="id")
+            quants = self.env["stock.quant"].search(location_domain, order="id")
```
sbidoul: As discussed let's move these 2 queries out of the loop, in case `location_is_empty` is computed for may locations at once.

sbidoul: Question: does `OUT_MOVE_LINE_DOMAIN` apply to quants too?

stock_storage_type/models/stock_location.py
```diff
@@ -301,22 +314,38 @@ def _compute_location_will_contain_lot_ids(self):
...
-                if not rec.location_is_empty:
-                    rec.location_is_empty = True
-                continue
+        # No restriction should apply on customer/supplier/...
+        # locations and we don't need to compute is empty
+        # if there is no limit on the location
+        only_empty_locations = self.filtered(
+            lambda l: not l._should_compute_location_is_empty()
+        )
+        only_empty_locations.update({"location_is_empty": True})
```
sbidoul, line 324:
```suggestion
        only_empty_locations.filter(lambda l: not l.location_is_empty).update({"location_is_empty": True})
```
avoid write if not required (there was this comment before) ?

sbidoul (approved):
Looks good (code review)

stock_storage_type/models/stock_location.py
sbejaoui: > Question: does `OUT_MOVE_LINE_DOMAIN` apply to quants too?

no

sbejaoui:
cc/ @phschmidt

stock_storage_type/models/stock_location.py
```diff
@@ -141,6 +142,19 @@ class StockLocation(models.Model):
...
+            self._cr.execute(
+                """
+                CREATE INDEX stock_move_line_location_state_index
+                ON
+                    stock_move_line (location_id, state)
+                WHERE
+                    (state IS NULL OR state NOT IN ('cancel', 'done'))
+                """
+            )
+        return super().init()
```
sbidoul: @sbejaoui no need to return, and I think its better to call super().init() first.

stock_storage_type/models/stock_location.py
```diff
@@ -301,22 +315,38 @@ def _compute_location_will_contain_lot_ids(self):
...
-                    rec.location_is_empty = True
-                continue
+        # No restriction should apply on customer/supplier/...
+        # locations and we don't need to compute is empty
+        # if there is no limit on the location
+        only_empty_locations = self.filtered(
+            lambda l: not l._should_compute_location_is_empty()
+        )
+        only_empty_locations.update({"location_is_empty": True})
+        records = self - only_empty_locations
```
jbaudoux, line 325: I would have kept the previous loop and built the recordset inside the loop to stay O(n) instead of this O(n²) line

```
locations_to_compute_ids = []
for rec in self:
  if not rec._should_compute_location_is_empty():
    ...
  else:
    locations_to_compute_ids.append(rec.id)
records = self.browse(locations_to_compute_ids)
```

stock_storage_type/models/stock_location.py
rousseldenis: In fact, the impact is really not significant

stock_storage_type/models/stock_location.py
rousseldenis: Performances tested. No impact

rousseldenis (approved):

PR #789, opened 2023-12-03
[FIX] access release channel code with sudo

sbejaoui:
users may not have system group, they can't access to the code field


sbidoul (approved):

lmignon (approved):

sbidoul:
The test failure should have been fixed on main, so merging.

/ocabot merge patch

PR #793, opened 2023-12-04
[14.0][IMP] stock_available_to_promise_release: Own field for release policy

mt-software-de:
~Because the picking policy of the picking is already set from the sale.order, we should not use the sale.order's policy anymore. Otherwise we are not able to change the policy in a later state anymore.~

We need a separate field to define the release_policy otherwise we are not able to change it in a later state.


mt-software-de:
cc @jbaudoux @sebalix

jbaudoux (changes requested):
It is currently taking the policy from the SO because you can force the policy on the picking type.
The use case is have a SO in asap policy, configure ship as all at once, configure to create BO at release. With this setup, you ensure you don't ship anything to the customer before everything is packed.
If your customer agreement changes why don't you change the policy on the SO?

mt-software-de:
> It is currently taking the policy from the SO because you can force the policy on the picking type. The use case is have a SO in asap policy, configure ship as all at once, configure to create BO at release. With this setup, you ensure you don't ship anything to the customer before everything is packed. If your customer agreement changes why don't you change the policy on the SO?

Because the picking policy on the sale.order is readonly after confirmation.
And because the picking policy of the stock.picking is already set from the so with this piece of code. https://github.com/odoo/odoo/blob/62ca62ac8256fe73271077f165687724e30c860c/addons/sale_stock/models/sale_order.py#L559
I think it is totally fine to remove this otherwise you are not able to change the policy anymore.

jbaudoux:
when the stock.picking policy is set by https://github.com/OCA/wms/tree/14.0/stock_picking_type_shipping_policy, which is what we heavily use, you need to rely on the original sale shipping policy

mt-software-de:
> when the stock.picking policy is set by https://github.com/OCA/wms/tree/14.0/stock_picking_type_shipping_policy, which is what we heavily use, you need to rely on the original sale shipping policy

OK but this means WE are now Not able to Change the Policy. I think this is Not good. Could we Not Change the behavior for outgoing Transfers from a Sale Order to Not Use the Policy from the Type.

mt-software-de:
> when the stock.picking policy is set by https://github.com/OCA/wms/tree/14.0/stock_picking_type_shipping_policy, which is what we heavily use, you need to rely on the original sale shipping policy

I looked once again into to this, the policy is only set by stock_picking_type_shipping_policy if the stock.picking.type is configured with force_. Otherwise it stays like it is. So i think it is fine to use the policy of the picking, because if someone actively decides to set force_* on the a outgoing picking type its fine to use it. IMO

jbaudoux:
@mt-software-de If you want to change this, we need a solution that is not a breaking change.
The sale policy is propagated on the procurement group. I'm fine to drop this method looking at the sales order as we have the info on the procurement group. However, we cannot take it from the picking as this is a transfer policy which may differ for the reason explained previously.
May I suggest then to also modify https://github.com/OCA/wms/blob/14.0/stock_available_to_promise_release/models/stock_picking.py#L71 to have an explicit new field on the picking called release_policy and I would make it a related to the procurement group policy.

mt-software-de:
After a short meeting, with @jbaudoux where we explained me the use case of having a different policy for release as for the picking. If we want to release asap (when there is not enough stock), but keep after releasing the picking_policy on the new created transfer as "when all products are ready". 
Therefore we need two separate fields. 
EDIT:
~The solution would be to make the move_type of the procurement.group editable in the transfer view.~
Of course we can not use the move_type of the procurement.group this i will also change the shipping_policy of the new generated picking.

stock_available_to_promise_release/migrations/14.0.2.5.0/post-migrate.py
```diff
@@ -0,0 +1,30 @@
...
+                stock_move as m
+            ON
+                m.picking_id = p.id
+            JOIN
+                procurement_group as g
+            ON
+                m.group_id = g.id
+        ) as sub
+        WHERE
+            stock_picking.id = sub.id
```
jbaudoux:
```suggestion
            release_policy = procurement_group.move_type
        FROM
            procurement_group
        WHERE
            stock_picking.group_id = procurement_group.id
```

jbaudoux (approved):
LGTM, thanks

Can you simplify the migration query?

stock_available_to_promise_release/migrations/14.0.2.5.0/post-migrate.py
mt-software-de: oh yes indeed the group_id is stored. Thxs

mt-software-de:
> LGTM, thanks
> 
> Can you simplify the migration query?

done

sebalix (approved):

mt-software-de:
A rebase is done, its now mergeable

PR #796, opened 2023-12-06
[14.0] shopfloor_mobile_base: force nav item counter update

JuMiSanAr:
The navigation drawer counters were not being updated properly when going through the full workflow: they were only updated on opening / closing the drawer, when loading a new scenario or when refreshing the page.

With this change, we make sure to reach out to the backend with every operation to ensure we have up-to-date figures.

ref: cos-4245 & cos-4328

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/main.js
+++ b/shopfloor_mobile_base/static/wms/src/main.js
@@ -77,7 +77,6 @@ new Vue({
             global_state_key: "",
             loading: false,
             loading_msg_custom: "",
-            loadingMenu: false,
             appconfig: null,
         };
         _.merge(data, config_registry.generare_data_keys());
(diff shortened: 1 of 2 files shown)


shopfloor_mobile_base/static/wms/src/main.js
```diff
@@ -231,9 +235,20 @@ new Vue({
                     "SERVICE-CTX-PROFILE-ID": this.profile.id,
                 },
             });
-            return odoo.call("menu").then(function (result) {
-                self.appmenu = result.data;
-            });
+            if (!this.loadingMenu) {
```
simahawk: As discussed together: pls use a specific event to trigger this only when needed.

simahawk (approved):

TDu (approved):

PR #797, opened 2023-12-06
[FIX] stock_release_channel: Fix enqueued jobs tests / stock_release_channel_process_end_time: Freeze

rousseldenis:
- [x] As queue_job has changed enqueued_jobs property behavior (void the property after performing jobs), we need to adapt tests.

See https://github.com/OCA/queue/commit/832dd9b96bdc0bc7ddd064e1e86f9d400d99a97d

- [x] In stock_release_channel_process_end_time: Freeze the channel process end time too


rousseldenis:
@simahawk @sebalix This influences only tests, so, I will merge in order to get other PR's green.

PR #802, opened 2023-12-19
[14.0]  shopfloor: improve qty picker

JuMiSanAr:
- Write the qty and Package type text in bold for the type to pick
- Instead of color-coding the boxes (since this is usually used to indicate what type of information it is), introduce a small icon next to the package type to show what it is.

image

ref: cos-4220

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -153,6 +153,7 @@ class DataAction(Component):
             "id",
             ("packaging_type_id:name", lambda rec, fname: rec.packaging_type_id.name),
             ("packaging_type_id:code", lambda rec, fname: rec.packaging_type_id.code),
+            ("packaging_type_id:shopfloor_icon", self._packaging_icon_data),
             "qty",
         ]
 
@@ -315,6 +316,17 @@ class DataAction(Component):
             self._packaging_parser,
             multi=True,
         )
+
+    def _packaging_icon_data(self, rec, field):
+        icon_data = {"alt_text": rec.packaging_type_id.name}
+        if not rec.packaging_type_id.shopfloor_icon:
+            return icon_data
+        icon_data[
+            "url"
+        ] = "/web/image/product.packaging.type/{}/shopfloor_icon/30x30".format(
+            rec.packaging_type_id.id
+        )
+        return icon_data
 
     def _product_supplier_code(self, rec, field):
         supplier_info = fields.first(
--- a/shopfloor/models/product_packaging_type.py
+++ b/shopfloor/models/product_packaging_type.py
@@ -1,4 +1,5 @@
-# Part of Odoo. See LICENSE file for full copyright and licensing details.
+# Copyright 2023 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import fields, models
 
@@ -6,4 +7,8 @@ from odoo import fields, models
 class ProductPackagingType(models.Model):
     _inherit = "product.packaging.type"
 
-    icon = fields.Image()
+    shopfloor_icon = fields.Binary(
+        help="Icon to be displayed in the frontend qty picker. "
+        "Its final size will be a 30x30 image, "
+        "which means that square-shaped icons will work best."
+    )
--- a/shopfloor/models/product_packaging_type.py
+++ b/shopfloor/models/product_packaging_type.py
@@ -1,0 +1,9 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class ProductPackagingType(models.Model):
+    _inherit = "product.packaging.type"
+
+    icon = fields.Image()
--- a/shopfloor/views/product_packaging_type.xml
+++ b/shopfloor/views/product_packaging_type.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_product_packaging_type_form" model="ir.ui.view">
+        <field name="name">shopfloor.product.packaging.type.form</field>
+        <field name="model">product.packaging.type</field>
+        <field
+            name="inherit_id"
+            ref="product_packaging_type.view_product_packaging_type_form"
+        />
+        <field name="arch" type="xml">
+            <field name="name" position="after">
+                <div>
+                    <field name="icon" widget="image" width="90" />
+                </div>
+            </field>
+        </field>
+    </record>
+</odoo>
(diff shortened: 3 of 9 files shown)


shopfloor/models/product_packaging_type.py
```diff
@@ -0,0 +1,9 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
simahawk: bad header

shopfloor/models/product_packaging_type.py
```diff
@@ -0,0 +1,9 @@
...
+
+from odoo import fields, models
+
+
+class ProductPackagingType(models.Model):
+    _inherit = "product.packaging.type"
+
+    icon = fields.Binary()
```
simahawk:
```suggestion
    shopfloor_icon = fields.Binary(help="Give some hints on the image size to avoid having too heavy images here")
```

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -280,9 +280,16 @@ export var PackagingQtyPicker = Vue.component("packaging-qty-picker", {
...
                         <span class="qty-todo">/ {{ qty_todo_by_pkg[pkg.id] }}</span>
                     </v-col>
-                    <v-col>
-                        <div class="pkg-name"> {{ pkg[pkgNameKey] }}</div>
-                        <div v-if="contained_packaging[pkg.id]" class="pkg-qty">(x{{ contained_packaging[pkg.id].qty }} {{ contained_packaging[pkg.id].pkg.name }})</div>
+                    <v-col cols="4">
+                        <div :class="qty_by_pkg[pkg.id] > 0 ? 'pkg-name font-weight-bold' : 'pkg-name'"> {{ pkg[pkgNameKey] }}</div>
+                        <div v-if="contained_packaging[pkg.id]" :class="qty_by_pkg[pkg.id] > 0 ? 'pkg-qty font-weight-bold' : 'pkg-qty'">(x{{ contained_packaging[pkg.id].qty }} {{ contained_packaging[pkg.id].pkg.name }})</div>
+                    </v-col>
+                    <v-col cols="2" v-if="_.result(contained_packaging[pkg.id], 'pkg.icon.url', '')">
```
simahawk: IMO the icon should stay before the input... but that's probably only a matter of taste

shopfloor/actions/data.py
```diff
@@ -316,6 +317,15 @@ def _product_packaging(self, rec, field):
             multi=True,
         )
 
+    def _packaging_icon_data(self, rec, field):
+        icon_data = {"alt_text": "Icon for packaging " + rec.packaging_type_id.name}
```
simahawk:
```suggestion
        icon_data = {"alt_text": rec.packaging_type_id.name}
```

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
simahawk: You should append `/30x30` to the URL to get a resized image.

simahawk (changes requested):

shopfloor/views/product_packaging_type.xml
```diff
@@ -0,0 +1,18 @@
...
+    <record id="view_product_packaging_type_form" model="ir.ui.view">
+        <field name="name">shopfloor.product.packaging.type.form</field>
+        <field name="model">product.packaging.type</field>
+        <field
+            name="inherit_id"
+            ref="product_packaging_type.view_product_packaging_type_form"
+        />
+        <field name="arch" type="xml">
+            <field name="name" position="after">
+                <div>
```
TDu, line 12: Is the enclosing `<div>` needed ?

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
JuMiSanAr: I did but it doesn't change anything in the frontend, the image adjusts to the size of the row. This is why I'm setting its size in a CSS class.

Regarding the icon, I put it to the right so that it's closer to the name (I think it's more readable like this), but it's only a preference too.

shopfloor/views/product_packaging_type.xml
JuMiSanAr: This is to move this field to the next line (if I don't add it, the image appears on the same row as the name of the packaging in the form).

shopfloor/actions/data.py
```diff
@@ -316,6 +317,15 @@ def _product_packaging(self, rec, field):
             multi=True,
         )
 
+    def _packaging_icon_data(self, rec, field):
+        icon_data = {"alt_text": rec.packaging_type_id.name}
+        if not rec.packaging_type_id.shopfloor_icon:
+            return icon_data
+        icon_data["url"] = "/web/image/product.packaging.type/{}/shopfloor_icon/".format(
```
simahawk: I still don't see the size here. If you don't use it, you have chances to load a bigger image.
What you do on the front is ok (forcing the width) but here we should try to load the smallest image possible.

JuMiSanAr: Makes sense, I've updated it :+1:

simahawk (approved):

jbaudoux:
We should align with what @lmignon did on v16

TDu:
> We should align with what @lmignon did on v16

Do you have a link to a pr ?

lmignon (approved):

lmignon:
In V60 we've a more conservative approach (IMO better) We highlight the expected unit into the picker when it's not a ''unit"

Here is the code of our extension addon

```js
import {PackagingQtyPicker} from "/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js";

const computed = PackagingQtyPicker.extendOptions.computed;
const qty_color_computed = computed.qty_color;

computed.qty_color = function () {
  const result = qty_color_computed.bind(this)();
  const font_style = "font-weight:950; font-size: 200%; color:rgb(102, 0, 0)";
  const new_result = result + ";" + font_style;
  return new_result;
};

computed.unit_color = function () {
  const font_style = "font-weight:950; font-size: 200%; color:rgb(102, 0, 0)";
  const is_not_unit = this.uom.id != 1; // Id of unit of measure "unit" is 1.
  const background_style = is_not_unit
    ? "background-color: orangered!important"
    : "background-color:transparent; ";

  const result = ";" + font_style + ";" + background_style;
  return result;
};

const template = PackagingQtyPicker.extendOptions.template;

/* Replace the multiline
    <v-col>
        {{ unit_uom.name }}
    </v-col>
  with
   <v-col :style="unit_color">
      {{ unit_uom.name }}
  </v-col>
  */
const new_template = template.replace(
  /<v-col>\s*{{\s*unit_uom.name\s*}}\s*<\/v-col>/,
  '<v-col :style="unit_color">{{ unit_uom.name }}</v-col>'
);

PackagingQtyPicker.extendOptions.template = new_template;

```

IMO it takes less place on the screen and avoid the display of too mush information.

lmignon (changes requested):
approval too quickly. What do you think about the approach suggested in the comment below?

PR #803, opened 2023-12-19
[14.0][FIX] stock_available_to_promise_release: Add logical dependency

TDu:
Having `stock_picking_restrict_cancel_with_orig_move` as a dependency is not necessary but it makes sense because of its utility. If not we would need to created a glue module to fix the dependency order on relation to the `_action_cancel` of stock.move.


stock_available_to_promise_release/__manifest__.py
```diff
@@ -8,7 +8,15 @@
...
     "category": "Stock Management",
-    "depends": ["stock"],
+    "depends": [
+        "stock",
+        # OCA/stock-logistics-workflow
+        # This module is not needed per se. But when it is installed it will conflict
+        # on the `_action_cancel` of the stock.move model, if executed first.
+        # As the module prevent breaking a chain of moves, It is a logical dependency
+        # to have.
+        "stock_picking_restrict_cancel_with_orig_move",
```
sebalix, line 18: I agree, what do you think @jbaudoux ?

jbaudoux (changes requested):
The issue is that cancelation is not properly handled at the moment. This must be tackled first.
You can't add that module as a dependency as it does not provide a valid solution neither. It is to restrictive and it does not support moves merging & splitting.

The only proper way at the moment is to unrelease the outgoing moves (that you backported from v16)
action_cancel on outgoing moves should be catched to call unrelease.

stock_available_to_promise_release/__manifest__.py
jbaudoux: I don't ! :)

PR #804, opened 2023-12-20
[IMP] improve logging

sbidoul:
- Better use of logger.warn, so tools like sentry recognize these messages as the same one.
- Translate message.
- Also, resolve a couple of deprecations.


sbidoul:
@lmignon @rousseldenis

lmignon (approved):

rvalyi (approved):

sbidoul:
No functional changes and the part I touched has test coverage.

PR #806, opened 2023-12-28
[14.0] shopfloor_mobile: display vendor code in single_pack_transfer card

JuMiSanAr:
Screenshot from 2023-12-28 14-12-05


sebalix:
@JuMiSanAr Is it possible to indent a bit to the right the vendor code? It could be difficult for the user to read all these info if everything is at the same level. What do you think?

sebalix:
Updated to indent a bit the vendor code, thanks @JuMiSanAr 
image

sebalix (approved):

PR #808, opened 2024-01-05
[16.0][PORT] stock_available_to_promise_release : fw-port changes from 14.0

jbaudoux:
cc @sbejaoui @lmignon @sebalix @mt-software-de

PR #815, opened 2024-01-17
[16.0][FIX] stock_available_to_promise_release: Refactor release_ready compute

jbaudoux:
Fix missing migration of last commit in https://github.com/OCA/wms/pull/808

@RodrigoBM @lmignon @sbejaoui @mt-software-de

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -445,7 +445,7 @@ class StockMove(models.Model):
         procurement_requests = []
         released_moves = self.env["stock.move"]
         # Ensure the release_ready field is correctly computed
-        self.invalidate_cache(["release_ready"])
+        self.invalidate_recordset(["release_ready"])
         for move in self:
             if not move.release_ready:
                 continue
--- a/stock_available_to_promise_release/models/stock_picking.py
+++ b/stock_available_to_promise_release/models/stock_picking.py
@@ -69,21 +69,17 @@ class StockPicking(models.Model):
 
     @api.depends(
         "move_type",
-        "move_lines.ordered_available_to_promise_qty",
-        "move_lines.need_release",
-        "move_lines.state",
+        "move_ids.ordered_available_to_promise_qty",
+        "move_ids.need_release",
+        "move_ids.state",
     )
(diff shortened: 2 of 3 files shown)


RodrigoBM (approved):

RodrigoBM:
Why does the test give error if it is corrected?

mt-software-de:
> Why does the test give error if it is corrected?

Good question. And why it was not raised here https://github.com/OCA/wms/pull/808

RodrigoBM:
> > Why does the test give error if it is corrected?
> 
> Good question. And why it was not raised here #808

I think the CI is running v16.0 code and not PR code. is it possible @sbidoul?

sbidoul:
It's not normal indeed that stock_available_to_promise_release is installed in that preparation step. I'll have a look.

sbidoul:
I see. It's a case of circular dependencies between repos.

Something here in OCA/wms depends on `stock_move_auto_assign_auto_release` (in OCA/stock-logistics-warehouse), which itself depends on `stock_available_to_promise_release` here in OCA/wms. This causes odoo-addon-stock_available_to_promise_release to be installed from PyPI in the preparation step. This would normally not be an issue, but since the released version does not even install we are stuck.

Strictly speaking it's a bug in OCA/oca-ci. I'll think about a solution.

sbidoul:
> It's not normal indeed that stock_available_to_promise_release is installed in that preparation step.

So to be precise, it is normal that it is installed in the preparation step, but we need to install the version of this repo.

jbaudoux:
@sbidoul Thanks for the analysis. Can we already merge this PR? The tests are ok in local.

sbidoul:
Merged manually to unblock. I'll need some time to test a proper fix for the CI.

RodrigoBM:
> Merged manually to unblock. I'll need some time to test a proper fix for the CI.

But @sbidoul  if this is already merged why is it still giving error in [my PR](https://github.com/OCA/stock-logistics-warehouse/pull/1903). I think there is something that I don't understand about CI!

Thank you very much for your time!!

sbidoul:
When we merge manually, we need to wait for the nightly build.

PR #818, opened 2024-01-18
[14.0][IMP] stock_available_to_promise_release: Multi warehouse capability

mt-software-de:
Since now the stock_available_to_promise_release doesn't care if a move belongs to the one warehouse or to an other. 
It always computes the previous_promised_qty and ordered_available_to_promise_qty without considering the warehouse of a move.


mt-software-de:
@jbaudoux and @simahawk -> Let's start a discussion? :-)

This PR will fail because of the test test_picking_priority (tests/test_reservation.py)
Because here a picking is created for the default output location which doesn't belong to a warehouse,
with my change we can not longer release moves which do not belong to a warehouse. 

I don't think this is quite good.
Since we can configure routes and rules to replenish a need (without warehouse) from a location (with warehouse) it should be still possible to do this.

Currently i add stock_helper as new dependency, to get the closest warehouse.
But in a normal case the warehouse_id on a move should be set i we doesn't need this logic.

Was there a reason why you didn't added the possibility two release with two independent warehouses?

jbaudoux:
Yes we need to change to support multi-warehouses and multi-companies. I'll review your proposal with @sebalix

sebalix (commented):
We need to be backwards compatible here:
- [current] ability to consider other warehouses when computing the available qty to promise
- [default] consider only the current warehouse of the move to compute its available qty to promise

I was thinking of a new configuration field on the route (alongside the release checkbox option) where we could list extra warehouses to consider for moves belonging to this route.

Use case: having a warehouse B used to replenish warehouse A, and when we compute the available qty to promise we want to consider both. So on the delivery route of the warehouse A, we could add B as extra warehouse to consider.

What do you think?

mt-software-de:
> We need to be backwards compatible here:
> 
>     * [current] ability to consider other warehouses when computing the available qty to promise
> 
>     * [default] consider only the current warehouse of the move to compute its available qty to promise
> 
> 
> I was thinking of a new configuration field on the route (alongside the release checkbox option) where we could list extra warehouses to consider for moves belonging to this route.
> 
> Use case: having a warehouse B used to replenish warehouse A, and when we compute the available qty to promise we want to consider both. So on the delivery route of the warehouse A, we could add B as extra warehouse to consider.
> 
> What do you think?

Thx, i think this would be the best solution. We wouldn't need any new dependency and it is clear which warehouse to use.
Also if move without a warehouse. I will change this.

sebalix:
@mt-software-de We checked the use-case in details with @jbaudoux about what I said above, no need to work on this feature, it doesn't work like I said (in details we put the location of B under A/Stock, so we do not need to develop the feature I explained above).
So for now, we can simply restrict the computation based on the warehouse.

mt-software-de:
> @mt-software-de We checked the use-case in details with @jbaudoux about what I said above, no need to work on this feature, it doesn't work like I said (in details we put the location of B under A/Stock, so we do not need to develop the feature I explained above). So for now, we can simply restrict the computation based on the warehouse.

@jbaudoux and @sebalix Conversely, this means that you cannot release any move if no warehouse is set.

mt-software-de:
@jbaudoux  and @sebalix this PR is now ready for review.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -215,33 +215,50 @@ def _previous_promised_qty_sql_lateral_where(self):
...
+            warehouse
+        )
         params.update(lateral_params)
         query = self._previous_promised_qty_sql_main_query().format(
             lateral_where=lateral_where
         )
         return query, params
 
+    def _group_by_warehouse(self):
+        return groupby(self, lambda m: m.warehouse_id)
```
sebalix, line 235:
```suggestion
        return groupby(self, lambda m: m.warehouse_id or m.location_id._get_closest_warehouse())
```
Why not using this then?

jbaudoux: Seems legit as the warehouse is not required on the move

sebalix: We could still have moves MTS without warehouse set on them, so without this fallback value we won't include such moves in the computation of the previous promised qty.

mt-software-de: they are included if move has a location which is a child of warehouse.view_location_id

sebalix: My bad, it's the other way around:
If you want to compute the previous promised qty on a MTS move without warehouse, the current implementation will return 0, right?
That probably why you said we cannot release moves without warehouse with this change? (if so, that's problematic)

mt-software-de: Yes i will return 0, but there shouldn't be a mts without warehouse. The warehouse gets set by the stock.rule run https://github.com/odoo/odoo/blob/6b9b598fbb29d0bd22172dd346f9b36fae43e0bd/addons/stock/models/stock_rule.py#L321

jbaudoux: If we go this way, we need to explain it in the README

jbaudoux:
Note: merge as major as there is a change that could require rule reconfiguration

stock_available_to_promise_release/models/stock_move.py
sebalix: Indeed, as `warehouse_id` is not a required field on the rule. We do have some MTS moves without warehouse for a customer, v14 as well (not  a lot but it happens).

mt-software-de: I can in fact re-add the closest warehouse. But it will not catch everything.
If there would be a rule without warehouse and the location is not located under a warehouse i will return 0. 
For v14 we would need stock_helper as dependency, for Version >= 16 we don't need it anymore because of https://github.com/odoo/odoo/pull/118128

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -322,63 +386,8 @@ def _compute_ordered_available_to_promise(self):
...
-                move.ordered_available_to_promise_uom_qty = max(
-                    min(uom_promised, move.product_uom_qty),
-                    0.0,
-                )
-                move.ordered_available_to_promise_qty = max(
-                    min(real_promised, move.product_qty),
-                    0.0,
-                )
+        for move, values in moves._get_ordered_available_to_promise().items():
+            move.write(values)
```
sebalix: We should not use `write` in compute method but `update` I think.
Compute methods could process draft records not yet saved in DB (like when encoding a new record from an Odoo form, compute acts like an onchange).

stock_available_to_promise_release/models/stock_move.py
sebalix: yes sure, but it is unfortunate to get a move to release with a location outside of any warehouse, it's not something we should support I think

stock_available_to_promise_release/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Stock Available to Promise Release",
-    "version": "14.0.2.4.0",
+    "version": "14.0.2.5.0",
```
sebalix: It's the job of the OCA bot to increase this version number, unless you have a migration script to play for what you ship

mt-software-de: Shouldn't this be considered as minor change? Also if there is no migration?

jbaudoux: From function point of view and as existing configuration can be impacted, it is for me a major change

stock_available_to_promise_release/readme/DESCRIPTION.rst
```diff
@@ -32,3 +32,8 @@ the remaining qty is split and attached to a new picking that can be released la
 
 Important: if the "Stock reservation horizon" is set, the qty is calculated
 on moves which have an expected date not beyond $today + $horizon (in days).
+
+Important: Moves without a warhouse set are not supported.
```
jbaudoux:
```suggestion
Important: Moves without a warehouse set are not supported.
```

stock_available_to_promise_release/readme/DESCRIPTION.rst
```diff
@@ -32,3 +32,8 @@ the remaining qty is split and attached to a new picking that can be released la
...
+If the warehouse is not set on the move, the "Ordered Available to Promise" Qty
```
jbaudoux:
```suggestion
If the warehouse is not set on the move, the "Ordered Available to Promise" quantity 
```

stock_available_to_promise_release/readme/DESCRIPTION.rst
```diff
@@ -32,3 +32,8 @@ the remaining qty is split and attached to a new picking that can be released la
 
 Important: if the "Stock reservation horizon" is set, the qty is calculated
 on moves which have an expected date not beyond $today + $horizon (in days).
+
+Important: Moves without a warhouse set are not supported.
+If the warehouse is not set on the move, the "Ordered Available to Promise" Qty 
+will always be 0. The warehouse gets set by the Stock rule, if the the Rule
+is not assigned to a warehouse, also the move will not end up with a warehouse.
```
jbaudoux:
```suggestion
will always be 0. It will also not be considered as consuming existing stock.
The warehouse gets set by the Stock rule, if the the Rule
is not assigned to a warehouse, also the move will not end up with a warehouse.
```

stock_available_to_promise_release/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Stock Available to Promise Release",
-    "version": "14.0.2.5.0",
+    "version": "14.0.2.6.0",
```
jbaudoux: don't set it yourself, this will be changed by the runbot

stock_available_to_promise_release/models/stock_move.py
sebalix: @mt-software-de @jbaudoux so what do we do about this? The fallback on closest warehouse hasn't been added.
For 16.0 I confirm that I could not found MTS moves without warehouse in one of our DB, so the current code works as expected for this version.

sebalix (approved):

stock_available_to_promise_release/models/stock_move.py
mt-software-de: For me it is fine, to merge without the fallback. On our project is isn't a problem.
In the most cases i think there is no MTS move which needs a release without a warehouse.

jbaudoux (approved):

PR #820, opened 2024-01-22
[16.0][FIX] stock_dynamic_routing : avoid set state cancel to original picking with canceled_by_routing

alv97:
When new picking is cancelled, original picking with canceled_by_routing = True stay all time to cancel, we add new check to see if have movements to mark that flag to false


alv97:
Waiting to Merge fix test on https://github.com/OCA/wms/pull/828

PR #821, opened 2024-01-23
[14.0] stock_available_to_promise_release: backport some commits from 16.0

sebalix:
An attempt to reduce the diff between 14.0 and 16.0 but it's quite difficult as the commit history between 14.0 and 16.0 starts to be really different.


jbaudoux (approved):
LGTM

PR #824, opened 2024-01-23
[14.0][FIX] stock_available_to_promise_release: Imp on chained moves cancel

TDu:
On move lines cancelation, the unrelease of the related transfers is being done. And it works well.

But when the module `stock_picking_restrict_cancel_with_orig_move` from stock-logistics-workflow is installed, they can conflict on the `_action_cancel` function from stock.move.line.

If the implementation of that function is first being called on this other module, an error will be raised before the unrelease of the moves can be done.

Both modules only depend on `stock`. This typically happens with chained moves.

The solution to also call the release of operation on the `action_cancel` of the stock picking, fixes the problem. Altough it is a litlle redundant with the implementation on stock move.

### DIFF
--- a/stock_available_to_promise_release/models/stock_picking.py
+++ b/stock_available_to_promise_release/models/stock_picking.py
@@ -190,3 +190,7 @@ class StockPicking(models.Model):
         processing has already started will be ignored
         """
         self.mapped("move_lines").unrelease(safe_unrelease=safe_unrelease)
+
+    def action_cancel(self):
+        self.filtered(lambda p: p.state not in ("done", "cancel")).unrelease()
+        return super().action_cancel()


jbaudoux:
Isn't it better to just mark `stock_picking_restrict_cancel_with_orig_move` as deprecated ?

TDu:
> Isn't it better to just mark `stock_picking_restrict_cancel_with_orig_move` as deprecated ?

I don't know I would think that module has its use as well.

mt-software-de (approved):
LGTM

jbaudoux (changes requested):
`action_cancel` on stock picking is calling `_action_cancel` on stock moves which is already doing the unrelease. So this looks useless.
The issue may be on the test in stock move to see if it can be unreleased. There have been improvement in v16 for this.

TDu:
> `action_cancel` on stock picking is calling `_action_cancel` on stock moves which is already doing the unrelease. So this looks useless. The issue may be on the test in stock move to see if it can be unreleased. There have been improvement in v16 for this.

I does indeed. But the explanation of why is in the commit message...

PR #828, opened 2024-01-24
[16.0] Fix build regarding `stock_available_to_promise_release` and `stock_release_channel_batch_mode_commercial_partner`

sebalix:
- Invalidate the cache of `stock.picking.release_ready` field when we put some qty in stock to recompute it.
- Fix test of `stock_release_channel_batch_mode_commercial_partner`

### DIFF
--- a/stock_release_channel_batch_mode_commercial_partner/tests/test_channel_release_batch.py
+++ b/stock_release_channel_batch_mode_commercial_partner/tests/test_channel_release_batch.py
@@ -31,6 +31,7 @@ class TestChannelReleaseBatch(ChannelReleaseCase):
 
     def test_release_auto_group_commercial_partner_no_next_batch(self):
         self.channel.batch_mode = "group_commercial_partner"
-        self.pickings.need_release = False  # cheat for getting the right condition
+        # cheat for getting the right condition
+        self.pickings.move_ids.need_release = False
         action = self.channel.release_next_batch()
         self._assert_action_nothing_in_the_queue(action)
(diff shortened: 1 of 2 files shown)


jbaudoux (approved):

stock_release_channel_batch_mode_commercial_partner/tests/test_channel_release_batch.py
```diff
@@ -31,6 +31,7 @@ def test_release_auto_group_commercial_partner(self):
 
     def test_release_auto_group_commercial_partner_no_next_batch(self):
         self.channel.batch_mode = "group_commercial_partner"
-        self.pickings.need_release = False  # cheat for getting the right condition
+        # cheat for getting the right condition
+        self.pickings.move_ids.need_release = False
```
jbaudoux, line 35: This will ensure there is nothing to release in the next call without cheating...
```suggestion
        action = self.channel.release_next_batch()

```

jbaudoux (changes requested):

stock_release_channel_batch_mode_commercial_partner/tests/test_channel_release_batch.py
sebalix: Sadly the pickings in the channel doesn't share the same commercial partner, we need to call `release_next_batch` two times to empty the channel.
So better to keep the cheat in place, this will work every time for what we want to test here.

jbaudoux (approved):

stock_release_channel_batch_mode_commercial_partner/tests/test_channel_release_batch.py
jbaudoux: ok then

PR #829, opened 2024-01-26
[14.0][IMP] shopfloor_delivery_shipment: add sales grouping

TDu:


shopfloor_delivery_shipment/services/delivery_shipment.py
```diff
@@ -593,46 +582,87 @@ def _data_for_content_to_load_from_pickings(self, shipment_advice):
             pickings_partially_loaded = loaded_lines.picking_id
             domain += [("picking_id", "in", pickings_partially_loaded.ids)]
         move_lines = self.env["stock.move.line"].search(domain)
-        return self._prepare_data_for_content(move_lines)
+        return self._prepare_data_for_content(move_lines, False)
```
mmequignon: I would make this arg named, so we can understand what this False means.
Also, I would remove the plural in the name.
```suggestion
        return self._prepare_data_for_content(move_lines, group_by_sale=False)
```
I don't know what default value is the most relevant here. Up to you.

shopfloor_delivery_shipment/services/delivery_shipment.py
```diff
@@ -593,46 +582,87 @@ def _data_for_content_to_load_from_pickings(self, shipment_advice):
...
-        if picking:
-            move_lines = self._find_move_lines_to_process_from_picking(
-                shipment_advice, picking
-            )
-        elif location:
-            move_lines = self._find_move_lines_from_location(shipment_advice, location)
-        return self._prepare_data_for_content(move_lines)
-
-    def _prepare_data_for_content(self, move_lines):
+        empty_group_name = " "
```
mmequignon:
```python
>>> json.dumps({"package_level_1": list(range(10)), False: list(range(10))})
'{"package_level_1": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], "false": [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]}'
```
Can't you use False as the empty key? I don't like `" "`

shopfloor_delivery_shipment/services/delivery_shipment.py
```diff
@@ -593,46 +582,87 @@ def _data_for_content_to_load_from_pickings(self, shipment_advice):
...
+        if picking:
+            move_lines = self._find_move_lines_to_process_from_picking(
+                shipment_advice, picking
+            )
+        elif location:
+            move_lines = self._find_move_lines_from_location(shipment_advice, location)
+        group_by_sales = bool(location)
+        return self._prepare_data_for_content(move_lines, group_by_sales)
+
+    def _prepare_data_for_content(self, move_lines, group_by_sales):
```
simahawk:
```suggestion
    def _prepare_data_for_content(self, move_lines, group_by_sale=True):
```

shopfloor_delivery_shipment/services/delivery_shipment.py
simahawk: I agree. Even `null`.

simahawk: BTW what's the main reason to do this server side?
In theory this is simply a fronted requirement and we already have grouping functions client side (eg: group by location). See in `wms_utils`
Unless there's a real benefit I would move everything frontend side.

simahawk (changes requested):

shopfloor_delivery_shipment/services/delivery_shipment.py
mmequignon: well, as i understand, the sorting has to consider already loaded shipments.
Is this information already available in the frontend?
Maybe it's easier to handle this from the backend.

shopfloor_delivery_shipment/services/delivery_shipment.py
TDu: done

shopfloor_delivery_shipment/services/delivery_shipment.py
TDu: I agree that " " group name was not best.
About using False or None. Having different value type has key will fail when using `json.dumps` and the `sort_key` argument like in https://github.com/OCA/rest-framework/blob/e8ce61565003639f31623354b952e5aee2df02f3/rest_log/components/service.py#L26

Anyway a None value will be transformed to a "null" string when the json is being dumped. So I set the group name to "null".

TDu:
About doing grouping and sorting server side vs client side.
To me the less client side code we have the better.
The data is being parsed in the backend anyway so if it needs some default grouping sorting why not doing it then ?

For this special case, we could add a field in the package level data to store the sales information. But then half of the problem would be fixed because we want to also have a sorting as well on which sales order has been started to be processed or not.

mmequignon:
> the less client side code we have the better

Agreed.

shopfloor_delivery_shipment/services/delivery_shipment.py
```diff
@@ -593,46 +582,87 @@ def _data_for_content_to_load_from_pickings(self, shipment_advice):
...
+        sales_started = []
+        if group_by_sale:
+            sales_started = move_lines.shipment_advice_id.loaded_picking_ids.sale_id
+            # Grouping by sales (the packages) sort starting from last position
+            #  2 the lines NOT in a sales order whose loading has started.
+            #  1 the lines in a sales being loaded and the line is done.
+            #  0 the lines in a sales being loaded and not yet done
+            move_lines = move_lines.sorted(
+                lambda ml: 2
+                if ml.move_id.sale_line_id.order_id not in sales_started
```
sebalix, line 642: We need to add `sale_stock` as dependency, Shopfloor doesn't depend on it, better to limit this change to `shopfloor_delivery_shipment`.

sebalix: Seems that `shopfloor_reception` has also tests depending on `sale` (while the module doesn't really depend on this module to work), this would need some tests rewrite. Hopefully some modules in OCA/wms are installing `sale` so tests are running well, but isolated they should crash.

simahawk:
> To me the less client side code we have the better.

Very bold statement :smile: 
And what is the rationale for this? Personal taste only? :)

The point is that this is not consistent with what has been done in the past.
This is really something only related to the way you display info on the screen.
Not on how data has to be handled server side.

Now, I understand you can be annoyed if someone ask you to refactor what you have done server side / client side.... yet, from my POV is not a good idea to break this design decision. Especially w/o proper motivation if not "I like it more server side".  Not to mention that this is adding more not needed complexity server side, whereas the only change that would be needed is to add the sale order ref.

TDu:
> > To me the less client side code we have the better.
> 
> Very bold statement 😄 And what is the rationale for this? Personal taste only? :)
> 
> The point is that this is not consistent with what has been done in the past. This is really something only related to the way you display info on the screen. Not on how data has to be handled server side.
> 
> Now, I understand you can be annoyed if someone ask you to refactor what you have done server side / client side.... yet, from my POV is not a good idea to break this design decision. Especially w/o proper motivation if not "I like it more server side". Not to mention that this is adding more not needed complexity server side, whereas the only change that would be needed is to add the sale order ref.

I am not sure what in my answer made you think that I was annoyed by your comment ? I don't see any feeling being shown in there.

Now about my comment

>  the less front end code the better.

 My thoughts about this is (in relation to this specific project) that the backend code is well tested compared to the front end and defacto much easier to maintain in the long run.
Also your mention of not breaking some design decision without clearly stating what those decisions are is not very helpful.

Anyway for this specific case adding a key (sales.order) within the package levels would not completely resolve the issue because the intend as specified in my previous comment

> For this special case, we could add a field in the package level data to store the sales information. But then half of the problem would be fixed because we want to also have a sorting as well on which sales order has been started to be processed or not.

The part not solved, would be sorting the sales order that have already been started before the one the user has not yet started to load.
Of course it could be done with some added key in the sales order, but I am afraid that would beak another design rule.

shopfloor_delivery_shipment/services/delivery_shipment.py
TDu: The module `shipment_advice` depends on `delivery` which depends on `sale_stock`

simahawk:
> I am not sure what in my answer made you think that I was annoyed by your comment ? I don't see any feeling being shown in there.

Why annoyed? "bold" does not mean that you are annoyed... I was just fooling you because you were not bringing any strong technical motivation to back your statement. No feelings involved :wink: :hugs: 
However, for my personal improvement, let me know if this way of saying might imply other meanings :pray: 

> Now about my comment
> 
> > the less front end code the better.
> 
> My thoughts about this is (in relation to this specific project) that the backend code is well tested compared to the front end and defacto much easier to maintain in the long run. Also your mention of not breaking some design decision without clearly stating what those decisions are is not very helpful.

Well, I'm not sure it is written somewhere (I can check that) but I can tell you that when we 1st designed this stack both me and Guewen agreed on leaving such tasks to the front (presenting things vs gathering data).
And you can see this everywhere in the code (as far as possible).
Regarding test coverage: that's not a good reason. If that's a problem, we have to address that, instead of working around it.

> Anyway for this specific case adding a key (sales.order) within the package levels would not completely resolve the issue because the intend as specified in my previous comment
> 
> > For this special case, we could add a field in the package level data to store the sales information. But then half of the problem would be fixed because we want to also have a sorting as well on which sales order has been started to be processed or not.
> 
> The part not solved, would be sorting the sales order that have already been started before the one the user has not yet started to load. Of course it could be done with some added key in the sales order, but I am afraid that would beak another design rule.

You can sort AND group client side. Why do you think this is not feasible/desirable?

PR #831, opened 2024-01-26
[IMP] stock_storage_type: Don't consider quant with quantities <= 0

rousseldenis:
As Odoo does not void directly a quant after a full quant transfer (quant remains with quantity == 0), consider that quantity when computing the fields to avoid mixed products or mixed lots


rousseldenis:
@jbaudoux

jbaudoux (approved):

PR #833, opened 2024-02-01
[FIX] stock_release_channel: Don't assign first a release channel if it will be filtered

rousseldenis:
The release channel should be assigned at the end of all evaluation rules including the last filtering function.


sbejaoui (approved):

rousseldenis:
@jbaudoux

jbaudoux (approved):

PR #834, opened 2024-02-08
[16.0][IMP] stock_available_to_promise_release: Unrelease when merging moves

lmignon:
When moves are merged into an outgoing picking and the product_uom_qty is updated it's important to unrelease the move since the release process has been done for the previous qty. This case occurs for example when the qty is modified for a line on a confirmed SO.

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -765,3 +765,56 @@ class StockMove(models.Model):
         values = super()._get_new_picking_values()
         values["release_policy"] = values["move_type"]
         return values
+
+    def write(self, vals):
+        released_moves = self.browse()
+        if self.env.context.get("in_merge_mode") and "product_uom_qty" in vals:
+            # when a move is merged, we need to unrelease it if the quantity
+            # is changed and the move is unreleasable
+            released_moves = self.filtered(lambda m: m._is_unreleaseable())
+            # a change on the product_uom_qty on a released move with quantity
+            # partially done should not be possible. The 'safe_unrelease' flag
+            # is set to False to ensure this case is checked. Nevertheless,
+            # we should never reach this point as the merge candidates are
+            # filtered out in the method _update_candidate_moves_list to never
+            # merge releaseable moves with partially done quantity.
+            released_moves.unrelease(safe_unrelease=False)
+        ret = super().write(vals)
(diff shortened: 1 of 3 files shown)


lmignon:
cc @rousseldenis @jbaudoux

flotho (approved):
Code review
Principle understood
LGTM

rousseldenis (approved):
Code review

TDu (approved):

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -704,3 +704,29 @@ def _get_new_picking_values(self):
...
+            new_qty = vals["product_uom_qty"]
+            to_unrelease = self.filtered(
+                lambda m: float_compare(
+                    m.product_uom_qty,
+                    new_qty,
+                    precision_rounding=m.product_uom.rounding,
+                )
+                != 0
+            )
+            to_unrelease.unrelease()
```
jbaudoux: You should not unrelease if it is not allowed (like started)
```suggestion
            to_unrelease.unrelease(safe_unrelease=True)
```

jbaudoux (changes requested):

rousseldenis:
@jbaudoux

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -704,3 +704,29 @@ def _get_new_picking_values(self):
...
+            new_qty = vals["product_uom_qty"]
+            to_unrelease = self.filtered(
+                lambda m: float_compare(
+                    m.product_uom_qty,
+                    new_qty,
+                    precision_rounding=m.product_uom.rounding,
+                )
+                != 0
+            )
+            to_unrelease.unrelease(safe_unrelease=True)
```
jbaudoux: You need to check move `_is_unreleaseable()` first.
If it cannot be unreleased, we must raise an error. For this, you need `safe_unrelease=False`
If you unrelease, I would release again after the merge

jbaudoux (changes requested):

jbaudoux:
ping @lmignon

lmignon:
> ping @lmignon

I'm on it.

lmignon:
@jbaudoux Your remarks are taken into account now.

jbaudoux (approved):

PR #835, opened 2024-02-08
[14.0][IMP] shopfloor: add option to exclude package types in product info

JuMiSanAr:
Depends on:
- https://github.com/OCA/product-attribute/pull/1523

Qty picker:

Settings:

ref: cos-4389

### DIFF
--- a/shopfloor_vendor_packaging/__init__.py
+++ b/shopfloor_vendor_packaging/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import actions
--- a/shopfloor_vendor_packaging/__manifest__.py
+++ b/shopfloor_vendor_packaging/__manifest__.py
@@ -1,0 +1,19 @@
+# Copyright 2024 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Shopfloor Vendor Packaging",
+    "summary": "Manage shopfloor behavior for vendor packaging",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Inventory",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "data": [
+        "views/shopfloor_menu.xml",
+    ],
+    "depends": ["shopfloor", "product_packaging_type_vendor"],
+    "auto-install": True,
+}
--- a/shopfloor_vendor_packaging/models/shopfloor_menu.py
+++ b/shopfloor_vendor_packaging/models/shopfloor_menu.py
@@ -1,0 +1,29 @@
+# Copyright 2024 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+DISPLAY_VENDOR_PACKAGING_HELP = """
+    When enabled, the user will be able to see the vendor packagings
+    in the frontend, e.g. in the qty picker.
+"""
+
+
+class ShopfloorMenu(models.Model):
+    _inherit = "shopfloor.menu"
+
+    display_vendor_packaging = fields.Boolean(
+        string="Display vendor packagings",
+        default=False,
+        help=DISPLAY_VENDOR_PACKAGING_HELP,
+    )
+    display_vendor_packaging_is_possible = fields.Boolean(
+        compute="_compute_display_vendor_packaging_is_possible",
+    )
+
+    @api.depends("scenario_id")
+    def _compute_display_vendor_packaging_is_possible(self):
+        for menu in self:
+            menu.display_vendor_packaging_is_possible = menu.scenario_id.has_option(
+                "display_vendor_packaging"
+            )
--- a/setup/shopfloor_reception_vendor_packaging/odoo/addons/shopfloor_reception_vendor_packaging
+++ b/setup/shopfloor_reception_vendor_packaging/odoo/addons/shopfloor_reception_vendor_packaging
@@ -1,0 +1,1 @@
+../../../../shopfloor_reception_vendor_packaging
--- a/setup/shopfloor_reception_vendor_packaging/setup.py
+++ b/setup/shopfloor_reception_vendor_packaging/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/setup/shopfloor_vendor_packaging/odoo/addons/shopfloor_vendor_packaging
+++ b/setup/shopfloor_vendor_packaging/odoo/addons/shopfloor_vendor_packaging
@@ -1,0 +1,1 @@
+../../../../shopfloor_vendor_packaging
--- a/setup/shopfloor_vendor_packaging/setup.py
+++ b/setup/shopfloor_vendor_packaging/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/shopfloor_reception_vendor_packaging/README.rst
+++ b/shopfloor_reception_vendor_packaging/README.rst
@@ -1,0 +1,1 @@
+wait for the bot
--- a/shopfloor_reception_vendor_packaging/__init__.py
+++ b/shopfloor_reception_vendor_packaging/__init__.py
@@ -1,0 +1,2 @@
+from . import services
+from .post_init_hook import post_init_hook
--- a/shopfloor_reception_vendor_packaging/__manifest__.py
+++ b/shopfloor_reception_vendor_packaging/__manifest__.py
@@ -1,0 +1,17 @@
+# Copyright 2024 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Shopfloor Reception Vendor Packaging",
+    "summary": "Manage shopfloor reception behavior for vendor packaging",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Inventory",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "depends": ["shopfloor_reception", "shopfloor_vendor_packaging"],
+    "auto_install": True,
+    "post_init_hook": "post_init_hook",
+}
--- a/shopfloor_reception_vendor_packaging/post_init_hook.py
+++ b/shopfloor_reception_vendor_packaging/post_init_hook.py
@@ -1,0 +1,32 @@
+# Copyright 2024 Camptocamp SA (http://www.camptocamp.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import json
+import logging
+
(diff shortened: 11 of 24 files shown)


shopfloor/models/res_company.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    shopfloor_excluded_product_packaging_ids = fields.Many2many(
```
simahawk: Why not on the shopfloor app?

shopfloor/actions/data.py
```diff
@@ -309,9 +309,17 @@ def _product_parser(self):
             ("seller_ids:supplier_code", self._product_supplier_code),
         ]
 
-    def _product_packaging(self, rec, field):
+    def _product_packaging(self, rec, field, apply_excluded=True):
```
simahawk: is `apply_excluded` used anywhere?

simahawk (changes requested):

shopfloor/actions/data.py
JuMiSanAr: No, I added it here in case we ever need to bypass this new logic.

shopfloor/models/res_company.py
JuMiSanAr: I thought of it, but it seemed to me like a very specific setting compared to the rest of the settings in the shopfloor app (i.e. this doesn't really impact the app as a whole, but a specific part of the shopfloor logic).

I can change it if you prefer.

simahawk: Thinking again on this: wouldn't make more sense to have a flag on pkg types?
CC @jbaudoux

mmequignon: I think I would've added a flag at packaging to determine if the package is a vendor package or not.
This way we can filter this depending on the scenario / menu.
T'would probably make it easier to do smth like that `_product_packaging(rec, ..., show_vendor_pkgs=False)`.
show_vendor_pkgs taking the value of `menu.show_vendor_pkgs`.

jbaudoux: I think it is better to have this concept of external package outside shopfloor as we also have a module to compute qties on a move by package type. We also don't want to include vendor pallet on non incoming moves.
I like @mmequignon proposal. The flag must be at packaging type and the menu could configure if we include them or not. For instance, on the reception, we could be interested to see the quantity of the vendor pallet.

JuMiSanAr: I've created a new module to enable this option here:
https://github.com/OCA/product-attribute/pull/1523

JuMiSanAr: I have now updated this PR to use the option in the PR in my last comment above.

Now the vendor packagings will not be displayed by default, and I enabled it in reception as discussed.

mmequignon: I believe reception is an example, we need to determine where we want to see those packaging.

shopfloor/actions/data.py
```diff
@@ -310,8 +326,16 @@ def _product_parser(self):
         ]
 
     def _product_packaging(self, rec, field):
+        display_vendor_packagings = rec.env.context["display_vendor_packagings"]
+        packagings = rec.packaging_ids
+        if display_vendor_packagings:
+            packagings = packagings.filtered(lambda x: x.qty)
+        else:
+            packagings = packagings.filtered(
+                lambda x: x.qty and not x.packaging_type_id.is_vendor_packaging
```
jbaudoux: This requires a new dependency to https://github.com/OCA/product-attribute/pull/1523 that you forgot to add.

shopfloor/actions/data.py
```diff
@@ -310,8 +326,16 @@ def _product_parser(self):
         ]
 
     def _product_packaging(self, rec, field):
+        display_vendor_packagings = rec.env.context["display_vendor_packagings"]
```
jbaudoux:
```suggestion
        display_vendor_packagings = rec.env.context.get("display_vendor_packagings")
```

jbaudoux (changes requested):
We should not add new dependencies in shopfloor. So those changes should move to a new module shopfloor_vendor_packaging (auto install glue between shopfloor and https://github.com/OCA/product-attribute/pull/1523)

I thought you could transmit the option to the front and let the front manage the exclusion at https://github.com/OCA/wms/blob/14.0/shopfloor_mobile/static/wms/src/wms_utils.js#L328
without a global variable in the context. All those overrides to manage context seem fragile.

JuMiSanAr:
> We should not add new dependencies in shopfloor. So those changes should move to a new module shopfloor_vendor_packaging (auto install glue between shopfloor and [OCA/product-attribute#1523](https://github.com/OCA/product-attribute/pull/1523))
> 
> I thought you could transmit the option to the front and let the front manage the exclusion at https://github.com/OCA/wms/blob/14.0/shopfloor_mobile/static/wms/src/wms_utils.js#L328 without a global variable in the context. All those overrides to manage context seem fragile.

I've moved this to the new `shopfloor_vendor_packaging` module.

Regarding trasmitting the option to the front: as far as I understood it (and as far as I see it), we should not let the frontend decide these things. The qty picker will in the end display whatever the backend sends.

mmequignon (changes requested):
This will be valid only if this module is installed.
I believe this has its place in `shopfloor`, adding a default behavior: do not display vendor packages.
Then, we add this option on the menus we want.

JuMiSanAr:
> This will be valid only if this module is installed. I believe this has its place in `shopfloor`, adding a default behavior: do not display vendor packages. Then, we add this option on the menus we want.

This module will be installed automatically when shopfloor is installed right? At least that's what I understand that happens when the `auto-install` flag is set to True.

mmequignon (changes requested):
Add unit tests please :pray:

mmequignon:
> > This will be valid only if this module is installed. I believe this has its place in `shopfloor`, adding a default behavior: do not display vendor packages. Then, we add this option on the menus we want.
> 
> This module will be installed automatically when shopfloor is installed right? At least that's what I understand that happens when the `auto-install` flag is set to True.

well, yes, you're right, I missed this.

@jbaudoux @simahawk any insight about this?

My thoughts are that shopfloor is meant implement incoming/outgoing/internal flows.
To me it makes sense that this is a core feature.
on the other hand, that would be another dependency.
Thanks!

shopfloor_reception/__manifest__.py
```diff
@@ -1,15 +1,15 @@
...
+    "version": "14.0.2.7.0",
     "development_status": "Beta",
     "category": "Inventory",
     "website": "https://github.com/OCA/wms",
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "maintainers": ["mmequignon", "JuMiSanAr"],
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["shopfloor"],
+    "depends": ["shopfloor", "shopfloor_vendor_packaging"],
```
JuMiSanAr: Move to its own `shopfloor_reception_vendor_packaging`

shopfloor_vendor_packaging/models/shopfloor_menu.py
```diff
@@ -0,0 +1,29 @@
...
+
+class ShopfloorMenu(models.Model):
+    _inherit = "shopfloor.menu"
+
+    display_vendor_packagings = fields.Boolean(
+        string="Display vendor packagings",
+        default=False,
+        help=DISPLAY_VENDOR_PACKAGINGS_HELP,
+    )
+    display_vendor_packagings_is_possible = fields.Boolean(
```
mmequignon: no plurals in names :pray:

shopfloor_vendor_packaging/models/shopfloor_menu.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2024 Camptocamp SA (http://www.camptocamp.com)
```
mmequignon, line 1: file not imported. Missing `__init__.py`

shopfloor_vendor_packaging/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from . import actions
```
mmequignon, line 2: from .
```suggestion
from . import actions
from . import models
```
you would've noticed if there were tests :)

mmequignon (changes requested):

shopfloor_reception/migrations/14.0.2.7.0/post-migration.py
```diff
@@ -0,0 +1,37 @@
...
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    reception_scenario = env["shopfloor.scenario"].search([("name", "=", "Reception")])
```
simahawk:
```suggestion
    reception_scenario = env["shopfloor.scenario"].search([("key", "=", "reception")])
```

shopfloor_reception/migrations/14.0.2.7.0/post-migration.py
```diff
@@ -0,0 +1,37 @@
...
+    options = scenario.options
+    options["display_vendor_packagings"] = True
+    options_edit = json.dumps(options or {}, indent=4, sort_keys=True)
+    scenario.write({"options_edit": options_edit})
+    _logger.info("Option display_vendor_packagings added to scenario Zone Picking")
+
+
+def _enable_option_in_menus(menus):
+    for menu in menus:
+        menu.display_vendor_packagings = True
```
simahawk:
```suggestion
        menus.display_vendor_packagings = True
```

shopfloor_reception/services/reception.py
```diff
@@ -710,9 +710,13 @@ def _data_for_stock_pickings(self, pickings, with_lines=False):
         ]
 
     def _data_for_move_lines(self, lines, **kw):
+        if self.work.menu.display_vendor_packagings:
+            kw["display_vendor_packagings"] = True
```
```suggestion
            kw["display_vendor_packagings"] = self.work.menu.display_vendor_packagings
```

shopfloor_vendor_packaging/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "category": "Inventory",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "data": [
+        "views/shopfloor_menu.xml",
+    ],
+    "depends": ["shopfloor", "product_packaging_type_vendor"],
+    "auto-install": True,
```
simahawk, line 18:
```suggestion
    "auto_install": True,
```
it won't work otherwise

simahawk (changes requested):

simahawk:
> > > This will be valid only if this module is installed. I believe this has its place in `shopfloor`, adding a default behavior: do not display vendor packages. Then, we add this option on the menus we want.
> > 
> > 
> > This module will be installed automatically when shopfloor is installed right? At least that's what I understand that happens when the `auto-install` flag is set to True.
> 
> well, yes, you're right, I missed this.
> 
> @jbaudoux @simahawk any insight about this?
> 
> My thoughts are that shopfloor is meant implement incoming/outgoing/internal flows. To me it makes sense that this is a core feature. on the other hand, that would be another dependency. Thanks!

I don't think is a core feature, but I don't know if this is should be restricted to reception only.
My take: if we are unsure today, let's keep it like this and then we'll see.
@jbaudoux WDYT?

jbaudoux (approved):

jbaudoux (commented):
I approve on the principle but you'll have to squash some commits and split with one commit per module

simahawk (approved):

jbaudoux:
Can you rebase ?

JuMiSanAr:
> Can you rebase ?

Done

PR #836, opened 2024-02-09
[16.0][IMP] stock_release_channel_process_end_time: date_deadline

lmignon:
Add a new option on the route definition to allow the use of the release channel's process end date as date deadline on stock moves created when releasing deliveries. This is required to allow the merge of generated moves. By default pulled moves keep the date deadline of their destination moves. If different moves for the same product from different sale orders are grouped into the same delivery, you expect that the system will generate only one pick operation. It's only possible if the generated pulled moves have the same date deadline.


lmignon:
cc @jbaudoux

stock_release_channel_process_end_time/models/stock_picking.py
```diff
@@ -81,3 +83,16 @@ def _after_release_set_expected_date(self):
...
+        # used by the stock.rule to initialize the move date_deadline
+        pickings_by_process_end_date = defaultdict(self.browse)
+        for rec in self:
+            pickings_by_process_end_date[rec.release_channel_id.process_end_date] |= rec
+        # release pickings by process_end_date
+        for process_end_date, pickings in pickings_by_process_end_date.items():
+            pickings = pickings.with_context(
+                release_channel_process_end_date=process_end_date
+            )
+            super(StockPicking, pickings).release_available_to_promise()
```
jbaudoux: Instead of using a context value, can't you override `_prepare_procurement_values` on `stock.move` ?

Sth like:
```
if self.picking_type_code == "outgoing" and self.picking_id.release_channel_id.process_end_date:
  values["date_deadline"] = process_end_date
```

stock_release_channel_process_end_time/models/stock_route.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Camptocamp (https://www.camptocamp.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class StockRoute(models.Model):
+    _inherit = "stock.route"
+
+    propagate_process_end_date_as_move_date_deadline = fields.Boolean(
```
jbaudoux: We already have option `stock_release_use_channel_end_date` in the settings. Can't we use that one? I don't think we need a new one.
I would at least keep the parameters there and not spread them.

stock_release_channel_process_end_time/readme/DESCRIPTION.rst
```diff
@@ -1,4 +1,14 @@
-This module allows to set an end time on a release channel that will be transmitted on
-related stock pickings (on scheduled date) when the channel awakes.
+When managing deliveries through a release channel two main issues are faced:
```
jbaudoux:
```suggestion
This module allows to configure an end time on a release channel that will become the channel end date when the channel awakes. This process end date specify when all the work in the channel should be finalized.
Note: ensure to configure the timezone on the address of the warehouse to have the right time to datetime conversion.

The channel process end date can be propagated to the generated pickings in order to solve two challenges:
```

stock_release_channel_process_end_time/readme/DESCRIPTION.rst
```diff
@@ -1,4 +1,14 @@
...
+  manage the deliveries in the same order as the one expected by the planned
+  release channels.
+
+- You expect to set as deadline of your released move operations the process
+  end date of the release channel. This is useful to ensure that move created
+  when releasing deliveries get the same deadline as the one set on the
+  release channel. This is also required to allow the merge of move operations
+  generated for the same product, location in the same stock picking.
+
+This module solves these issues.
```
jbaudoux: no issue, only challenges :)
```suggestion
```

jbaudoux (changes requested):

stock_release_channel_process_end_time/models/stock_picking.py
lmignon: It was my first thought. I finally choose  this approach to make the 'propagate' configurable on the route. I'll simplify all the approach to only rely oh the option into the global settings.

jbaudoux (approved):
Thanks for the changes

rousseldenis (approved):
Code review

jbaudoux:
@lmignon Can you rebase?

sbejaoui:
> @lmignon Can you rebase?

done

PR #837, opened 2024-02-13
[14.0][IMP] shopfloor: mark lines as sf_checkout_done when handled in backend

JuMiSanAr:
This PR overrides the put_in_pack method from stock.picking
so that it marks lines as shopfloor_checkout_done
when putting them in a pack in the backend.

This is done to make sure that these lines are displayed
in the shopfloor frontend even if they're handled in the backend.

ref: cos-4343

### DIFF
--- a/shopfloor/models/stock_picking.py
+++ b/shopfloor/models/stock_picking.py
@@ -116,3 +116,14 @@ class StockPicking(models.Model):
         )
         assigned_moves._action_assign()
         return new_picking.id
+
+    def _put_in_pack(self, move_line_ids, create_package_level=True):
+        """
+        Marks the corresponding move lines as 'shopfloor_checkout_done'
+        when the package is created in the backend.
+
+        """
+        new_package = super()._put_in_pack(move_line_ids, create_package_level)
+        lines = move_line_ids.filtered(lambda p: p.result_package_id == new_package)
+        lines.write({"shopfloor_checkout_done": True})
+        return new_package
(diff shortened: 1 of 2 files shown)


sebalix (commented):
Why a dedicated module?
IMO this should be part of the `shopfloor` module that brings the `checkout` scenario.

Even if the use of this `sf_checkout_done` flag is not elegant, this mess is introduced by `shopfloor` and has to be handled by the same module. This extra module doesn't bring a real value itself.

ping @simahawk @jbaudoux

JuMiSanAr:
> Why a dedicated module? IMO this should be part of the `shopfloor` module that brings the `checkout` scenario.
> 
> Even if the use of this `sf_checkout_done` flag is not elegant, this mess is introduced by `shopfloor` and has to be handled by the same module. This extra module doesn't bring a real value itself.
> 
> ping @simahawk @jbaudoux

Updated :+1:

jbaudoux (approved):

shopfloor/models/stock_picking.py
```diff
@@ -116,3 +116,16 @@ def split_assigned_move_lines(self, move_lines=None):
...
+        """
+        Marks the corresponding move lines as 'shopfloor_checkout_done'
+        when the package is created in the backend.
+
+        """
+        new_package = super()._put_in_pack(move_line_ids, create_package_level)
+        lines = self.move_line_ids.filtered(
+            lambda p: p.result_package_id == new_package
+        )
+        lines.write({"shopfloor_checkout_done": True})
```
jbaudoux: I think this is more correct:
```suggestion
        lines = move_line_ids.filtered(
            lambda p: p.result_package_id == new_package
        )
        lines.write({"shopfloor_checkout_done": True})
```

mmequignon (commented):
what's the point to this field now that, it is set to `True` even from the backend?

simahawk (approved):

simahawk:
> what's the point to this field now that, it is set to `True` even from the backend?

This is a good remark. However, the questionable part is probably just the name, rather than the function. 
We still need a way to mark single lines as done w/o relying on other values (state, qty, etc).

PR #838, opened 2024-02-13
[16.0] [sale_]stock_release_channel_partner_by_date: set a release channel for a specific delivery address and date

sebalix:
This module integrates the release channels with sales.

Thanks to it, a sale user is able to force the release channel to use in deliveries, ignoring the customer preferences.

TODO:
- [x] TZ compliance
- [x] add a glue module between `sale_stock_release_channel_partner_by_date` and `stock_release_channel_delivery` to restrict the channel selection from the SO based on the carrier

cc @jbaudoux


santostelmo (approved):

sale_stock_release_channel/models/stock_release_channel.py
```diff
@@ -0,0 +1,23 @@
...
+    @api.model
+    def assign_release_channel(self, picking):
+        # Override to propagate the release channel from order to picking
+        picking.ensure_one()
+        if picking.picking_type_id.code != "outgoing" or picking.state in (
+            "cancel",
+            "done",
+        ):
+            return
+        sale_channel = picking.sale_id.release_channel_id
```
jbaudoux: This is not compatible with https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_picking_group_by_partner_by_carrier/models/stock_picking.py#L14

cyrilmanuel: @sebalix @jbaudoux  could you check this point ??

sebalix: I will during the next sprint

jbaudoux: The forced channel must be propagated to the procurement group and managed like the carrier here https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_picking_group_by_partner_by_carrier/models/stock_move.py#L101

jbaudoux (commented):
I suggest this refactor to keep a separation between sales and channel assignment on deliveries:
* On the SO, allow to define a preferred channel for a customer for a specific date (and carrier). By default, it is the channels criteria that allows to assign a delivery inside the channel. However, if an exception is defined for the partner, it is the chosen channel that will be used. The channel will be valid for all deliveries of the partner on that date until it is delivered (the exception is then removed).
* If you cancel the SO, remove the partner-channel exception if there is no more deliveries planned for this partner on the SO expected date. The exception has no reason to remain.
* The preferred channel is valid for all SO on that date. If the partner had no channel in the plan on that date, it becomes an additional valid delivery date for the partner. You can then choose that date as a commitment date on any other SO for that partner without raising a sales exception. You should then see on the SO that it will be delivered by that channel.

Pro:
* This will then be compatible with `stock_picking_group_by_partner_by_carrier_by_date`.
* The SO only allows to manage partner delivery exception. The channel assignment remains only a matter of channel & partner config.
* The exception is valid until the delivery, not afterwards when you wake-up again the channel.
* You don't keep a delivery forced in a channel while the channel is asleep

FYI @lmignon @sbejaoui @rousseldenis

sale_stock_release_channel/models/stock_release_channel.py
jbaudoux: See suggested refactor

stock_release_channel/models/stock_preferred_release_channel.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class StockPreferredReleaseChannel(models.Model):
+    _name = "stock.preferred.release.channel"
```
rousseldenis: @sebalix @jbaudoux Why not having put this in a separate 'stock_release_channel_preferred_channel' module ? I'd like introducing new features separately to:

- lower the code
- identify a functionality more clearly

jbaudoux: Then we should move partner_ids out of `stock_release_channel` module and put all this into a `stock_release_channel_partner` module. Because currently, we can define a fixed list of partners on a channel and here we are just adding a validity date on this list.

sebalix: It was in the balance... first I wanted to merge the existing M2M between partners and channels to have only one relation between them, with an optional date set on each line. But for @jbaudoux it was best to separate the two things are they are not really the same at the end.
Now the reason to keep this in the current `stock_release_channel` was to avoid an additional glue module like:
- `sale_stock_release_channel`
- `sale_stock_release_channel_preferred_channel`

And keep only the first one (as it is its main purpose: "forcing" a release channel from an SO). Or we could put `stock_release_channel_preferred_channel` as a dependency in `sale_stock_release_channel`.

If you and @jbaudoux prefer to get these things separated I'll go this way, you tell me.

EDIT: didn't refresh the page and just saw @jbaudoux comment, I like it

sale_stock_release_channel/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
...
+            rec.preferred_release_channel_id = False
+            date_ = rec.commitment_date or rec.expected_date
+            delivery_date = date_ and date_.date()
+            if not delivery_date:
+                continue
+            pref_rc = self.env["stock.preferred.release.channel"].search(
+                [
+                    ("partner_id", "=", rec.partner_shipping_id.id),
+                    ("date", "=", delivery_date),
+                ]
```
jbaudoux: You need a glue to integrate the carrier. The SO carrier must belong to the channel carriers or the channel must accept any carrier.
And suggest delivery-carrier/delivery_auto_refresh module to have the carrier immediately set on the SO.

sale_stock_release_channel/views/sale_order.xml
```diff
@@ -0,0 +1,40 @@
...
+          <tree editable="top">
+            <field name="partner_id" invisible="1" force_save="1" />
+            <field
+                            name="release_channel_id"
+                            options="{'no_create': True}"
+                            domain="parent.warehouse_id and ['|', ('warehouse_id', '=', parent.warehouse_id), ('warehouse_id', '=', False)]"
+                        />
+            <field name="date" />
+          </tree>
+        </field>
```
jbaudoux: Isn't it too much info on the SO to display all preferred/specific channels ?
What about displaying only `preferred_release_channel_id` and making an inverse method to set a new exception on the date of the SO?
As it relates to the delivery date, I would have displayed that after the date.

stock_release_channel/models/stock_picking.py
```diff
@@ -106,9 +107,23 @@ def _find_release_channel_possible_candidate(self):
             .sorted(key=lambda r: (not bool(r.partner_ids), r.sequence))
         )
 
+    def _get_preferred_release_channel_domain(self):
+        assert self.date_deadline
+        return [
+            ("partner_id", "=", self.partner_id.id),
+            ("date", "=", self.date_deadline.date()),
```
jbaudoux: We are using the `scheduled_date` here https://github.com/OCA/wms/blob/16.0/stock_release_channel_process_end_time/models/stock_release_channel.py#L92
Also we should not look in the past. I would use `max(scheduled_date, now)`
And pay attention to timezone convertion

stock_release_channel/models/stock_picking.py
```diff
@@ -117,7 +132,19 @@ def _get_release_channel_possible_candidate_domain(self):
...
             ("warehouse_id", "=", self.picking_type_id.warehouse_id.id),
-            "|",
-            ("partner_ids", "=", False),
-            ("partner_ids", "in", self.partner_id.ids),
         ]
+        # Look for a preferred release channel at first
+        preferred_rc_domain = None
+        if self.date_deadline:
+            preferred_rc = self.env["stock.preferred.release.channel"].search(
+                self._get_preferred_release_channel_domain(), limit=1
```
jbaudoux: don't limit to 1. There could be multiple results that would be filtered out by the other criteria (warehouse, carrier, ...)

stock_release_channel/models/stock_preferred_release_channel.py
jbaudoux: I thought it would be safer to have the specific/preferred channel in a separate model to not touch the existing relation for the common partner channels. Then on the partner, you see the common channels, the common delivery days, and we add a list of exceptions (channel + date)

stock_release_channel/models/stock_preferred_release_channel.py
```diff
@@ -0,0 +1,43 @@
...
+        ),
+    ]
+
+    @api.autovacuum
+    def _gc_preferred_release_channels(self):
+        pref_rcs = self.search(self._gc_preferred_release_channels_domain())
+        pref_rcs.unlink()
+
+    def _gc_preferred_release_channels_domain(self):
+        return [("date", "<", fields.Date.today())]
```
jbaudoux: tz unsafe
I think it is fine to autovacuum old data (maybe after x months)
What is necessary, it is to remove (archive?) the preferred/specific channel when you put it asleep

stock_release_channel/views/res_partner.xml
```diff
@@ -18,6 +18,25 @@
...
+                            colspan="2"
+                        >
+                          <tree editable="top">
+                            <field name="partner_id" invisible="1" force_save="1" />
+                            <field
+                                    name="release_channel_id"
+                                    options="{'no_create': True}"
+                                />
+                            <field name="date" />
+                          </tree>
```
jbaudoux: I prefer m2m_tags with a nice name_get

jbaudoux (changes requested):
I'm not fan of the name preferred. It is not a preference. It is an exception for a specific day. What about having the new feature in a new module `stock_release_channel_partner_by_date` ?

stock_release_channel/models/res_partner.py
```diff
@@ -15,3 +15,9 @@ class ResPartner(models.Model):
         column2="channel_id",
         string="Release Channels",
     )
+    preferred_release_channel_ids = fields.One2many(
```
jbaudoux: What about
```suggestion
    release_channel_by_date_ids = fields.One2many(
```

stock_release_channel/models/res_partner.py
```diff
@@ -15,3 +15,9 @@ class ResPartner(models.Model):
...
+        comodel_name="stock.preferred.release.channel",
+        inverse_name="partner_id",
+        string="Preferred Release Channels by delivery date",
+        help="Preferred Release Channel for a delivery date.",
```
jbaudoux:
```suggestion
        string="Additional Release Channels",
        help="Additional release channels for a specific delivery date.",
```

stock_release_channel/models/stock_preferred_release_channel.py
jbaudoux: What about naming the model `stock.release.channel.partner.date` ?
And the module `stock_release_channel_partner_by_date`

stock_release_channel/models/stock_preferred_release_channel.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class StockPreferredReleaseChannel(models.Model):
+    _name = "stock.preferred.release.channel"
+    _description = "Preferred Release Channel for a delivery address"
```
jbaudoux:
```suggestion
    _description = "Additional release channel for a delivery address and date"
```

stock_release_channel/views/res_partner.xml
jbaudoux: Additional release channels: 27/06/2024 channelABC, 29/06/2024 channelXYZ

stock_release_channel/models/stock_preferred_release_channel.py
sebalix: @rousseldenis @jbaudoux code moved into two new modules:

- `stock_release_channel_partner_by_date`
- `sale_stock_release_channel_partner_by_date`

sale_stock_release_channel/views/sale_order.xml
sebalix: Done

sale_stock_release_channel/models/stock_release_channel.py
sebalix: Refactored

sebalix:
@rousseldenis I fixed tests you wrote last year in `stock_storage_type`, not sure why it failed (seems the Python implementation of `assertAlmostEqual` changed over time?) nor if the fix is the right one, your review on it is very welcome.

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+            rec.release_channel_id = channel_date.release_channel_id
+
+    def _inverse_release_channel_id(self):
+        channel_date_model = self.env["stock.release.channel.partner.date"]
+        for rec in self:
+            if not rec.release_channel_id:
+                continue
+            if not rec.partner_shipping_id:
+                raise UserError(
+                    _("A customer and shipping address has to be set " "first please.")
```
jbaudoux:
```suggestion
                    _("A customer and shipping address has to be set first please.")
```

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+                )
+            date_ = rec.commitment_date or rec.expected_date
+            delivery_date = date_ and date_.date()
```
jbaudoux: Should be in the tz of the partner_shipping_id

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+            # Remove any existing specific channel entry
+            # for the current delivery address and date
+            existing_channel_date = channel_date_model.search(
+                [
+                    ("partner_id", "=", rec.partner_shipping_id.id),
+                    ("date", "=", delivery_date),
+                ]
+            )
+            existing_channel_date.unlink()
+            if rec.release_channel_id:
```
jbaudoux:
```suggestion
            if not rec.release_channel_id:
                existing_channel_date.unlink()
            elif existing_channel_date:
                existing_channel_date.release_channel_id = rec.release_channel_id.id
            else:
```

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,70 @@
...
+                        "release_channel_id": rec.release_channel_id.id,
+                        "partner_id": rec.partner_shipping_id.id,
+                        "date": delivery_date,
+                    }
+                )
+
+    def _get_release_channel_partner_date_domain(self):
+        self.ensure_one()
+        date_ = self.commitment_date or self.expected_date
+        delivery_date = date_ and date_.date()
```
jbaudoux: Same code as above. Extract in a separate method and make it tz friendly (the tz of the partner_shipping_id)

stock_release_channel_partner_by_date/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 Camptocamp
+# Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
```
jbaudoux:
```suggestion
# Copyright 2024 Camptocamp
# Copyright 2024 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
```

stock_release_channel_partner_by_date/models/stock_picking.py
```diff
@@ -0,0 +1,46 @@
...
+        domain = super()._get_release_channel_possible_candidate_domain()
+        # Look for a specific release channel at first
+        specific_rc_domain = None
+        if self.scheduled_date:
+            specific_rcs = self._get_release_channel_partner_dates()
+            if specific_rcs:
+                specific_rc_domain = [("id", "in", specific_rcs.release_channel_id.ids)]
+        if specific_rc_domain:
+            domain = expression.AND([domain, specific_rc_domain])
+        return domain
```
jbaudoux, line 59: This doesn't look right. If you define on channel B of warehouse CHIC that it has to be used on the scheduled date for partner P, then when you have a picking for warehouse WH and a channel A on that warehouse, you should still match that channel A

stock_release_channel/models/stock_picking.py
```diff
@@ -106,9 +106,23 @@ def _find_release_channel_possible_candidate(self):
             .sorted(key=lambda r: (not bool(r.partner_ids), r.sequence))
         )
 
+    def _get_release_channel_possible_candidate_domain_partners(self):
```
jbaudoux: You have a partner domain also in the module `stock_release_channel_geoengine` that needs to hook this new method

jbaudoux (changes requested):

sale_stock_release_channel_partner_by_date/models/sale_order.py
sebalix: It is still wobbly IMO, as we could also remove the release channel from the SO and/or remove the delivery date, and we could still keep a specific config entry which should not be used anymore.

Maybe we should switch to a normal M2O field and only create/remove such specific channel exceptions when confirming/canceling a SO. What do you think?

stock_release_channel_partner_by_date/models/stock_picking.py
sebalix: You mean we need to filter specific release channels with the warehouse of the current picking, right?
```python
        if self.scheduled_date:
            specific_rcs = self._get_release_channel_partner_dates().filtered(
                lambda o: o.release_channel_id.warehouse_id == self.picking_type_id.warehouse_id
            )
            if specific_rcs:
                [...]
```

jbaudoux: Not only the warehouse. In my opinion, all criteria except the partner criteria must remain valid. Could be the warehouse, picking types, carriers.
A possibility would be to collect all possible candidates excluding partner criteria. Then combine this first with partner dates criteria, and as fallback combine with partner criteria (the one from base module or from geoengine)

sale_stock_release_channel_partner_by_date/models/sale_order.py
jbaudoux: Yes, that sounds better to do that on SO confirmation.

I would keep the compute method and add the depends on the commitment date (but not the expected date).

So you show pre-existing exception. You can set a new one. But if you change the commitment date, as you search for already existing exception, you show if there is one for the date, otherwise it get's empty. User is free to set again a new exception for the date.
If you keep open a draft SO without commitment date but with a channel, and re-open it one day later, the chosen channel remains (as you don't have a depends on expected date) but the exception is not yet committed as SO is not confirmed. When the SO is confirmed, the exception is updated or created.

If you have computed a channel, I think we should not be able to change or remove the exception. I would make it readonly. Then optionally add an action to remove or modify the existing exception. We could add caution in this.

stock_release_channel_partner_by_date/models/stock_picking.py
sebalix: But this is already handled by the call to `super()` there:
```python
    def _get_release_channel_possible_candidate_domain(self):
        domain = super()._get_release_channel_possible_candidate_domain()
```
Warehouse, carriers etc are in the domain, and we add our specific channel entries on top of that.
Regarding the exclusion of partner criteria, this is addressed as well: a hook `_inject_possible_candidate_domain_partners` has been added for this purpose which returns `False` in case we have specific channel entries set.
Am I missing something?

jbaudoux: `_inject_possible_candidate_domain_partners` is not considering the other criteria to decide to inject or not. Like the example I gave you initially. `specific_rcs` will always be `True` and so inject will return `False`. So you won't anymore inject the partner domain on warehouse WH which is wrong. It should be only on warehouse CHIC

stock_release_channel/models/stock_picking.py
sebalix: done

sale_stock_release_channel_partner_by_date/models/sale_order.py
sebalix: PR reworked to register a specific channel on SO confirmation + removing it on cancellation if no other orders are using it.

stock_release_channel_partner_by_date/models/stock_picking.py
sebalix: Done, to achieve this the method computing the domain for channel candidates has been rewritten to merge two kind of domains: channel and picking domains: 11fdcb37ab51dff8870c4818df7ad141c3f92e51

stock_release_channel/models/stock_picking.py
```diff
@@ -106,18 +106,43 @@ def _find_release_channel_possible_candidate(self):
...
+
+        Return a boolean.
+        """
+        return True
+
+    def _get_release_channel_possible_candidate_domain(self):
+        self.ensure_one()
+        channel_domain = self._get_release_channel_possible_candidate_domain_channel()
+        picking_domain = self._get_release_channel_possible_candidate_domain_picking()
+        domain = channel_domain + picking_domain
```
jbaudoux: a bit cleaner 
```suggestion
        domain = expression.AND([channel_domain, picking_domain])
```

stock_release_channel/models/stock_picking.py
```diff
@@ -106,18 +106,43 @@ def _find_release_channel_possible_candidate(self):
...
+        if self._inject_possible_candidate_domain_partners():
+            domain.extend(
+                self._get_release_channel_possible_candidate_domain_partners()
+            )
```
jbaudoux: same here
```suggestion
            expression.AND([
                domain,
                self._get_release_channel_possible_candidate_domain_partners()
            ])
```
or build domains and return `expression.AND(domains)`

stock_release_channel/models/stock_picking.py
```diff
@@ -106,18 +106,43 @@ def _find_release_channel_possible_candidate(self):
...
+        return [
             "|",
             ("picking_type_ids", "=", False),
             ("picking_type_ids", "in", self.picking_type_id.ids),
             "|",
             ("warehouse_id", "=", False),
             ("warehouse_id", "=", self.picking_type_id.warehouse_id.id),
+        ]
+
+    def _get_release_channel_possible_candidate_domain_partners(self):
```
jbaudoux: as you used singular or channel and picking domains
```suggestion
    def _get_release_channel_possible_candidate_domain_partner(self):
```

stock_release_channel_partner_by_date/models/stock_picking.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _get_release_channel_partner_date_domain(self):
+        assert self.scheduled_date
+        return [
+            ("partner_id", "=", self.partner_id.id),
+            ("date", "=", max(self.scheduled_date.date(), fields.Date.today())),
```
jbaudoux: This is not in the right TZ.
I think you'll need something very similar to `scheduled_date_prior_to_channel_end_date_search` to apply proper TZ in the search

jbaudoux: There is a simpler TZ solution here https://github.com/OCA/wms/pull/838/files#diff-c7f5e9d16e4e9d14049bfabd4366d325e0add04cc2ed847cb49c60a3f8d42d9aL18

sale_stock_release_channel/models/sale_order.py
sebalix: I added glue module `sale_stock_release_channel_partner_by_date_delivery`

stock_release_channel_partner_by_date/models/stock_picking.py
sebalix: done

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _get_release_channel_id_depends(self):
+        return ["commitment_date"]
+
+    @api.depends(lambda o: o._get_release_channel_id_depends())
+    def _compute_release_channel_id(self):
+        for rec in self:
+            domain = rec._get_release_channel_partner_date_domain()
+            if not domain:
+                continue
```
jbaudoux: I think this is useless

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,139 @@
...
+        return res
+
+    def _create_release_channel_partner_date(self):
+        self.ensure_one()
+        model = self.env["stock.release.channel.partner.date"]
+        if self.state != "sale" or not self.release_channel_id:
+            return model
+        channel_dates = (
+            self.release_channel_id.release_channel_partner_date_ids.filtered_domain(
+                self._get_release_channel_partner_date_domain()
```
jbaudoux: with active_test=False
and if inactive, set active instead of recreating an entry

sale_stock_release_channel_partner_by_date/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
+* On a sale order form, open the tab *Other info*
+* In the *Delivery* section, you can set the field *Force Release Channel*.
```
jbaudoux:
```suggestion
* In the *Delivery* section, you can set the field *Release Channel*.
```

sale_stock_release_channel_partner_by_date/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
...
+  The release channels you can select are filtered based on the warehouse you
+  could have selected in the same *Delivery* section.
```
jbaudoux, line 4: This sounds good idea but I don't see that in the code

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,139 @@
...
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    release_channel_id = fields.Many2one(
+        comodel_name="stock.release.channel",
+        compute="_compute_release_channel_id",
```
jbaudoux, line 12: It's missing the domain on the warehouse

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _get_release_channel_partner_date_domain(self):
+        self.ensure_one()
+        delivery_date = self._get_delivery_date()
+        if not delivery_date:
+            return
+        return [
+            ("partner_id", "=", self.partner_shipping_id.id),
+            ("date", "=", delivery_date),
+        ]
```
jbaudoux, line 74: it's missing the warehouse in the domain

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    release_channel_id = fields.Many2one(
+        domain=(
+            "carrier_id and "
+            "['|', ('carrier_ids', '=', False), ('carrier_ids', 'in', [carrier_id])] "
+            "or []"
```
jbaudoux: i think you can just write
```suggestion
            "[('carrier_ids', 'in', [False, carrier_id])]"
```

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def _get_release_channel_id_depends(self):
+        depends = super()._get_release_channel_id_depends()
+        depends.append("carrier_id")
+        return depends
+
+    def _get_release_channel_partner_date_domain(self):
+        domain = super()._get_release_channel_partner_date_domain()
+        if self.carrier_id:
+            domain.append(("release_channel_id.carrier_ids", "in", self.carrier_id.ids))
```
jbaudoux: Is it taking also the release channels without carrier ?

jbaudoux (changes requested):
The warehouse got lost in the domains

sale_stock_release_channel_partner_by_date/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Release Channels with Sales",
+    "summary": "Release channels integration with Sales",
+    "version": "16.0.1.0.0",
+    "development_status": "Beta",
+    "license": "AGPL-3",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
```
jbaudoux: :)
```suggestion
    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
```

sale_stock_release_channel_partner_by_date/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Sébastien Alix <sebastien.alix@camptocamp.com>
```
jbaudoux, line 1:
```suggestion
* Sébastien Alix <sebastien.alix@camptocamp.com>

Design
~~~~~~

* Jacques-Etienne Baudoux <je@bcim.be>
```

sale_stock_release_channel_partner_by_date_delivery/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Release Channels with Sales - Delivery",
+    "summary": "Filters channels on sales based on selected carrier.",
+    "version": "16.0.1.0.0",
+    "development_status": "Beta",
+    "license": "AGPL-3",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
```
jbaudoux:
```suggestion
    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
```

sale_stock_release_channel_partner_by_date/models/sale_order.py
sebalix: Without it that will return the result of `search([])`, not sure this is what we want. To be rewritten probably

sale_stock_release_channel_partner_by_date/tests/test_sale_release_channel.py
```diff
@@ -0,0 +1,59 @@
...
+        self.env["stock.release.channel"].assign_release_channel(picking_out)
+        self.assertEqual(picking_out.release_channel_id, self.test_channel)
+        # Cancelling the order removes the specific channel entry
+        order.with_context(disable_cancel_warning=True).action_cancel()
+        channel_date = order._get_release_channel_partner_date()
+        self.assertFalse(channel_date)
+
+    def test_sale_computed_release_channel(self):
+        # Create a specific channel entry from an order, and check if it is
+        # automatically used by a second order with the same delivery date.
```
jbaudoux, line 43: test also it is not automatically used if the specific entry relates to a channel for another warehouse

sale_stock_release_channel_partner_by_date/readme/USAGE.rst
sebalix: Lost during the refactoring I guess yes...

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
sebalix: I think I tried but it wasn't returning the expected result. I'll check again.

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
sebalix: Good point, I didn't know what was the expected result with such glue module.
To fix, we should match such release channels as well.

sebalix: Done

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
sebalix: Proposed solution doesn't work.
I changed the computation of this domain returned to the view, not sure it is the best approach but I could not think of a best one :man_shrugging:

sale_stock_release_channel_partner_by_date/models/sale_order.py
sebalix: Fixed

sale_stock_release_channel_partner_by_date/models/sale_order.py
sebalix: Done

sale_stock_release_channel_partner_by_date/models/sale_order.py
jbaudoux: No because you already test it here https://github.com/OCA/wms/pull/838/files#diff-c052d7cf6444a81e0f15b713218cdcca8711e655000bc6b76a2ee1034e2f0e9cR64
It's just weird to see a test on a domain here that in fact you don't use here

jbaudoux:
LGTM. Would be nice to have this additional test to ensure it also works in multi-warehouses : https://github.com/OCA/wms/pull/838#discussion_r1713317357

sale_stock_release_channel_partner_by_date/tests/test_sale_release_channel.py
sebalix: Done

jbaudoux (approved):
Looking good :)
For me you can squash all the fixups

sale_stock_release_channel_partner_by_date/models/sale_order.py
sebalix: We want to skip the update of `release_channel_id` field if there is no domain (or no delivery date available), removing this check + `continue` will override the `release_channel_id` field value everytime, it's not what we want.

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,160 @@
...
+                continue
+            channel_date = rec.release_channel_partner_date_id
+            rec.release_channel_id = channel_date.release_channel_id
+
+    @api.depends(
+        "state",
+        "partner_shipping_id",
+        "commitment_date",
+        "expected_date",
+        "release_channel_id",
```
jbaudoux:
```suggestion
        "warehouse_id",
```

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,160 @@
...
+        parts = ", ".join(self._release_channel_id_domain_parts())
+        return f"[{parts}]"
+
+    def _release_channel_id_domain_parts(self):
+        return [
+            "'|', ('warehouse_id', '=', False), ('warehouse_id', '=', warehouse_id)",
+        ]
+
+    def _get_release_channel_id_depends(self):
+        return ["commitment_date"]
```
jbaudoux:
```suggestion
        return ["commitment_date", "partner_shipping_id", "warehouse_id"]
```

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -0,0 +1,160 @@
...
+    def _get_release_channel_partner_date(self):
+        self.ensure_one()
+        model = self.env["stock.release.channel.partner.date"]
+        domain = self._get_release_channel_partner_date_domain()
+        return domain and model.search(domain, limit=1) or model
+
+    def _get_release_channel_partner_date_domain(self):
+        self.ensure_one()
+        delivery_date = self._get_delivery_date()
+        if not delivery_date:
```
jbaudoux:
```suggestion
        if not delivery_date or not self.partner_shipping_id or not self.warehouse_id:
```

sale_stock_release_channel_partner_by_date/models/sale_order.py
jbaudoux: let's use 
`if not delivery_date or not self.partner_shipping_id or not self.warehouse_id:`
instead then

jbaudoux (approved):

PR #839, opened 2024-02-16
[14.0][IMP] shopfloor, stock_picking_partner_note: display picking info notes in cluster_picking and checkout scenarios

JuMiSanAr:
Depends on:
- https://github.com/OCA/stock-logistics-workflow/pull/1504

Stock picking partner notes are notes that are pre-defined for each customer
and can be displayed for each picking.
With this PR, we send them to the frontend for the cluster picking scenario and display them where needed.

This PR also deprecated `shopfloor_packing_info` in favor of `stock_picking_partner_note`, which is now used for both checkout and cluster picking.

note_in_cluster_picking

ref: cos-4214

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -98,7 +98,6 @@ class Checkout(Component):
         return {
             "selected_move_lines": self._data_for_move_lines(lines.sorted()),
             "picking": self.data.picking(picking),
-            "packing_info": self._data_for_packing_info(picking),
             "no_package_enabled": not self.options.get("checkout__disable_no_package"),
             # Used by inheriting module
             "package_allowed": True,
@@ -110,14 +109,6 @@ class Checkout(Component):
             data=self._data_response_for_select_package(picking, lines),
             message=message,
         )
-
-    def _data_for_packing_info(self, picking):
-        """Return the packing information
-
-        Intended to be extended.
-        """
-        # TODO: This could be avoided if included in the picking parser.
-        return ""
 
     def _response_for_select_dest_package(self, picking, move_lines, message=None):
         packages = picking.mapped("move_line_ids.result_package_id").filtered(
--- a/shopfloor/tests/test_checkout_list_delivery_packaging.py
+++ b/shopfloor/tests/test_checkout_list_delivery_packaging.py
@@ -111,7 +111,6 @@ class CheckoutListDeliveryPackagingCase(CheckoutCommonCase, CheckoutSelectPackag
                 "selected_move_lines": [
                     self._move_line_data(ml) for ml in selected_lines.sorted()
                 ],
-                "packing_info": self.service._data_for_packing_info(self.picking),
                 "no_package_enabled": not self.service.options.get(
                     "checkout__disable_no_package"
                 ),
--- a/shopfloor/tests/test_checkout_scan_package_action.py
+++ b/shopfloor/tests/test_checkout_scan_package_action.py
@@ -172,7 +172,6 @@ class CheckoutScanPackageActionCase(CheckoutCommonCase, CheckoutSelectPackageMix
             data={
                 "picking": self.data.picking(picking),
                 "selected_move_lines": self.data.move_lines(selected_lines),
-                "packing_info": self.service._data_for_packing_info(picking),
                 "no_package_enabled": not self.service.options.get(
                     "checkout__disable_no_package"
                 ),
--- a/shopfloor_mobile/static/wms/src/css/main.css
+++ b/shopfloor_mobile/static/wms/src/css/main.css
@@ -71,3 +71,10 @@ I tested only w/ checkout/select_package for now
     top: 5px;
     right: 5px;
 }
+
+/*
+Split the lines correctly if there are multiple lines separated by \n in a v-text.
+*/
+.split-text-lines {
+    white-space: pre-line;
+}
--- a/shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
+++ b/shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
@@ -34,7 +34,7 @@ const ClusterPicking = {
                 v-on:cancel="state.on_cancel"
                 />
             <v-alert type="info" tile v-if="state_is('start_line') && state.data.picking.note" class="packing-info">
-                <p v-text="state.data.picking.note" />
+                <p class="split-text-lines" v-text="state.data.picking.note" />
             </v-alert>
             <batch-picking-line-detail
                 v-if="state_in(['start_line', 'scan_destination', 'change_pack_lot', 'stock_issue'])"
--- a/shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
+++ b/shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
@@ -33,6 +33,9 @@ const ClusterPicking = {
                 v-on:confirm="state.on_confirm"
                 v-on:cancel="state.on_cancel"
                 />
+            <v-alert type="info" tile v-if="state_is('start_line') && state.data.picking.note" class="packing-info">
+                <p v-text="state.data.picking.note" />
+            </v-alert>
             <batch-picking-line-detail
                 v-if="state_in(['start_line', 'scan_destination', 'change_pack_lot', 'stock_issue'])"
                 :line="state.data"
--- a/shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
+++ b/shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
@@ -41,11 +41,12 @@ def populate_stock_picking_note__packing(env):
                 "note_type_id": packing_note_type.id,
             }
         )
-        partner = env["res.partner"].search(
-            [("shopfloor_packing_info_id", "=", packing_info.id)], limit=1
+        partners = env["res.partner"].search(
+            [("shopfloor_packing_info_id", "=", packing_info.id)]
         )
-        if partner:
-            partner.stock_picking_note_ids |= note
+        if partners:
+            for partner in partners:
+                partner.stock_picking_note_ids |= note
 
 
 def migrate(cr, version):
--- a/shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
+++ b/shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
@@ -1,0 +1,56 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+# NOTE: This module is deprecated in favor of stock_picking_partner_note.
+# Here, we migrate the data from shopfloor.packing.info to stock.picking.note.
+
+
+def setup_stock_picking_note__packing(env):
+    """Create a new packing note type named 'packing'."""
+
+    _logger.info("Create a new picking note type named 'packing'")
+    packing_note_type = env["stock.picking.note.type"].search(
+        [("name", "=", "packing")], limit=1
+    )
+    if not packing_note_type:
+        packing_note_type = env["stock.picking.note.type"].create({"name": "packing"})
+    return packing_note_type
+
+
+def populate_stock_picking_note__packing(env):
+    """Migrate data from shopfloor.packing.info to stock.picking.note of type 'packing'.
+
+    We also update the stock_picking_note_ids of the partners
+    based on their existing shopfloor_packing_info_id values."""
+
+    _logger.info(
+        "Migrate data from shopfloor.packing.info to stock.picking.note of type 'packing'"
+    )
+    packing_note_type = setup_stock_picking_note__packing(env)
+    existing_packing_infos = env["shopfloor.packing.info"].search([])
+    for packing_info in existing_packing_infos:
+        note = env["stock.picking.note"].create(
+            {
+                "name": packing_info.text,
+                "note_type_id": packing_note_type.id,
+            }
+        )
+        partner = env["res.partner"].search(
+            [("shopfloor_packing_info_id", "=", packing_info.id)], limit=1
+        )
+        if partner:
+            partner.stock_picking_note_ids |= note
+
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    populate_stock_picking_note__packing(env)
--- a/shopfloor_packing_info/readme/DESCRIPTION.rst
+++ b/shopfloor_packing_info/readme/DESCRIPTION.rst
@@ -1,3 +1,7 @@
+DEPRECATED: This module is deprecated in favor of stock_picking_partner_note,
+which is now used in shopfloor to display notes in the app.
+
+
 Adds the option to choose predefined messages for the
 parameter "Checkout Packing Info", that is set on Customers
 and reused in other places, mainly pickings. These new messages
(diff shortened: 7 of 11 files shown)


shopfloor_packing_info/services/checkout.py
```diff
@@ -10,7 +10,7 @@ class CheckoutExt(Checkout):
 
     def _data_for_packing_info(self, picking):
         res = super()._data_for_packing_info(picking)
-        if picking.picking_type_id.shopfloor_display_packing_info:
+        if picking.note:
```
JuMiSanAr: @jbaudoux I don't see how this is supposed to work:
shouldn't we keep the current version with `shopfloor_packing_info` in the `select_package` screen, and use `picking.note` in the screen where we select a picking?

shopfloor_packing_info/views/stock_picking_views.xml
```diff
@@ -10,23 +10,15 @@
                 position="before"
             >
                 <group name="shopfloor">
-                    <field name="shopfloor_display_packing_info" invisible="1" />
```
JuMiSanAr: If I understand correctly, `shopfloor_display_packing_info` will be dropped entirely, do we really want this?

jbaudoux: Yes the purpose is to decrease shopfloor complexity.
This gets replaced by the picking note that can be populated with stock_picking_partner_note

jbaudoux:
I think you better deprecate this module and update shppfloor to send the picking note to the front and display it in the front
You don't need to depend on stock_picking_partner_note as this is just populating the note on the picking

JuMiSanAr:
> I think you better deprecate this module and update shppfloor to send the picking note to the front and display it in the front You don't need to depend on stock_picking_partner_note as this is just populating the note on the picking

@jbaudoux 
Ok, I've added a note in de `DESCRIPTION.rst` file in `shopfloor_packing_note` (let me know if there's a better way to mark a module as deprecated) and I've also pushed to this PR a new `shopfloor_stock_picking_partner_note` module. Let me know if it looks ok to you.

shopfloor_stock_picking_partner_note/services/checkout.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class CheckoutExt(Checkout):
+    _inherit = "base.shopfloor.process"
+
+    def _data_for_packing_info(self, picking):
+        res = super()._data_for_packing_info(picking)
+        note = picking.note
+        if note:
+            res += "{}{}".format(os.linesep, note).strip()
```
jbaudoux: Why don't you put that in shopfloor module ? The `note` field is provided by `stock` module, you don't need a new dependency.

I guess you need to do the same in the cluster picking & zone picking scenario (cos-4214)

jbaudoux (changes requested):

shopfloor_packing_info/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,7 @@
+DEPRECATED: This module is deprecated in favor of stock_picking_partner_note,
+which is now used in shopfloor to display notes in the app.
```
jbaudoux, line 2: You could add in this module the dependency to `stock_picking_partner_note` and the migration script to move data to the new model

shopfloor_stock_picking_partner_note/services/checkout.py
JuMiSanAr: Because we're using the added functionality implemented in the stock_picking_partner_note` module right?
`https://github.com/OCA/stock-logistics-workflow/pull/1504/files#diff-9ec7eee25885dbe0455eef41843ed79f5fbfd8c8579b90c09f05754ee93fef2cR11` 

Regarding the two other scenarios, this method is not used and I don't see any packing-related info already implemented there, that's why I didn't touch it.
If we want it there too, I think that's a different topic and could be implemented at a later stage.

jbaudoux: One module is filling the note, shopfloor is displaying the note. But they don't depend on each other. You could fill the note differently.

JuMiSanAr:
@jbaudoux I've updated the module with your remarks, but now I'm wondering: will the notes be correctly recomputed each time we change the partner? Because the field introduced in `stock_picking_partner_note` is stored...

shopfloor/services/checkout.py
```diff
@@ -100,22 +100,14 @@ def _response_for_select_package(self, picking, lines, message=None):
             data={
                 "selected_move_lines": self._data_for_move_lines(lines.sorted()),
                 "picking": self.data.picking(picking),
-                "packing_info": self._data_for_packing_info(picking),
+                "packing_info": self.data.packing_info(picking),
```
jbaudoux: picking note is already in `_picking_parser`, can't we drop `packing_info` ?

shopfloor/actions/data.py
```diff
@@ -369,3 +371,13 @@ def _get_operation_progress(self, domain):
...
         }
+
+    @ensure_model("stock.picking")
+    def packing_info(self, record):
+        """Return the packing information"""
+
+        record.ensure_one()
+        note = record.note
+        if note:
+            return "{}{}".format(os.linesep, note).strip()
```
jbaudoux: I don't get this. You prefix the note by a linesep and then remove it?

shopfloor/services/cluster_picking.py
```diff
@@ -402,6 +402,7 @@ def _data_move_line(self, line, **kw):
         data["package_dest"] = None
         data["batch"] = self.data.picking_batch(batch)
         data["picking"] = self.data.picking(picking)
+        data["packing_info"] = self.data.packing_info(picking)
```
jbaudoux: don't need this, the note is already in data["picking"]

shopfloor_mobile/static/wms/src/scenario/cluster_picking.js
```diff
@@ -30,6 +30,9 @@ const ClusterPicking = {
                 v-on:confirm="state.on_confirm"
                 v-on:cancel="state.on_cancel"
                 />
+            <v-alert type="info" tile v-if="state_is('start_line') && state.data.packing_info" class="packing-info">
+                <p v-text="state.data.packing_info" />
```
jbaudoux: Use data.picking.note

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,64 @@
...
+
+# NOTE: This module is deprecated in favor of stock_picking_partner_note.
+# Here, we migrate the data from shopfloor.packing.info to stock.picking.note.
+
+
+def setup_stock_picking_note__picking(env):
+    """Create a new picking note type named 'picking'."""
+
+    _logger.info("Create a new picking note type named 'picking'")
+    env["stock.picking.note.type"].create({"name": "picking"})
```
jbaudoux: Existing data was only for packing.
You should not create a picking note type here

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def setup_stock_picking_note__packing(env):
+    """Create a new packing note type named 'packing'."""
+
+    _logger.info("Create a new picking note type named 'packing'")
+    env["stock.picking.note.type"].create({"name": "packing"})
```
jbaudoux:
```suggestion
    packing_note_type = env["stock.picking.note.type"].search(
        [("name", "=", "packing")], limit=1
    )
    if not packing_note_type:
      packing_note_type = env["stock.picking.note.type"].create({"name": "packing"})
```

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,64 @@
...
+
+    We also update the stock_picking_note_ids of the partners
+    based on their existing shopfloor_packing_info_id values."""
+
+    _logger.info(
+        "Migrate data from shopfloor.packing.info to stock.picking.note of type 'packing'"
+    )
+    packing_note_type = env["stock.picking.note.type"].search(
+        [("name", "=", "packing")], limit=1
+    )
```
jbaudoux: maybe merge setup_stock_picking_note__packing here

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,64 @@
...
+    existing_packing_infos = env["shopfloor.packing.info"].search([])
+    for packing_info in existing_packing_infos:
+        env["stock.picking.note"].create(
+            {
+                "name": packing_info.text,
+                "note_type_id": packing_note_type.id,
+            }
+        )
+    partners = env["res.partner"].search([])
```
jbaudoux:
```suggestion
    partners = env["res.partner"].search([("shopfloor_packing_info_id", "!=", False)])
```

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,64 @@
...
+        [("name", "=", "packing")], limit=1
+    )
+    existing_packing_infos = env["shopfloor.packing.info"].search([])
+    for packing_info in existing_packing_infos:
+        env["stock.picking.note"].create(
+            {
+                "name": packing_info.text,
+                "note_type_id": packing_note_type.id,
+            }
+        )
```
jbaudoux: keep mapping between old record and new record

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,64 @@
...
+    partners = env["res.partner"].search([])
+    for partner in partners:
+        note = env["stock.picking.note"].search(
+            [("name", "=", partner.shopfloor_packing_info_id.text)]
+        )
```
jbaudoux: Use mapping instead

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,64 @@
...
+        if note:
+            partner.stock_picking_note_ids = partner.stock_picking_note_ids | note
```
jbaudoux:
```suggestion
            partner.stock_picking_note_ids |= note
```

jbaudoux (changes requested):

jbaudoux:
> @jbaudoux I've updated the module with your remarks, but now I'm wondering: will the notes be correctly recomputed each time we change the partner? Because the field introduced in `stock_picking_partner_note` is stored...

Yes, there is a depends on partner: https://github.com/OCA/stock-logistics-workflow/pull/1504/files#diff-9ec7eee25885dbe0455eef41843ed79f5fbfd8c8579b90c09f05754ee93fef2cR13

shopfloor/services/cluster_picking.py
```diff
@@ -1418,7 +1418,10 @@ def _states(self):
         """
         return {
             "confirm_start": self._schema_for_batch_details,
-            "start_line": self._schema_for_single_line_details,
+            "start_line": dict(
+                self._schema_for_single_line_details,
+                packing_info={"type": "string", "nullable": True},
```
jbaudoux: I guess this should be dropped

shopfloor/tests/test_checkout_list_delivery_packaging.py
```diff
@@ -111,7 +111,7 @@ def test_list_delivery_packaging_not_available(self):
                 "selected_move_lines": [
                     self._move_line_data(ml) for ml in selected_lines.sorted()
                 ],
-                "packing_info": self.service._data_for_packing_info(self.picking),
+                "packing_info": self.service.data.packing_info(self.picking),
```
jbaudoux: I guess this should be dropped

shopfloor/tests/test_checkout_scan_package_action.py
```diff
@@ -172,7 +172,7 @@ def test_scan_package_action_scan_package_keep_source_package_error(self):
             data={
                 "picking": self.data.picking(picking),
                 "selected_move_lines": self.data.move_lines(selected_lines),
-                "packing_info": self.service._data_for_packing_info(picking),
+                "packing_info": self.service.data.packing_info(picking),
```
jbaudoux: I guess this should be dropped

jbaudoux (commented):
I don't see the display of the note in the zone picking

JuMiSanAr:
> I don't see the display of the note in the zone picking

The task descriptions says it's not applicable for zone picking right?

"In shopfloor cluster picking (n.a. for zone picking), display the respective picking info..."

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+# NOTE: This module is deprecated in favor of stock_picking_partner_note.
```
simahawk, line 10: We should move this upgrade into shopfloor and ignore this module. Just keep the msg that is deprecated.
As we migrate the data we can flag it as to be uninstalled.

simahawk (changes requested):
LG overall but the mig step and the dep.

shopfloor_packing_info/migrations/14.0.1.2.0/post-migrate.py
JuMiSanAr: @jbaudoux what do you think?

simahawk: Nevermind about this. However, IMO we should either uninstall the module automatically or at least add a post load hook to log deprecation.

jbaudoux: The script is in this module for those who had it installed. This should not move to shopfloor

simahawk (approved):

jbaudoux (approved):
More feature with less code, isn't it wonderful !

Please split last commit. One commit for shopfloor, and one for shopfloor_mobile

simahawk:
@JuMiSanAr can you rebase?

JuMiSanAr:
Failing tests should be fixed with: 
https://github.com/OCA/stock-logistics-workflow/pull/1543

rousseldenis:
@JuMiSanAr Don't forget to have a meaningful PR's description. It should at least mention the module you modify in order to help reviewing. Thanks

shopfloor_mobile/static/wms/src/css/main.css
```diff
@@ -71,3 +71,10 @@ I tested only w/ checkout/select_package for now
     top: 5px;
     right: 5px;
 }
+
+/*
+Split the lines correctly in the picking note if there are multiple lines
+*/
+.picking-note {
```
simahawk: instead of using a specific class, define a generic one that you can reuse anywhere no matter what you need to render.

Eg: `split-text-lines` or similar.

simahawk (changes requested):
image

JuMiSanAr:
> image

This test fails because we depend on `stock_picking_partner_note`, which currently has `note` as an Html field. As soon as we merge https://github.com/OCA/stock-logistics-workflow/pull/1543#pullrequestreview-1992070362 , the test should pass.

simahawk:
> > image
> 
> This test fails because we depend on `stock_picking_partner_note`, which currently has `note` as an Html field. As soon as we merge [OCA/stock-logistics-workflow#1543 (review)](https://github.com/OCA/stock-logistics-workflow/pull/1543#pullrequestreview-1992070362) , the test should pass.

Then you should add that PR as a tmp test req :wink

PR #842, opened 2024-02-23
[16.0] stock_release_channel: multi-company support

sebalix:


stock_release_channel/models/stock_picking.py
```diff
@@ -84,6 +84,7 @@ def _get_release_channel_possible_candidate_domain(self):
         self.ensure_one()
         return [
             ("state", "!=", "asleep"),
+            ("company_id", "=", self.picking_type_id.company_id.id),
```
jbaudoux:
```suggestion
            ("company_id", "=", self.company_id.id),
```

stock_release_channel/models/stock_release_channel.py
```diff
@@ -52,6 +60,7 @@ class StockReleaseChannel(models.Model):
         domain="warehouse_id"
         " and [('warehouse_id', '=', warehouse_id), ('code', '=', 'outgoing')]"
         " or [('code', '=', 'outgoing')]",
+        check_company=True,
```
jbaudoux: You cannot use both domain and check_company on the field. You need to add `("company_id", "in", (company_id, False))` in the domain

stock_release_channel/models/stock_release_channel.py
```diff
@@ -209,6 +218,7 @@ class StockReleaseChannel(models.Model):
         column2="partner_id",
         string="Partners",
         context={"active_test": False},
+        domain="['|', ('company_id', '=', company_id), ('company_id', '=', False)]",
```
jbaudoux: Here you can use
```suggestion
        check_company=True,
```

jbaudoux (changes requested):
I would also add on:
* stock_release_channel
  * picking_ids `check_company=True` 
* shipment_advice
  * dock_id : `domain="[('warehouse_id', '=', warehouse_id)]"`
  * planned_move_ids `check_company=True`
  * loaded_move_lines_ids `check_company=True`
  * loaded_move_lines_without_package_ids `check_company=True`

stock_release_channel/models/stock_picking.py
sebalix: `<stock.picking>.company_id` is not required, while it is on `<stock.picking.type>`

jbaudoux: But you are processing a picking, so better use its company.
I guess they forgot the required=True and we can assume there will always be one

jbaudoux (approved):
LG. Thanks. Can you keep one commit per module ?

Do you do the implementation also in the transport repo?

jbaudoux:
@lmignon Is checking the toursolver failing test case. He fixed it in transport repo. I don't know why it is still complaining here

jbaudoux:
> @lmignon Is checking the toursolver failing test case. He fixed it in transport repo. I don't know why it is still complaining here

fixed here https://github.com/OCA/wms/pull/928

lmignon (approved):
LGTM (Code review only)

sebalix:
@jbaudoux there is one commit per module yes. I'll do the changes on `shipment_advice` yes

sebalix:
https://github.com/OCA/stock-logistics-transport/pull/146

PR #843, opened 2024-02-23
[14.0][IMP] shopfloor: apply zero_check only if product is not a consumable

JuMiSanAr:
Consumables don't need to be taken into account when applying the zero_check.

ref: cos-4403

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -776,8 +776,11 @@ class ClusterPicking(Component):
                 qty_done=quantity,
             )
         move_line.write({"qty_done": quantity, "result_package_id": bin_package.id})
-
-        zero_check = move_line.picking_id.picking_type_id.shopfloor_zero_check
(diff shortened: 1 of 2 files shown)


shopfloor/services/cluster_picking.py
```diff
@@ -776,8 +776,11 @@ def scan_destination_pack(self, picking_batch_id, move_line_id, barcode, quantit
                 qty_done=quantity,
             )
         move_line.write({"qty_done": quantity, "result_package_id": bin_package.id})
-
-        zero_check = move_line.picking_id.picking_type_id.shopfloor_zero_check
+        # Only apply zero check if the product is not a consumable.
+        zero_check = (
+            move_line.product_id.type != "consu"
```
mmequignon: i'd say we only apply this check if product type is `product`
There's also service products.
Odoo or another module might add more.

mmequignon (changes requested):

simahawk (approved):

PR #847, opened 2024-02-26
[16.0] Add 'stock_release_channel_block_backorder'

sebalix:
Linked to:
- #848

cc @rousseldenis @jbaudoux


stock_release_channel_block_backorder/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+            if picking.blocked_for_channel_assignation:
+                picking.blocked_for_channel_assignation_label = _("Blocked")
+            else:
+                picking.blocked_for_channel_assignation_label = ""
+
+    def _create_backorder(self):
+        backorders = super()._create_backorder()
+        # Block backorder moves if qty is not available
+        for move in backorders.move_ids:
+            move.delivery_requires_other_lines = move.product_qty_unavailable > 0
```
jbaudoux:
```suggestion
            move.delivery_requires_other_lines = not move.picking_id.release_ready
```

stock_release_channel_block_backorder/tests/test_stock_picking.py
```diff
@@ -0,0 +1,108 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+# from odoo import Command
```
jbaudoux: drop

stock_release_channel_block_backorder/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+Block delivery of unavailable products on backorders.
+
+A blocked delivey cannot be assigned to a release channel.
+To unblock such delivery a new order has to be created, or the user has to
```
jbaudoux: Should be noted that you likely need the stock_picking_group_by_partner_by_carrier/ module

stock_release_channel_block_backorder/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Release Channel - Block Unavailable Product",
```
jbaudoux:
```suggestion
    "name": "Stock Release Channel - Block Backorder",
```

stock_release_channel_block_backorder/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "summary": """Block delivery of unavailable products on backorders.""",
```
jbaudoux:
```suggestion
    "summary": """Block delivery of unavailable backorders.""",
```

stock_release_channel_block_backorder/models/stock_picking.py
sebalix: So I could drop `product_qty_unavailable` field as well

jbaudoux: Like you did seems better but you could drop the computed
```suggestion
def _block_on_backorder(self)
  return not move.ordered_available_to_promise_qty

            move.delivery_requires_other_lines = move._block_on_backorder```
using float_compare

stock_release_channel_block_backorder/readme/DESCRIPTION.rst
sebalix: added

sebalix:
Build fixed here: https://github.com/OCA/wms/pull/849

yvaucher (approved):

sebalix:
After some analysis, I would like to bring some changes to this module:

- rename `delivery_requires_other_lines` to `release_blocked` to be agnostic about why it's blocked and how to unblock it
- do not mention in help/tooltips that additional moves should be added into the transfer to unblock it automatically: it doesn't work this way in Odoo std (each SO has its own procurement group which is a key element to find an existing transfer to put newly created moves in), the behavior expected by this text is brought by other OCA modules
- integrate `release_blocked` flag of move in the computation of `release_ready`, so a blocked move cannot be released at all
- rename the module `stock_release_channel_block` or even `stock_available_to_promise_release_block` as:
  - the `release_blocked` flag will be part of the `release_ready` computation, so not really linked to the channel mechanism
  - we could block any deliveries (including backorders,  those will be blocked automatically, add an option on the picking type for that maybe?)

sebalix:
Closing in favor of: https://github.com/OCA/wms/pull/865

PR #850, opened 2024-02-28
[16.0][FIX] stock_available_to_promise_release: fix cache and failing tests

jbaudoux:
_compute_ordered_available_to_promise has no depends set. Any field depending on computed values must invalidate cache before reading.

cc @sebalix @mt-software-de @rousseldenis

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -238,6 +238,9 @@ class StockMove(models.Model):
         )
         return query, params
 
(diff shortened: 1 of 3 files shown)


mt-software-de:
LG. Could you just please update the docstring of the methods and explaining why a invalidate cache is needed.

mt-software-de (approved):

jbaudoux:
> LG. Could you just please update the docstring of the methods and explaining why a invalidate cache is needed.

done

lmignon (approved):

PR #851, opened 2024-02-29
[16.0][FIX] stock_release_channel_process_end_time: releasable picking computation

jbaudoux:
Fix failing test on `stock_release_channel` when this module is installed

Ensure computation is bound to related channel.
Fix timezone conversion and compare dates.
Don't break base behavior when no process_end_date is set.

cc @lmignon @rousseldenis @sbejaoui @sebalix


lmignon (approved):
LGTM (Code review). Thank you @jbaudoux for these fixes.

jbaudoux:
Test added for the fixed use case

lmignon (approved):

sebalix (approved):

PR #856, opened 2024-03-01
[16.0][IMP] Release : set expected date

jbaudoux:
* stock_available_to_promise_release
    * Set date before release, so much easier, no need to extract chain of moves and go back and forth the chain
    * Don't modify other moves of the picking. If picking contains moves from other SO, all the other pickings in the chain of moves where also updated, not only the one just released (cc @mt-software-de) 
    * Add hook to get the date
    * Allow to receive a predefined date to set on the pickings
* stock_release_channel
   * Add hook to get date from channel
   * Add support to channel change. Propagate new date
* stock_release_channel_process_end_time
    * Use new provided hooks which fixes date propagation; No need to travel the chain of moves
    * Now works properly when the module that propagates the channel on all pickings of the chain is not installed
    * Channel change and update date is now already managed by stock_release_channel module. Just use the provided hook

cc @sebalix @lmignon @sbejaoui @mt-software-de

### DIFF
--- a/stock_release_channel/models/stock_move.py
+++ b/stock_release_channel/models/stock_move.py
@@ -25,3 +25,11 @@ class StockMove(models.Model):
         if pickings:
             pickings._delay_assign_release_channel()
         return moves
+
+    def _release_get_expected_date(self):
+        """Return the new scheduled date of a single delivery move"""
+        channel = self.picking_id.release_channel_id
+        expected_date = channel and channel._get_expected_date()
+        if not expected_date:
+            return super()._release_get_expected_date()
+        return expected_date
--- a/stock_release_channel/models/stock_picking.py
+++ b/stock_release_channel/models/stock_picking.py
@@ -15,7 +15,33 @@ class StockPicking(models.Model):
         ondelete="restrict",
         copy=False,
         tracking=True,
+        inverse="_inverse_release_channel_id",
     )
+
+    def _inverse_release_channel_id(self):
+        """When the release channel is modified, update the expected date
+
+        Set new date on all moves in the chain"""
+        for delivery in self:
+            if not delivery.last_release_date:
+                continue
+            if delivery.picking_type_code != "outgoing":
+                continue
+            channel = delivery.release_channel_id
+            new_date = channel and channel._get_expected_date()
+            if not new_date:
+                continue
+            moves = delivery.move_ids.filtered(
+                lambda m: m.state not in ("cancel", "done")
+            )
+            move_chain_ids = []
+            while moves:
+                move_chain_ids += moves.ids
+                moves = moves.move_orig_ids.filtered(
+                    lambda m: m.state not in ("cancel", "done")
+                )
+            all_moves = moves.browse(move_chain_ids)
+            all_moves._release_set_expected_date(new_date)
 
     def _delay_assign_release_channel(self):
         for picking in self:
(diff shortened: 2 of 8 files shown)


stock_release_channel_process_end_time/models/stock_release_channel.py
```diff
@@ -100,3 +94,17 @@ def _field_picking_domains(self):
                 new_domain.append(criteria)
         res["count_picking_release_ready"] = new_domain
         return res
+
+    def _get_expected_date(self):
+        """Return the new date to set on move chain"""
```
sebalix: Better to use comment instead of docstrings when overriding methods (this overrides original docstrings in IDE and if the method is triggered by a job, the description of the job = the last docstring applied on the method, which is often misleading).

sebalix (commented):
Minor comment, LG

stock_release_channel/models/stock_move.py
```diff
@@ -25,3 +25,13 @@ def _action_confirm(self, merge=True, merge_into=False):
...
+
+    def _release_get_expected_date(self):
+        """Return the new scheduled date of a single delivery move"""
+        channel = self.picking_id.release_channel_id
+        if not channel:
+            return super()._release_get_expected_date()
+        expected_date = channel._get_expected_date()
+        if not expected_date:
+            return super()._release_get_expected_date()
+        return expected_date
```
mt-software-de:
```suggestion
        channel = self.picking_id.release_channel_id
        expected_date = channel and channel._get_expected_date()
        if not expected_date:
            return super()._release_get_expected_date()
        return expected_date
```

stock_release_channel/models/stock_picking.py
```diff
@@ -15,8 +15,34 @@ class StockPicking(models.Model):
...
+                continue
+            moves = delivery.move_ids.filtered(
+                lambda m: m.state not in ("cancel", "done")
+            )
+            move_chain_ids = []
+            while moves:
+                move_chain_ids += moves.ids
+                moves = moves.move_orig_ids.filtered(
+                    lambda m: m.state not in ("cancel", "done")
+                )
```
mt-software-de, line 42:
```suggestion
            moves = delivery.move_ids
            move_chain_ids = []
            while moves:
                moves = moves.filtered(
                    lambda m: m.state not in ("cancel", "done")
                )
                if not moves:
                    break
                move_chain_ids += moves.ids
                moves = moves.move_orig_ids
```

mt-software-de (changes requested):
Thanks LG. 
Only some minor remarks.

stock_release_channel/models/stock_picking.py
```diff
@@ -15,8 +15,34 @@ class StockPicking(models.Model):
...
+    def _inverse_release_channel_id(self):
+        """When the release channel is modified, update the expected date
+
+        Set new date on all moves in the chain"""
+        for delivery in self:
+            if not delivery.last_release_date:
+                continue
+            if delivery.picking_type_code != "outgoing":
+                continue
+            channel = delivery.release_channel_id
```
santostelmo, line 30:
```suggestion
            channel = delivery.release_channel_id
            if not channel:
                continue
```
Here it is mandatory to have a channel assigned. `_get_expected_date` expects a record

santostelmo (changes requested):

stock_release_channel/models/stock_picking.py
jbaudoux: it's longer and more complex.... ;)

stock_release_channel/models/stock_move.py
jbaudoux: changed

stock_release_channel/models/stock_picking.py
jbaudoux: fixed

lmignon (approved):
LGTM (Code review)

mt-software-de (approved):

PR #862, opened 2024-03-08
[16.0] copier repo update

nilshamerlinck:
to avoid the discrepancy of maintainer-tools version between:
- the pre-commit hook (old)
- the oca bot (up-to-date)

which leads to pre-commit re-adding the xml header in `static/description.index.html` (while we don't want it anymore, see [here](https://github.com/OCA/maintainer-tools/pull/596))


StefanRijnhart:
Obsoleted by https://github.com/OCA/wms/commit/ca437abf70940c5ddd90856bd5d5a7a191b8c975. Thanks though!

PR #819, opened 2024-01-21
[16.0][ADD] stock_release_channel_shipment_advice_deliver

jbaudoux:
This module adds an action to the release channel to automate the delivery of its shippings through shipment advices.

Depends on:
* https://github.com/OCA/stock-logistics-transport/pull/114

@sbejaoui I refactored a bit the code to drop the dependency on:
* stock_picking_started : stock_available_to_promise_release already does the check on printed
* on propagate release channel : picking_chain_ids is more accurate https://github.com/OCA/wms/blob/16.0/stock_release_channel/models/stock_release_channel.py#L125
* process end date: I could not figure out why this was a dependency
* stock_picking_backorder_reason needs to be updated in the button_validate

cc @sebalix @simahawk @lmignon

PR #864, opened 2024-03-14
[IMP] stock_available_to_promise_release: set move to need release after pick cancel

sbejaoui:
if we manually cancel one of picking chain we set the dest moves to need_release so they can be released again

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -482,6 +482,15 @@ class StockMove(models.Model):
     def _action_cancel(self):
         # Unrelease moves that must be, before canceling them.
         self.unrelease()
+        # if we manually cancel one of picking chain we set the dest moves
+        # to need_release so they can be released again
+        if not self.env.context.get("cancel_due_to_unrelease"):
+            for dest_moves in self._get_chained_moves_iterator("move_dest_ids"):
+                need_release_moves = dest_moves.filtered(
+                    lambda m: m.state not in ("draft", "cancel", "done")
+                    and not m.need_release
+                )
+                need_release_moves.need_release = True
         super()._action_cancel()
         self.write({"need_release": False})
         return True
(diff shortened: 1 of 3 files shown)


lmignon (approved):
Makes sense (Code review only)

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -396,6 +396,15 @@ def _action_cancel(self):
...
         moves_to_unrelease.unrelease()
+        # if we manually cancel one of picking chain we set the dest moves
+        # to need_release so they can be released again
+        if not self.env.context.get("cancel_due_to_unrelease"):
+            for dest_moves in self._get_chained_moves_iterator("move_dest_ids"):
+                need_release_moves = dest_moves.filtered(
+                    lambda m: m.state not in ("draft", "cancel", "done")
+                    and not m.need_release
+                )
+                need_release_moves.need_release = True
```
jbaudoux, line 493: You cannot set the complete move as releasable. Only the canceled quantity can be releasable.
If you have a ship of 10 units. You pick 6 and you cancel the backorder of 4, then you cannot mark the 10 units back as releasable.

You also forgot to manage the case where it's more steps than pick-ship, like pick-pack-ship. Seems you put the pack as need_release. Anyway, that's more complicate than that.

jbaudoux (changes requested):
There is also the case where you don't create a backorder on the pick when you process a partial quantity. In this case, I don't think action_cancel is called. I'll see to solve that one

jbaudoux:
> There is also the case where you don't create a backorder on the pick when you process a partial quantity. In this case, I don't think action_cancel is called. I'll see to solve that one

Fixed here https://github.com/OCA/wms/pull/819#issuecomment-2044713224 in commit https://github.com/OCA/wms/commit/948637a525db6afd193ee365e40ba4fc13a5a16c

jbaudoux:
> > There is also the case where you don't create a backorder on the pick when you process a partial quantity. In this case, I don't think action_cancel is called. I'll see to solve that one
> 
> Fixed here [#819 (comment)](https://github.com/OCA/wms/pull/819#issuecomment-2044713224) in commit [948637a](https://github.com/OCA/wms/commit/948637a525db6afd193ee365e40ba4fc13a5a16c)

And here https://github.com/OCA/wms/pull/905

stock_available_to_promise_release/models/stock_move.py
lmignon: @jbaudoux I'm on it. A first test suggests that this code introduces malicious behavior if a partial quantity is picked and the release process is called again on the ship. The new pick is generated for the original quantity without taking into account what has already been prepared.

PR #865, opened 2024-03-14
[16.0][ADD] stock_available_to_promise_release_block: block release of operations

sebalix:
Aims to replace https://github.com/OCA/wms/pull/847

**TODO:**
- [x] README


mmequignon (approved):
not much to say! :blush:

stock_available_to_promise_release_block/models/stock_move.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _get_release_ready_depends(self):
+        depends = super()._get_release_ready_depends()
+        depends.append("release_blocked")
+        return depends
+
+    def _is_release_ready(self):
+        # Override to not release moves that are blocked
+        is_release_ready = super()._is_release_ready()
+        return is_release_ready and not self.release_blocked
```
jbaudoux: I would have first checked self.release_blocked before calling super

stock_available_to_promise_release_block/models/stock_picking.py
```diff
@@ -0,0 +1,72 @@
...
+            rec.block_release_allowed = (
+                release_enabled
+                and not rec.release_blocked
+                and rec.state in ("confirmed", "waiting")
+            )
+
+    @api.depends("move_ids", "move_ids.release_blocked")
+    def _compute_release_blocked(self):
+        for rec in self:
+            rec.release_blocked = all(rec.move_ids.mapped("release_blocked"))
```
jbaudoux:
```suggestion
            shipping_policy = rec._get_shipping_policy()
            if shipping_policy == "one":
                rec.release_blocked = any(rec.move_ids.mapped("release_blocked"))
            else:
                rec.release_blocked = all(rec.move_ids.mapped("release_blocked"))
```

jbaudoux (changes requested):
Just a small remark, otherwise looks good

stock_available_to_promise_release_block/models/stock_picking.py
sebalix: done

stock_available_to_promise_release_block/models/stock_move.py
sebalix: done

stock_available_to_promise_release_block/models/stock_picking.py
```diff
@@ -0,0 +1,76 @@
...
+        compute="_compute_release_blocked",
+        store=True,
+    )
+    release_blocked_label = fields.Char(
+        string="Release Blocked",
+        compute="_compute_release_blocked_label",
+    )
+
+    @api.depends(
+        "move_ids.rule_id.available_to_promise_defer_pull",
```
jbaudoux:
```suggestion
        "need_release",
```

stock_available_to_promise_release_block/models/stock_picking.py
```diff
@@ -0,0 +1,76 @@
...
+        "release_blocked",
+        "state",
+    )
+    def _compute_block_release_allowed(self):
+        for rec in self:
+            rules = rec.move_ids.rule_id
+            release_enabled = any(rules.mapped("available_to_promise_defer_pull"))
+            rec.block_release_allowed = (
+                release_enabled
```
jbaudoux:
```suggestion
            rec.block_release_allowed = (
                rec.need_release
```

jbaudoux (changes requested):
This should prevent to block a delivery if all has been released already

stock_available_to_promise_release_block/models/stock_move.py
```diff
@@ -0,0 +1,37 @@
...
+        # Override to not release moves that are blocked
+        if not self.release_blocked:
+            return super()._is_release_ready()
+        return False
+
+    def _blocked_on_backorder(self):
+        return float_is_zero(
+            self.ordered_available_to_promise_qty,
+            precision_rounding=self.product_uom.rounding,
+        )
```
jbaudoux: I would always return True
If you decide to not release everything, the backorder should anyway be blocked in all cases
I would still keep the method as it could be inherited and modified for MTO products

jbaudoux (changes requested):

stock_available_to_promise_release_block/models/stock_move.py
sebalix: So no need for an extra option for that. By default we block the backorders.

jbaudoux: yep

jbaudoux (changes requested):
Can you also add the blocked info in the transfer allocation list view ?

From a OUT, when we click on the release ready smartbutton, we see the status of each move. We should also see here which one is blocked

sebalix:
@jbaudoux all points have been addressed

jbaudoux (approved):

jbaudoux:
@rousseldenis @lmignon Module has been ported to OCA and refactored to be more generic in the handling of what is blocked and manual blocking/unblocking of move or picking.
This module https://github.com/OCA/wms/pull/867 adds the handling from SO
What has not been ported is the automatic unblocking when new moves are added but this can be added on top of this module and `stock_picking_group_by_partner_by_carrier`

lmignon (approved):
great @sebalix LGTM (Code review only)

PR #873, opened 2024-03-19
[14.0][FIX] s_a_promise_release: Reset last_release_date on unrelease

TDu:
When a transfer has been unreleased the `last_release_date` is not reset.
This breaks the counters and the list of records (To Release, To Do) on the related channel.

Now with this change the field name/label is incoherent...

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -635,6 +635,7 @@ class StockMove(models.Model):
                 move_names=move_names,
             )
             picking.message_post(body=body)
+            picking.last_release_date = False
 
     def _split_origins(self, origins):
         """Split the origins of the move according to the quantity into the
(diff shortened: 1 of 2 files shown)


jbaudoux (approved):

jbaudoux:
cc @rousseldenis @lmignon

jbaudoux:
@TDu Can you fwd port to 16.0 ?

rousseldenis:
@TDu A little test should be welcome

mmequignon (approved):

stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -192,3 +192,6 @@ def unrelease(self, safe_unrelease=False):
         processing has already started will be ignored
         """
         self.mapped("move_lines").unrelease(safe_unrelease=safe_unrelease)
+        self.filtered(lambda pick: pick.need_release).update(
+            {"last_release_date": False}
+        )
```
jbaudoux: This should be on the move otherwise if you unrelease the moves, this won't be triggered.

jbaudoux (changes requested):

stock_available_to_promise_release/models/stock_picking.py
TDu: I have updated the fix

mt-software-de (approved):

rousseldenis:
@TDu Could you rebase ?

jbaudoux (approved):

PR #882, opened 2024-04-03
[14.0][FIX] shopfloor_reception: Ensure state assign of remaining move

mt-software-de:
If there is not all processed for one move, we have to ensure that the state of the remaining move is still `assigned`.
Otherwise the last remaining qty of this move will be not split into a own transfer.

https://github.com/OCA/wms/blob/c9cf770e08eae63574f7dd7c5dcbc238e6e00fe0/shopfloor_reception/services/reception.py#L1297
https://github.com/OCA/wms/blob/c9cf770e08eae63574f7dd7c5dcbc238e6e00fe0/shopfloor/models/stock_move.py#L100

cc @jbaudoux @TDu @mmequignon

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1305,7 +1305,6 @@ class Reception(Component):
         new_move = move.create(split_move_vals)
         new_move.move_line_ids = selected_line
         new_move._action_confirm(merge=False)
-        new_move._recompute_state()
         new_move._action_assign()
         # Set back the quantity to do on one of the lines
         line = fields.first(
@@ -1318,7 +1317,8 @@ class Reception(Component):
                 move.product_uom_qty, line[0].product_uom_id
             )
             line.product_uom_qty = move_quantity
-        move._recompute_state()
+        else:
+            move._action_assign()
         new_move.extract_and_action_done()
 
     def set_destination(
--- a/shopfloor_reception/tests/test_set_destination.py
+++ b/shopfloor_reception/tests/test_set_destination.py
@@ -199,3 +199,42 @@ class TestSetDestination(CommonCase):
         self.assertEqual(picking.state, "assigned")
         # So the quantity left to do on the current move has decreased
         self.assertEqual(move.product_uom_qty, 9)
+
+    def test_auto_posting_partial_new_pack(self):
+        self.menu.sudo().auto_post_line = True
+        picking = self._create_picking()
+        selected_move_line = picking.move_line_ids.filtered(
+            lambda l: l.product_id == self.product_a
+        )
+        package_name = "FooBar"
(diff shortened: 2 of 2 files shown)


mt-software-de:
@TDu @mmequignon @jbaudoux 
why did we choose to call the recompute_state?
I think that's not necessary since we calling action_assign immediately afterwards.
https://github.com/OCA/wms/blob/c9cf770e08eae63574f7dd7c5dcbc238e6e00fe0/shopfloor_reception/services/reception.py#L1303

mt-software-de:
ping @jbaudoux @TDu

jbaudoux (commented):
It looks good according to the test but we need to be careful because all tests seem to rely on the fact that move lines are created by the picking type configuration.
In our case, the picking type is configured to not create move lines. The main difference is that the move line is created by the user and the quantity todo is 0. This could have an impact on partial processing.
Can you extend the tests to cover our real usage of the module ?

jbaudoux:
Cc @rousseldenis

mt-software-de:
> It looks good according to the test but we need to be careful because all tests seem to rely on the fact that move lines are created by the picking type configuration.
> In our case, the picking type is configured to not create move lines. The main difference is that the move line is created by the user and the quantity todo is 0. This could have an impact on partial processing.
> Can you extend the tests to cover our real usage of the module ?

I try to extend the Test this week. Then it could be merged and Forward ported.

mt-software-de:
> It looks good according to the test but we need to be careful because all tests seem to rely on the fact that move lines are created by the picking type configuration. In our case, the picking type is configured to not create move lines. The main difference is that the move line is created by the user and the quantity todo is 0. This could have an impact on partial processing. Can you extend the tests to cover our real usage of the module ?

I checked it again and i am now a little confused. I think the test reflects exactly what's happening. A purchase is created and confirmed, which creates a transfer with moves and move lines. The move line is then only partially processed, which creates a new completed move. The old move must then reflect the remaining quantity with the state `assigned`. Or am i wrong?

jbaudoux:
> > It looks good according to the test but we need to be careful because all tests seem to rely on the fact that move lines are created by the picking type configuration. In our case, the picking type is configured to not create move lines. The main difference is that the move line is created by the user and the quantity todo is 0. This could have an impact on partial processing. Can you extend the tests to cover our real usage of the module ?
> 
> I checked it again and i am now a little confused. I think the test reflects exactly what's happening. A purchase is created and confirmed, which creates a transfer with moves and move lines. The move line is then only partially processed, which creates a new completed move. The old move must then reflect the remaining quantity with the state `assigned`. Or am i wrong?

@mt-software-de The question was what will happen if the picking type is configured to not create move lines which is your use case. As there are no tests with this configuration, you better ensure there are no regression

PR #888, opened 2024-04-10
[IMP] stock_available_to_promise_release: Improve migration/init script to initialize 'release_policy' field with default values

rousseldenis:
The upgrade script should initialize the value with the sql default instead updating every existing record (maybe billions).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+# generated from manifests external_dependencies
+openupgradelib
--- a/stock_available_to_promise_release/__init__.py
+++ b/stock_available_to_promise_release/__init__.py
@@ -1,3 +1,2 @@
-from . import models
-from . import wizards
-from .hooks import pre_init_hook
+from . import models, wizards
+from .hooks import init_release_policy, pre_init_hook
--- a/stock_available_to_promise_release/__manifest__.py
+++ b/stock_available_to_promise_release/__manifest__.py
@@ -25,4 +25,5 @@
     "application": False,
     "development_status": "Beta",
     "pre_init_hook": "pre_init_hook",
+    "external_dependencies": {"python": ["openupgradelib"]},
 }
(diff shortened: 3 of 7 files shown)


rousseldenis:
@jbaudoux @mt-software-de

rousseldenis:
This is the magic:

image

mt-software-de:
Many thanks yes this was missing. 
Why did not set the release_policy as it is done here?
https://github.com/OCA/wms/blob/111d87780457fa740dc0cc5609c47ab26be5916d/stock_available_to_promise_release/migrations/16.0.3.1.0/post-migrate.py#L5

rousseldenis:
> Many thanks yes this was missing. Why did not set the release_policy as it is done here?
> 
> https://github.com/OCA/wms/blob/111d87780457fa740dc0cc5609c47ab26be5916d/stock_available_to_promise_release/migrations/16.0.3.1.0/post-migrate.py#L5

Here, I initialize the field with the default value, so, it should be done in pre script.

I've put it also in pre-init as the module can be installed on huge database too

mt-software-de:
> > Many thanks yes this was missing. Why did not set the release_policy as it is done here?
> > https://github.com/OCA/wms/blob/111d87780457fa740dc0cc5609c47ab26be5916d/stock_available_to_promise_release/migrations/16.0.3.1.0/post-migrate.py#L5
> 
> Here, I initialize the field with the default value, so, it should be done in pre script.
> 
> I've put it also in pre-init as the module can be installed on huge database too


> > Many thanks yes this was missing. Why did not set the release_policy as it is done here?
> > https://github.com/OCA/wms/blob/111d87780457fa740dc0cc5609c47ab26be5916d/stock_available_to_promise_release/migrations/16.0.3.1.0/post-migrate.py#L5
> 
> Here, I initialize the field with the default value, so, it should be done in pre script.
> 
> I've put it also in pre-init as the module can be installed on huge database too

Yes i have seen this, but why we don't remove the post_migration
and moving the post_migration part into the new init method?

jbaudoux (approved):

jbaudoux:
replaced by https://github.com/OCA/wms/pull/931

PR #890, opened 2024-04-10
[16.0][FIX] stock_available_to_promise_release: Don't post message if no picking on unrelease

rousseldenis:


stock_available_to_promise_release/tests/common.py
```diff
@@ -55,7 +55,7 @@ def _create_picking_chain(cls, wh, products=None, date=None, move_type="direct")
         if products is None:
             products = []
 
-        group = cls.env["procurement.group"].create(
+        cls.group = cls.env["procurement.group"].create(
```
sebalix: Not sure having this assignation at the class level inside a method which can be called on demand is a good thing.

stock_available_to_promise_release/tests/test_unrelease.py
```diff
@@ -162,3 +162,26 @@ def test_auto_unrelease_on_backorder(self):
         self.assertTrue(
             all(m.procure_method == "make_to_order" for m in backorder_ship.move_ids)
         )
+
+    def test_cancel_procurement(self):
+        """
+        Simulate a quantity cancel
+        """
+        values = {
+            "group_id": self.group,
```
sebalix, line 171: Follow-up of my comment above, if you need the procurement group, why not declaring it in `setUpClass` after the call of `_create_picking_chain`:
```python
cls.shipping = cls._out_picking(
    cls._create_picking_chain(
        cls.wh, [(cls.product1, 5)], date=datetime(2019, 9, 2, 16, 0)
    )
)
cls.group = cls.shipping.group_id
```

rousseldenis:
@sebalix @jbaudoux

stock_available_to_promise_release/tests/test_unrelease.py
sebalix: @rousseldenis with my previous comment I wanted to highlight that only `setup` and `setupClass` methods should be allowed to assign names on `self/cls`, while helper methods like `_create_picking_chain` or `_create_procurement_group` should only returns a result without updating the "environment" outside of the method like `cls.group` here (which could trigger side-effects if this `cls.group` is overridden by calling a second time `_create_picking_chain`).

This follows a bit the pure/impure functions borrowed from functional programming, but for the user this also respect the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

I won't block on this, it can be refactored later if needed!

sebalix (approved):

jbaudoux (approved):

jbaudoux (approved):

PR #892, opened 2024-04-19
[ADD] sale_shopfloor_reception_refund_return

vnikolayev1:
- Added module for sale order that marks created return as to refund, sets sale order line.


vnikolayev1:
@mt-software-de

PR #896, opened 2024-04-30
[16.0][IMP] sale_stock_available_to_promise_release: show/hide availability on SO reports

victorvermot:
Add parameter to make the column Available on SO reports invisible

Results
- Setting enabled
OCAAB
- With setting activated
OCAAB1
- Setting disabled
OCAAB2

### DIFF
--- a/sale_stock_available_to_promise_release/models/res_company.py
+++ b/sale_stock_available_to_promise_release/models/res_company.py
@@ -1,0 +1,13 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    stock_sale_report_display_is_available = fields.Boolean(
+        help="Show Available column in SO and quotation reports",
+        default=True,
+    )
--- a/sale_stock_available_to_promise_release/models/res_config_settings.py
+++ b/sale_stock_available_to_promise_release/models/res_config_settings.py
@@ -1,0 +1,13 @@
+# Copyright 2023 Trobz
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    stock_sale_report_display_is_available = fields.Boolean(
+        related="company_id.stock_sale_report_display_is_available",
+        readonly=False,
+    )
--- a/sale_stock_available_to_promise_release/reports/sale_order.xml
+++ b/sale_stock_available_to_promise_release/reports/sale_order.xml
@@ -9,7 +9,7 @@
     >
       <xpath expr="//th[@name='th_description']" position="after">
         <th
-                t-if="doc.state != 'draft'"
+                t-if="doc.state != 'draft' and doc.company_id.stock_sale_report_display_is_available"
                 name="th_is_available"
                 class="text-right nowrap"
             >Available</th>
@@ -17,7 +17,7 @@
 
       <xpath expr="//td[@name='td_name']" position="after">
           <td
-                t-if="doc.state != 'draft'"
+                t-if="doc.state != 'draft' and doc.company_id.stock_sale_report_display_is_available"
                 name="td_is_available"
                 class="nowrap text-right"
             >
--- a/sale_stock_available_to_promise_release/views/res_config_settings.xml
+++ b/sale_stock_available_to_promise_release/views/res_config_settings.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="res_config_settings_view" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="stock.res_config_settings_view_form" />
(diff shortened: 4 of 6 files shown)


sale_stock_available_to_promise_release/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    display_is_available = fields.Boolean(
```
santostelmo: IMO This should be true by default and related to the company. You maybe want to show this field only in some companies

sale_stock_available_to_promise_release/reports/sale_order.xml
```diff
@@ -9,15 +9,15 @@
     >
       <xpath expr="//th[@name='th_description']" position="after">
         <th
-                t-if="doc.state != 'draft'"
+                t-if="doc.state != 'draft' and doc.env['ir.config_parameter'].sudo().get_param('sale_stock_available_to_promise_release.display_is_available')"
```
santostelmo:
```suggestion
                t-if="doc.state != 'draft' and doc.company_id.display_is_available"
```
As mentioned above if you configure by company then you can do something like that.

sale_stock_available_to_promise_release/models/res_config_settings.py
```diff
@@ -0,0 +1,15 @@
...
+        related="company_id.display_is_available",
+        help="Show Available column in SO and quotation reports",
+        readonly=False,
+        config_parameter="sale_stock_available_to_promise_release.display_is_available",
```
santostelmo: Do you need config parameter ?

cyrilmanuel (approved):

santostelmo (approved):

sale_stock_available_to_promise_release/models/res_config_settings.py
victorvermot: you're right I removed it

leemannd (approved):

sale_stock_available_to_promise_release/models/res_company.py
```diff
@@ -0,0 +1,13 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    display_is_available = fields.Boolean(
```
sebalix: The name of this field is too generic for `res.company` and could conflict with another field name easily with all modules installed on an Odoo instance.
What about `sale_stock_report_display_is_available` or `stock_sale_report_display_is_available` (as all other settings are prefixed with `stock_` in the parent module).

sale_stock_available_to_promise_release/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    display_is_available = fields.Boolean(
+        related="company_id.display_is_available",
+        help="Show Available column in SO and quotation reports",
```
sebalix: I think you can remove the `help` here, it'll come from the original field automatically.

sale_stock_available_to_promise_release/views/res_config_settings.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="res_config_settings_view" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field
+            name="inherit_id"
+            ref="stock.res_config_settings_view_form"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='annual_inventory_date']" position="after">
```
sebalix: There is already a section introduced by the parent module to host settings related to the release features: https://github.com/OCA/wms/blob/16.0/stock_available_to_promise_release/views/res_config_settings.xml#L13C64-L13C81
Better to place the new setting in this section.

sebalix (changes requested):

sebalix (approved):

jbaudoux (approved):

sale_stock_available_to_promise_release/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    stock_sale_report_display_is_available = fields.Boolean(
+        help="Show Available column in SO and quotation reports",
+        default=True,
```
jbaudoux, line 12: Shouldn't this be `False` by default ?

cc @sebalix

sebalix: Before this PR the column was always displayed, we are only adding an option here to hide it, but by default we keep it displayed as before. No?

jbaudoux: all good then :)

jbaudoux:
@victorvermot Can you drop `[16.0][ADD]: ` from the commit message ?

PR #897, opened 2024-04-30
[16.0][IMP] sale_stock_available_to_promise_release_block block multiple order lines

santostelmo:
When we block multiple order lines we get the following traceback:

```
ValueError: <class 'ValueError'>: "Expected singleton: stock.move(53171, 53172)" while evaluating
'if records:\n        records.move_ids.action_block_release()'
```

### DIFF
--- a/sale_stock_available_to_promise_release_block/views/sale_order_line.xml
+++ b/sale_stock_available_to_promise_release_block/views/sale_order_line.xml
@@ -42,7 +42,8 @@
     <field name="state">code</field>
     <field name="code">
     if records:
-        records.move_ids.action_unblock_release()
+      for move in records.move_ids:
+        move.action_unblock_release()
     </field>
   </record>
 
@@ -54,7 +55,8 @@
     <field name="state">code</field>
     <field name="code">
     if records:
-        records.move_ids.action_block_release()
+      for move in records.move_ids:
+        move.action_block_release()
     </field>
   </record>
 


cyrilmanuel (approved):

sale_stock_available_to_promise_release_block/views/sale_order_line.xml
```diff
@@ -42,7 +42,8 @@
     <field name="state">code</field>
     <field name="code">
     if records:
-        records.move_ids.action_unblock_release()
+      for move in records.move_ids:
+        move.action_unblock_release()
```
sebalix, line 46: No need to change this, the `action_unblock_release` has no singleton issue, see:
https://github.com/OCA/wms/blob/16.0/stock_available_to_promise_release_block/models/stock_move.py#L42-L44

sale_stock_available_to_promise_release_block/views/sale_order_line.xml
```diff
@@ -54,7 +55,8 @@
     <field name="state">code</field>
     <field name="code">
     if records:
-        records.move_ids.action_block_release()
+      for move in records.move_ids:
+        move.action_block_release()
```
sebalix, line 59: Why not changing instead the implementation of `action_block_release` in the parent module instead?
Here https://github.com/OCA/wms/blob/16.0/stock_available_to_promise_release_block/models/stock_move.py#L37-L40
To update with:
```python
    def action_block_release(self):
        """Block the release."""
        for move in self:
            if move.need_release:
                move.release_blocked = True
```

santostelmo: It's also an option

sale_stock_available_to_promise_release_block/views/sale_order_line.xml
santostelmo: What I see in your link is 
```
    def action_unblock_release(self):
        """Unblock the release."""
        self.release_blocked = False
```
We can't unblock multiple records. Right ?

sale_stock_available_to_promise_release_block/views/sale_order_line.xml
santostelmo: The PR in the module `stock_available_to_promise` https://github.com/OCA/wms/pull/898

santostelmo:
That was solved hre: https://github.com/OCA/wms/pull/898

PR #898, opened 2024-05-03
[16.0][FIX] stock_available_to_promise_release_block multi block

santostelmo:
Allow to block multiple stock moves

### DIFF
--- a/stock_available_to_promise_release_block/models/stock_move.py
+++ b/stock_available_to_promise_release_block/models/stock_move.py
@@ -36,8 +36,9 @@ class StockMove(models.Model):
 
     def action_block_release(self):
         """Block the release."""
-        if self.need_release:
-            self.release_blocked = True
+        for move in self:
+            if move.need_release:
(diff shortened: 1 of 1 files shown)


stock_available_to_promise_release_block/models/stock_move.py
```diff
@@ -36,9 +36,11 @@ def _blocked_on_backorder(self):
...
-            self.release_blocked = True
+        for move in self:
+            if move.need_release:
+                move.release_blocked = True
 
     def action_unblock_release(self):
         """Unblock the release."""
-        self.release_blocked = False
+        for move in self:
+            move.release_blocked = False
```
sebalix: No need to change this implementation, `records.field = value` already set `value` on `field` for all `records`.
The singleton error appears only when you try to access a non-relational field from a collection of records (not when writing them), e.g. `if records.name`

sebalix (approved):

PR #902, opened 2024-05-17
[16.0][IMP] shopfloor: Zone picking no destination package required

lmignon:
Add an option to skip the packing step when processing a line into the zone picking scenario. Thanks to this option the processing of a line can be done by processing the location, set the qty and scan the destination even if the product is not into a package.

* [x] includes #900


shopfloor/data/shopfloor_scenario_data.xml
```diff
@@ -20,7 +20,9 @@
     "unload_package_at_destination": true,
     "multiple_move_single_pack": true,
     "no_prefill_qty": true,
-    "scan_location_or_pack_first": true
+    "scan_location_or_pack_first": true,
+    "allow_alternative_destination_package": true,
+    "no_packing": true
```
jbaudoux: I don't like negative options. And as we already have `destination_package` (in fact there is already 2 names for the same thing: pack & package), I would prefer
```suggestion
    "require_destination_package": true
```

jbaudoux (changes requested):

lmignon:
@jbaudoux Code changed

rousseldenis:
@jbaudoux

jbaudoux (approved):

lmignon:
@jbaudoux rebased to fix conflicts. Could be merged....

jbaudoux:
@lmignon there is a small pre-commit issue

lmignon:
@jbaudoux pre-commit is now :green_circle: :smirk:

jbaudoux:
cc @sebalix @simahawk @TDu

TDu (approved):

rousseldenis (approved):

PR #903, opened 2024-05-22
[16.0][IMP] sale_stock_available_to_promise_release mto availability_status

santostelmo:
When stock becomes available for `is_mto` product,, sale order line availability status should reflect that status.
Currently nothing happens in that circumstances as the status always stays in ` on_order` even if there's stock available.

Also standard is already providing a computed field `is_mto` => `_on_order_route` can be dropped.

### DIFF
--- a/sale_stock_available_to_promise_release/models/sale_order_line.py
+++ b/sale_stock_available_to_promise_release/models/sale_order_line.py
@@ -26,13 +26,6 @@ class SaleOrderLine(models.Model):
         digits="Product Unit of Measure", compute="_compute_availability_status"
     )
 
-    def _on_order_route(self):
-        self.ensure_one()
-        mto_route = self.env.ref("stock.route_warehouse0_mto")
-        product_is_mto = mto_route in self.product_id.route_ids
-        line_is_mto = mto_route == self.route_id
-        return product_is_mto or line_is_mto
-
     @api.depends(
         "move_ids.ordered_available_to_promise_uom_qty",
         "product_id.route_ids",
@@ -73,11 +66,8 @@ class SaleOrderLine(models.Model):
         # required values
         product = self.product_id
         rounding = product.uom_id.rounding
-        # on_order product
-        if self._on_order_route():
-            availability_status = "on_order"
         # Fully available
-        elif (
+        if (
             product.type == "service"
             or float_compare(
                 available_qty, self.product_uom_qty, precision_rounding=rounding
@@ -90,6 +80,9 @@ class SaleOrderLine(models.Model):
         elif float_compare(available_qty, 0, precision_rounding=rounding) == 1:
             availability_status = "partial"
             delayed_qty = self.product_uom_qty - available_qty
+        # On order product
+        elif self.is_mto:
+            availability_status = "on_order"
         # No stock
         elif float_is_zero(available_qty, precision_rounding=rounding):
             product_replenishment_date = product._get_next_replenishment_date()


sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -90,6 +80,9 @@ def _get_availability_data(self):
         elif float_compare(available_qty, 0, precision_rounding=rounding) == 1:
             availability_status = "partial"
             delayed_qty = self.product_uom_qty - available_qty
+        # On order product
+        elif self.is_mto:
+            availability_status = "on_order"
         # No stock
         elif float_is_zero(available_qty, precision_rounding=rounding):
```
jbaudoux, line 87:
```suggestion
        else:
```

sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -90,6 +80,9 @@ def _get_availability_data(self):
         elif float_compare(available_qty, 0, precision_rounding=rounding) == 1:
             availability_status = "partial"
             delayed_qty = self.product_uom_qty - available_qty
+        # On order product
+        elif self.is_mto:
+            availability_status = "on_order"
```
jbaudoux, line 85:
```suggestion
```

sale_stock_available_to_promise_release/models/sale_order_line.py
```diff
@@ -98,6 +91,7 @@ def _get_availability_data(self):
                 availability_status = "restock"
                 expected_availability_date = product_replenishment_date
                 delayed_qty = self.product_uom_qty
+                # on_order product
```
jbaudoux:
```suggestion
            # On order product
            elif self.is_mto:
                availability_status = "on_order"
```

jbaudoux (changes requested):
I would place it even after "restock"

santostelmo:
> I would place it even after "restock"

But then if you have a unavailable mto product, you end up with restock status. It should be `on_order` in this case no ?

sale_stock_available_to_promise_release/models/sale_order_line.py
jbaudoux: drop this comment

jbaudoux:
> > I would place it even after "restock"
> 
> But then if you have a unavailable mto product, you end up with restock status. It should be `on_order` in this case no ?

You're right. Then drop the misplaced comment :)

sale_stock_available_to_promise_release/models/sale_order_line.py
santostelmo: done

jbaudoux (approved):

jbaudoux:
@sebalix @mmequignon We did this change. Ok for you ?

sale_stock_available_to_promise_release/models/sale_order_line.py
mmequignon, line 85: shouldn't this be checked first ?
We might have MTO sale order lines availability status set to partial here.

santostelmo: It depends what we want to show to the user first. I would show 1st partial (we are able to deliver some qty), 2nd on_order ( we have no quantity available and need to fully order it)

mmequignon: If mto, it's on order, period, whether you have stock or not.

mmequignon: If user wants smth else, he changes the route on the so line, and that's it.

santostelmo: Sale person wants to know if product is available to ship or if action is required to order it. Otherwise this status is useless for him.

mmequignon: If you want to display availability, then product shouldn't be MTO.
@jbaudoux @santostelmo can we discuss this ? I'm not sure to understand what we're solving here...

jbaudoux: @mmequignon For standard MTO products, it's on order when SO is draft but the status will update when the product is received and available.
Also, in case of MTO as MTS (MTO route that generates an orderpoint with min/max=0), then if you have stock because someone returned an MTO product, then you see if it is available when placing the SO

jbaudoux:
@mmequignon good for you?

cyrilmanuel (approved):

sebalix (approved):

PR #904, opened 2024-05-23
[16.0][FIX] stock_available_to_promise_release : cancellation must unrelease

jbaudoux:
Before the change, when you cancel a delivery move, if it cannot be unreleased, it was skipping the unrelease but still canceling the move. You then end-up with a canceled OUT but with a PICK still in to do.

I also improved the error message to show all forbidden deliveries and their list of related blocking pickings (no matter the amount of steps) 

cc @TDu @sebalix @rousseldenis @lmignon

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -112,25 +112,38 @@ class StockMove(models.Model):
         )
 
     def _check_unrelease_allowed(self):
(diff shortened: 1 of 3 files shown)


rousseldenis (approved):
LGTM

lmignon (approved):
Thank you for the fix.  LGTM (Code review)

santostelmo (approved):

PR #908, opened 2024-06-05
[IMP] stock_release_channel_propagate_channel_picking: Propagate when assigning too

rousseldenis:
As internal pickings could have generated a backorder and outgoing picking too without unrelease (e.g.: long term channels - no sleep), the channel is void on those internal backorders. So, at outgoing picking channel assignation, propagate the channel to the still existing origin pickings.


sebalix (approved):
LG

lmignon (approved):
LGTM (Code review)

jbaudoux (commented):
Can you drop changes to OCA generated files?

rousseldenis:
> Can you drop changes to OCA generated files?

Will be hard. But that does not bother as these changes reflect also changes of the module.

jbaudoux:
> > Can you drop changes to OCA generated files?
> 
> Will be hard. But that does not bother as these changes reflect also changes of the module.

@rousseldenis It's annoying when porting the commits to other version

stock_release_channel_propagate_channel_picking/models/stock_move.py
```diff
@@ -13,6 +13,19 @@ def _after_release_update_chain(self):
...
+    def move_origins_to_propagate_release_channel(self):
+        """
+        Get the origin moves for those release channel should be updated:
+
+        - Don't take into account 'cancel' and 'done' moves
+        - Don't take into account moves' picking that already have a release channel
+        """
+        return self.mapped("move_orig_ids").filtered(
+            lambda m: m.state not in ("cancel", "done")
+            and not m.picking_id.release_channel_id
```
jbaudoux, line 26: Why? If you modify the channel on the ougoing move, you should still update the pickings.
I would exclude the outgoing moves (in case of inter-wh flow)

jbaudoux (changes requested):

PR #909, opened 2024-06-13
[IMP] shopfloor, zone picking: Allow partial transfer when scanning a location

sbejaoui:
In the zone picking scenario, operators can choose to pick less than the reserved quantity. Previously, the system required them, in this case, to place the goods in a package first.

This restriction was unnecessary, as it forced operators to perform an extra step that they do not actually do in real life. We trust them to know what they're doing, so let's keep it simple.

With this fix, if a user scans a location, they can move the products directly to the scanned location.

### DIFF
--- a/shopfloor/services/zone_picking.py
+++ b/shopfloor/services/zone_picking.py
@@ -1169,43 +1169,40 @@ class ZonePicking(Component):
             )
 
         extra_message = ""
-        if moving_full_quantity:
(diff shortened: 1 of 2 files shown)


lmignon (changes requested):
@sbejaoui you should modify the PR https://github.com/OCA/wms/pull/902 where the option to control us a destination package is required or not has been introduced.

PR #912, opened 2024-07-01
[16.0] shopfloor cluster_picking: Custom sorting by device

lmignon:
With this change, in the cluster picking scenario,  you can now provide a custom code on a stock device that will be used to order the operation when this device is used


rousseldenis (approved):
LGTM

jbaudoux (approved):

PR #917, opened 2024-07-05
[16.0] shopfloor: location_content_transfer: get_work additional domain and sort key.

lmignon:
The initial aim of this PR is to enable the 'location_content_transfer' scenario to be configured more precisely. 
It aims to:

* allow additional criteria to be configured to be applied when the user searches for work to be done (example of use: a menu for processing urgent transfers only)
* allow you to configure a different sort to be applied to the operations to be performed than the one based on priority. (by location, for example)
 
When analysing the code, I found that there was an inconsistency between the code used to create the menu counters and the one used to find the transactions in the 'location_content_transfer' scenario. The code in the latter scenario was functionally identical to the one used for menus and zone_picking. Harmonisation work has therefore been carried out to eliminate this inconsistency.

As result, a generic solution has been put in place for scenarios using the 'search_move_line' action to find the operations to be carried out, so that the menus can be used to configure an additional domain to be applied by default and the method for sorting operations.

### DIFF
--- a/shopfloor/__manifest__.py
+++ b/shopfloor/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Shopfloor",
     "summary": "manage warehouse operations with barcode scanners",
-    "version": "16.0.2.2.0",
+    "version": "16.0.2.2.1",
     "development_status": "Beta",
     "category": "Inventory",
     "website": "https://github.com/OCA/wms",
--- a/shopfloor/actions/move_line_search.py
+++ b/shopfloor/actions/move_line_search.py
@@ -20,9 +20,9 @@ class MoveLineSearch(Component):
             self.work, "picking_types", self.env["stock.picking.type"].browse()
         )
 
-    def _search_move_lines_by_location_domain(
+    def _search_move_lines_domain(
         self,
-        locations,
+        locations=None,
         picking_type=None,
         package=None,
         product=None,
@@ -31,17 +31,22 @@ class MoveLineSearch(Component):
         picking_ready=True,
         # When True, adds the package in the domain even if the package is False
         enforce_empty_package=False,
+        enforce_picking_types=True,
     ):
         domain = [
-            ("location_id", "child_of", locations.ids),
             ("qty_done", "=", 0),
             ("state", "in", ("assigned", "partially_available")),
         ]
-        if picking_type:
-            # auto_join in place for this field
-            domain += [("picking_id.picking_type_id", "=", picking_type.id)]
-        elif self.picking_types:
+        picking_types = (
+            picking_type or self.picking_types
+            if enforce_picking_types
+            else picking_type
+        )
+        if picking_types:
             domain += [("picking_id.picking_type_id", "in", self.picking_types.ids)]
+        locations = locations or picking_types.default_location_src_id
+        if locations:
+            domain += [("location_id", "child_of", locations.ids)]
         if package or package is not None and enforce_empty_package:
             domain += [("package_id", "=", package.id if package else False)]
         if product:
@@ -58,9 +63,9 @@ class MoveLineSearch(Component):
             domain += [("picking_id.state", "=", "assigned")]
         return domain
 
-    def search_move_lines_by_location(
+    def search_move_lines(
         self,
-        locations,
+        locations=None,
         picking_type=None,
         package=None,
         product=None,
@@ -70,10 +75,11 @@ class MoveLineSearch(Component):
         sort_keys_func=None,
         picking_ready=True,
         enforce_empty_package=False,
+        enforce_picking_types=True,
     ):
         """Find lines that potentially need work in given locations."""
         move_lines = self.env["stock.move.line"].search(
-            self._search_move_lines_by_location_domain(
+            self._search_move_lines_domain(
                 locations,
                 picking_type,
                 package,
@@ -82,6 +88,7 @@ class MoveLineSearch(Component):
                 match_user=match_user,
                 picking_ready=picking_ready,
                 enforce_empty_package=enforce_empty_package,
+                enforce_picking_types=enforce_picking_types,
             )
         )
         sort_keys_func = sort_keys_func or self._sort_key_move_lines(order)
--- a/shopfloor/actions/move_line_search.py
+++ b/shopfloor/actions/move_line_search.py
@@ -88,25 +88,38 @@ class MoveLineSearch(Component):
         move_lines = move_lines.sorted(sort_keys_func)
         return move_lines
 
-    @staticmethod
-    def _sort_key_move_lines(order):
+    def _sort_key_assigned_to_current_user(self, line):
+        user_id = line.shopfloor_user_id.id or line.picking_id.user_id.id or None
+        # Determine sort priority
+        # Priority 0: Assigned to the current user
+        # Priority 1: Not assigned to any user
+        # Priority 2: Assigned to a different user
+        if user_id == self.env.uid:
+            priority = 0
+        elif user_id is None:
+            priority = 1
+        else:
+            priority = 2
+        return (priority,)
+
+    def _sort_key_move_lines(self, order=None):
         """Return a sorting function to order lines."""
 
         if order == "priority":
             # make prority negative to keep sorting ascending
-            return lambda line: (
+            return lambda line: self._sort_key_assigned_to_current_user(line) + (
                 -int(line.move_id.priority or "0"),
                 line.move_id.date,
                 line.move_id.id,
             )
         elif order == "location":
-            return lambda line: (
+            return lambda line: self._sort_key_assigned_to_current_user(line) + (
                 line.location_id.shopfloor_picking_sequence or "",
                 line.location_id.name,
                 line.move_id.date,
                 line.move_id.id,
             )
-        return lambda line: line
+        return self._sort_key_assigned_to_current_user
 
     def counters_for_lines(self, lines):
         # Not using mapped/filtered to support simple lists and generators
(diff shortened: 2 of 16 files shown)


lmignon:
cc @jbaudoux @rousseldenis

lmignon:
This way of doing things is not the most performant one. It requires to browse all the candidate pickings and related stock.move.line for ordering the operations and return the location of the first one into the list. The proposed implementation should not alter the current performance since the initial search was previously done without 'limit' criteria and all the related `stock.move.line` were loaded by the system due to the orm's pre-fetching mechanism. Another approach might have been not to allow a sorting method to be entered in the menu config and to adapt the code to sort via the orm. In our case, we wanted to order based on the source location of the operation. This implies a join between the stock_picking and stock_move_line tables for sorting. Such a join is only possible by manipulating the orm's low-level classes. This approach makes the code more complex, but also offers the advantage of using these low-level classes to obtain the information required in 1 SQL query. If you think it's better to go this way, here's what some of the code might look like https://github.com/acsone/wms/commit/04d37a0ac19f1e06d23285eed50d904fadb4babf

shopfloor/actions/move_line_search.py
```diff
@@ -31,49 +67,58 @@ def _search_move_lines_by_location_domain(
         picking_ready=True,
         # When True, adds the package in the domain even if the package is False
         enforce_empty_package=False,
+        enforce_picking_types=True,
```
jbaudoux: You are not enforcing the picking types in the domain when its empty. That's not inline with `enforce_empty_package`.
I would drop this new argument

shopfloor/actions/move_line_search.py
```diff
@@ -31,49 +67,58 @@ def _search_move_lines_by_location_domain(
...
         ]
-        if picking_type:
-            # auto_join in place for this field
-            domain += [("picking_id.picking_type_id", "=", picking_type.id)]
-        elif self.picking_types:
+        picking_types = (
+            picking_type or self.picking_types
+            if enforce_picking_types
+            else picking_type
+        )
```
jbaudoux:
```suggestion
        picking_types = picking_type or self.picking_types
```

shopfloor/actions/move_line_search.py
```diff
@@ -82,38 +127,80 @@ def search_move_lines_by_location(
...
+        order = order or self.sort_order
         sort_keys_func = sort_keys_func or self._sort_key_move_lines(order)
         move_lines = move_lines.sorted(sort_keys_func)
         return move_lines
 
-    @staticmethod
-    def _sort_key_move_lines(order):
+    def _sort_key_move_lines(self, order=None):
         """Return a sorting function to order lines."""
+        if order == "none":
```
jbaudoux: ?
```suggestion
        if order is None:
```

shopfloor/actions/move_line_search.py
```diff
@@ -82,38 +127,80 @@ def search_move_lines_by_location(
...
+        # Determine sort priority
+        # Priority 0: Assigned to the current user
+        # Priority 1: Not assigned to any user
+        # Priority 2: Assigned to a different user
+        if user_id == self.env.uid:
+            priority = 0
+        elif user_id is None:
+            priority = 1
+        else:
+            priority = 2
```
jbaudoux, line 187:
```suggestion
            priority = -1
        else:
            priority = -2
```

shopfloor/services/location_content_transfer.py
```diff
@@ -252,24 +235,16 @@ def _create_moves_from_location(self, location):
...
-                lambda line: line.qty_done < line.reserved_uom_qty
-            )
-            location = fields.first(move_lines).location_id
-            if location:
-                break
+        move_lines = self.search_move_line.search_move_lines(match_user=True)
+        for move_line in move_lines:
+            if move_line.location_id:
+                return move_line.location_id
         return location
```
jbaudoux: `location_id` is a required field. You can just do
```suggestion
        return fields.first(move_lines).location_id
```

shopfloor/static/description/index.html
```diff
@@ -8,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
```
jbaudoux: Can you skip the generation of auto generated files?

jbaudoux (changes requested):
Some remarks

shopfloor/actions/move_line_search.py
lmignon: We want those assigned to current user first then the one unassigned ...
image

shopfloor/actions/move_line_search.py
lmignon: @jbaudoux This new argument is required to allow to check if some operations are already planned into a location when you ask for work on a location with the location_content_transfer scenario configured. If you filter on the picking type at this stage, you will not get the message 'You cannot move this using this menu.'  but you will receive 'New move lines cannot be assigned: canceled'

jbaudoux: But the naming of the argument is confusing because it looks like the other existing enforce argument but reacts differently

lmignon: I am open to any suggestion to change the name of this parameter :smirk:

rousseldenis:
@jbaudoux

shopfloor/services/zone_picking.py
```diff
@@ -429,7 +429,7 @@ def _find_location_move_lines(
         enforce_empty_package=False,
     ):
         """Find lines that potentially need work in given locations."""
-        return self.search_move_line.search_move_lines_by_location(
+        return self.search_move_line.search_move_lines(
             locations or self.zone_location,
             picking_type=picking_type or self.picking_type,
```
jbaudoux, line 434: Already managed by `search_move_lines`
```suggestion
            picking_type=picking_type,
```

shopfloor/actions/move_line_search.py
```diff
@@ -31,49 +67,58 @@ def _search_move_lines_by_location_domain(
...
-            # auto_join in place for this field
-            domain += [("picking_id.picking_type_id", "=", picking_type.id)]
-        elif self.picking_types:
+        picking_types = (
+            picking_type or self.picking_types
+            if enforce_picking_types
+            else picking_type
+        )
+        if picking_types:
             domain += [("picking_id.picking_type_id", "in", self.picking_types.ids)]
```
jbaudoux:
```suggestion
            domain += [("picking_id.picking_type_id", "in", picking_types.ids)]
```

shopfloor/services/location_content_transfer.py
```diff
@@ -356,7 +328,10 @@ def scan_location(self, barcode):  # noqa: C901
                 message=self.msg_store.cannot_move_something_in_picking_type()
             )
 
-        move_lines = self._find_location_move_lines_from_scan_location(location)
+        move_lines = self.search_move_line.search_move_lines(
+            locations=location, match_user=True, enforce_picking_types=False
```
jbaudoux: We should still respect the picking types of the menu
```suggestion
            locations=location, match_user=True,
```

shopfloor/actions/move_line_search.py
jbaudoux: I still don't understand why we need such argument

shopfloor/actions/move_line_search.py
jbaudoux: I still think this is enough

jbaudoux (changes requested):
I still don't get the purpose of `enforce_picking_types` :/

lmignon:
@jbaudoux If I remove the possibility to search for picking in an other picking type than the one provided by the menu if specified, I'm no more able to support the use case in the location_content_transfer scenario when content has different picking type than menu, and the menu is configured with the option to unreserve. In such a case, content must be unreserved, new moves created and the previous content re-reserved. That why I need this new `enforce_picking_types` parameter to also uses the search method in such a case where I don't want to enforce the picking types to the those configured on the menu or passed to the method.

lmignon:
@jbaudoux Your last comment is now implemented into the last commit.

shopfloor/actions/move_line_search.py
jbaudoux: I think you forgot that one

shopfloor/services/zone_picking.py
jbaudoux: what about this one ?

jbaudoux:
> @jbaudoux Your last comment is now implemented into the last commit.

@lmignon nearly. There are still 2 unresolved comments related to your last changes. Should be obvious

lmignon:
@jbaudoux done

shopfloor/__manifest__.py
```diff
@@ -6,11 +6,11 @@
 {
     "name": "Shopfloor",
     "summary": "manage warehouse operations with barcode scanners",
-    "version": "16.0.2.2.0",
+    "version": "16.0.2.2.1",
     "development_status": "Beta",
     "category": "Inventory",
     "website": "https://github.com/OCA/wms",
-    "author": "Camptocamp, BCIM, Akretion, Odoo Community Association (OCA)",
+    "author": "Camptocamp, BCIM, Akretion, ACSONE SA/NV, Odoo Community Association (OCA)",
```
jbaudoux: Can you drop this change from this PR?

lmignon: I will remove the commit with this change. I'm a little bit surprised that with all the work we do on this addon we can't put ACSONE into the authors list.

lmignon: commit removed

jbaudoux (approved):

PR #925, opened 2024-08-22
[IMP] shopfloor: Improve message for location content transfer items

rousseldenis:
In case of content transfer line by line, user wants to be recalled which origin location he comes from if he wants to transfer partially and put the remaining products back to source (e.g.: no more space on destination location)

@phschmidt @jbaudoux @sbejaoui @lmignon 

An easy one :sweat_smile:

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -637,10 +637,14 @@ class MessageAction(Component):
             "body": _("Transfer {} complete").format(picking.name),
         }
 
-    def location_content_transfer_item_complete(self, location_dest):
+    def location_content_transfer_item_complete(self, location_src, location_dest):
         return {
             "message_type": "success",
-            "body": _("Content transfer to {} completed").format(location_dest.name),
+            "body": _(
+                "Content line transferred from %(location_name)s to %(location_dest_name)s",
+                location_name=location_src.name,
+                location_dest_name=location_dest.name,
+            ),
         }
 
     def location_content_transfer_complete(self, location_src, location_dest):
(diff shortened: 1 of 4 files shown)


jbaudoux (changes requested):
I'm not in favor of this change. The source location is marked as empty when you pick from there. If you partially process a quantity, you will still have to process the remaining quantity to the asked location. If you want to cancel remaining quantity and put back at source, there should be an explicit action with the right steps (and checks)

rousseldenis:
> The source location is marked as empty when you pick from there.

What is your use case where source location is empty???

jbaudoux:
> > The source location is marked as empty when you pick from there.
> 
> What is your use case where source location is empty???

Someone takes from over-stock location N1 and starts unloading to stock location S2. Someone else moves from input location I3 to the empty N1.

rousseldenis:
@jbaudoux This change has nothing to do with the flow itself.

It adds the source location in the message. Nothing more.

shopfloor/actions/message.py
```diff
@@ -581,10 +581,14 @@ def transfer_complete(self, picking):
             "body": _("Transfer {} complete").format(picking.name),
         }
 
-    def location_content_transfer_item_complete(self, location_dest):
+    def location_content_transfer_item_complete(self, location_src, location_dest):
         return {
             "message_type": "success",
-            "body": _("Content transfer to {} completed").format(location_dest.name),
+            "body": _(
+                "Content line transferred from %(location_name)s to %(location_dest_name)s",
```
jbaudoux, line 644: What about
```suggestion
                "Transferred from %(location_name)s to %(location_dest_name)s",
```
or
```suggestion
                "Product %(product_name) transferred from %(location_name)s to %(location_dest_name)s",
```
and
```suggestion
                "Package %(package_name) transferred from %(location_name)s to %(location_dest_name)s",
```

jbaudoux (approved):
I see the location source is displayed when last content is processed in `location_content_transfer_complete`
So it makes sense to display it here also :)

But can you do better than `Content line` ?

sbejaoui (approved):

rousseldenis:
Merging this as very old. Improvement later.

/ocabot merge minor

PR #927, opened 2024-09-05
[14.0][IMP] stock_reception_screen_measuring_device: measure smaller packages

mmequignon:
When goods are received, triggers measurements also for smaller packagings that weren't ordered or received.

This behavior is not triggered when received packagings are not the same as ordered packagings.

### DIFF
--- a/stock_reception_screen_measuring_device/migrations/14.0.2.0.0/post-migration.py
+++ b/stock_reception_screen_measuring_device/migrations/14.0.2.0.0/post-migration.py
@@ -1,0 +1,25 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
+def populate_new_field(env):
+    records_to_update = (
+        env["stock.reception.screen"].search([("current_step", "!=", "done")]).exists()
+    )
+    _logger.info(
+        "Set smaller_package_has_missing_dimensions on ongoing reception screens"
+    )
+    records_to_update._compute_smaller_package_has_missing_dimensions()
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    populate_new_field(env)
--- a/stock_reception_screen_measuring_device/migrations/14.0.2.0.0/pre-migration.py
+++ b/stock_reception_screen_measuring_device/migrations/14.0.2.0.0/pre-migration.py
@@ -1,0 +1,29 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def create_and_populate_new_fields(cr):
+    cr.execute(
+        """
+        ALTER TABLE stock_reception_screen
+        ADD COLUMN IF NOT EXISTS smaller_package_has_missing_dimensions BOOLEAN;
+        """
+    )
+    # Set value to False for done reception screens.
+    # Otherwise, let the ORM do its job in post migration
+    _logger.info("Set smaller_package_has_missing_dimensions on done reception screens")
+    cr.execute(
+        """
+        UPDATE stock_reception_screen
+        SET smaller_package_has_missing_dimensions = FALSE
+        WHERE current_step = 'done';
+        """
+    )
+
+
+def migrate(cr, version):
+    create_and_populate_new_fields(cr)
--- a/stock_reception_screen_measuring_device/models/stock_reception_screen.py
+++ b/stock_reception_screen_measuring_device/models/stock_reception_screen.py
@@ -18,6 +18,12 @@ class StockReceptionScreen(models.Model):
         store=True,
         help="Indicates if the package have any measurement missing.",
     )
+    smaller_package_has_missing_dimensions = fields.Boolean(
+        "Smaller Package Requires Measures?",
+        compute="_compute_smaller_package_has_missing_dimensions",
+        store=True,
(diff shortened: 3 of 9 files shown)


stock_reception_screen_measuring_device/migrations/14.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,29 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def create_and_populate_new_fields(cr):
+    cr.execute(
+        """
+        ALTER TABLE stock_reception_screen
+        ADD COLUMN IF NOT EXISTS smaller_package_has_missing_dimensions BOOLEAN;
+        """
+    )
```
sebalix, line 15: Not blocking, but there are `odoo.tools.sql.create_column` helper for that.

sebalix (approved):

stock_reception_screen_measuring_device/models/stock_reception_screen.py
```diff
@@ -30,13 +36,18 @@ class StockReceptionScreen(models.Model):
...
+    @api.depends(
+        "current_move_product_id.packaging_ids.measuring_device_id",
+    )
     def _compute_scan_requested(self):
         for record in self:
-            record.scan_requested = (
-                record.product_packaging_id
-                and record.product_packaging_id.measuring_device_id
-            )
+            # TODO
```
simahawk: TODO what?

simahawk (approved):
LG

stock_reception_screen_measuring_device/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+    _logger.info(
+        "Set smaller_package_has_missing_dimensions on ongoing reception screens"
+    )
+    records_to_update._compute_smaller_package_has_missing_dimensions()
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
```
grindtildeath, line 24: You can use `@openupgrade.migrate()` as decorator to get an environment as parameter instead of `cr` (not blocking)

stock_reception_screen_measuring_device/models/stock_reception_screen.py
```diff
@@ -30,13 +36,17 @@ class StockReceptionScreen(models.Model):
...
-            record.scan_requested = (
-                record.product_packaging_id
-                and record.product_packaging_id.measuring_device_id
-            )
+            all_product_packagings = record.current_move_product_id.packaging_ids
+            record.scan_requested = False
+            for packaging in all_product_packagings:
+                if packaging.measuring_device_id:
+                    record.scan_requested = True
+                    break
```
grindtildeath, line 48:
```suggestion
            record.scan_requested = any(packaging.measuring_device_id for packaging in all_product_packagings)
```

stock_reception_screen_measuring_device/models/stock_reception_screen.py
```diff
@@ -71,29 +95,62 @@ def _compute_package_has_missing_dimensions(self):
...
+            ("packaging_length", "=", 0),
+            ("width", "=", 0),
+            ("height", "=", 0),
+            ("max_weight", "=", 0),
+        ]
+
+    def _get_smaller_package_without_dimensions(self):
+        self.ensure_one()
+        domain = self._get_smaller_package_without_dimensions_domain()
+        return self.env["product.packaging"].search(domain, order="qty desc", limit=1)
```
grindtildeath, line 141: Looks weird, if you want the "smaller" package, shouldn't you order by `qty asc` with the limit set to 1 ?

grindtildeath (approved):
Looks good, small remarks

stock_reception_screen_measuring_device/models/stock_reception_screen.py
simahawk: @mmequignon ping

stock_reception_screen_measuring_device/models/stock_reception_screen.py
```diff
@@ -30,13 +36,17 @@ class StockReceptionScreen(models.Model):
         store=True,
     )
 
-    @api.depends("product_packaging_id", "product_packaging_id.measuring_device_id")
+    @api.depends(
+        "current_move_product_id.packaging_ids.measuring_device_id",
```
sebalix, line 39: We could trigger a lot of computation with such depends knowing that `scan_requested` field is `store=True`, especially each time we call `_measuring_device_assign`/`_measuring_device_release` methods.

IMO this `scan_requested` field could be set `store=False` as it's only used to display a button on the view. So this field will be computed only when needed, avoiding recomputation of its value on all `stock.reception.screen` records.

stock_reception_screen_measuring_device/models/stock_reception_screen.py
mmequignon: no, we measure from the bigger to the smallest
bix box -> medium box -> small box

stock_reception_screen_measuring_device/models/stock_reception_screen.py
mmequignon: Even if this is nicer to read, I think the for loop implementation is faster, because it does not evaluate the value every   record.

stock_reception_screen_measuring_device/models/stock_reception_screen.py
mmequignon: We want the bigger package among the smaller packages missing measurements.

stock_reception_screen_measuring_device/models/stock_reception_screen.py
simahawk: @mmequignon this sounds like a good point

simahawk: fixed

PR #928, opened 2024-09-24
[FIX] stock_release_channel_shipment_advice_toursolver: Tests

lmignon:
Fix tests: The method called to get results from toursolver has changed with https://github.com/OCA/stock-logistics-transport/pull/141


lmignon:
ping @jbaudoux

jbaudoux (approved):
Thanks for the quick fix

PR #854, opened 2024-03-01
[14.0][ADD] stock_service_level_route

petrus-v:
Based on #841 this allows to restrict route by stock level service

PR #929, opened 2024-09-25
[ADD] stock_service_level_route: restrict route by service level

petrus-v:
supersede #854 

This allow to manage route by service level. Let's imagine you have in the same sale channel different level of service for the same product and want a different logistic route.


alexandregaldeano (approved):
LGTM

barkat-matthias (approved):
LGTM

PR #932, opened 2024-10-10
[16.0] stock_picking_batch_creation: Allows to split picking 

lmignon:
A new option 'Split picking exceeding the limits' on the wizard allow, when the system is not able to find a picking that fits the criteria to create the batch, to lower the criteria by removing those based on the volume, weight and number of lines. If a picking is found and you allow to split it, the system will try to split the picking so that the new picking fits the criteria and can be added to the batch.

Depends on:
* [x] https://github.com/OCA/stock-logistics-workflow/pull/1741

### DIFF
--- a/stock_picking_batch_creation/wizards/make_picking_batch.py
+++ b/stock_picking_batch_creation/wizards/make_picking_batch.py
@@ -1,6 +1,6 @@
 # Copyright 2021 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
+import logging
 import math
 import threading
 from collections import defaultdict
@@ -13,7 +13,10 @@ from ..exceptions import (
     NoPickingCandidateError,
     NoSuitableDeviceError,
     PickingCandidateNumberLineExceedError,
+    PickingSplitNotPossibleError,
 )
+
+_logger = logging.getLogger(__name__)
 
 
 class MakePickingBatch(models.TransientModel):
@@ -279,7 +282,29 @@ class MakePickingBatch(models.TransientModel):
         ).action_apply()
         return picking
 
-    def _get_first_picking(self, no_limit=False):
+    def _is_picking_exceeding_limits(self, picking):
+        """Check if the picking exceeds the limits of the available devices.
+
+        :param picking: the picking to check
+        """
+        nbr_lines, volume, weight = self._get_picking_max_dimensions()
+        return (
+            picking.nbr_picking_lines > nbr_lines
+            or picking.volume > volume
+            or picking.weight > weight
+        )
+
+    def _get_first_picking(self, raise_if_not_found=False):
+        """Get the first picking to add to the batch.
+
+        If the split_picking_exceeding_limits is set, we try to find the first picking
+        without taking into account the limit on the number of lines and we split it
+        if it exceeds the limits. If the split is not possible, we raise an error.
+
+        Otherwise, we try to find the first picking taking into account the limit on the
+        number of lines.
+        """
+        no_limit = self.split_picking_exceeding_limits
         domain = self._get_picking_domain_for_first(no_nbr_lines_limit=no_limit)
         if not no_limit:
             device_domains = []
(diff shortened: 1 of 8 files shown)


stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -260,15 +261,73 @@ def _execute_search_pickings(self, domain, limit=None):
...
+    def _split_first_picking_for_limit(self, picking):
+        nbr_lines, volume, weight = self._get_picking_max_dimensions()
+        wizard = self.env["stock.split.picking"].with_context(active_ids=picking.ids)
+        wizard.create(
+            {
+                "mode": "dimensions",
+                "max_nbr_lines": nbr_lines,
+                "max_volume": volume,
+                "max_weight": weight,
+            }
```
jbaudoux, line 281:
```suggestion
    # In base split module
    def _prepare_prepare_split_picking_wizard(self):
       return {
         "mode": self.split_picking_exceeding_limits_mode
       }

    # In split_dimension module
    def _prepare_prepare_split_picking_wizard(self):
       res = super()._prepare_prepare_split_picking_wizard()
       if self.split_picking_exceeding_limits_mode == "dimensions":
         nbr_lines, volume, weight = self._get_picking_max_dimensions()
         res.update({
                "mode": "dimensions",
                "max_nbr_lines": nbr_lines,
                "max_volume": volume,
                "max_weight": weight,
          })
        return res

    # In base split module
    def _split_first_picking_for_limit(self, picking):
        wizard = self.env["stock.split.picking"].with_context(active_ids=picking.ids)
        wizard.create(self._prepare_split_picking_wizard())
```

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -260,15 +261,73 @@ def _execute_search_pickings(self, domain, limit=None):
...
+    def _is_picking_exceeding_limits(self, picking):
+        """Check if the picking exceeds the limits of the available devices.
+
+        :param picking: the picking to check
+        """
+        nbr_lines, volume, weight = self._get_picking_max_dimensions()
+        return (
+            picking.nbr_picking_lines > nbr_lines
+            or picking.volume > volume
+            or picking.weight > weight
```
jbaudoux, line 294:
```suggestion
    # In base split module
    def _is_picking_exceeding_limits(self, picking):
      return False

    # In split_dimension module
    def _is_picking_exceeding_limits(self, picking):
      if self.split_picking_exceeding_limits_mode == "dimensions":
        nbr_lines, volume, weight = self._get_picking_max_dimensions()
        return (
            picking.nbr_picking_lines > nbr_lines
            or picking.volume > volume
            or picking.weight > weight
        )
      return super()._is_picking_exceeding_limits(picking)
```

jbaudoux (approved):
Looks good.

The device can also define a limit of bins. I'll need this feature in v18.

As commented below, this needs to be split in multiple modules to not add new dependencies

jbaudoux (changes requested):
Reviewing this again, I think we should split this into multiple modules.
The existing module should not depend on new modules (there are quite some modules to split by dimension).

I would move the split to a new module `stock_picking_batch_creation_split` defining `split_picking_exceeding_limits` and the mode `split_picking_exceeding_limits_mode` as a selection offering `move` and depending on `stock_split_picking`.
**Edit:** I think this `move` mode is useless. Maybe `split_picking_exceeding_limits` can be dropped and only the selection `split_picking_exceeding_limits_mode` kept with a default behavior as `no split`.

Then a module `stock_picking_batch_creation_split_dimension` that declares the mode `by dimension` and depends on `stock_split_picking_dimension`.

Then I can add a module `stock_picking_batch_creation_split_bins` that declares the mode `by number of bins` and would depend on `stock_split_picking_nbr_bins`.

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -260,15 +261,73 @@ def _execute_search_pickings(self, domain, limit=None):
...
+        # if the split_picking_exceeding_limits is set to True.
+        if (
+            picking
+            and self.split_picking_exceeding_limits
+            and self._is_picking_exceeding_limits(picking)
+        ):
+            split_picking = self._split_first_picking_for_limit(picking)
+            if not split_picking and raise_if_not_found:
+                raise PickingSplitNotPossibleError(picking)
+            picking = split_picking
```
jbaudoux, line 329: Move to `stock_picking_batch_creation_split` module

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -260,15 +261,73 @@ def _execute_search_pickings(self, domain, limit=None):
...
+        wizard = self.env["stock.split.picking"].with_context(active_ids=picking.ids)
+        wizard.create(
+            {
+                "mode": "dimensions",
+                "max_nbr_lines": nbr_lines,
+                "max_volume": volume,
+                "max_weight": weight,
+            }
+        ).action_apply()
+        return picking
```
jbaudoux, line 283: There is something not right about sorting and the way the picking is spit. In the split you push the remaining lines in a backorder having a new id. So on the next batch creation, as the pickings are sorted by id, you pushed that backorder to the end of all pickings to select while it should remain the next one in the list. So you will work on each SO one by one and then work on the backorders instead of doing a complete SO and then start the next SO.
The issue relies in the way you did the split. You should get here back the list of pickings (split orders) and use that instead of the origin picking

lmignon:
It's really a lot of changes and if we keep the PR open for the next 6 months others changes will be asked for sure.....

jbaudoux:
@ivantodorovich will perform the requested changes so that this PR can be finalized and merged and migrated to v18.0

ivantodorovich (approved):

jbaudoux (approved):
I discussed this with Ivan and we will keep this in a single module like you did. The move of selected moves to a new split order will be changed in v18

ivantodorovich:
The depended PR got merged.

@lmignon could you remove the test dependency commit on this one?
I belive we can merge it

lmignon:
@ivantodorovich test dependency removed

PR #933, opened 2024-10-10
[16.0][IMP] shopfloor_batch_automatic_creation: Allows to split picking

lmignon:
depends on:

* [ ] #932 
* [ ] https://github.com/OCA/stock-logistics-workflow/pull/1741


jbaudoux (approved):
LG

PR #937, opened 2024-10-14
[16.0][FIX] stock_available_to_promise_release: normal outgoing moves are mergeable

lmignon:
outgoing moves from unreleasable pickings are mergeable by default. This fix is required to avoid to disalbe the merge of moves from outgoing picking when the module is installed but the associated stock.rule is not configured to defer the pull of not available qty

ping @jbaudoux

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -57,8 +57,9 @@ class StockMove(models.Model):
 
     @api.depends("need_release", "rule_id", "rule_id.available_to_promise_defer_pull")
     def _compute_unrelease_allowed(self):
+        user_is_allowed = self.env.user.has_group("stock.group_stock_user")
         for move in self:
-            unrelease_allowed = move._is_unreleaseable()
+            unrelease_allowed = user_is_allowed and move._is_unreleaseable()
             if unrelease_allowed:
                 iterator = move._get_chained_moves_iterator("move_orig_ids")
                 next(iterator)  # skip the current move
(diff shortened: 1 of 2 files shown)


stock_available_to_promise_release/models/stock_move.py
```diff
@@ -787,7 +787,9 @@ def write(self, vals):
     def _is_mergeable(self):
         self.ensure_one()
         return self.state not in ("done", "cancel") and (
-            self.picking_type_id.code != "outgoing" or self.unrelease_allowed
+            self.picking_type_id.code != "outgoing"
+            or not self._is_unreleaseable()
+            or self.unrelease_allowed
```
```suggestion
            not self._is_unreleaseable()
            or self.unrelease_allowed
```

jbaudoux (commented):
By the way, shouldn't `user_is_allowed = self.env.user.has_group("stock.group_stock_user")` move to `unrelease_allowed` instead of `_is_unreleasable()` ?

lmignon:
> By the way, shouldn't `user_is_allowed = self.env.user.has_group("stock.group_stock_user")` move to `unrelease_allowed` instead of `_is_unreleasable()` ?

I share the same felling. I change the code.

lmignon:
@jbaudoux changes done

jbaudoux (approved):
Great :)

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -57,8 +57,9 @@ class StockMove(models.Model):
 
     @api.depends("need_release", "rule_id", "rule_id.available_to_promise_defer_pull")
     def _compute_unrelease_allowed(self):
+        user_is_allowed = self.env.user.has_group("stock.group_stock_user")
         for move in self:
-            unrelease_allowed = move._is_unreleaseable()
+            unrelease_allowed = move._is_unreleaseable() and user_is_allowed
```
rousseldenis: @lmignon Minor comment, I would have checked user_is_allowed before evaluating all other fields on move

rousseldenis (approved):

stock_available_to_promise_release/models/stock_move.py
lmignon: Good catch. Code fixed.

sbejaoui (approved):

PR #938, opened 2024-10-17
[IMP] stock_storage_type: Allow to re-apply putaway rules on computed location for a selected move line

rousseldenis:
We must ensure the same destination is computed at each time.

During recomputation, this:

- Allows to keep the same destination if it is still allowed
- Allows to change the destination if it is not allowed anymore

- [x] Improve the writing of domains appending with `AND()`

### DIFF
--- a/stock_storage_type/models/stock_storage_category_capacity.py
+++ b/stock_storage_type/models/stock_storage_category_capacity.py
@@ -1,6 +1,7 @@
 # Copyright 2022 ACSONE SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 from odoo import _, api, fields, models
+from odoo.osv.expression import AND, OR
 
 
 class StorageCategoryProductCapacity(models.Model):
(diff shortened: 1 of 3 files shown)


rousseldenis:
@jbaudoux @lmignon @phschmidt

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -95,7 +96,30 @@ def _domain_location_storage_type(self, candidate_locations, quants, products):
         ]
         # Build the domain using the 'allow_new_product' field
         if self.allow_new_product == "empty":
-            location_domain.append(("location_is_empty", "=", True))
+            # We should include the destination location of the current
+            # stock move line to avoid excluding it if already selected
```
lmignon, line 118: @rousseldenis Can you improve your comment to describe the use case / context? It's hard to figure out the purpose of this code when reading the comment.

lmignon (changes requested):

stock_storage_type/models/stock_storage_category_capacity.py
rousseldenis: Improved. Thanks

lmignon (approved):

PR #939, opened 2024-10-17
[16.0][IMP] stock_available_to_promise_release: Improved release process ro…

lmignon:
…bustness

Previous to this change, if the release process was 'forced' multiple time on an already released move, the system issued new procumrement rules without thaking into account the qty already released or processed. As side effect, the qty into the picking operation was higher than the expected qty to deliver. We now take into account the qty already released into the release process

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -562,12 +562,16 @@ class StockMove(models.Model):
 
         # Pull the released moves
         for move in released_moves:
+            qty_to_release = move._get_qty_to_release()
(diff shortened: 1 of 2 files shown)


lmignon:
ping @jbaudoux @TDu @sebalix @rousseldenis

sbejaoui (approved):

jbaudoux (changes requested):
This complexity should not be necessary because by definition a move is either in need_release or not. When in need_release, the qty to release is the product_uom_qty, and when not in need_release there is nothing to release.

This looks like a workaround for a bug non identified.

Not requesting changes, just not in favor of merging this

PR #941, opened 2024-10-28
[16.0][ADD] stock_release_channel_split_picking: Glue module

lmignon:
This module is a glue module that integrates the ***stock_release_channel*** module with the ***stock_picking_split*** module. It ensures that when a stock picking is split, the new picking is linked to the same release channel as the original picking.


lmignon:
ping @rousseldenis @sbejaoui

sbejaoui (approved):
LGTM

stock_release_channel_split_picking/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Split Picking With Release Channel",
```
jbaudoux:
```suggestion
    "name": "Stock Split Picking With Release Channel Propagate Picking",
```

stock_release_channel_split_picking/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "summary": """Glue module to preserve the release channel on picking split""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/wms",
+    "depends": [
+        "stock_release_channel",
```
jbaudoux:
```suggestion
        "stock_release_channel_propagate_channel_picking",
```

jbaudoux (changes requested):
Rename module name and fix deps

lmignon:
ping @rousseldenis @jbaudoux

jbaudoux (commented):
There are several places where a split order is created (like in shopfloor) and that split module is not used.
I think that issue should be solved in the base release channel module by removing the copy=False and ensuring the channel is removed when explicitly required like on backorder creation.
That would be more robust.

PR #943, opened 2024-10-29
[16.0][ADD] stock_available_to_promise_release_restrict_lot

sbejaoui:
When both stock_available_to_promise_release and stock_restrict_lot modules are installed, the calculation of quantities available to promise doesn't properly account for lot restrictions. This results in a priority conflict, where higher-priority moves that are restricted to a specific lot receive available quantities from any lot, bypassing their lot restriction.

This glue module enhances stock move allocation by:
- Allowing stock moves to respect both priority and specific lot allocations.
- Ensuring that available quantities are promised according to move priority, but only when the lot matches the restriction.


stock_available_to_promise_release_restrict_lot/models/stock_move.py
```diff
@@ -0,0 +1,81 @@
...
+            "|",
+            "&",
+            ("priority", "=", self.priority),
+            ("date_priority", "<", self.date_priority),
+            "&",
+            ("priority", "=", self.priority),
+            "&",
+            ("date_priority", "=", self.date_priority),
+            ("id", "<", self.id),
+        ]
```
lmignon: This domain is really hard to read. The use of the AND and OR operators from the expressions module should improve readability ant therefore maintenance.

lmignon (changes requested):
A little comment about readability...

stock_available_to_promise_release_restrict_lot/models/stock_move.py
```diff
@@ -0,0 +1,85 @@
...
+from odoo.osv.expression import AND, OR
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    def _previous_promised_qty_sql_lateral_where(self, warehouse):
+        sql, params = super()._previous_promised_qty_sql_lateral_where(warehouse)
+        sql += "AND m.restrict_lot_id is null"
```
jbaudoux, line 14: I would not do this. If you place a move without restrict lot, any move before should reserve a qty no matter they are restricted to a lot or not.
Example:
Move1 for lotA
Move2 for any lot. Should consider the previous Move1 taking from stock

stock_available_to_promise_release_restrict_lot/models/stock_move.py
```diff
@@ -0,0 +1,85 @@
...
+                        ],
+                        [
+                            ("priority", "=", self.priority),
+                            ("date_priority", "=", self.date_priority),
+                            ("id", "<", self.id),
+                        ],
+                    ]
+                ),
+            ]
+        )
```
jbaudoux, line 40: Seems to miss quite a few things in this domain: need_release, outgoing, picking type, locations...

jbaudoux (changes requested):
I wonder how this can work

PR #947, opened 2024-11-15
[14.0][FIX] shopfloor: compatibility location barcode with barcodes_generator_location

clementmbr:
When using wms/shopfloor with [stock-logistics-barcode/barcodes_generator_location](https://github.com/OCA/stock-logistics-barcode/tree/14.0/barcodes_generator_location), the location's view is getting unusable because the module barcodes_generator_location is awaiting to have the barcode field in a dedicated group like in [stock-logistics-barcode/stock_barcodes](https://github.com/OCA/stock-logistics-barcode/blob/14.0/stock_barcodes/views/stock_location_views.xml#L12C13-L16C21) :
image

This PR creates a dedicated group for the barcode field, making **the 2 modules shopfloor and barcodes_generator_location compatible**:
image

cc @Kev-Roche

### DIFF
--- a/shopfloor/views/stock_location.xml
+++ b/shopfloor/views/stock_location.xml
@@ -5,12 +5,11 @@
         <field name="model">stock.location</field>
         <field name="inherit_id" ref="stock.view_location_form" />
         <field name="arch" type="xml">
-            <!-- show barcode because it's handy -->
-            <xpath expr="//field[@name='location_id']/.." position="after">
-                <label for="barcode" />
-                <field name="barcode" />
-            </xpath>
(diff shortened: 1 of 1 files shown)


sebalix:
Thanks for your PR.
IMO what we are missing here is a simple module `stock_location_barcode` with [this change](https://github.com/OCA/stock-logistics-barcode/blob/14.0/stock_barcodes/views/stock_location_views.xml), and make it a dependency of both `shopfloor` and `stock_barcodes`.

Checked in 18.0: `barcode` field still not re-introduced on std view.

PR #950, opened 2024-11-20
[14.0][FIX] shopfloor: cluster_picking do not create a new move line

mt-software-de:
When a wrong product is scanned, while setting the qty_done. 
The qty done is not changed and no response was returned. 
This leads to the error, that a new move line is created. 

Now it would be returning that this scanned product is not in the picking
and will not create a new move line.

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -696,16 +696,32 @@ class ClusterPicking(Component):
         if not product:
             packaging = search.packaging_from_scan(barcode)
             product = packaging.product_id
-        if product and move_line.product_id == product:
-            quantity += packaging.qty or 1.0
-            response = self._response_for_scan_destination(move_line, qty_done=quantity)
-            return response
+        if product:
+            if move_line.product_id == product:
+                quantity += packaging.qty or 1.0
+                response = self._response_for_scan_destination(
+                    move_line, qty_done=quantity
+                )
+                return response
+            return self._response_for_scan_destination(
+                move_line,
+                message=self.msg_store.wrong_record(product),
+                qty_done=quantity,
+            )
         # Handle barcode of a lot
         lot = search.lot_from_scan(barcode)
-        if lot and move_line.lot_id == lot:
-            quantity += 1.0
-            response = self._response_for_scan_destination(move_line, qty_done=quantity)
-            return response
+        if lot:
+            if move_line.lot_id == lot:
+                quantity += 1.0
+                response = self._response_for_scan_destination(
+                    move_line, qty_done=quantity
+                )
+                return response
+            return self._response_for_scan_destination(
+                move_line,
+                message=self.msg_store.wrong_record(lot),
+                qty_done=quantity,
+            )
         return response
 
     def scan_destination_pack(self, picking_batch_id, move_line_id, barcode, quantity):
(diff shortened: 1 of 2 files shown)


mt-software-de:
@TDu could you review this one?

TDu (approved):

TDu (commented):
Much better indeed :+1:

mt-software-de:
@jbaudoux could you please review this as well?

rousseldenis (approved):

jbaudoux (changes requested):
It doesn't look correct yet.
If you have 2 lines for the same product but different lots, it's like it is increasing both lines;
Those barcode searches do not look very GS1 friendly. It should call `find` with the list of handlers (product, packaging, lot)

mt-software-de:
> It doesn't look correct yet.
> If you have 2 lines for the same product but different lots, it's like it is increasing both lines;
> Those barcode searches do not look very GS1 friendly. It should call `find` with the list of handlers (product, packaging, lot)

What do you mean? I think my change is correct. The method in generell could be refactored but then also the other places where the different find methods are used.

jbaudoux:
@mt-software-de can you remind me at which stage in the process this method is called?

mt-software-de:
> @mt-software-de can you remind me at which stage in the process this method is called?

This method is called when you already selected a move line and try to increase the qty done. You can now either scan the package, product or lot to increase it.

shopfloor/services/cluster_picking.py
```diff
@@ -696,10 +696,19 @@ def _set_destination_pack_update_quantity(self, move_line, quantity, barcode):
...
+        if product:
+            if move_line.product_id == product:
+                quantity += packaging.qty or 1.0
+                response = self._response_for_scan_destination(
+                    move_line, qty_done=quantity
+                )
+                return response
+            return self._response_for_scan_destination(
+                move_line,
+                message=self.msg_store.product_not_found_in_pickings(),
```
jbaudoux: I think this is more suitable
```suggestion
                message=self.msg_store.wrong_record(product),
```

jbaudoux (commented):
Can you also then adapt the lot section below with the same logic?

mt-software-de:
> Can you also then adapt the lot section below with the same logic?

Yes of course this makes sense.

mt-software-de:
> Can you also then adapt the lot section below with the same logic?

done

shopfloor/tests/test_cluster_picking_scan_destination_no_prefill_qty.py
```diff
@@ -76,15 +77,16 @@ def test_scan_destination_pack_increment_with_wrong_product(self):
...
+        self.assertFalse(new_move_line)
         expected_qty_done = qty_done
         self.assert_response(
             response,
             next_state="scan_destination",
             data=self._line_data(line, qty_done=expected_qty_done),
-            message=self.service.msg_store.bin_not_found_for_barcode(
-                self.product_b.barcode
-            ),
+            message=self.service.msg_store.product_not_found_in_pickings(),
```
jbaudoux:
```suggestion
            message=self.service.msg_store.wrong_record(product),
```

jbaudoux (approved):

PR #954, opened 2024-11-27
[16.0][FIX] sale_stock_release_channel_partner_by_date: Avoid release channel reset

santostelmo:
Avoid release channel reset when creating a SO.

Steps to replicate:

1. create a new sale order
2. select a customer
3. select a commitment date (Delivery Date) in the future that does not propose any release channel
4. select a release channel (any)
5. save

Results: on save, release channel is reset to empty
Expected behavior:  on save, the release channel is kept

### DIFF
--- a/sale_stock_release_channel_partner_by_date/models/sale_order.py
+++ b/sale_stock_release_channel_partner_by_date/models/sale_order.py
@@ -42,6 +42,8 @@ class SaleOrder(models.Model):
             if not rec._check_release_channel_partner_date_requirements():
                 continue
             channel_date = rec.release_channel_partner_date_id
+            if not channel_date:
+                continue
             rec.release_channel_id = channel_date.release_channel_id
 
     @api.depends(
--- a/sale_stock_release_channel_partner_by_date/tests/test_sale_release_channel.py
+++ b/sale_stock_release_channel_partner_by_date/tests/test_sale_release_channel.py
@@ -1,6 +1,8 @@
 # Copyright 2024 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
+from dateutil.relativedelta import relativedelta
+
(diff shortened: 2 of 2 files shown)


sebalix (commented):
A test to reproduce the issue is welcome

sebalix:
I fixed the PR title, the commit should be fixed too (wrong module mentioned)

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -42,7 +42,8 @@ def _compute_release_channel_id(self):
             if not rec._check_release_channel_partner_date_requirements():
                 continue
             channel_date = rec.release_channel_partner_date_id
-            rec.release_channel_id = channel_date.release_channel_id
+            if channel_date:
```
sebalix, line 45:
```suggestion
            if not rec.release_channel_id and channel_date:
```
? Not sure, to check

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -42,7 +42,8 @@ def _compute_release_channel_id(self):
...
+                rec.release_channel_id = channel_date.release_channel_id
```
sebalix:
```suggestion
            forced_release_channel_id = self.env.context.get("order_release_channel_id", {}).get(rec.id)
            if not channel_date and forced_release_channel_id:
                rec.release_channel_id = forced_release_channel_id
                continue
            rec.release_channel_id = channel_date.release_channel_id
```

santostelmo: That implementation doesn't seem to be possible:
1. I don't know why but context variable passed in create order_release_channel_id never reach `_compute_release_channel_id`
2. We can't know the id of the record before creating it.

PR #955, opened 2024-11-29
[16.0][FIX] sale_stock_release_channel_partner_by_date*: fix computation of release channel

sebalix:
Aims to supersede #954

### DIFF
--- a/sale_stock_release_channel_partner_by_date/models/sale_order.py
+++ b/sale_stock_release_channel_partner_by_date/models/sale_order.py
@@ -42,7 +42,8 @@ class SaleOrder(models.Model):
             if not rec._check_release_channel_partner_date_requirements():
                 continue
             channel_date = rec.release_channel_partner_date_id
-            rec.release_channel_id = channel_date.release_channel_id
+            if channel_date:
+                rec.release_channel_id = channel_date.release_channel_id
 
     @api.depends(
         "state",
--- a/sale_stock_release_channel_partner_by_date/models/sale_order.py
+++ b/sale_stock_release_channel_partner_by_date/models/sale_order.py
@@ -61,7 +61,11 @@ class SaleOrder(models.Model):
         self.ensure_one()
         model = self.env["stock.release.channel.partner.date"]
         domain = self._get_release_channel_partner_date_domain()
-        return domain and model.search(domain, limit=1) or model
+        return (
+            domain
+            and model.with_context(active_test=False).search(domain, limit=1)
+            or model
+        )
 
     def _get_release_channel_partner_date_domain(self):
         self.ensure_one()
--- a/sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
+++ b/sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
@@ -22,11 +22,23 @@ class SaleOrder(models.Model):
 
     def _get_release_channel_partner_date_domain(self):
         domain = super()._get_release_channel_partner_date_domain()
-        if domain and self.carrier_id:
-            carrier_domain = [
-                "|",
-                ("release_channel_id.carrier_ids", "in", self.carrier_id.ids),
-                ("release_channel_id.carrier_ids", "=", False),
-            ]
+        if domain:
+            if self.carrier_id:
+                carrier_domain = [
+                    ("release_channel_id.carrier_ids", "in", self.carrier_id.ids),
+                ]
+            else:
+                carrier_domain = [
+                    ("release_channel_id.carrier_ids", "=", False),
+                ]
             domain = expression.AND([domain, carrier_domain])
         return domain
+
+    def _compute_release_channel_id(self):
+        # pylint: disable=missing-return
+        super()._compute_release_channel_id()
+        for rec in self:
+            # Selected release channel and carrier have to be compatible
+            if rec.release_channel_id.carrier_ids and rec.carrier_id:
+                if rec.carrier_id not in rec.release_channel_id.carrier_ids:
+                    rec.release_channel_id = False
--- a/sale_stock_release_channel_partner_by_date_delivery/tests/test_sale_release_channel.py
+++ b/sale_stock_release_channel_partner_by_date_delivery/tests/test_sale_release_channel.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from odoo import fields
+from odoo.tests.common import Form
 
 from odoo.addons.sale_stock_release_channel_partner_by_date.tests.common import (
     SaleReleaseChannelCase,
@@ -22,9 +23,18 @@ class TestSaleReleaseChannel(SaleReleaseChannelCase):
             }
         )
         cls.carrier_channel.action_wake_up()
+        cls.carrier2_channel = cls.default_channel.copy(
+            {
+                "name": "Test with carrier2",
+                "sequence": 10,
+                "carrier_ids": [(6, 0, cls.carrier2.ids)],
+            }
+        )
+        cls.carrier2_channel.action_wake_up()
 
-    def test_sale_order_with_wrong_carrier(self):
+    def test_sale_order_without_carrier_with_channel_date(self):
         delivery_date = fields.Datetime.now()
+        # Create one specific channel not matching the SO regarding the carrier
         channel_date_model = self.env["stock.release.channel.partner.date"]
         channel_date_model.create(
             {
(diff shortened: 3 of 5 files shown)


sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -42,7 +42,8 @@ def _compute_release_channel_id(self):
             if not rec._check_release_channel_partner_date_requirements():
                 continue
             channel_date = rec.release_channel_partner_date_id
-            rec.release_channel_id = channel_date.release_channel_id
+            if channel_date:
```
jbaudoux, line 45: Maybe you can check here that the warehouse is compatible with the currently selected channel? Like you did for the carrier.

sebalix: Yes indeed. But I have still one open questions about this (you can check the test where there are FIXME in this PR):

If we have configured a channel **X** (with a carrier) and **Y** (no carrier), and have a specific channel date set for **X**.

Currently if we do not define a carrier on the SO, the computed release channel wil be taken from the configured specific channel (as it's the only one that exists). So the SO will show the release channel **X**.

Then we validate the SO, and we trigger the channel assignment on the delivery: it is now assigned to the release channel **Y** (of course, there was no carrier defined, I guess it fallbacks on **Y** because of that).

IMO this is not what the user will expect.

How should we handle SO without carrier (same question for SO without warehouse) regarding release channels?

sale_stock_release_channel_partner_by_date_delivery/tests/test_sale_release_channel.py
```diff
@@ -22,31 +23,41 @@ def setUpClass(cls):
...
                 "release_channel_id": self.carrier_channel.id,
                 "date": delivery_date.date(),
             }
         )
-        # With wrong carrier set: order doesn't detect the specific channel for carrier
         order = self._create_sale_order(date=delivery_date)
-        order.carrier_id = self.carrier2
-        self.assertFalse(order.release_channel_id)
+        self.assertFalse(order.carrier_id)
+        self.assertEqual(order.release_channel_id, self.carrier_channel)
```
jbaudoux: This looks wrong to me. If the channel has a carrier restriction and the order has no carrier, you cannot end-up with the channel auto assigned to the SO

sale_stock_release_channel_partner_by_date/models/sale_order.py
jbaudoux: i think you should not include False for warehouse/carrier in `_get_release_channel_partner_date_domain`. The warehouse should be set by a default value. For the carrier, it is set by delivery_auto_refresh. So if one of them is empty, then we should not show a channel that has a restriction that is not respected

sebalix: I agree (I don't remember why we searched on carrier_id=False in the initial implementation, we did that together and on purpose :thinking: ).

sebalix: For the warehouse, while we could state it is always set on the SO (I could not find some without WH in customers DBs), it is however not mandatory on the release channel. We should consider these release channels as elligible, and thus keep our check on warehouse_id=False in the domain.

sale_stock_release_channel_partner_by_date_delivery/tests/test_sale_release_channel.py
sebalix: Fixup pushed

santostelmo (approved):

Ricardoalso (approved):
LGTM

jbaudoux (approved):

jbaudoux (commented):
@sebalix Can you make one commit per module ?

jbaudoux:
@sebalix Can you make one commit per module ? (reminder)

sebalix:
@jbaudoux done, and I pushed two others changes (last two commits), can you do a pass on them?

sale_stock_release_channel_partner_by_date/models/sale_order.py
```diff
@@ -60,7 +61,11 @@ def _get_release_channel_partner_date(self):
         self.ensure_one()
         model = self.env["stock.release.channel.partner.date"]
         domain = self._get_release_channel_partner_date_domain()
-        return domain and model.search(domain, limit=1) or model
+        return (
+            domain
+            and model.with_context(active_test=False).search(domain, limit=1)
+            or model
+        )
```
jbaudoux, line 68: Like in https://github.com/OCA/wms/pull/1035/files#diff-6beda63adf25626bf49fb26b9d1e07118cfb4ca798340d82212b771edd37107aL29
```suggestion
         return (
             model
             .with_context(active_test=False)
             .search(self._release_channel_partner_date_domain)
             .filtered(
                lambda o: o.release_channel_id.filtered_domain(
                    self._release_channel_possible_candidate_domain_base
                )
            )
```
using: https://github.com/OCA/wms/pull/1036/files#diff-1c58ad6a38530714dcc008ed080e2d6baf8ee44c01f54a064a92b24589276a6dR12

and in `_get_release_channel_partner_date_domain` drop the warehouse part of the domain.
and then you don't need to extend in _delivery module.

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
```diff
@@ -22,11 +22,23 @@ def _get_release_channel_id_depends(self):
...
-            ]
+        if domain:
+            if self.carrier_id:
+                carrier_domain = [
+                    ("release_channel_id.carrier_ids", "in", self.carrier_id.ids),
+                ]
+            else:
+                carrier_domain = [
+                    ("release_channel_id.carrier_ids", "=", False),
+                ]
```
jbaudoux, line 33: I think it was correct before. The channel is valid for the configured carrier otherwise it's valid for any carrier.
Like I've done here https://github.com/OCA/wms/pull/1036/files#diff-f5d9018dab2ec5bc651d62b3b894bb66d32fe9a443004318faf1fd240ac4f622R13

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
```diff
@@ -22,11 +22,23 @@ def _get_release_channel_id_depends(self):
...
             domain = expression.AND([domain, carrier_domain])
         return domain
+
+    def _compute_release_channel_id(self):
```
jbaudoux, line 37: don't get why this is necessary. Only carrier compatible channels can be computed

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
jbaudoux: At least it is not in line with the release channel domain on the SO as you can choose a release channel without carrier
https://github.com/OCA/wms/pull/955/files#diff-f730c0d0b31786d57639f65a74a612fce6747ed263018bd094c72fdad146ed42R14

sale_stock_release_channel_partner_by_date/models/sale_order.py
jbaudoux: agreed

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
jbaudoux: if carrier, first domain was correct
but it was indeed lacking the else case

Moved to `sale_stock_release_channel_delivery` here https://github.com/OCA/wms/pull/1036/files#diff-f5d9018dab2ec5bc651d62b3b894bb66d32fe9a443004318faf1fd240ac4f622R18

sale_stock_release_channel_partner_by_date_delivery/models/sale_order.py
jbaudoux: Fixed with:
* The `release_channel_partner_date_id` was not emptied when the carrier was changed.
  * https://github.com/OCA/wms/pull/1039/files#diff-c052d7cf6444a81e0f15b713218cdcca8711e655000bc6b76a2ee1034e2f0e9cR56
* The compatibility can now be managed properly in base module
  * https://github.com/OCA/wms/pull/1039/files#diff-c052d7cf6444a81e0f15b713218cdcca8711e655000bc6b76a2ee1034e2f0e9cR51

jbaudoux:
Replaced by:
* https://github.com/OCA/wms/pull/1038
* https://github.com/OCA/wms/pull/1039

jbaudoux (changes requested):
superseded for bug fixing

PR #957, opened 2024-12-04
[16.0][FIX] stock_available_to_promise_release: fix view

jbaudoux:
Drop label change on date label on stock move form view as the odoo standard view changed in v14 to display label according to state

* In odoo v13 : https://github.com/odoo/odoo/blob/13.0/addons/stock/views/stock_move_views.xml#L345
* since odoo v14, date is with `nolabel` : https://github.com/odoo/odoo/blob/16.0/addons/stock/views/stock_move_views.xml#L333

@rousseldenis @sebalix @i-vyshnevska @cyrilmanuel


jbaudoux:
@sbejaoui Can you drop a review?

sbejaoui (approved):

PR #714, opened 2023-08-08
[14.0] Add shopfloor_gs1

sebalix:
Followup of #521 rebased on top of 14.0 to fix runboat.

PR #966, opened 2025-01-06
[16.0][MIG] shopfloor_gs1

rousseldenis:
From:

- https://github.com/OCA/wms/pull/714

Depends on:

- #1001


shopfloor_gs1/utils.py
```diff
@@ -0,0 +1,88 @@
...
+        return True
+
+    @classmethod
+    def _get_value(cls, ai_value: GS1ElementString):
+        """
+        We must return the date value if set
+        We must return the formatted GTIN if set
+        as the library will transform the GTIN 14 into
+        a GTIN 13 one. That one is the barcode
+        set in Odoo.
```
simahawk, line 43:
```suggestion
        """
        Determine the final value for the given parsed element.

        Returns

        * date value if set
        * formatted GTIN if set
         
        The library will transform the GTIN 14 into
        a GTIN 13, which is the one used by Odoo.
```

simahawk (commented):
LGTM 
@rousseldenis can you rebase and cleanup? 🙏

simahawk:
One more thing: as the v14 PR was never merged, IMO we can squash the fixup commit ;)

shopfloor_gs1/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["simahawk", "sebalix"],
+    "license": "AGPL-3",
+    "depends": ["shopfloor"],
+    "external_dependencies": {
+        "python": [
+            # >= 2.3.0 required to use 'GS1Message.parse_hri' method
+            # and next version 3.0.0 has been refactored bringing
+            # incompatibility issues (to check later).
+            "biip"
```
simahawk, line 21: the pin of the lib is lost here. It's mandatory as there's no other 2.x version that is suitable

nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (approved):

PR #971, opened 2025-01-13
[14.0] stock_available_to_promise_release: Allow unrelease processed qties

mmequignon:
This PR adds a new option on stock routes, allowing to unrelease a delivery even if some internal operations (e.g. pick) are already done, creating a return move to the stock location.

### DIFF
--- a/stock_available_to_promise_release/models/stock_location_route.py
+++ b/stock_available_to_promise_release/models/stock_location_route.py
@@ -7,6 +7,16 @@ from odoo import fields, models
 class Route(models.Model):
     _inherit = "stock.location.route"
 
+    allow_unrelease_return_done_move = fields.Boolean(
+        string="Reverse done transfer on cancellation",
+        default=False,
+        help=(
+            "If checked, unreleasing the delivery may create a new inverse "
+            "internal operation on the last done pulled transfer. "
+            "Otherwise, you won't be able to unrelease as soon as one of "
+            "the pulled transfer is done"
+        ),
+    )
     available_to_promise_defer_pull = fields.Boolean(
         string="Release based on Available to Promise",
         default=False,
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -5,11 +5,12 @@
 import itertools
 import logging
 import operator as py_operator
+from collections import defaultdict
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 from odoo.osv import expression
-from odoo.tools import date_utils, float_compare, float_round, groupby
+from odoo.tools import date_utils, float_compare, float_is_zero, float_round, groupby
 
 _logger = logging.getLogger(__name__)
 
@@ -96,12 +97,22 @@ class StockMove(models.Model):
         """
         self.ensure_one()
         pickings = origin_moves.mapped("picking_id")
-        if pickings.filtered("printed"):
-            # The picking is printed, we can't unrelease the move
-            # because the processing of the origin moves is started.
+        ongoing_pickings = pickings.filtered(
+            lambda p: p.printed and p.state not in ("cancel", "done")
+        )
+        if ongoing_pickings:
+            # The picking is either printed or done/cancelled.
+            # We can't unrelease the move because the processing of the origin moves
+            # is either started or done.
             return False
+        # If allow_unrelease_on_cancel is not checked, only release assigned/waiting
+        # moves. If checked, also unrelease done moves.
+        if self.rule_id.allow_unrelease_return_done_move:
+            unreleasable_states = ("cancel",)
+        else:
+            unreleasable_states = ("done", "cancel")
         origin_moves = origin_moves.filtered(
-            lambda m: m.state not in ("done", "cancel")
+            lambda m: m.state not in unreleasable_states
         )
         origin_qty_todo = sum(origin_moves.mapped("product_qty"))
         return (
@@ -603,6 +614,81 @@ class StockMove(models.Model):
             yield moves
             moves = moves.mapped(chain_field)
 
+    def _return_quantity_in_stock(self, qty_to_return_per_move):
+        """Return a quantity from a list of moves.
+
+        The quantity to return is in the product uom"""
+        moves_to_return = self.browse([m_id for m_id in qty_to_return_per_move.keys()])
+        moves_per_type = groupby(moves_to_return, lambda m: m.picking_type_id)
+        for picking_type, moves_list in moves_per_type:
+            moves = self.browse().union(*moves_list)
+            pickings = moves.picking_id
+            if not pickings:
+                continue
+            return_type = picking_type.return_picking_type_id
+            wiz_values = {
+                "picking_id": fields.first(pickings).id,
+                "location_id": return_type.default_location_dest_id.id,
+            }
+            product_return_moves = []
+            if not return_type:
+                message = _(
+                    "The operation %(picking_names)s is done and cannot be returned",
+                    picking_names=", ".join(pickings.mapped("name")),
+                )
+                raise UserError(message)
+            for move in moves:
+                # Cannot return an unprocessed move
+                if move.state != "done":
+                    continue
+                product = move.product_id
+                uom = product.uom_id
+                qty_to_return = qty_to_return_per_move.get(move.id, 0)
+                # Cannot return 0 qty
+                if float_is_zero(qty_to_return, precision_rounding=uom.rounding):
+                    continue
+                return_move_vals = {
+                    "product_id": product.id,
+                    "quantity": qty_to_return,
+                    "uom_id": uom.id,
+                    "move_id": move.id,
+                }
+                product_return_moves.append((0, 0, return_move_vals))
+            if product_return_moves:
+                wiz_values["product_return_moves"] = product_return_moves
+                return_wiz = self.env["stock.return.picking"].create(wiz_values)
+                return_wiz.create_returns()
+        return True
+
+    def _unrelease_set_returnable_qty_per_move(
+        self, qty_to_return, qty_to_return_per_move
+    ):
+        returnable_qty = 0
+        for move in self:
+            rounding = move.product_id.uom_id.rounding
+            # As a move might have multiple dest ids, we might have
+            # already planned to return a few units already.
+            # Get it, and deduce it from the returnable qty
+            move_qty_planned = qty_to_return_per_move.get(move.id, 0)
+            # A move might already have return moves linked to it, deduce their quantity
+            move_returned_qty = sum(
+                move.returned_move_ids.filtered(lambda m: m.state != "cancel").mapped(
+                    "product_qty"
+                )
+            )
+            move_returnable_qty = min(
+                qty_to_return, move.product_qty - move_returned_qty - move_qty_planned
+            )
+            if float_is_zero(move_returnable_qty, precision_rounding=rounding):
+                continue
+            # Update the quantity
+            qty_to_return_per_move[move.id] += move_returnable_qty
+            qty_to_return -= move_returnable_qty
+            returnable_qty += move_returnable_qty
+            if float_is_zero(qty_to_return, precision_rounding=rounding):
+                break
+        return returnable_qty
+
     def unrelease(self, safe_unrelease=False):
         """Unrelease unreleasavbe moves
 
@@ -614,29 +700,74 @@ class StockMove(models.Model):
             moves_to_unrelease = self.filtered("unrelease_allowed")
         moves_to_unrelease._check_unrelease_allowed()
         moves_to_unrelease.write({"need_release": True})
-        impacted_picking_ids = set()
 
+        qty_to_return_per_move = defaultdict(float)
         for move in moves_to_unrelease:
+            rounding = move.product_id.uom_id.rounding
+            # When a move is returned, it is going straight to WH/Stock,
+            # skipping all intermediate zones (pick/pack).
+            # That is why we need to keep track of qty returned along the way.
+            # We do not want to return the same goods at each step.
+            # At a given step (pick/pack/ship), qty to return is
+            # move.product_uom_qty - cancelled_qty_at_step - already returned qties
+            qty_to_unrelease = move.product_qty
+            qty_returned_for_move = 0
             iterator = move._get_chained_moves_iterator("move_orig_ids")
-            moves_to_cancel = self.env["stock.move"]
+            moves_to_cancel_for_move = self.env["stock.move"]
             # backup procure_method as when you don't propagate cancel, the
             # destination move is forced to make_to_stock
             procure_method = move.procure_method
             next(iterator)  # skip the current move
             for origin_moves in iterator:
-                origin_moves = origin_moves.filtered(
+                qty_to_cancel = qty_to_unrelease - qty_returned_for_move
+                if float_is_zero(qty_to_cancel, precision_rounding=rounding):
+                    break
+                todo_origin_moves = origin_moves.filtered(
                     lambda m: m.state not in ("done", "cancel")
                 )
-                if origin_moves:
-                    origin_moves = move._split_origins(origin_moves)
-                    impacted_picking_ids.update(origin_moves.mapped("picking_id").ids)
+                qty_canceled = 0
+                if todo_origin_moves:
+                    moves_to_cancel = move._split_origins(
+                        todo_origin_moves, qty=qty_to_cancel
+                    )
                     # avoid to propagate cancel to the original move
-                    origin_moves.write({"propagate_cancel": False})
-                    # origin_moves._action_cancel()
-                    moves_to_cancel |= origin_moves
-            moves_to_cancel._action_cancel()
+                    moves_to_cancel.write({"propagate_cancel": False})
+                    moves_to_cancel_for_move |= moves_to_cancel
+                    qty_canceled = sum(moves_to_cancel.mapped("product_qty"))
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_to_return = qty_to_cancel - qty_canceled
+                if float_is_zero(qty_to_return, precision_rounding=rounding):
+                    continue
+                done_moves = origin_moves.filtered(lambda m: m.state == "done")
+                if not move.rule_id.allow_unrelease_return_done_move:
+                    # Without allow_unrelease_return_done_move enabled,
+                    # only moves that aren't done can be unreleased.
+                    msg_args = {
+                        "move_name": move.name,
+                        "done_move_names": ", ".join(done_moves.mapped("name")),
+                    }
+                    message = _(
+                        (
+                            "You cannot unrelease the move %(move_name)s "
+                            "because some origin moves %(done_move_names)s are done"
+                        ),
+                        **msg_args
+                    )
+                    raise UserError(message)
+                # Multiple pickings can satisfy a move
+                # -> len(move.move_orig_ids.picking_id) > 1
+                # Group done_moves per picking, and create returns
+                returnable_qty = done_moves._unrelease_set_returnable_qty_per_move(
+                    qty_to_return, qty_to_return_per_move
+                )
+                qty_returned_for_move += returnable_qty
+
+            moves_to_cancel_for_move._action_cancel()
             # restore the procure_method overwritten by _action_cancel()
             move.procure_method = procure_method
+        self._return_quantity_in_stock(qty_to_return_per_move)
         moves_to_unrelease.write({"need_release": True})
         for picking, moves in itertools.groupby(
             moves_to_unrelease, lambda m: m.picking_id
@@ -649,14 +780,15 @@ class StockMove(models.Model):
             picking.message_post(body=body)
             picking.last_release_date = False
 
-    def _split_origins(self, origins):
+    def _split_origins(self, origins, qty=None):
         """Split the origins of the move according to the quantity into the
         move and the quantity in the origin moves.
 
         Return the origins for the move's quantity.
         """
         self.ensure_one()
-        qty = self.product_qty
+        if not qty:
+            qty = self.product_qty
         # Unreserve goods before the split
         origins._do_unreserve()
         rounding = self.product_uom.rounding
--- a/stock_available_to_promise_release/tests/test_unrelease_cancel.py
+++ b/stock_available_to_promise_release/tests/test_unrelease_cancel.py
@@ -1,0 +1,238 @@
+# Copyright 2025 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+from datetime import datetime
+
+from .common import PromiseReleaseCommonCase
+
+
+class TestAvailableToPromiseReleaseCancel(PromiseReleaseCommonCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        cls.wh.delivery_steps = "pick_pack_ship"
+        cls._update_qty_in_location(cls.loc_bin1, cls.product1, 50.0)
+        cls._update_qty_in_location(cls.loc_bin1, cls.product2, 50.0)
+
+        delivery_route = cls.wh.delivery_route_id
+        ship_rule = delivery_route.rule_ids.filtered(
+            lambda r: r.location_id == cls.loc_customer
+        )
+        cls.loc_output = ship_rule.location_src_id
+        pack_rule = delivery_route.rule_ids.filtered(
+            lambda r: r.location_id == cls.loc_output
+        )
+        cls.loc_pack = pack_rule.location_src_id
+        pick_rule = delivery_route.rule_ids.filtered(
+            lambda r: r.location_id == cls.loc_pack
+        )
+        cls.pick_type = pick_rule.picking_type_id
+        cls.pack_type = pack_rule.picking_type_id
+
+        cls.picking_chain = cls._create_picking_chain(
+            cls.wh, [(cls.product1, 10)], date=datetime(2019, 9, 2, 16, 0)
+        )
+        cls.ship_picking = cls._out_picking(cls.picking_chain)
+        cls.pack_picking = cls._prev_picking(cls.ship_picking)
+        cls.pick_picking = cls._prev_picking(cls.pack_picking)
+
+        # Why is this not working when creating picking after enabling this setting?
+        delivery_route.write(
+            {
+                "available_to_promise_defer_pull": True,
+                "allow_unrelease_return_done_move": True,
+            }
+        )
+        cls.ship_picking.release_available_to_promise()
+        cls.cleanup_type = cls.env["stock.picking.type"].create(
+            {
+                "name": "Cancel Cleanup",
+                "default_location_dest_id": cls.loc_stock.id,
+                "sequence_code": "CCP",
+                "code": "internal",
+            }
+        )
+        cls.pick_type.return_picking_type_id = cls.cleanup_type
+        cls.pack_type.return_picking_type_id = cls.cleanup_type
+
+    @classmethod
+    def _get_cleanup_picking(cls):
+        return cls.env["stock.picking"].search(
+            [("picking_type_id", "=", cls.cleanup_type.id)]
+        )
+
+    def test_unrelease_picked(self):
+        # In this case, we should get 1 return picking from
+        # WH/PACK to WH/STOCK
+        self._deliver(self.pick_picking)
+        self.ship_picking.unrelease()
+        self.assertTrue(self.ship_picking.need_release)
+        self.assertEqual(self.pack_picking.state, "cancel")
+        self.assertEqual(self.pick_picking.state, "done")
+        cancel_picking = self._get_cleanup_picking()
+        self.assertEqual(len(cancel_picking), 1)
+        self.assertEqual(cancel_picking.location_id, self.loc_pack)
+        self.assertEqual(cancel_picking.location_dest_id, self.loc_stock)
+
+    def test_unrelease_packed(self):
+        # In this case, we should get 1 return picking from
+        # WH/OUT to WH/STOCK
+        self._deliver(self.pick_picking)
+        self._deliver(self.pack_picking)
+        self.ship_picking.unrelease()
+        self.assertTrue(self.ship_picking.need_release)
+        self.assertEqual(self.pack_picking.state, "done")
+        self.assertEqual(self.pick_picking.state, "done")
+        cancel_picking = self._get_cleanup_picking()
+        self.assertEqual(len(cancel_picking), 1)
+        self.assertEqual(cancel_picking.location_id, self.loc_output)
+        self.assertEqual(cancel_picking.location_dest_id, self.loc_stock)
+
+    def test_unrelease_picked_partial(self):
+        qty_picked = [(self.product1, 5.0)]
+        self._deliver(self.pick_picking, product_qty=qty_picked)
+        pick_backorder = self._get_backorder_for_pickings(self.pick_picking)
+        self.assertTrue(pick_backorder)
+        self.ship_picking.unrelease()
+        self.assertTrue(self.ship_picking.need_release)
+        self.assertEqual(self.pack_picking.state, "cancel")
+        self.assertEqual(self.pick_picking.state, "done")
+        cancel_picking = self._get_cleanup_picking()
+        # In the end, we cancelled 5 units for the pick backorder, and returned
+        # 5 units from pack -> stock
+        self.assertEqual(pick_backorder.state, "cancel")
+        self.assertEqual(cancel_picking.location_id, self.loc_pack)
+        self.assertEqual(cancel_picking.location_dest_id, self.loc_stock)
+        self.assertEqual(cancel_picking.move_lines.product_uom_qty, 5.0)
+
+    def test_unrelease_packed_partial(self):
+        self._deliver(self.pick_picking)
+        qty_packed = [(self.product1, 5.0)]
+        self._deliver(self.pack_picking, product_qty=qty_packed)
+        pack_backorder = self._get_backorder_for_pickings(self.pack_picking)
+        self.assertTrue(pack_backorder)
+        self.ship_picking.unrelease()
+        self.assertTrue(self.ship_picking.need_release)
+        self.assertEqual(self.pack_picking.state, "done")
+        self.assertEqual(self.pick_picking.state, "done")
+        cancel_pickings = self._get_cleanup_picking()
+        self.assertEqual(len(cancel_pickings), 2)
+        # In the end, we cancelled 5 units for the pack backorder, returned
+        # 5 units from pack -> stock, and 5 units from output -> stock
+        pack_cancel = cancel_pickings.filtered(lambda p: p.location_id == self.loc_pack)
+        ship_cancel = cancel_pickings.filtered(
+            lambda p: p.location_id == self.loc_output
+        )
+        self.assertEqual(pack_cancel.move_lines.product_uom_qty, 5.0)
+        self.assertEqual(ship_cancel.move_lines.product_uom_qty, 5.0)
+
+    @classmethod
+    def put_in_pack(cls, move):
+        # is it necessary to create stock moves?
+        move._action_assign()
+        pack = cls.env["stock.quant.package"].create({"name": move.name})
+        move.move_line_ids.result_package_id = pack
+        return pack
+
+    def test_unrelease_multiple_moves_same_product(self):
+        # Create a picking with twice the same move
+        product_qty = [
+            (self.product1, 20),
+        ]
+        picking_chain = self._create_picking_chain(self.wh, products=product_qty)
+        ship_picking = self._out_picking(picking_chain)
+        ship_picking.release_available_to_promise()
+        # Creating a second move. Both moves thave the same origin (pack.move_line)
+        self.env["stock.move"].create(ship_picking.move_lines._split(4))
+        pack_picking = self._prev_picking(ship_picking)
+        pick_picking = self._prev_picking(pack_picking)
+        self._deliver(pick_picking)
+        self._deliver(pack_picking)
+        ship_picking.unrelease()
+        cancel_pickings = self._get_cleanup_picking()
+        self.assertEqual(cancel_pickings.move_lines.product_qty, 20)
+
+    def test_unrelease_packed_multi(self):
+        # Pick and pack 2 pickings, unrelease both before shipping
+        # Both have same picking types, goods should be returned
+        # to stock in the same picking
+        ship_no_pack = self.ship_picking
+        pack_no_pack = self.pack_picking
+        pick_no_pack = self.pick_picking
+        # The new picking chain will have packages
+        product_qty = [(self.product1, 10), (self.product2, 10)]
+        picking_chain = self._create_picking_chain(self.wh, products=product_qty)
+        ship_with_pack = self._out_picking(picking_chain)
+        ship_with_pack.release_available_to_promise()
+        pack_with_pack = self._prev_picking(ship_with_pack)
+        pick_with_pack = self._prev_picking(pack_with_pack)
+        # Process pick pickings
+        self._deliver(pick_with_pack)
+        self._deliver(pick_no_pack)
+        # put pack moves in packages on pack_with_pack,
+        pack_moves = pack_with_pack.move_lines
+        pack_move1 = pack_moves.filtered(lambda m: m.product_id == self.product1)
+        pack_move2 = pack_moves.filtered(lambda m: m.product_id == self.product2)
+        self.put_in_pack(pack_move1)
+        self.put_in_pack(pack_move2)
+        # Process pack pickings
+        self._deliver(pack_with_pack)
+        self._deliver(pack_no_pack)
+        # unrelease both ship pickings at once
+        (ship_with_pack | ship_no_pack).unrelease()
+        cancel_pickings = self._get_cleanup_picking()
+        # We should have 1 return picking only
+        self.assertEqual(len(cancel_pickings), 1)
+        # We should have 3 moves
+        cancel_moves = cancel_pickings.move_lines
+        self.assertEqual(len(cancel_moves), 3)
+        # We should have:
+        # - 1 move for product1 without pack
+        # - 1 move for product1 with pack
+        # - 1 move for product2 with pack
+        cancel_move1_no_pack = cancel_moves.filtered(
+            lambda m: m.product_id == self.product1 and not m.move_line_ids.package_id
+        )
+        cancel_move1_with_pack = cancel_moves.filtered(
+            lambda m: m.product_id == self.product1 and m.move_line_ids.package_id
+        )
+        cancel_move2_with_pack = cancel_moves.filtered(
+            lambda m: m.product_id == self.product2 and m.move_line_ids.package_id
+        )
+        self.assertTrue(cancel_move1_no_pack)
+        self.assertEqual(cancel_move1_no_pack.product_qty, 10)
+        self.assertTrue(cancel_move1_with_pack)
+        self.assertEqual(cancel_move1_with_pack.product_qty, 10)
+        self.assertTrue(cancel_move2_with_pack)
+        self.assertEqual(cancel_move2_with_pack.product_qty, 10)
+
+    def test_return_quantity_in_stock(self):
+        move_model = self.env["stock.move"]
+        pack_move = self.pack_picking.move_lines
+        # process pick and pack, so pack is done and returnable
+        self._deliver(self.pick_picking)
+        self._deliver(self.pack_picking)
+        # Using empty_recordsets doesn't raises an exception and doesn't create
+        # a return picking
+        empty_args = {}
+        move_model._return_quantity_in_stock(empty_args)
+        self.assertFalse(self._get_cleanup_picking())
+        # Adding a move with no quantity
+        empty_args = {pack_move.id: 0}
+        move_model._return_quantity_in_stock(empty_args)
+        self.assertFalse(self._get_cleanup_picking())
+        # Adding a quantity should create a return picking
+        valid_args = {pack_move.id: 5}
+        move_model._return_quantity_in_stock(valid_args)
+        return_picking = self._get_cleanup_picking()
+        self.assertEqual(return_picking.move_lines.product_qty, 5)
+
+    def test_unrelease_shipped(self):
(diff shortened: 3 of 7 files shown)


stock_available_to_promise_release/models/stock_location_route.py
```diff
@@ -7,6 +7,14 @@
 class Route(models.Model):
     _inherit = "stock.location.route"
 
+    allow_unrelease_on_cancel = fields.Boolean(
+        string="Reverse done transfer on cancellation",
+        default=False,
+        help=(
+            "If checked, this option will create new inverse internal operations "
+            "on done internal operations"
```
jbaudoux:
```suggestion
            "If checked, unreleasing the delivery may create a new return operation "
            "on the last done pulled transfer. Otherwise, you won't be able to unrelease as soon as one of the pulled transfer is done"
```

stock_available_to_promise_release/models/stock_location_route.py
```diff
@@ -7,6 +7,14 @@
 class Route(models.Model):
     _inherit = "stock.location.route"
 
+    allow_unrelease_on_cancel = fields.Boolean(
```
jbaudoux: The fact that the action cancel triggers unrelease is not the matter here.
It is the fact that unrelease creates a return of done moves.
```suggestion
    allow_unrelease_return_done_move = fields.Boolean(
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -622,6 +670,30 @@ def unrelease(self, safe_unrelease=False):
                     origin_moves.write({"propagate_cancel": False})
                     # origin_moves._action_cancel()
                     moves_to_cancel |= origin_moves
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_cancelled_at_step = sum(origin_moves.mapped("product_uom_qty"))
```
jbaudoux: Compute in the uom of the product as each move could have a different uom
```suggestion
                qty_cancelled_at_step = sum(origin_moves.mapped("product_qty"))
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,13 +644,22 @@ def unrelease(self, safe_unrelease=False):
         impacted_picking_ids = set()
 
         for move in moves_to_unrelease:
+            # When a move is returned, it is going straight to WH/Stock,
+            # skipping all intermediate zones (pick/pack).
+            # That is why we need to keep track of qty returned along the way.
+            # We do not want to return the same goods at each step.
+            # At a given step (pick/pack/ship), qty to return is
+            # move.product_uom_qty - cancelled_qty_at_step - already returned qties
+            qty_to_unrelease = move.product_uom_qty
```
jbaudoux: In the uom of the product as each move could have a different uom
```suggestion
            qty_to_unrelease = move.product_qty
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,13 +644,22 @@ def unrelease(self, safe_unrelease=False):
...
+            qty_returned_for_move = 0
             iterator = move._get_chained_moves_iterator("move_orig_ids")
             moves_to_cancel = self.env["stock.move"]
             # backup procure_method as when you don't propagate cancel, the
             # destination move is forced to make_to_stock
             procure_method = move.procure_method
             next(iterator)  # skip the current move
             for origin_moves in iterator:
+                done_moves = origin_moves.filtered(lambda m: m.state == "done")
                 origin_moves = origin_moves.filtered(
```
jbaudoux: When calling `_split_origins`, you need to give the remaining quantity that needs to be split
`move._split_origins(origin_moves, qty_to_unrelease)`


def _split_origins(self, origins, **qty**):
  ~~qty = self.product_qty~~

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,13 +644,22 @@ def unrelease(self, safe_unrelease=False):
...
 
         for move in moves_to_unrelease:
+            # When a move is returned, it is going straight to WH/Stock,
+            # skipping all intermediate zones (pick/pack).
+            # That is why we need to keep track of qty returned along the way.
+            # We do not want to return the same goods at each step.
+            # At a given step (pick/pack/ship), qty to return is
+            # move.product_uom_qty - cancelled_qty_at_step - already returned qties
+            qty_to_unrelease = move.product_uom_qty
+            qty_returned_for_move = 0
```
jbaudoux, line 714: I would drop this and just decrease `qty_to_unrelease` on each iteration

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -622,6 +670,30 @@ def unrelease(self, safe_unrelease=False):
...
                     moves_to_cancel |= origin_moves
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_cancelled_at_step = sum(origin_moves.mapped("product_uom_qty"))
+                qty_to_return_at_step = (
+                    qty_to_unrelease - qty_cancelled_at_step - qty_returned_for_move
+                )
+                # TODO check move.rule_id.allow_unrelease_on_cancel ?
+                if not qty_to_return_at_step:
```
jbaudoux: use float_compare

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -622,6 +670,30 @@ def unrelease(self, safe_unrelease=False):
...
                     # origin_moves._action_cancel()
                     moves_to_cancel |= origin_moves
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_cancelled_at_step = sum(origin_moves.mapped("product_uom_qty"))
+                qty_to_return_at_step = (
+                    qty_to_unrelease - qty_cancelled_at_step - qty_returned_for_move
+                )
+                # TODO check move.rule_id.allow_unrelease_on_cancel ?
```
jbaudoux: No need before the if nothing to return
but after I would raise an error if we need to return but the option is not enabled

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -622,6 +670,30 @@ def unrelease(self, safe_unrelease=False):
...
+                if not qty_to_return_at_step:
+                    continue
+                # Multiple pickings can satisfy a move
+                # -> len(move.move_orig_ids.picking_id) > 1
+                # Group done_moves per picking, and create returns
+                groups_iterator = groupby(done_moves, key=lambda m: m.picking_id)
```
jbaudoux: in `origin_moves = move._split_origins(origin_moves)`, I would put the result in another variable so that original `origin_moves` is not lost. Then you can define here `done_moves`

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -622,6 +670,30 @@ def unrelease(self, safe_unrelease=False):
...
+                for __, moves_list in groups_iterator:
+                    moves_to_return = self.browse([move.id for move in moves_list])
+                    returned_qty = moves_to_return._return_quantity_in_stock(
+                        qty_to_return_at_step
+                    )
+                    qty_returned_for_move += returned_qty
```
jbaudoux: I would drop this

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -622,6 +670,30 @@ def unrelease(self, safe_unrelease=False):
...
+                    qty_to_return_at_step -= returned_qty
+                    if qty_to_return_at_step == 0:
+                        break
+
```
jbaudoux, line 766: and do this outside of for loop
`qty_to_unrelease = qty_cancelled_at_step`

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -622,6 +670,30 @@ def unrelease(self, safe_unrelease=False):
...
+                # Group done_moves per picking, and create returns
+                groups_iterator = groupby(done_moves, key=lambda m: m.picking_id)
+                for __, moves_list in groups_iterator:
+                    moves_to_return = self.browse([move.id for move in moves_list])
+                    returned_qty = moves_to_return._return_quantity_in_stock(
+                        qty_to_return_at_step
+                    )
+                    qty_returned_for_move += returned_qty
+                    qty_to_return_at_step -= returned_qty
+                    if qty_to_return_at_step == 0:
```
jbaudoux: use float_compare

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
             yield moves
             moves = moves.mapped(chain_field)
 
+    def _return_quantity_in_stock(self, quantity_to_return):
+        picking = self.picking_id
+        picking.ensure_one()
+
+        returned_quantity = 0
+        # create return
+        return_type = picking.picking_type_id.return_picking_type_id
```
jbaudoux: nice to have: raise error if no return_type configured

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
...
+        wiz_values = {
+            "picking_id": picking.id,
+            "original_location_id": picking.location_dest_id.id,
```
jbaudoux: This is not necessary. Otherwise, it should be
```suggestion
            "original_location_id": picking.location_id.id,
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                continue
+            move_qty = min(quantity_to_return, move.quantity_done)
+            return_move_vals = {
+                "product_id": move.product_id.id,
+                "quantity": move_qty,
+                "uom_id": move.product_id.uom_id.id,
+                "move_id": move.id,
+            }
+            product_return_moves.append((0, 0, return_move_vals))
+            returned_quantity += move_qty
```
jbaudoux: The `quantity_to_return` should decrease on each iteration

jbaudoux (changes requested):
Need to adapt README as well

jbaudoux:
cc @rousseldenis @lmignon

stock_available_to_promise_release/models/stock_move.py
mmequignon: Well, it's done in upper method
https://github.com/OCA/wms/pull/971/files#diff-fca2e9865c2d75a9793dbe699bffdf976606d996b69d09cacd25660f2926e998R693
Thing is that it's not nice to modify an argument inside a method, which is why I'm adding a new variable `returned_qty` which I compare with arg.
In the end, I return the number of units that were returned, and deduct this number in the wrapping method.
This is where the `quantity_to_return` counter is.

jbaudoux: Yes but I mean this line `move_qty = min(quantity_to_return, move.quantity_done)`
you forgot to substract the already processed qty by previous iterations. Should be `move_qty = min(quantity_to_return - returned_quantity, move.quantity_done)`

stock_available_to_promise_release/models/stock_move.py
mmequignon: You want me to modify the implementation of `_split_origins` ?

stock_available_to_promise_release/models/stock_move.py
mmequignon: We have to talk about this, because there's actually 2 counters here.
let's consider this case:
Picked 20 units, packed 10 units.
When unreleasing, we unrelease pack first, returning 10 units back to stock.
Then, we unrelease pick, and we have to be aware of those 10 units we just returned, otherwise we will return 20 units, and this is bad.
Which is why I'm having 2 counters, a global one, and a per move one.
I might not understand what you mean, let's discuss this.

stock_available_to_promise_release/models/stock_move.py
jbaudoux: there is no choice since you may decrease the quantity on each iteration

stock_available_to_promise_release/models/stock_move.py
mmequignon: Well, we need to have this compatible with both
If we have 2 routes one with setting enabled, one with setting disabled, do we want to raise an exception?
To me we shouldn't raise an exception, but do nothing if setting is disabled.
What do you think?

jbaudoux: If you multiple routes, only one is taken into consideration. There is only one rule_id on the move. Only that one matters

jbaudoux: if the setting is disabled, then you should not reach the part where you need to return a move. This was already checked beforehand in the check method. Nevertheless, it is good to raise same error here for safety

stock_available_to_promise_release/models/stock_move.py
jbaudoux: @mmequignon Can you drop this line?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
...
+        # create return
+        return_type = picking.picking_type_id.return_picking_type_id
+        wiz_values = {
+            "picking_id": picking.id,
+            "original_location_id": picking.location_dest_id.id,
+            "location_id": return_type.default_location_dest_id.id,
+        }
+        product_return_moves = []
+        for move in self:
+            if not move.state == "done":
```
```suggestion
            if move.state != "done":
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                continue
+            move_qty = min(quantity_to_return - returned_quantity, move.quantity_done)
```
jbaudoux: Process real quantities and not uom quantities. You can use `product_qty` as the move is done
```suggestion
            move_qty = min(quantity_to_return - returned_quantity, move.product_qty)
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
...
+            return_move_vals = {
+                "product_id": move.product_id.id,
+                "quantity": move_qty,
+                "uom_id": move.product_id.uom_id.id,
+                "move_id": move.id,
+            }
+            product_return_moves.append((0, 0, return_move_vals))
+            returned_quantity += move_qty
+            if returned_quantity == quantity_to_return:
```
jbaudoux: use float compare

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
             yield moves
             moves = moves.mapped(chain_field)
 
+    def _return_quantity_in_stock(self, quantity_to_return):
```
```suggestion
    def _return_quantity_in_stock(self, quantity_to_return):
        """Return a quantity from a list of moves.
        
        The quantity to return is in the product uom"""
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,39 @@ def _get_chained_moves_iterator(self, chain_field):
...
+        picking = self.picking_id
+        picking.ensure_one()
```
```suggestion
        picking.ensure_one()
        move.product_id.ensure_one()
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,24 +644,66 @@ def unrelease(self, safe_unrelease=False):
...
                 if origin_moves:
-                    origin_moves = move._split_origins(origin_moves)
-                    impacted_picking_ids.update(origin_moves.mapped("picking_id").ids)
+                    qty_to_split = qty_to_unrelease - qty_returned_for_move
+                    # TODO find new name
+                    moves_to_cancel = move._split_origins(
+                        origin_moves, qty=qty_to_split
+                    )
+                    impacted_picking_ids.update(
+                        moves_to_cancel.mapped("picking_id").ids
```
jbaudoux: No need for mapped
```suggestion
                        moves_to_cancel.picking_id.ids
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,24 +644,66 @@ def unrelease(self, safe_unrelease=False):
...
+                    )
                     # avoid to propagate cancel to the original move
-                    origin_moves.write({"propagate_cancel": False})
+                    moves_to_cancel.write({"propagate_cancel": False})
                     # origin_moves._action_cancel()
```
jbaudoux: remove this obsolete commented line

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,24 +644,66 @@ def unrelease(self, safe_unrelease=False):
...
+                # Multiple pickings can satisfy a move
+                # -> len(move.move_orig_ids.picking_id) > 1
+                # Group done_moves per picking, and create returns
+                groups_iterator = groupby(done_moves, key=lambda m: m.picking_id)
+                for __, moves_list in groups_iterator:
+                    moves_to_return = self.browse([move.id for move in moves_list])
+                    returned_qty = moves_to_return._return_quantity_in_stock(
+                        qty_to_return_at_step
+                    )
+                    qty_returned_for_move += returned_qty
```
jbaudoux: I would move this before the block just after the `if float_is_zero`
`qty_returned_for_move += qty_to_return_at_step`

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,24 +644,66 @@ def unrelease(self, safe_unrelease=False):
...
+                done_moves = origin_moves.filtered(lambda m: m.state == "done")
                 origin_moves = origin_moves.filtered(
                     lambda m: m.state not in ("done", "cancel")
                 )
+                qty_cancelled_at_step = 0
                 if origin_moves:
-                    origin_moves = move._split_origins(origin_moves)
-                    impacted_picking_ids.update(origin_moves.mapped("picking_id").ids)
+                    qty_to_split = qty_to_unrelease - qty_returned_for_move
+                    # TODO find new name
```
jbaudoux: You can drop this TODO, you fixed it already

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,24 +644,66 @@ def unrelease(self, safe_unrelease=False):
...
+            qty_returned_for_move = 0
             iterator = move._get_chained_moves_iterator("move_orig_ids")
-            moves_to_cancel = self.env["stock.move"]
+            moves_to_cancel_for_move = self.env["stock.move"]
             # backup procure_method as when you don't propagate cancel, the
             # destination move is forced to make_to_stock
             procure_method = move.procure_method
             next(iterator)  # skip the current move
             for origin_moves in iterator:
+                done_moves = origin_moves.filtered(lambda m: m.state == "done")
```
jbaudoux: Move this variable down where it is used

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,24 +644,66 @@ def unrelease(self, safe_unrelease=False):
...
+                    moves_to_cancel_for_move |= moves_to_cancel
+                    qty_cancelled_at_step = sum(moves_to_cancel.mapped("product_qty"))
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_to_return_at_step = (
+                    qty_to_unrelease - qty_cancelled_at_step - qty_returned_for_move
+                )
+                if float_is_zero(qty_to_return_at_step, precision_rounding=rounding):
+                    continue
```
```suggestion
                if float_is_zero(qty_to_return_at_step, precision_rounding=rounding):
                    continue
                elif not move.rule_id.allow_unrelease_return_done_move:
                    raise UserError(_(f"You cannot unrelease the move {move.name} because some origin moves ', '.join(done_moves.mapped("name"))} are done"))
```

stock_available_to_promise_release/models/stock_move.py
```suggestion
        return_type = picking.picking_type_id.return_picking_type_id
        if not return_type:
            raise UserError(_(f"The operation {picking.name} is done and cannot be returned"))
```

jbaudoux (commented):
Some minor remarks

stock_available_to_promise_release/models/stock_move.py
jbaudoux: without using f-string to not give code in the translations

stock_available_to_promise_release/models/stock_move.py
jbaudoux: without using f-string to not give code in the translations

stock_available_to_promise_release/models/stock_move.py
mmequignon: I don't think we should.
Break only breaks this loop, but this `qty_returned_for_move` must be up to date for the outter one, which relies on this variable to determine how many remaining units has to be returned.
Maybe I'm missing something.

stock_available_to_promise_release/tests/test_unrelease_cancel.py
mt-software-de, line 1: This new test is not in the __init__.py

mt-software-de (changes requested):

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -605,24 +663,77 @@ def unrelease(self, safe_unrelease=False):
...
+                            "You cannot unrelease the move %(move_name)s "
+                            "because some origin moves %(done_move_names)s are done"
+                        ),
+                        **msg_args
+                    )
+                    raise UserError(message)
+                # Multiple pickings can satisfy a move
+                # -> len(move.move_orig_ids.picking_id) > 1
+                # Group done_moves per picking, and create returns
+                groups_iterator = groupby(done_moves, key=lambda m: m.picking_id)
```
jbaudoux: As shopfloor can create split order, we should create one return per picking type to ensure that all products are in the same return picking. And if the products are in packages, then we face a bug in odoo reservation
```suggestion
                groups_iterator = groupby(done_moves, key=lambda m: m.picking_type_id)
```

jbaudoux (changes requested):

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,58 @@ def _get_chained_moves_iterator(self, chain_field):
...
+        rounding = uom.rounding
+
+        returned_quantity = 0
+        # If no return type set on the picking type, we cannot know where
+        # to return goods at.
+        # Raise an exception.
+        return_type = picking.picking_type_id.return_picking_type_id
+        if not return_type:
+            message = _(
+                "The operation picking_name} is done and cannot be returned",
```
jbaudoux: typo
```suggestion
                "The operation {picking_name} is done and cannot be returned",
```

stock_available_to_promise_release/models/stock_move.py
jbaudoux: I tested and it's already better. Still we end-up with one return per move because of the loop on the out moves to unrelease.
The solution would be to not call `moves_to_return._return_quantity_in_stock` here but record each move and quantity to return by picking type. Then at the end, outside the out moves loop, you can call `moves_to_return._return_quantity_in_stock` for all the moves & qty per picking type.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -604,27 +657,82 @@ def unrelease(self, safe_unrelease=False):
...
+                            "because some origin moves %(done_move_names)s are done"
+                        ),
+                        **msg_args
+                    )
+                    raise UserError(message)
+                # Multiple pickings can satisfy a move
+                # -> len(move.move_orig_ids.picking_id) > 1
+                # Group done_moves per picking, and create returns
+                groups_iterator = groupby(done_moves, key=lambda m: m.picking_id)
+                for picking, moves_list in groups_iterator:
```
```suggestion
                for picking_type, moves_list in groups_iterator:
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -604,27 +657,82 @@ def unrelease(self, safe_unrelease=False):
...
+                    moves_to_return = self.browse([move.id for move in moves_list])
+                    qty_to_return_per_move = moves_to_return._get_returneable_qty_per_move(qty_to_return_at_step)
+                    returneable_qty = sum(qty_to_return_per_move.values())
+                    moves_to_return_per_pick.setdefault(picking, {}).update(qty_to_return_per_move)
```
```suggestion
                    moves_to_return_per_pick.setdefault(picking_type, {}).update(qty_to_return_per_move)
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,64 @@ def _get_chained_moves_iterator(self, chain_field):
...
+            return_type = picking_type.return_picking_type_id
+            wiz_values = {
+                "picking_id": picking.id,
+                "location_id": return_type.default_location_dest_id.id,
+            }
+            product_return_moves = []
+            if not return_type:
+                message = _(
+                    "The operation {picking_name} is done and cannot be returned",
+                    picking_name=picking.name,
```
```suggestion
                    picking_name=", ".join(moves.picking_id.mapped("name")),
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,64 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                product_return_moves.append((0, 0, return_move_vals))
+            if product_return_moves:
+                wiz_values["product_return_moves"] = product_return_moves
+                return_wiz = self.env["stock.return.picking"].create(wiz_values)
+                return_wiz.create_returns()
+        return True
+
+    def _get_returneable_qty_per_move(self, qty_to_return):
+        res = {}
+        for move in self.filtered(lambda m: m.state == "done"):
```
jbaudoux: you are already processing done moves
```suggestion
        for move in self:
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +597,64 @@ def _get_chained_moves_iterator(self, chain_field):
...
+            rounding = move.product_id.uom_id.rounding
+            returneable_qty = min(qty_to_return, move.product_qty)
```
jbaudoux: As multiple out moves could have the same orig move and it's looping on out moves, it could be that the orig move was already considered in the previous loop
So you need to decrease `move.product_qty` by what is already counted in moves_to_return_per_type for that move

jbaudoux (changes requested):

stock_available_to_promise_release/models/stock_move.py
mmequignon: This method is helpful and could be used elsewhere, we need to add some restrictions.
Reusability at the cost of an extra check is good, IMO.
I'll drop this if you insist, but I kind of disagree.

stock_available_to_promise_release/models/stock_move.py
mmequignon: Just pushed a fix for that

jbaudoux: I don't see it

jbaudoux:
can you rebase to include https://github.com/OCA/wms/pull/993 ?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -591,6 +598,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+    def _return_quantity_in_stock(self, qty_to_return_per_move):
+        """Return a quantity from a list of moves.
+
+        The quantity to return is in the product uom"""
+        moves_to_return = self.browse([m_id for m_id in qty_to_return_per_move.keys()])
+        moves_per_type = groupby(moves_to_return, lambda m: m.picking_type_id)
+        for picking_type, moves_list in moves_per_type:
+            moves = self.browse()
+            for move in moves_list:
+                moves |= move
```
```suggestion
            moves = self.browse().concat(*moves_list)
```

mmequignon:
@jbaudoux done

jbaudoux (approved):
Looking good now :)

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -100,8 +101,14 @@ def _is_unrelease_allowed_on_origin_moves(self, origin_moves):
             # The picking is printed, we can't unrelease the move
             # because the processing of the origin moves is started.
             return False
+        # If allow_unrelease_on_cancel is not checked, only release assigned/waiting
```
jbaudoux, line 108: In the test above, you now have to exclude the cancel/done when checking if anything printed

jbaudoux (changes requested):

stock_available_to_promise_release/models/stock_move.py
mmequignon: Shouldn't we use the same states as for origin_moves after?
```python3
        # If allow_unrelease_on_cancel is not checked, only release assigned/waiting
        # moves. If checked, also unrelease done moves.
        if self.rule_id.allow_unrelease_return_done_move:
            unreleasable_states = ("cancel",)
        else:
            unreleasable_states = ("done", "cancel")
```

jbaudoux: no, printed is only relevant for open transfers

jbaudoux: it's not right your check as it can be done but not printed.

mmequignon: fixed

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -96,12 +97,22 @@ def _is_unrelease_allowed_on_origin_moves(self, origin_moves):
...
-        if pickings.filtered("printed"):
-            # The picking is printed, we can't unrelease the move
-            # because the processing of the origin moves is started.
+        unreleasable_pickings = pickings.filtered(
+            lambda p: p.printed and p.state not in ("cancel", "done")
+        )
+        if unreleasable_pickings:
+            # The picking is either printed or done/cancelled.
+            # We can't unrelease the move because the processing of the origin moves
+            # is either started or done.
```
```suggestion
        ongoing_pickings = pickings.filtered(
            lambda p: p.printed and p.state not in ("cancel", "done")
        )
        if ongoing_pickings:
            # The picking is printed and not done/cancelled. We can't unrelease.
```

jbaudoux (commented):
Variable naming and comment is confusing as it is NOT unreleasable

Code is correct

stock_available_to_promise_release/models/stock_move.py
mmequignon: I agree

jbaudoux (approved):

jbaudoux:
@mt-software-de Can you update your review?

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -616,27 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
                 )
-                if origin_moves:
-                    origin_moves = move._split_origins(origin_moves)
-                    impacted_picking_ids.update(origin_moves.mapped("picking_id").ids)
+                qty_cancelled_at_step = 0
+                if todo_origin_moves:
+                    qty_to_split = qty_to_unrelease - qty_returned_for_move
+                    moves_to_cancel = move._split_origins(
+                        todo_origin_moves, qty=qty_to_split
+                    )
```
mt-software-de: Could it happen that qty_to_split is zero? If so the qty_to_split would be set to the move's product_qty in method _split_origins. I think that would not be correct or?
```suggestion
                    qty_to_split = qty_to_unrelease - qty_returned_for_move
                    if not qty_to_split:
                        continue
                    moves_to_cancel = move._split_origins(
                        todo_origin_moves, qty=qty_to_split
                    )
```

mt-software-de (changes requested):

stock_available_to_promise_release/models/stock_move.py
jbaudoux: i would also rename `qty_to_split` into `qty_to_cancel` for clarity and do that check as first step in the loop. If nothing left to cancel, you can even `break`

```
            for origin_moves in iterator:
                qty_to_cancel = qty_to_unrelease - qty_returned_for_move
                if float_compare(qty_to_cancel, 0, precision_rounding=move.product_uom.rounding) <= 0:
                    break
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -616,27 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
+                    impacted_picking_ids.update(moves_to_cancel.picking_id.ids)
```
jbaudoux: doesn't look used

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -616,27 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
             next(iterator)  # skip the current move
             for origin_moves in iterator:
-                origin_moves = origin_moves.filtered(
+                todo_origin_moves = origin_moves.filtered(
                     lambda m: m.state not in ("done", "cancel")
                 )
-                if origin_moves:
-                    origin_moves = move._split_origins(origin_moves)
-                    impacted_picking_ids.update(origin_moves.mapped("picking_id").ids)
+                qty_cancelled_at_step = 0
```
jbaudoux: en_US... :)
```suggestion
                qty_canceled_at_step = 0
```

jbaudoux: or even just `qty_canceled`

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -616,27 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
-            moves_to_cancel._action_cancel()
+                    moves_to_cancel.write({"propagate_cancel": False})
+                    moves_to_cancel_for_move |= moves_to_cancel
+                    qty_cancelled_at_step = sum(moves_to_cancel.mapped("product_qty"))
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_to_return_at_step = (
+                    qty_to_unrelease - qty_cancelled_at_step - qty_returned_for_move
+                )
```
jbaudoux: not changing anything, just more readable
```suggestion
                qty_to_return = qty_to_cancel - qty_canceled
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -616,27 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
+                done_moves = origin_moves.filtered(lambda m: m.state == "done")
+                if float_is_zero(qty_to_return_at_step, precision_rounding=rounding):
+                    continue
```
jbaudoux, line 742: Move that test one line up just after computing the variable and before building `done_moves`

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -616,27 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
+                    moves_to_cancel.write({"propagate_cancel": False})
+                    moves_to_cancel_for_move |= moves_to_cancel
+                    qty_cancelled_at_step = sum(moves_to_cancel.mapped("product_qty"))
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_to_return_at_step = (
+                    qty_to_unrelease - qty_cancelled_at_step - qty_returned_for_move
+                )
+                done_moves = origin_moves.filtered(lambda m: m.state == "done")
```
jbaudoux: for clarity as you have `todo_origin_moves`
```suggestion
                done_origin_moves = origin_moves.filtered(lambda m: m.state == "done")
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,79 @@ def _get_chained_moves_iterator(self, chain_field):
...
+            )
+            if float_is_zero(move_returneable_qty, precision_rounding=rounding):
+                continue
+            # Update the quantity
+            qty_to_return_per_move[move.id] += move_returneable_qty
+            qty_to_return -= move_returneable_qty
+            returneable_qty += move_returneable_qty
+            if float_is_zero(qty_to_return, precision_rounding=rounding):
+                break
+        return returneable_qty
```
jbaudoux: We have to return the complete `qty_to_return` otherwise there is an issue
```suggestion
            if float_is_zero(qty_to_return, precision_rounding=rounding):
                break
        else:
             raise UserError("Unrelease cannot return the required quantity")
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -616,27 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
+                        **msg_args
+                    )
+                    raise UserError(message)
+                # Multiple pickings can satisfy a move
+                # -> len(move.move_orig_ids.picking_id) > 1
+                # Group done_moves per picking, and create returns
+                returneable_qty = done_moves._get_returneable_qty_per_move(
+                    qty_to_return_at_step, qty_to_return_per_move
+                )
+                qty_returned_for_move += returneable_qty
```
jbaudoux: I would rename `_get_...` method name as it is not getting anything but more setting something
```suggestion
                done_origin_moves._unrelease_set_qty_to_return_per_move(
                    qty_to_return(_at_step), qty_to_return_per_move
                )
                qty_returned_for_move += qty_to_return(_at_step)
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,79 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                continue
+            return_type = picking_type.return_picking_type_id
+            wiz_values = {
+                "picking_id": fields.first(pickings).id,
+                "location_id": return_type.default_location_dest_id.id,
+            }
+            product_return_moves = []
+            if not return_type:
+                message = _(
+                    "The operation {picking_names} is done and cannot be returned",
```
```suggestion
                    "The operation %(picking_names)s is done and cannot be returned",
```

jbaudoux (changes requested):
I left some small remarks mainly for code clarity

stock_available_to_promise_release/models/stock_move.py
mmequignon: your version isn't accurate, as it is possible that a current step `qty_to_return(_at_step) > sum(done_origin_moves.mapped(product_qty))`
In this case, `returneable_qty < qty_to_return(_at_step)` and the difference would be returned at next iteration.

The main goal is to get `returneable_qty` in order to keep `qty_returned_for_move` up to date, but qty_to_return isn't right.
I don't mind the naming of the method, though.

stock_available_to_promise_release/models/stock_move.py
jbaudoux: it's maybe not a good idea my proposal in case someone cancel the pack before the ship

jbaudoux (approved):
@mt-software-de LG now

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+            if product_return_moves:
+                wiz_values["product_return_moves"] = product_return_moves
+                return_wiz = self.env["stock.return.picking"].create(wiz_values)
+                return_wiz.create_returns()
+        return True
+
+    def _unrelease_set_returneable_qty_per_move(
+        self, qty_to_return, qty_to_return_per_move
+    ):
+        returneable_qty = 0
```
mt-software-de:
```suggestion
        returnable_qty = 0
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+            )
+            if float_is_zero(move_returneable_qty, precision_rounding=rounding):
+                continue
+            # Update the quantity
+            qty_to_return_per_move[move.id] += move_returneable_qty
+            qty_to_return -= move_returneable_qty
+            returneable_qty += move_returneable_qty
+            if float_is_zero(qty_to_return, precision_rounding=rounding):
+                break
+        return returneable_qty
```
mt-software-de:
```suggestion
        return returnable_qty
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                    "move_id": move.id,
+                }
+                product_return_moves.append((0, 0, return_move_vals))
+            if product_return_moves:
+                wiz_values["product_return_moves"] = product_return_moves
+                return_wiz = self.env["stock.return.picking"].create(wiz_values)
+                return_wiz.create_returns()
+        return True
+
+    def _unrelease_set_returneable_qty_per_move(
```
mt-software-de:
```suggestion
    def _unrelease_set_returnable_qty_per_move(
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+        self, qty_to_return, qty_to_return_per_move
+    ):
+        returneable_qty = 0
+        for move in self:
+            rounding = move.product_id.uom_id.rounding
+            # As a move might have multiple dest ids, we might have
+            # already planned to return a few units already.
+            # Get it, and deduce it from the returneable qty
```
mt-software-de:
```suggestion
            # Get it, and deduce it from the returnable qty
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+            move_qty_planned = qty_to_return_per_move.get(move.id, 0)
+            # A move might already have return moves linked to it, deduce their quantity
+            move_returned_qty = sum(
+                move.returned_move_ids.filtered(lambda m: m.state != "cancel").mapped(
+                    "product_qty"
+                )
+            )
+            move_returneable_qty = min(
```
mt-software-de:
```suggestion
            move_returnable_qty = min(
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                qty_to_return, move.product_qty - move_returned_qty - move_qty_planned
+            )
+            if float_is_zero(move_returneable_qty, precision_rounding=rounding):
```
mt-software-de:
```suggestion
            if float_is_zero(move_returnable_qty, precision_rounding=rounding):
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                continue
+            # Update the quantity
+            qty_to_return_per_move[move.id] += move_returneable_qty
+            qty_to_return -= move_returneable_qty
+            returneable_qty += move_returneable_qty
```
mt-software-de:
```suggestion
            qty_to_return_per_move[move.id] += move_returnable_qty
            qty_to_return -= move_returnable_qty
            returneable_qty += move_returnable_qty
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -614,29 +700,74 @@ def unrelease(self, safe_unrelease=False):
...
+                        **msg_args
+                    )
+                    raise UserError(message)
+                # Multiple pickings can satisfy a move
+                # -> len(move.move_orig_ids.picking_id) > 1
+                # Group done_moves per picking, and create returns
+                returneable_qty = done_moves._unrelease_set_returneable_qty_per_move(
+                    qty_to_return, qty_to_return_per_move
+                )
+                qty_returned_for_move += returneable_qty
```
mt-software-de:
```suggestion
                returnable_qty = done_moves._unrelease_set_returnable_qty_per_move(
                    qty_to_return, qty_to_return_per_move
                )
                qty_returned_for_move += returnable_qty
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
             moves = moves.mapped(chain_field)
 
+    def _return_quantity_in_stock(self, qty_to_return_per_move):
+        """Return a quantity from a list of moves.
+
+        The quantity to return is in the product uom"""
+        moves_to_return = self.browse([m_id for m_id in qty_to_return_per_move.keys()])
+        moves_per_type = groupby(moves_to_return, lambda m: m.picking_type_id)
+        for picking_type, moves_list in moves_per_type:
+            moves = self.browse().concat(*moves_list)
```
mt-software-de: It is better to use union, concat would list a same move twice. 
```suggestion
            moves = self.browse().union(*moves_list)
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -603,6 +614,81 @@ def _get_chained_moves_iterator(self, chain_field):
...
+                    picking_names=", ".join(pickings.mapped("name")),
+                )
+                raise UserError(message)
+            for move in moves:
+                # Cannot return an unprocessed move
+                if move.state != "done":
+                    continue
+                product = move.product_id
+                uom = product.uom_id
+                qty_to_return = qty_to_return_per_move.get(move.id, 0)
```
mt-software-de, line 646: Since qty_to_return_per_move is a defaultdict, you don't need `get`
```suggestion
                qty_to_return = qty_to_return_per_move[move.id]
```

mt-software-de (changes requested):
Only some minor remarks and some typos. 
Otherwise than that it looks great. Thx

stock_available_to_promise_release/models/stock_move.py
mmequignon: Hmmm, I see what you mean, but I don't fully agree.
Yes, it is a `defaultdict` here, but nothing prevents another module to use this method, for instance.
`self.env["stock.move"].browse([1,2,3])._return_quantity_in_stock({1: 10, 2: 10})`
This is valid, kinda useful, and would raise an exception.

mmequignon: ping @jbaudoux

mt-software-de: Fine for me.

mt-software-de (approved):

PR #972, opened 2025-01-14
[IMP][14.0] shopfloor delivery: Better messages

mmequignon:
Now that return moves are possibly being created when a picking is unreleased, display a nice message to the user when trying to process a product that should be returned in stock.

Based on https://github.com/OCA/wms/pull/971

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -216,6 +216,15 @@ class MessageAction(Component):
     def already_done(self):
         return {"message_type": "info", "body": _("Operation already processed.")}
 
+    def transfer_cancelled(self):
+        return {
+            "message_type": "info",
+            "body": _(
+                "Transfer has been cancelled. "
+                "This cannot be processed using this scenario"
+            ),
+        }
+
     def move_already_done(self):
         return {"message_type": "warning", "body": _("Move already processed.")}
 
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -961,3 +961,13 @@ class MessageAction(Component):
             "message_type": "error",
             "body": _("Unable to find a line with the same product but different lot."),
         }
+
+    def picking_type_is_return(self, picking):
+        body = _("Reserved for %(picking_type)s %(picking_name)s") % {
+            "picking_type": picking.picking_type_id.name,
+            "picking_name": picking.name,
+        }
+        return {
+            "message_type": "error",
+            "body": body,
+        }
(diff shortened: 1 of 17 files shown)


shopfloor/actions/message.py
```diff
@@ -216,6 +216,15 @@ def confirm_pack_moved(self):
     def already_done(self):
         return {"message_type": "info", "body": _("Operation already processed.")}
 
+    def transfer_cancelled(self):
```
jbaudoux, line 219: prefer en_US

```suggestion
    def transfer_canceled(self):
```

shopfloor/actions/message.py
```diff
@@ -216,6 +216,15 @@ def confirm_pack_moved(self):
...
+        return {
+            "message_type": "info",
+            "body": _(
+                "Transfer has been cancelled. "
```
jbaudoux, line 223:
```suggestion
                "Transfer has been canceled. "
```

shopfloor/services/service.py
```diff
@@ -46,21 +46,34 @@ def search_move_line(self):
...
+            return self.msg_store.stock_picking_not_available(fields.first(pickings))
+
+    def _check_picking_processible(self, pickings, states=("assigned",)):
+        """Check if given pickings can be processed"""
+        message = self._check_picking_status(pickings, states=states)
+        if message:
+            return message
+        message = self._check_picking_type(pickings)
+        if message:
+            return message
```
jbaudoux, line 76: You should first check the picking type and then call _check_picking_status on the pickings with the right picking type.
Maybe _check_picking_type can return the list of filtered pickings and raise error if that list is empty.
Then call _check_picking_status on that result.

This is to ensure that you check the status on the pickings to process and not on a picking with a wrong picking type.

jbaudoux (changes requested):
Can you also drop the commits from the other PR from this one and add a `test-requirement.txt` to depend on the PR

shopfloor/services/service.py
jbaudoux: And extract `if not pickings.exists()` from `_check_picking_status` as it's not a status. This should be the first test. Then test picking type. And then test picking status.

jbaudoux: And extract `if not pickings.exists()` from `_check_picking_status` as it's not a status. This should be the first test. Then test picking type. And then test picking status.

mmequignon:
closed in favor of closed in favor of https://github.com/OCA/wms/pull/987

PR #973, opened 2025-01-15
[14][IMP] shopfloor: Ensure destination location on moves for allow move create

mt-software-de:
Ensure that the destination gets also set on the move. When a move is created within the single pack or single product transfer scenario, the move ends up with the default destination location of the operation type. 
To ensure that the destination location of the move gets also set, we need to set it also on the move not only on the move.line. 

cc @jbaudoux @i-vyshnevska


jbaudoux (commented):
What's the purpose of sync_src_location ?

mt-software-de:
> What's the purpose of sync_src_location ?
I updated the description.

jbaudoux:
> Without syncing the src location, the moves will end up with the warehouse view location as src location and with actual scanned location as dest location.

@mt-software-de Isn't it something that should be tackled at move creation? When you scan the source location and product/package to move, the move should be created on that location and not on the picking type source location. Same if you plan the moves, should be planned on the right source location.

mt-software-de:
> > Without syncing the src location, the moves will end up with the warehouse view location as src location and with actual scanned location as dest location.
> 
> @mt-software-de Isn't it something that should be tackled at move creation? When you scan the source location and product/package to move, the move should be created on that location and not on the picking type source location. Same if you plan the moves, should be planned on the right source location.

I checked it again the source location was never be the problem. I was on allow move create the dest location on the move which was missing. And only for the single product transfer because there are no package level.

shopfloor/actions/stock.py
```diff
@@ -228,3 +229,26 @@ def no_putaway_available(self, picking_types, move_lines):
...
+    def _lock_lines(self, lines):
+        self._actions_for("lock").for_update(lines)
+
+    def _set_destination_on_lines(
+        self, lines, location_dest, unload=False, set_src_location=False
+    ):
+        lines_with_package_level = lines.package_level_id.move_line_ids
+        lines_without_package_level = lines - lines_with_package_level
+        if lines_with_package_level:
+            lines_with_package_level.package_level_id.location_dest_id = location_dest
```
jbaudoux, line 242: Add a comment that odoo propagates this to the move and move lines

shopfloor/actions/stock.py
```diff
@@ -228,3 +229,26 @@ def no_putaway_available(self, picking_types, move_lines):
         # when no putaway is found, the move line destination stays the
         # default's of the picking type
         return any(line.location_dest_id in base_locations for line in move_lines)
+
+    def _lock_lines(self, lines):
+        self._actions_for("lock").for_update(lines)
+
+    def _set_destination_on_lines(
+        self, lines, location_dest, unload=False, set_src_location=False
```
jbaudoux: unload and set_src_location are obsolete

jbaudoux (commented):
Looking better. Just a few left overs

jbaudoux:
cc @rousseldenis

jbaudoux (approved):

mt-software-de:
@jbaudoux @rousseldenis @mmequignon  Could we move forward with this?

jbaudoux:
@mt-software-de can you make one commit per module ?

mt-software-de:
> @mt-software-de can you make one commit per module ?

done

PR #977, opened 2025-01-30
[16.0] shopfloor: checkout scan change lot - port of #712

jbaudoux:
Port of:
* https://github.com/OCA/wms/pull/712

cc @sebalix

### DIFF
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -574,7 +574,7 @@ class Checkout(Component):
         self, picking, selection_lines, lot, prefill_qty=1, **kw
     ):
         message = None
-        lines = selection_lines.filtered(lambda l: l.lot_id == lot)
+        lines = self._picking_lines_by_lot(picking, selection_lines, lot)
         if not lines:
             change_package_lot = self._actions_for("change.package.lot")
             if not kw.get("confirm_lot"):
@@ -661,6 +661,10 @@ class Checkout(Component):
             check_lot=False,
             message=message,
         )
+
+    def _picking_lines_by_lot(self, picking, selection_lines, lot):
+        """Control filtering of selected lines by given lot."""
+        return selection_lines.filtered(lambda l: l.lot_id == lot)
 
     def _change_lot_response_handler_ok(self, move_line, message=None):
         return message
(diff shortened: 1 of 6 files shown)


shopfloor/services/checkout.py
```diff
@@ -1749,6 +1826,7 @@ def _schema_stock_picking_details(self):
             group_lines_by_location={"type": "boolean"},
             show_oneline_package_content={"type": "boolean"},
             need_confirm_pack_all={"type": "string"},
+            need_confirm_lot={"type": "integer"},
```
sebalix: The v14 PR was adding `nullable: True` here, removed on purpose?

EDIT: ok you switched from None to False everywhere

shopfloor/tests/test_checkout_scan_line.py
```diff
@@ -248,17 +251,51 @@ def test_scan_line_error_product_not_in_picking(self):
...
+        # Second scan to confirm the change of lot
+        response = self.service.dispatch(
+            "scan_line",
+            params={
+                "picking_id": picking.id,
+                "barcode": lot.name,
+                "confirm_lot": lot.id,
+            },
+        )
+        message = self.msg_store.lot_replaced_by_lot(previous_lot, lot)
```
sebalix, line 282: Can you explain me why we do not have anymore a draft inventory generated here?

sebalix (approved):
LG (compared with v14 PR)

shopfloor/services/checkout.py
sebalix: I do not remember why it was put as `integer+nullable` instead of boolean... like if there was a trap there.

EDIT: was done in this commit: https://github.com/OCA/wms/pull/712/commits/b55b89a648b1958f9d34413edcac1756e46b08c8 which hasn't been ported

shopfloor/tests/test_checkout_scan_line.py
jbaudoux: You don't have inventory in odoo v16 anymore. You can only flag the quant or location to be inventoried without more detail

shopfloor/services/checkout.py
jbaudoux: This explains the errors I was getting. I included that missing commit and reduced my mig commit to its minimum

PR #983, opened 2025-02-14
[14.0] shopfloor checkout: Better messages

mmequignon:
Based on https://github.com/OCA/wms/pull/972
Only review shopfloor_checkout commits here, thanks <3

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -216,6 +216,15 @@ class MessageAction(Component):
     def already_done(self):
         return {"message_type": "info", "body": _("Operation already processed.")}
 
+    def transfer_cancelled(self):
+        return {
+            "message_type": "info",
+            "body": _(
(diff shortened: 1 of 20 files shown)


shopfloor/actions/message.py
```diff
@@ -952,3 +983,13 @@ def lot_change_no_line_found(self):
             "message_type": "error",
             "body": _("Unable to find a line with the same product but different lot."),
         }
+
+    def picking_type_is_return(self, picking):
```
jbaudoux, line 987: The fact that it is a return is a specific use case. Make the method name more generic.
```suggestion
    def reserved_for_other_picking_type(self, picking):
```

jbaudoux (commented):
It looks good !

mmequignon:
closed in favor of https://github.com/OCA/wms/pull/987

PR #986, opened 2025-02-19
[FIX][14.0] shopfloor cluster_picking: fix line split

mmequignon:
At `scan_destination_pack` when a destination package is scanned without any qty processed (especially when no prefill quantity is enabled), we would end up with a new move line with qty = 0.

Instead, return an error to the user.

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -686,43 +686,65 @@ class ClusterPicking(Component):
 
     def _set_destination_pack_update_quantity(self, move_line, quantity, barcode):
         """Handle the done quantity increment on set_destination end point."""
-        response = None
         if not self.work.menu.no_prefill_qty:
-            return response
+            return None
+        handlers = {
+            "product": self._set_destination_update_quantity__by_product,
+            "packaging": self._set_destination_update_quantity__by_packaging,
+            "lot": self._set_destination_update_quantity__by_lot,
+            "none": self._set_destination_update_quantity__fallback,
+        }
+        search_result = self._set_destination_pack_update_quantity__find(
+            barcode, handlers.keys()
+        )
+        handler = handlers.get(
+            search_result.type, self._set_destination_update_quantity__fallback
+        )
+        return handler(move_line, search_result.record, quantity)
+
+    def _set_destination_pack_update_quantity__find(self, barcode, search_types):
         search = self._actions_for("search")
-        # Handle barcode of product or packaging
-        product = search.product_from_scan(barcode)
-        packaging = self.env["product.packaging"].browse()
-        if not product:
-            packaging = search.packaging_from_scan(barcode)
-            product = packaging.product_id
-        if product:
-            if move_line.product_id == product:
-                quantity += packaging.qty or 1.0
-                response = self._response_for_scan_destination(
-                    move_line, qty_done=quantity
-                )
-                return response
-            return self._response_for_scan_destination(
-                move_line,
-                message=self.msg_store.wrong_record(product),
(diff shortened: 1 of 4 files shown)


jbaudoux (approved):
LG

mt-software-de (approved):
LGTM

shopfloor/services/cluster_picking.py
```diff
@@ -795,6 +817,11 @@ def scan_destination_pack(self, picking_batch_id, move_line_id, barcode, quantit
                 },
                 qty_done=quantity,
             )
+
+        # Keeping second return value for compatibility
+        # Quantity already checked at _check_line_qty_processible
+        new_line, __ = move_line._split_qty_to_be_done(quantity)
```
mt-software-de, line 823: Could you also extend `test_scan_destination_pack_bin_not_found` and ensure that there is also no new stock.move.line created. 
Because by moving `_split_qty_to_be_done` you fixed also the problem, that there is a stock.move.line created, when a wrong barcode is scanned and you entered a lower qty as `product_uom_qty`.

mt-software-de (changes requested):

shopfloor/services/cluster_picking.py
```diff
@@ -686,43 +686,65 @@ def _scan_line_by_location(self, picking, move_line, location):
...
+        return self._response_for_scan_destination(
+            move_line,
+            message=self.msg_store.wrong_record(lot),
+            qty_done=quantity,
+        )
+
+    def _set_destination_update_quantity__fallback(
+        self, move_line, empty_rec, quantity
+    ):
+        return None
```
TDu, line 747: Should it not also return a response ?

mmequignon: No, response was initially set to None, and if no product or lot were found from barcode, None would be returned.
This is the trick used to do multiple things from a single shopfloor screen (increase quantities and set destination package).

shopfloor/services/cluster_picking.py
mt-software-de: ping @mmequignon

mmequignon: Yes sorry, forgot to answer, I'll do this ASAP

TDu (approved):
LG

shopfloor/services/cluster_picking.py
mt-software-de: @mmequignon could you extend the test?

mmequignon: @mt-software-de I tried to reproduce this behavior with 14.0, no new move line is created after `scan_destination_pack` is called.
```python
        all_lines = self.env["stock.move.line"].search([])
        line = self.one_line_picking.move_line_ids
        response = self.service.dispatch(
            "scan_destination_pack",
            params={
                "picking_batch_id": self.batch.id,
                "move_line_id": line.id,
                # this bin is used for the other picking
                "barcode": "⌿",
                "quantity": line.product_uom_qty,
            },
        )
        new_lines = self.env["stock.move.line"].search([("id", "not in", all_lines.ids)])
        self.assertFalse(new_lines)
```
What am I missing here?

mt-software-de: The quantity should be less than `line.product_uom_qty`. Only then the split is called.

jbaudoux: @mmequignon Can you check that ?

mt-software-de:
@jbaudoux could you reopen this one? And merge it, it is only missing the extension of the test. I can do this after this is merged.

PR #987, opened 2025-02-19
[14.0] shopfloor: better messages when transfer has been cancel and product is reserved by a return

mmequignon:
Attempt to make shopfloor messages more meaningful, especially when scanning an unexpected record.
The general idea is too avoid those `barcode not found` errors in cases where the barcode was found, but not relevant.

Relates to:
 -  https://github.com/OCA/wms/pull/971
 
Supersedes:
 - https://github.com/OCA/wms/pull/983
 - https://github.com/OCA/wms/pull/972

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -560,11 +560,9 @@ class MessageAction(Component):
             "body": _("Place it in {}?").format(location_name),
         }
 
-    def product_not_found_in_current_picking(self):
-        return {
-            "message_type": "error",
-            "body": _("Product is not in the current transfer."),
-        }
+    def product_not_found_in_current_picking(self, product):
+        body = _("Product %s is not in the current transfer.", product.name)
+        return {"message_type": "error", "body": body}
 
     def lot_mixed_package_scan_package(self):
         return {
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -961,3 +961,13 @@ class MessageAction(Component):
             "message_type": "error",
             "body": _("Unable to find a line with the same product but different lot."),
         }
+
+    def reserved_for_other_picking_type(self, picking):
+        body = _("Reserved for %(picking_type)s %(picking_name)s") % {
+            "picking_type": picking.picking_type_id.name,
+            "picking_name": picking.name,
+        }
+        return {
+            "message_type": "error",
+            "body": body,
+        }
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -216,6 +216,15 @@ class MessageAction(Component):
     def already_done(self):
         return {"message_type": "info", "body": _("Operation already processed.")}
 
+    def transfer_canceled(self):
+        return {
+            "message_type": "info",
+            "body": _(
+                "Transfer has been canceled. "
+                "This cannot be processed using this scenario"
+            ),
+        }
+
     def move_already_done(self):
         return {"message_type": "warning", "body": _("Move already processed.")}
 
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -468,6 +468,30 @@ class MessageAction(Component):
             "body": _("No transfer found for this product."),
         }
 
+    def transfer_not_found_for_barcode(self, barcode):
+        body = _("No transfer found for barcode %s", barcode)
+        return {
+            "message_type": "error",
+            "body": body,
+        }
+
+    def transfer_not_found_for_record(self, record):
+        model_mapping = {
+            "product.product": "product",
+            "stock.picking": "transfer",
+            "stock.quant.package": "package",
+            "product.packaging": "packaging",
+            "stock.location": "location",
+            "stock.production.lot": "lot",
+            "stock.move": "move",
+        }
+        model_name = model_mapping.get(record._name)
+        body = _("No transfer found for %s %s", model_name, record.name)
+        return {
+            "message_type": "error",
+            "body": body,
+        }
+
     def product_not_found_in_location_or_transfer(self, product, location, picking):
         return {
             "message_type": "error",
--- a/shopfloor/services/checkout.py
+++ b/shopfloor/services/checkout.py
@@ -471,21 +471,31 @@ class Checkout(Component):
             return self._response_for_summary(picking)
 
         # Search of the destination package
-        search_result = self._scan_line_find(picking, barcode)
-        result_handler = getattr(self, "_select_lines_from_" + search_result.type)
-        kw = {"confirm_pack_all": confirm_pack_all, "confirm_lot": confirm_lot}
-        return result_handler(picking, selection_lines, search_result.record, **kw)
+        handlers = {
+            "package": self._select_lines_from_package,
+            "product": self._select_lines_from_product,
+            "packaging": self._select_lines_from_packaging,
+            "lot": self._select_lines_from_lot,
+            "serial": self._select_lines_from_serial,
+            "delivery_packaging": self._select_lines_from_delivery_packaging,
+            "none": self._select_lines_from_none,
+        }
+        search_result = self._scan_line_find(picking, barcode, handlers.keys())
+        # setting scanned record as kwarg in order to make better logs.
+        # The reason for this is that from a product we might select various records
+        # and lose track of what was initially scanned. This forces us to display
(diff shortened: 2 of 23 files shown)


test-requirements.txt
```diff
@@ -1,2 +1,3 @@
 freezegun
 odoo_test_helper
+odoo14-addon-stock_available_to_promise_release @ git+https://github.com/OCA/wms@refs/pull/971/head#subdirectory=setup/stock_available_to_promise_release
```
jbaudoux: shopfloor doesn't depend on that module. Are new tests failing without that ?

shopfloor_delivery_shipment/services/delivery_shipment.py
```diff
@@ -208,13 +208,33 @@ def _scan_package(self, shipment_advice, package, picking, location):
...
+                move_lines.picking_id, shipment_advice
+            )
             if message:
                 return self._response_for_scan_document(
                     shipment_advice,
                     location=location,
                     message=message,
                 )
+            # Check that all move lines in package have a package level.
+            # This might happen if a sale order line is cancelled while
```
jbaudoux:
```suggestion
            # This might happen if a sale order line is canceled while
```

shopfloor/actions/message.py
```diff
@@ -216,6 +216,15 @@ def confirm_pack_moved(self):
     def already_done(self):
         return {"message_type": "info", "body": _("Operation already processed.")}
 
+    def transfer_canceled(self):
+        return {
+            "message_type": "info",
+            "body": _(
+                "Transfer has been cancelled. "
```
```suggestion
                "Transfer has been canceled. "
```

shopfloor/tests/test_delivery_scan_deliver.py
```diff
@@ -445,6 +460,112 @@ def test_scan_deliver_scan_product_alone_in_package_qty_one(self):
         move_lines = pick.move_lines.mapped("move_line_ids")
         self._test_scan_set_done_ok(move_lines, self.product_c.barcode, [1])
 
+    def test_scan_deliver_picking_cancelled(self):
```
jbaudoux:
```suggestion
    def test_scan_deliver_picking_canceled(self):
```

shopfloor/tests/test_delivery_scan_deliver.py
```diff
@@ -10,6 +10,21 @@ class DeliveryScanDeliverCase(DeliveryCommonCase):
...
+                {
+                    "name": "Cancel Cleanup",
+                    "default_location_src_id": cls.out_location.id,
+                    "default_location_dest_id": cls.stock_location.id,
+                    "sequence_code": "CCP",
+                    "code": "internal",
+                }
+            )
+        )
+        cls.picking_type.sudo().return_picking_type_id = cls.cleanup_type
```
jbaudoux: Do you need this line? It's like you manually create the cleanup transfer in the tests

shopfloor/tests/test_delivery_scan_deliver.py
```diff
@@ -445,6 +460,112 @@ def test_scan_deliver_scan_product_alone_in_package_qty_one(self):
...
+    def test_scan_deliver_picking_cancelled(self):
+        self.picking.action_cancel()
+        params = {"barcode": self.picking.name}
+        response = self.service.dispatch("scan_deliver", params=params)
+        self.assert_response_deliver(
+            response,
+            message=self.service.msg_store.transfer_canceled(),
+        )
+
+    def test_scan_deliver_return_package(self):
```
jbaudoux, line 493: Can you add the case where only one move of the package is reserved by the cleanup ?
Don't cancel the complete picking but only the move product_a. After that it's the same code.

shopfloor_delivery_shipment/tests/test_delivery_shipment_scan_document_package.py
```diff
@@ -146,6 +146,8 @@ def test_scan_document_package_already_loaded(self):
             self.service.data.package_levels(package_level),
         )
 
+    # def test_scan_document_package_partially_reserved(self):
+
```
jbaudoux: I guess you can repeat here the same tests from delivery
test_scan_deliver_return_package (full + partial) / product / packaging / lot

jbaudoux (commented):
Looks good. Just finish the tests for delivery and it should be good

jbaudoux (approved):

jbaudoux:
@mt-software-de for review

shopfloor/actions/message.py
```diff
@@ -459,6 +468,30 @@ def product_not_found_in_pickings(self):
             "body": _("No transfer found for this product."),
         }
 
+    def transfer_not_found_for_barcode(self, barcode):
+        body = _("No transfer found for barcode {}").format(barcode)
```
mt-software-de:
```suggestion
        body = _("No transfer found for barcode %s") % barcode
```

shopfloor/actions/message.py
```diff
@@ -459,6 +468,30 @@ def product_not_found_in_pickings(self):
...
+            "product.product": "product",
+            "stock.picking": "transfer",
+            "stock.quant.package": "package",
+            "product.packaging": "packaging",
+            "stock.location": "location",
+            "stock.production.lot": "lot",
+            "stock.move": "move",
+        }
+        model_name = model_mapping.get(record._name)
+        body = _("No transfer found for {} {}").format(model_name, record.name)
```
mt-software-de:
```suggestion
        body = _("No transfer found for %s %s") % (model_name, record.name)
```

shopfloor/actions/message.py
```diff
@@ -527,11 +560,9 @@ def place_in_location_ask_confirmation(self, location_name):
             "body": _("Place it in {}?").format(location_name),
         }
 
-    def product_not_found_in_current_picking(self):
-        return {
-            "message_type": "error",
-            "body": _("Product is not in the current transfer."),
-        }
+    def product_not_found_in_current_picking(self, product):
+        body = _("Product {} is not in the current transfer.").format(product.name)
```
mt-software-de:
```suggestion
        body = _("Product %s is not in the current transfer.") % product.name
```

shopfloor/services/checkout.py
```diff
@@ -1029,23 +1054,26 @@ def scan_package_action(self, picking_id, selected_line_ids, barcode):
...
+        }
+        search_result = self._scan_package_find(picking, barcode, handlers.keys())
+        handler = handlers.get(search_result.type, self._scan_package_action_from_none)
+        kwargs = {
+            "barcode": barcode,
+            "scanned_record": search_result.record,
+        }
+        return handler(picking, selected_lines, search_result.record, **kwargs)
 
     def _scan_package_find(self, picking, barcode, search_types=None):
```
mt-software-de: This is called now always with search_types
```suggestion
    def _scan_package_find(self, picking, barcode, search_types):
```

shopfloor/services/service.py
```diff
@@ -46,21 +63,41 @@ def search_move_line(self):
...
+    def _check_picking_status(self, pickings, states=("assigned",)):
+        """Checks if the picking exists, is already done or canceled."""
+        if not any(picking.state != "done" for picking in pickings):
+            return self.msg_store.already_done()
+        if not any(picking.state != "cancel" for picking in pickings):
+            return self.msg_store.transfer_canceled()
+        if not any(
+            picking.state in states for picking in pickings
+        ):  # the picking must be ready
+            return self.msg_store.stock_picking_not_available(fields.first(pickings))
```
mt-software-de, line 88: Why was this changed to `not any`?
For the done check, before it returned the message if any picking is done, now it returns the messages only if all pickings are done.

mt-software-de (changes requested):

shopfloor/services/service.py
jbaudoux: The concept is that we load all pickings from the source location, so we need to check there is at least one suitable picking to continue further. It must not be done, nor cancel, and with the right picking type.
For the done/cancel case, I think it can only occur if you start by scanning a specific picking otherwise it will be excluded from the initial search. So practically, it will always be a single picking. Nevertheless, the `not any` represent the best the purpose of the test

shopfloor/actions/message.py
jbaudoux: `        body = _("No transfer found for barcode %s", barcode)`

mt-software-de (approved):

PR #988, opened 2025-02-26
[18.0][MIG] stock_picking_type_shipping_policy

sebalix:
## Notes:
- previous `shipping_policy` field is replaced by `force_move_type` boolean field
- migration script added to set the std field `move_type` based on the value of previous `shipping_policy` + set `force_move_type` to True
- remove override of `stock.move` and switched to an override of `<stock.picking>._compute_move_type`
- DESCRIPTION of the module updated (BTW the mentioned module `stock_routing_operation` doesn't exist, I supposed it should be `stock_dynamic_routing`?)


sebalix:
@jbaudoux FYI in v18 we have a new field [`<stock.picking.type>.move_type`](https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_picking.py#L148-L151) (required), and the `stock.picking` takes in priority the `<procurement.group>.move_type`, otherwise fallback on the one defined in its `picking_type_id.move_type`: https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_picking.py#L715-L719

But this std code doesn't allow to force the `move_type` as the current module does, like ignoring the `group_id.move_type` and force to `picking_type_id.move_type`. For now I kept the field `<stock.picking.type>.shipping_policy` defined by the current module, while changing its label to not conflict with the new std "Shipping Policy" field. But maybe we could leverage the std one to add new section values in it, to explicitely "force" the value we want. Your POV, keeping two fields, or try to merge them in one?

stock_picking_type_shipping_policy/models/stock_picking_type.py
```diff
@@ -19,6 +19,7 @@ def _selection_shipping_policy(self):
         ]
 
     shipping_policy = fields.Selection(
+        string="Force Shipping Policy",
```
sebalix, line 10: Changed the user's label to not conflict with `move_type`/`Shipping Policy` std field

stock_picking_type_shipping_policy/tests/test_stock_picking_type_shipping_policy.py
```diff
@@ -28,35 +29,46 @@ def setUpClass(cls):
...
+        return moves_after - moves_before
+
+    def _validate_picking(self, picking):
+        picking.move_line_ids.write({"picked": True})
+        picking._action_done()
+
     def test_shipping_policy(self):
         self.pack_type.shipping_policy = "force_all_products_ready"
         self.pick_type.shipping_policy = "force_as_soon_as_possible"
-        # Create picking
```
sebalix: Rewrote this part to align with new default Odoo delivery flow that creates the Pick first, then Pack + Ship.
To ensure things are created as expected, we generate the initial move/picking through a procurement.

stock_picking_type_shipping_policy/models/stock_picking_type.py
jbaudoux: I would change it to a boolean as the selection is now standard on the picking type as you said

jbaudoux: and extend `_compute_move_type` on stock.picking to check that boolean to see if we should use the one from the picking type
https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_picking.py#L716

stock_picking_type_shipping_policy/models/stock_move.py
```diff
@@ -0,0 +1,22 @@
...
+        elif picking_type.shipping_policy == "force_all_products_ready":
+            return "one"
+        return None
+
+    def _get_new_picking_values(self):
+        res = super()._get_new_picking_values()
+        picking_type_move_type = self._shipping_policy_from_picking_type()
+        if picking_type_move_type:
+            res["move_type"] = picking_type_move_type
+        return res
```
jbaudoux: It looks obsolete to me now. It's replaced by a compute on stock.picking

jbaudoux (changes requested):

stock_picking_type_shipping_policy/models/stock_picking_type.py
sebalix: Done:

- previous `shipping_policy` field is replaced by `force_move_type` boolean field
- migration script added to set the std field `move_type` based on the value of previous `shipping_policy` + set `force_move_type` to True
- remove override of `stock.move` and switched to an override of `<stock.picking>._compute_move_type`
- DESCRIPTION of the module updated (BTW the mentioned module `stock_routing_operation` doesn't exist, I supposed it should be `stock_dynamic_routing`?)

sebalix: Should we rename the module now that Odoo has some kind of support of the shipping policy on operation types? Like `stock_picking_type_force_move_type` or `stock_picking_type_force_shipping_policy`?

stock_picking_type_shipping_policy/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,13 @@
...
+This module adds a Force Shipping Policy field on Operations Types to ensure
+transfers will take the policy according to their types, ignoring the one set
+on the Procurement Group.
+
+This is especially useful if you use a pick-pack-ship setup with the
+release of operation (`stock_available_to_promise_release`) module along
+side with the `stock_dynamic_routing` that may split operations by zone
+of the warehouse. In that case, you want to be sure the pack operations
+will wait all different picks to be processed before releasing the
+availability of the pack operation.
```
jbaudoux, line 13:
```suggestion
This is especially useful if you use a multi-steps setup (like pick-pack-ship) and you want the pick and/or pack operation to wait for all goods to be available no matter what was the sales shipping policy.
```

sebalix:
Moved to https://github.com/OCA/stock-logistics-workflow/pull/1888

stock_picking_type_shipping_policy/readme/DESCRIPTION.md
sebalix: Done in the new PR

PR #991, opened 2025-03-05
[14.0][IMP] stock_warehouse_flow: Update flow when carrier changes

mmequignon:
This PR aims to provide the possibility to update a flow on pickings when carrier is updated.

based on:
 - https://github.com/OCA/stock-logistics-workflow/pull/1895


stock_warehouse_flow/tests/test_warehouse_flow.py
```diff
@@ -8,6 +8,31 @@
...
+        move_ship = moves.filtered(lambda m: m.picking_type_id.code == "outgoing")
+        # When flow was applied to move, picking type set by odoo was saved
+        # in default_picking_type_id, and flow.to_picking_type_id is set
+        self.assertEqual(move_ship.picking_type_id, flow_pick_ship.to_picking_type_id)
+        default_type = self.env.ref("stock.picking_type_out")
+        self.assertEqual(move_ship.default_picking_type_id, default_type)
+        # Only pick_ship matches, because there's a constraint on the carrier
+        matching_flows = self.env["stock.warehouse.flow"]._search_for_move(move_ship)
+        self.assertEqual(matching_flows, flow_pick_ship)
+        # Drop the carrier from the procurement, only direct flow should match
```
sebalix:
```suggestion
        # Drop the carrier from the procurement, only ship only flow should match
```
as "direct" is used for shipping policy (confusing)

stock_warehouse_flow/tests/test_warehouse_flow.py
```diff
@@ -8,6 +8,31 @@
...
+        move_ship.group_id.carrier_id = False
+        matching_flows = self.env["stock.warehouse.flow"]._search_for_move(move_ship)
+        self.assertEqual(matching_flows, flow_ship_only)
+        # Drop default_picking_type_id, no flow is found, because
+        # now flow has Delivery Orders POST as from_picking_type_id
```
sebalix, line 31: Not sure to understand this comment. Current move operation type (which has been updated at the beginning) doesn't match any flows?

sebalix (commented):
LG as a first step, let's see how this will be used later (particularly how a flow will be re-applied on the existing move).

stock_warehouse_flow/tests/test_warehouse_flow.py
mmequignon: You create the move with picking type A.
All flows have `from_picking_type_id` == A.
If you apply any of those flows to the move, picking_type_id is updated to `to_picking_type_id` (B).
The only reason why it was matching before is because we made a backup of A in `default_picking_type_id`.
Drop this backup, it doesn't match any flow anymore.

stock_warehouse_flow_release/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/wms",
     "category": "Warehouse Management",
-    "version": "14.0.2.0.1",
+    "version": "14.0.2.1.0",
```
jbaudoux, line 9: I would also make auto_install False

And remove `(not anymore when it is confirmed)` from README

stock_warehouse_flow_delivery_refresh/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "version": "14.0.1.0.0",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/wms",
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "stock_warehouse_flow",
+        "stock_available_to_promise_release",
+    ],
```
jbaudoux, line 13: as you depend on alpha module stock_warehouse_flow
```suggestion
    ],
    "development_status": "Alpha",
```

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -538,6 +552,9 @@ def _apply_on_move(self, move, assign_picking=True):
         move.rule_id = rule
         if assign_picking:
             move._assign_picking()
+            # If all moves are moved to another picking, we can unlink the old one.
+            if not old_picking.move_lines:
+                old_picking.unlink()
```
jbaudoux: If this is triggered from the picking you have in open in form view, this will cause a reload of odoo.
If would just write state cancel on the old_picking (I don't know if this is not the case in standard or if it is draft)

jbaudoux (changes requested):

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -528,7 +536,13 @@ def _apply_on_move(self, move, assign_picking=True):
             return False
         logger.info("Applying flow '%s' on '%s'", self.name, move)
         rule = self._get_rule_from_delivery_route()
```
jbaudoux: As now the flow can be re-run multiple times (as we keep a copy of the default picking type), we should not re-run if we don't apply a new flow. Otherwise each time you reassign the move to a new picking which is awkward and useless
```suggestion
        rule = self._get_rule_from_delivery_route()
        if move.rule_id == rule:
            return
        logger.info("Applying flow '%s' on '%s'", self.name, move)
```

jbaudoux (changes requested):

stock_warehouse_flow/models/stock_warehouse_flow.py
jbaudoux: just tested, it is in draft
So replace with `old_picking.state = "cancel"`

shopfloor_delivery_shipment/tests/test_delivery_shipment_base.py
```diff
@@ -40,6 +40,7 @@ def setUpClassBaseData(cls, *args, **kwargs):
                     (cls.product_d, 10),
                 ],
             )
+            picking = picking.with_context(skip_align_group_carrier=True)
```
mt-software-de: Why is this necessary? There is no dependency from `shopfloor_delivery_shipment` to `delivery_procurement_group_carrier`.

stock_warehouse_flow/models/stock_warehouse_flow.py
```diff
@@ -538,6 +555,9 @@ def _apply_on_move(self, move, assign_picking=True):
         move.rule_id = rule
         if assign_picking:
             move._assign_picking()
+            # If all moves are moved to another picking, we can unlink the old one.
+            if not old_picking.move_lines:
+                old_picking.state = "cancel"
```
mt-software-de, line 560: Why not unlinking it then?

stock_warehouse_flow_delivery_refresh/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2025 Camptocamp SA, BCIM
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _stock_warehouse_flow_onchange_carrier(self):
```
mt-software-de: I think naming it without using `onchange_carrier` is more accurate, because this method doesn't know where it was called from, therefor it is enough to name it after what i actual does. 
```suggestion
    def _refresh_warehouse_flow(self):
```
or
```suggestion
    def _update_warehouse_flow(self):
```

stock_warehouse_flow_delivery_refresh/__manifest__.py
mt-software-de, line 1: Shouldn't it be a default behavior? Then it should be a `auto_install=True`

stock_warehouse_flow_delivery_refresh/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
...
+        flow_model = self.env["stock.warehouse.flow"]
+        moves_need_release = self.filtered(lambda m: m.picking_id.need_release)
+        for move in moves_need_release:
+            flow_model._search_and_apply_for_move(move, assign_picking=True)
```
mt-software-de: By moving the check against `need_release` into the for loop, you would not need a second iteration. 
Also the `need_release` is stored on the move it self. 
```suggestion
        for move in self:
            if  not move.need_release:
                continue
            flow_model._search_and_apply_for_move(move, assign_picking=True)
```

stock_warehouse_flow_delivery_refresh/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This is a glue module between Stock Warehouse Flow and Stock Available On Premise Release.
+
+This module applies a new Stock Warehouse Flow on Stock Moves when carrier changes on a Stock Picking
```
mt-software-de:
```suggestion
This module applies a new Stock Warehouse Flow on Stock Moves waiting for release when carrier changes on a Stock Picking
```

stock_warehouse_flow_release/models/stock_move.py
```diff
@@ -8,13 +8,6 @@
 class StockMove(models.Model):
     _inherit = "stock.move"
 
-    def _apply_flow_on_action_confirm(self):
-        # Override to not apply the flow configuration on moves to release.
-        # The flow will be applied on the release, not before.
-        if self.rule_id.route_id.available_to_promise_defer_pull:
-            return False
-        return super()._apply_flow_on_action_confirm()
-
```
mt-software-de: With this change it is called on confirm when the flag `disable_warehouse_flow_on_confirm` isn't set and called on release. Shouldn't it be checked on release now? If the flag is set or not. Otherwise it is called twice on confirm and on release. 

Why is the flag set by the migration, but not set as default on for routes with `available_to_promise_defer_pull`?

mt-software-de (changes requested):

stock_warehouse_flow/models/stock_warehouse_flow.py
mmequignon: https://github.com/OCA/wms/pull/991#discussion_r2003414564

shopfloor_delivery_shipment/tests/test_delivery_shipment_base.py
mmequignon: Yes, but if both are installed, it creates an incompatibility

stock_warehouse_flow_release/models/stock_move.py
mmequignon: It has been moved in `stock_warehouse_flow`
https://github.com/OCA/wms/pull/991/files#diff-531a3523042e74031943fc83a9176f2de25e3ac2be78e85d1edb27f92ecd669cR19-R23
We now have a dedicated option for this, rather than checking if `available_to_promise_defer_pull` is set on the route.
We also have a migration for this.

stock_warehouse_flow_delivery_refresh/models/stock_move.py
mmequignon: makes sense, thanks!

stock_warehouse_flow_delivery_refresh/__manifest__.py
mmequignon: I'll let @jbaudoux answer to this one, this was on the specs. We might have other customers that do not want this behavior, and in this case, it's easier to install the module than uninstalling an auto_install one.

stock_warehouse_flow_release/models/stock_move.py
mt-software-de: I think the `disable_warehouse_flow_on_confirm` should move into `stock_warehouse_flow_release` because only there you want to decide if you want to calc the flow on confirm or on release.
If you just want to disable the calc of the flow in the base addon you just don't need to configure it.
But thats not a blocker for me. 

But if you still want to keep `disable_warehouse_flow_on_confirm` on the base module, then i think there should be also a dedicated flag to allow the flow calculation on release. 

Otherwise the flow will be calculated always on release. 

In my opinon we should either 
- add a dedicated flag for this
or
- keep the addon as it is just removing the `auto_install` flag

cc @jbaudoux

stock_warehouse_flow_delivery_refresh/__manifest__.py
mt-software-de: fine for me for in both ways

shopfloor_delivery_shipment/tests/test_delivery_shipment_base.py
mt-software-de: ok. but in the most cases, this is an indication of buggy code.
I tested it really quick. I found the bug https://github.com/OCA/stock-logistics-workflow/pull/1895/files#r2011903437

mt-software-de:
```suggestion
```

mt-software-de (changes requested):

stock_warehouse_flow_release/models/stock_move.py
jbaudoux: @sebalix your opinion on this?

stock_warehouse_flow/models/stock_location_route.py
```diff
@@ -17,6 +17,7 @@ class StockLocationRoute(models.Model):
         compute="_compute_applicable_flow_ids",
         string="Applicable Flows",
     )
+    disable_warehouse_flow_on_confirm = fields.Boolean()
```
sebalix: Why not a selection field instead:
```suggestion
    apply_flow_on = fields.Selection(selection=[("on_confirm", "When move is confirmed")])
```
And the glue module between flow & release will add a new value:
```python
apply_flow_on = fields.Selection(selection_add=[("on_release", "When move is released")])
```
And code to adapt to check this option.

stock_warehouse_flow_release/models/stock_move.py
mt-software-de: https://github.com/OCA/wms/pull/991#discussion_r2012056885

stock_warehouse_flow/models/stock_location_route.py
mmequignon: done

stock_warehouse_flow_release/models/stock_move.py
mmequignon: done

stock_warehouse_flow_release/models/stock_move.py
```diff
@@ -8,19 +8,17 @@
...
+        return self.picking_type_id.code == "outgoing"
 
     def _before_release(self):
         # Apply the flow when releasing the move
         super()._before_release()
         FLOW = self.env["stock.warehouse.flow"]
         move_ids_to_release = []
-        for move in self:
+        for move in self.filtered(lambda m: m._apply_flow_on_release()):
             _move_ids_to_release = FLOW._search_and_apply_for_move(move).ids
```
mt-software-de: A second iteration is unnecessary if you move the check `_apply_flow_on_release` into the for loop. 
```suggestion
        for move in self:
            if not move._apply_flow_on_release():
                continue
            _move_ids_to_release = FLOW._search_and_apply_for_move(move).ids
```

mt-software-de (changes requested):
Minor change request

jbaudoux (approved):

stock_warehouse_flow_release/models/stock_move.py
mmequignon: it's done already, please target a line that changed when suggesting comment, so github can flag them as outdated :pray: 
Thanks!

mt-software-de: Where is it done?

jbaudoux: @mmequignon it's not showing as done in the diff...

mmequignon: My bad, I messed up while checking locally :facepalm: 
It's pushed now

mt-software-de (approved):

PR #996, opened 2025-03-24
[14.0][IMP] delivery_carrier_preference: Do not align group carrier

mmequignon:
As stated by @jbaudoux on this PR https://github.com/OCA/stock-logistics-workflow/pull/1895
After it is merged, procurement group carrier will be aligned when picking carrier will be updated.
This feature is incompatible with delivery_carrier_preference.
Add a context key to disable the feature.

### DIFF
--- a/delivery_carrier_preference/models/stock_picking.py
+++ b/delivery_carrier_preference/models/stock_picking.py
@@ -39,7 +39,9 @@ class StockPicking(models.Model):
                 }
             }
         else:
-            self.carrier_id = carrier
+            # Set a context key to not trigger a carrier change on the
+            # procurement group.
+            self.with_context(skip_align_group_carrier=True).carrier_id = carrier
 
     def get_preferred_carriers(self):
         # TODO Check possible conflicting settings between doc company and


jbaudoux (approved):

delivery_carrier_preference/models/stock_picking.py
```diff
@@ -39,7 +39,9 @@ def add_preferred_carrier(self):
                 }
             }
         else:
-            self.carrier_id = carrier
+            # Set a context key to not trigger a carrier change on the
+            # procurement group.
+            self.with_context(skip_align_group_carrier=True).carrier_id = carrier
```
mt-software-de, line 44: Why is there an issue with setting the carrier on the group? This will then end trigger the compute for a new stock.warehouse.flow.
@jbaudoux

jbaudoux: because it's done before creating the new proc group

mt-software-de: If the carrier should not be set on whole group just don't set it before creating the new proc group. 
This would make the context key unnecessary.

mt-software-de: At least a test would have been nice. :-)

PR #997, opened 2025-03-25
[16.0][IMP] stock_release_channel_partner_by_date: add check_late parameter on _get_release_channel_partner_date_domain

santostelmo:
Add optional parameter `check_late` `_get_release_channel_partner_date_domain`. It allows to search for release channels to process picking that are late => `max(self.scheduled_date, fields.Datetime.now())`
or for the effective scheduled_date => `self.scheduled_date`

### DIFF
--- a/stock_release_channel_partner_by_date/models/stock_picking.py
+++ b/stock_release_channel_partner_by_date/models/stock_picking.py
@@ -9,9 +9,12 @@ from odoo.osv import expression
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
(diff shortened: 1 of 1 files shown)


sebalix (approved):

jbaudoux (changes requested):
I don't understand. I don't see how this new argument is used and there is no test.

If you configure a channel for a partner on a specific date, why would you not deliver with this channel what was planned in the past ?

santostelmo:
Closing this PR. Finally we go in another direction

PR #998, opened 2025-03-25
[16.0][IMP] stock_release_channel_partner_by_date: enforce specific channel partner date

santostelmo:


stock_release_channel_partner_by_date/tests/test_release_channel_partner_date.py
```diff
@@ -53,3 +55,19 @@ def test_release_channel_sleep_archive_specific_date(self):
...
         self.assertFalse(channel_date.active)
+
+    def test_release_channel_on_specific_date_not_available(self):
+        """Test that when no release channel is available to satisfy
+        a specific partner date,no fallback release channel is
+        proposed."""
+        self._create_channel_partner_date(
+            self.delivery_date_channel,
+            self.partner,
+            fields.Datetime.now() - relativedelta(day=1),
```
sebalix:
```suggestion
            fields.Datetime.subtract(fields.Datetime.now() days=1),
```

stock_release_channel_partner_by_date/models/stock_picking.py
```diff
@@ -57,3 +57,17 @@ def _get_release_channel_possible_candidate_domain(self):
         if specific_rc_domain:
             domain = expression.AND([domain, specific_rc_domain])
         return domain
+
+    def _find_release_channel_possible_candidate(self):
+        channels = super()._find_release_channel_possible_candidate()
+        specific_channel = self.env["stock.release.channel.partner.date"].search(
+            self._get_release_channel_partner_date_domain()
```
jbaudoux, line 64: You are doing the same as the method above except that you don't respect the company and warehouse. It is not the right approach

jbaudoux (changes requested):

stock_release_channel_partner_by_date/models/stock_picking.py
jbaudoux: Here is a proposal https://github.com/OCA/wms/pull/999

I kept your test but added the warehouse as you changed the picking type so that it properly detects it's an exception valid for the picking's warehouse

jbaudoux: Note that the carrier must also be valid is you have `stock_release_channel_delivery`. That's why we use `_get_release_channel_possible_candidate_domain_picking()` and not only `_get_release_channel_partner_date_domain()`

sale_stock_release_channel_partner_by_date_delivery/tests/test_sale_release_channel.py
```diff
@@ -42,9 +42,9 @@ def test_sale_order_with_wrong_carrier(self):
         self.assertFalse(order._get_release_channel_partner_date())
         picking_out = order.picking_ids
         self.assertFalse(picking_out.release_channel_id)
-        # Then delivery gets the default channel
+        # Then delivery does not fallback in any other channel
         self.env["stock.release.channel"].assign_release_channel(picking_out)
-        self.assertEqual(picking_out.release_channel_id, self.default_channel)
+        self.assertFalse(picking_out.release_channel_id)
```
jbaudoux, line 47: This is a change of behavior.
If we set an exception, it is only valid for a defined carrier.
We shouldn't change that.

We should prevent to have an specific channel on the SO that is not in line with the SO carrier.

PR #999, opened 2025-03-27
[16.0][FIX] stock_release_channel_partner_by_date: properly detect the exception

jbaudoux:
Fixing compatibility with:
* stock_release_channel_shipment_lead_time
* stock_release_channel_partner_public_holidays
  * was already good
  * Added test for compatibility in:
    * https://github.com/OCA/wms/pull/1033

cc @sebalix @santostelmo

### DIFF
--- a/stock_release_channel/models/stock_picking.py
+++ b/stock_release_channel/models/stock_picking.py
@@ -108,19 +108,33 @@ class StockPicking(models.Model):
         )
 
     def _get_release_channel_possible_candidate_domain_channel(self):
+        """Domain for finding channel candidates based on channel.
+
+        The condition is defined by the channel.
+        """
         return [
             ("is_manual_assignment", "=", False),
             ("state", "!=", "asleep"),
-        ]
-
-    def _get_release_channel_possible_candidate_domain_picking(self):
-        return [
-            ("company_id", "=", self.company_id.id),
             "|",
             ("picking_type_ids", "=", False),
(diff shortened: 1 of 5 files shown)


santostelmo (approved):

stock_release_channel/models/stock_picking.py
```diff
@@ -111,16 +111,22 @@ def _get_release_channel_possible_candidate_domain_channel(self):
         return [
             ("is_manual_assignment", "=", False),
             ("state", "!=", "asleep"),
+            "|",
+            ("picking_type_ids", "=", False),
+            ("picking_type_ids", "in", self.picking_type_id.ids),
```
sebalix, line 120: Is it expected to use picking data in `_get_release_channel_possible_candidate_domain_channel` domain hook? I thought we designed it for basic criteria selection of the channel (like the ones already defined) without relying on current picking. For that we have `_get_release_channel_possible_candidate_domain_picking`.

sebalix (commented):
LG overall but still one question

stock_release_channel/models/stock_picking.py
jbaudoux: We need a domain to find out that the exception is valid for that warehouse & carrier but not picking type as it could be recomputed by the warehouse flow module.
I'll add docstring to those methods.

jbaudoux: I added the docstrings

sebalix (approved):

PR #1000, opened 2025-04-02
[14.0][ADD] stock_available_to_promise_release_shipment_advice

jbaudoux:
Prevent to unrelease when loaded in a shipment advice.

in case you allow to unrelease even when internal transfers are done. 

Depends on:
* https://github.com/OCA/wms/pull/971

@mmequignon @mt-software-de @rousseldenis


mt-software-de (approved):
LG

mmequignon (approved):

PR #1001, opened 2025-04-03
[16.0][IMP] shopfloor: Add a parser to searches

rousseldenis:
In some flows like GS1, a barcode representation can contain multiple content (e.g.: product, lot, expiration date, ...).

In current flows, we allow only one record return from the searches based on barcode scan.

In order to change the behavior when some extra data is found in barcodes, we add the parsed result in the search result.

e.g.: For reception flows, if we scan a GS1 and the lot is not found, we want to pre-fill expiration date for instance

### DIFF
--- a/shopfloor/actions/barcode_parser.py
+++ b/shopfloor/actions/barcode_parser.py
@@ -1,0 +1,44 @@
+# Copyright 2025 ACSONE SA/NV (https://www.acsone.eu)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo.addons.component.core import Component
+
+from ..actions.search import SearchAction
+
+
+class BarcodeResult:
+
+    __slots__ = ("type", "value", "raw")
+
+    def __init__(self, **kw) -> None:
+        for k in self.__slots__:
+            setattr(self, k, kw.get(k))
+
+
+class BarcodeParser(Component):
+    """
+    Some barcodes can have complex data structure
+    """
+
+    _name = "shopfloor.barcode.parser"
+    _inherit = "shopfloor.process.action"
+    _usage = "barcode"
+
+    def __init__(self, search_action: SearchAction):
+        # Get search action keys
+        self.search_action = search_action
+
+    @property
+    def _authorized_barcode_types(self):
+        return self.search_action._barcode_type_handler.keys()
+
+    def parse(self, barcode) -> list[BarcodeResult]:
+        """
+        This method will parse the barcode and return the
+        value with its type if determined.
+
+        Override this to implement specific parsing
+
+        """
+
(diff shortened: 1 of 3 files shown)


shopfloor/actions/search.py
```diff
@@ -57,6 +57,8 @@ def _barcode_type_handler(self):
             "packaging": self.packaging_from_scan,
             "delivery_packaging": self.delivery_packaging_from_scan,
             "origin_move": self.origin_move_from_scan,
+            # Extra data can be contained in barcodes
+            "expiration_date": self.dummy_from_scan,
```
simahawk: wouldn't be better to have `self.expiration_date_from_scan` ? Even if it does nothing it's already a good hook.

shopfloor/actions/barcode_parser.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class BarcodeParser(Component):
+    """
+    Some barcodes can have complex data structure
+    """
+
+    _name = "shopfloor.barcode.parser"
+    _inherit = "shopfloor.process.action"
+    _usage = "barcode"
```
simahawk, line 25: maybe I forgot how I implemented this machinery :sweat_smile:  but.. how is this supposed to kick in?

rousseldenis: @simahawk Arf, it seems a part of the code has been dismantled...

In fact, the idea is to pass the parser to the search() and to return the result which can be multi-part (with several information inside it).

Very useful in reception scenario where you want to fill in automatically several fields.

shopfloor/actions/search.py
rousseldenis: Done

shopfloor/actions/barcode_parser.py
rousseldenis: That is in the second commit I lost but recovered.

rousseldenis:
@sebalix @jbaudoux

rousseldenis:
@jbaudoux

rousseldenis:
@lmignon @simahawk

shopfloor/actions/search.py
```diff
@@ -188,3 +206,10 @@ def origin_move_from_scan(self, barcode, extra_domain=None):
         if extra_domain:
             outgoing_move_domain = AND([outgoing_move_domain, extra_domain])
         return model.search(outgoing_move_domain)
+
+    def dummy_from_scan(self, barcode):
+        return None
+
+    def expiration_date_from_scan(self, barcode):
+        # TODO
```
jbaudoux, line 214: still TODO ?

rousseldenis: That's following @simahawk 

For the time being, we don't search for expiration dates but we need to have a handler to get a parse result for dates from GS1

jbaudoux (approved):

PR #1003, opened 2025-04-15
[17.0][MIG] stock_service_level

damdam-s:
migrated from 14.0


petrus-v (approved):
Thanks for taking care of this migrations

stock_service_level/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# @author Pierre Verkest <pierreverkest84@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock service level",
+    "summary": "Manage stock route by service level",
+    "version": "17.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Warehouse Management",
+    "website": "https://github.com/OCA/wms",
+    "author": "Pierre Verkest <pierreverkest84@gmail.com>,"
```
jbaudoux, line 11: maybe just your name here or it should be Foodles?
```suggestion
    "author": "Pierre Verkest,"
```

stock_service_level/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2024 Foodles (http://www.foodles.co).
+# @author Pierre Verkest <pierreverkest84@gmail.com>
```
jbaudoux, line 2: Can you just keep copyright line? (same for other files)
You can put additional info in CREDITS file

damdam-s:
@jbaudoux as part of migration, I'm not sure we should do it during the technical migration process. 

@petrus-v any idea regarding this as you are the one being credited for this module 😄

stock_service_level/__manifest__.py
petrus-v: I'm fine with this suggestion

stock_service_level/__manifest__.py
petrus-v: We have already a lot of modules that use this pattern https://github.com/OCA/event/blob/15.0/event_sale_session/models/event_session.py#L3.

I'm fine if it's moved to a credit/author files and maybe with the initial author mentioned.

@jbaudoux is it an OCA convention?  do you know why prefer that other ways ?

Also I've the feeling it's not the best moment to change that in a migration PR which add risk to get no response from the initial author and freeze migration task.

Finally we should probably add pre-commit hook to inform developers on that practices and eventually migrate existing case.

my 2cts

jbaudoux: As your are the author, it's fine for me if it's fine for you :)
I'm not aware of conventions

nico-foodles (approved):

stock_service_level/tests/common.py
```diff
@@ -0,0 +1,67 @@
...
+        cls.product.categ_id.route_ids |= cls.pick_ship_route
+        cls.location_1 = cls.env["stock.location"].create(
+            {"name": "loc1", "location_id": cls.warehouse.lot_stock_id.id}
+        )
+        cls.location_2 = cls.env["stock.location"].create(
+            {"name": "loc2", "location_id": cls.warehouse.lot_stock_id.id}
+        )
+
+    def _update_product_stock(self, qty, location=None):
+        inventory = self.env["stock.inventory"].create(
```
nico-foodles, line 46: You can no longer do this in odoo 17

nico-foodles (changes requested):

stock_service_level/__manifest__.py
rousseldenis: @damdam-s Could you change this before we merge ?

PR #1004, opened 2025-04-15
[16.0][ADD] shopfloor_product_dimension

rousseldenis:


shopfloor_product_dimension/actions/detail.py
```diff
@@ -0,0 +1,26 @@
...
+class DataDetailAction(Component):
+    _inherit = "shopfloor.data.detail.action"
+
+    @property
+    def _product_dimension_parser(self):
+        return [
+            "product_length:length",
+            "product_height:height",
+            "product_width:width",
+            "weight",
```
lmignon, line 15: @rousseldenis Why not the weight uom?

rousseldenis: I still need to fix this as this introduce a dependency on product_logistics_uom and I'm doubting on adding it or not.

giarve: @rousseldenis go for it, it already is a dependency for 'shopfloor' base module :)

jbaudoux (approved):
LG

nicolas-delbovier-acsone (approved):
LGTM

PR #1005, opened 2025-04-15
[16.0][ADD] shopfloor_reception_product_dimension/mobile

rousseldenis:
Depends on:

- #1004


lmignon (changes requested):
IMO, the set dimension step should only be visible if no dimension is set on product. Otherwise, you'll require an additional click on each product you process

nicolas-delbovier-acsone (commented):
LGTM, can we re-open this?

Issue #976, opened 2025-01-29
[16.0] stock_storage_type : allow new product refactor


jbaudoux:
During migration to v16, some refactoring was required as some concepts had been integrated in odoo.

To support more advanced decision for "allow new product", an "allow new product" had been added on the capacity.

![Image](https://github.com/user-attachments/assets/8ccf8265-129d-4e91-8e45-3aafa899627c)

The standard "allow new product" on the storage category is not used.

This design forced to create several package types while conceptually it is still the same package type. Also you cannot express a proper condition based on what it being moved (like based on other product attribute or based on the quantity) as you are limited to the package type as unique condition.

I propose to drop that complexity on the capacity and use by default the odoo standard storage category "allow new product".
And then on the storage category, add a more advanced mode where you can define a storage category condition (like the storage location sequence condition) and an "allow new product".

cc @rousseldenis

sebalix:
New tab "Advanced Conditions" (new data model) on Storage Category form.
Two columns for this new data model:

- Condition (with a name and code to run, new data model or re-use existing one)
- Allow New Product (policy to apply)

PR #1006, opened 2025-04-18
[16.0][REF] stock_storage_type: remove `allow_new_product` field from storage capacity lines

sebalix:
Module is now using the standard `allow_new_product` field set on the storage category + allow to force another value based on "Allow New Product" conditions (Python code snippets).

Address #976 

## TODO:
- [x] remove `allow_new_product` field from capacity lines + adapt code & tests
- [x] add a new "Storage Condition" data model to display on form view, allowing to set a different `allow_new_product` value based on a condition (domain)
- [x] write a migration script to move old `allow_new_product` values from capacity lines to these new conditions (tested from a demo DB created with current 16.0 version)


sebalix:
Seems that these `shopfloor` tests are broken with these changes:

- ./shopfloor/tests/test_location_content_transfer_putaway.py
    - test_ignore_no_putaway_available
    - test_normal_putaway
- ./shopfloor/tests/test_single_pack_transfer_putaway.py
    - test_ignore_no_putaway_available

Checking

EDIT: should be fixed, I had to split the `allow_new_product` value in its own field, and the code snippet returning a boolean value telling if the condition is met or not, it was a bad idea to define the `allow_new_product` value directly in the code snippet, as the different computed+stored field on `stock.location` need to check all category conditions if one of them could make the location elligible (as it was the case before with capacities). Implementation updated accordingly.

stock_storage_type/models/stock_storage_category_condition.py
```diff
@@ -0,0 +1,157 @@
...
+    _inherit = "stock.storage.condition.mixin"
+    _name = "stock.storage.category.condition"
+    _description = "Stock Storage Category Condition"
+
+    def _selection_allow_new_product(self):
+        return self.env["stock.storage.category"]._fields["allow_new_product"].selection
+
+    allow_new_product = fields.Selection(
+        selection=_selection_allow_new_product, default="mixed", required=True
+    )
```
jbaudoux: This is not a condition, this is the rule you apply if the condition is True

jbaudoux (changes requested):
There is a model design issue.
You need an intermediate model `StockStorageCategoryAllowNewProduct` that defines the condition from `StockStorageCategoryAllowNewProductCond` and the allow_new_product rule

stock_storage_type/models/stock_quant.py
```diff
@@ -49,8 +49,15 @@ def _check_storage_capacities(self):
             lots_in_location = other_quants_in_location.mapped("lot_id")
             capacity_fails = []
             for capacity in allowed_capacities:
```
jbaudoux: allowed_capacities is already checked above, now you need to look at the conditions and its related allow_new_product to see if one matches.
Content of the loop should be the same, only fetch from the advanced allow new product table, the rule that applies and then fallback on the category allow_new_product value

Something like:
```
storage_category_id = location.computed_storage_category_id
allow_new_product = storage_category_id.allow_new_product
# check advanced mode with conditions
for advanced in storage_category_id.advanced_allow_new_product_ids:
   # eval condition
   if advanced.condition:
      allow_new_product = advanced.allow_new_product
      break
```

you can put this in `get_allow_new_product()`


```
# then do the checks for this allow_new_product (you don't need to loop anymore)
if allow_new_product == "empty" and other_quants_in_location:
...
```

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -26,26 +16,22 @@ class StorageCategoryProductCapacity(models.Model):
...
 
-    @api.model
     def _get_display_name_attributes(self):
-        """
-        Adds the storage capacity attributes to compose the display name
-        """
         attributes = super()._get_display_name_attributes()
-        value = self._fields["allow_new_product"].convert_to_export(
-            self.allow_new_product, self
+        value = self.storage_category_id._fields["allow_new_product"].convert_to_export(
```
jbaudoux: Do not add the value in the name if you have advanced_allow_new_product_ids defined on the category

sebalix:
@jbaudoux I updated the PR by adding the new intermediate data model, can you update your review?

stock_storage_type/models/stock_quant.py
```diff
@@ -49,8 +49,15 @@ def _check_storage_capacities(self):
...
+                allow_new_product = capacity.storage_category_id.get_allow_new_product(
```
jbaudoux: you can remove the for loop
```suggestion
            allow_new_product = location.computed_storage_category_id.get_allow_new_product(
```

stock_storage_type/models/stock_quant.py
```diff
@@ -49,8 +49,15 @@ def _check_storage_capacities(self):
...
+                    product=quant.product_id,
+                    storage_category_capacity=capacity,
```
jbaudoux:
```suggestion
                    storage_category=location.computed_storage_category_id,
```

stock_storage_type/models/stock_quant.py
```diff
@@ -49,8 +49,15 @@ def _check_storage_capacities(self):
...
+                    package_type=capacity.package_type_id,
```
jbaudoux:
```suggestion
                    package_type=package_type,
```

stock_storage_type/models/stock_storage_category.py
```diff
@@ -92,3 +97,31 @@ def _compute_max_weight_in_kg(self):
                 to_unit=uom_kg,
                 round=False,
             )
+
+    def get_allow_new_product(
+        self,
+        product,
+        storage_category_capacity=None,
```
jbaudoux:
```suggestion
        storage_category=None,
```

stock_storage_type/models/stock_storage_category.py
```diff
@@ -92,3 +97,31 @@ def _compute_max_weight_in_kg(self):
...
+
+        It first evaluates the conditions based on different criteria, and if no
+        value can be found among them it fallbacks on the category option value.
+        """
+        self.ensure_one()
+        for rule in self.allow_new_product_ids:
+            res = rule.condition_id.evaluate(
+                self,
+                product,
+                storage_category_capacity,
```
jbaudoux:
```suggestion
                storage_category,
```

stock_storage_type/models/stock_storage_category_allow_new_product_cond.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+    ]
+
+    def _default_code_snippet_docs(self):
+        return """
+        Available vars:
+        * condition (recordset)
+        * storage_category (recordset)
+        * product (recordset)
+        * storage_category_capacity (recordset)
```
jbaudoux: drop this
```suggestion
```

stock_storage_type/models/stock_storage_category_allow_new_product_cond.py
```diff
@@ -0,0 +1,147 @@
...
+
+        Must initialize a boolean 'result' variable set to True when condition is met
+
+        """
+
+    def _get_code_snippet_eval_context(
+        self,
+        storage_category,
+        product,
+        storage_category_capacity,
```
jbaudoux:
```suggestion
```

stock_storage_type/models/stock_storage_category_allow_new_product_cond.py
```diff
@@ -0,0 +1,147 @@
...
+        :returns: dict -- evaluation context given to safe_eval
+        """
+        self.ensure_one()
+        return {
+            "env": self.env,
+            "user": self.env.user,
+            "condition": self,
+            "storage_category": storage_category,
+            "product": product,
+            "storage_category_capacity": storage_category_capacity,
```
jbaudoux:
```suggestion
```

jbaudoux (commented):
looks nearly good

stock_storage_type/models/stock_storage_category_allow_new_product.py
```diff
@@ -0,0 +1,30 @@
...
+    def _selection_allow_new_product(self):
+        return self.env["stock.storage.category"]._fields["allow_new_product"].selection
+
+    storage_category_id = fields.Many2one(
+        comodel_name="stock.storage.category",
+        ondelete="cascade",
+        required=True,
+        index=True,
+    )
+    condition_id = fields.Many2one(
```
jbaudoux: allow to put multiple conditions (like on the location sequence), it's much more practical
```suggestion
    condition_ids = fields.Many2many(
```

stock_storage_type/models/stock_storage_category_capacity.py
sebalix: Done

stock_storage_type/models/stock_quant.py
sebalix: Done

stock_storage_type/models/stock_storage_category.py
sebalix: Not needed, `self` is already the current category

stock_storage_type/models/stock_storage_category_allow_new_product.py
sebalix: Done, and added a tooltip on condition fields telling that all conditions have to match to apply the rule.

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -60,7 +46,7 @@ def _compute_has_restrictions(self):
         for capacity in self:
             capacity.has_restrictions = any(
                 [
-                    capacity.allow_new_product != "mixed",
+                    capacity.storage_category_id.allow_new_product != "mixed",
                     capacity.storage_category_id.max_height,
                     capacity.storage_category_id.max_weight,
                 ]
```
jbaudoux: `has_restrictions` should move to category and the warning should also be displayed if any from allow_new_product_ids != mixed

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -94,11 +80,18 @@ def _domain_location_storage_type(self, candidate_locations, quants, products):
             ("computed_storage_category_id.capacity_ids", "in", self.ids),
         ]
         # Build the domain using the 'allow_new_product' field
-        if self.allow_new_product == "empty":
+        allow_new_product = self.storage_category_id.get_allow_new_product(
+            product=products,
+            storage_category_capacity=self,
```
jbaudoux: This method should move to category and also take the package type in argument. There is no need anymore to loop over capacities in the caller;

Also in `stock_location`, the methods are still using `storage_type` which should be renamed to `storage_capacity` for better understanding.

jbaudoux (changes requested):
stock location should be simplified as now all constrains are move to the category, so the capacity should be there just to find the right categories.
There is no need anymore to search for pertinent capacities, you can directly take pertinent categories. cf: `Capacity.search(pertinent_loc_s_t_domain)`

stock_storage_type/models/stock_storage_category.py
jbaudoux: but storage_category_capacity should not play a role here anymore, it shows there is still something to simplify

stock_storage_type/models/stock_storage_category_capacity.py
```diff
@@ -1,22 +1,12 @@
 # Copyright 2022 ACSONE SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
-from odoo import _, api, fields, models
+from odoo import api, fields, models
 
 
 class StorageCategoryProductCapacity(models.Model):
 
     _inherit = "stock.storage.category.capacity"
```
jbaudoux: it looks like you can move remaining stuffs here to category

stock_storage_type/models/stock_storage_category.py
```diff
@@ -98,6 +101,79 @@ def _compute_max_weight_in_kg(self):
...
+        A storage category has restrictions when it:
+            - does not accept mixed products
+            - or does not accept mixed lots
+            - or do have a maximum height set on its category
+            - or do have a maximum weight set on its category
+        """
+        for rec in self:
+            rec.has_restrictions = any(
+                [
+                    rec.allow_new_product != "mixed",
```
sebalix, line 121: Need to fix that by looping on `allow_new_product_ids` too :thinking:

stock_storage_type/models/stock_location.py
```diff
@@ -513,30 +507,30 @@ def select_first_allowed_location(self, package_type, quants, products):
         allowed = self.select_allowed_locations(package_type, quants, products, limit=1)
         return allowed
 
-    def _domain_location_storage_type_constraints(self, package_type, quants, products):
-        """Compute the domain for the location storage type which match the package
+    def _domain_storage_category_constraints(self, package_type, quants, products):
+        """Compute the domain for the storage category which match the package
         storage type
```
```suggestion
        type
```

stock_storage_type/models/stock_location.py
```diff
@@ -589,45 +583,31 @@ def select_allowed_locations(self, package_type, quants, products, limit=None):
         """
         # We have package who may be placed in a stock.location
         #
-        # 1. On the stock.location there are location_storage_type and on the
+        # 1. On the stock.location there are capacities and on the
```
```suggestion
        # 1. On the location there is a storage category that defines which package type is allowed. This is given by the storage capacities.
```

stock_storage_type/models/stock_location.py
```diff
@@ -589,45 +583,31 @@ def select_allowed_locations(self, package_type, quants, products, limit=None):
...
         # packages there are package_storage_type. Between both, there's a m2m
-        # who says which package ST can be placed in which location ST
+        # who says which package ST can be placed in which capacity.
         #
-        # 2. On a location_ST there are some additional restrictions: a -
-        # capacity (volume / height / weight) and b - properties (boolean
+        # 2. On a capacity and its category there are some additional restrictions:
```
```suggestion
        # 2. On a storage category there are some additional restrictions:
```

stock_storage_type/models/stock_location.py
```diff
@@ -589,45 +583,31 @@ def select_allowed_locations(self, package_type, quants, products, limit=None):
...
+        # a - capacity (volume / height / weight) and b - properties (boolean
         # flags: only empty, don't mix lots, don't mix products)
-        Capacity = self.env["stock.storage.category.capacity"]
+        Category = self.env["stock.storage.category"]
         _logger.debug(
             "select allowed location for package storage type %s (q=%s, p=%s)",
```
```suggestion
            "select allowed location for package type %s (q=%s, p=%s)",
```

stock_storage_type/models/stock_location.py
```diff
@@ -589,45 +583,31 @@ def select_allowed_locations(self, package_type, quants, products, limit=None):
...
-        # who says which package ST can be placed in which location ST
+        # who says which package ST can be placed in which capacity.
         #
-        # 2. On a location_ST there are some additional restrictions: a -
-        # capacity (volume / height / weight) and b - properties (boolean
+        # 2. On a capacity and its category there are some additional restrictions:
+        # a - capacity (volume / height / weight) and b - properties (boolean
         # flags: only empty, don't mix lots, don't mix products)
-        Capacity = self.env["stock.storage.category.capacity"]
+        Category = self.env["stock.storage.category"]
```
jbaudoux: maybe move where it's used

stock_storage_type/models/stock_location.py
```diff
@@ -589,45 +583,31 @@ def select_allowed_locations(self, package_type, quants, products, limit=None):
...
         _logger.debug(
             "select allowed location for package storage type %s (q=%s, p=%s)",
             package_type.name,
             quants,
             products.mapped("name"),
         )
         # 1: filter locations on compatible storage type
```
```suggestion
        # 1: find pertinent storage categories
```

jbaudoux (commented):
There are some comments and docstrings to update :)

`Filter allowed locations for a storage type`
to change to
`Filter allowed locations for a package type`

It looks good. I'll test it now.

stock_storage_type/models/stock_location.py
```diff
@@ -64,7 +64,7 @@ class StockLocation(models.Model):
         help="technical field: True if the location is empty "
         "and there is no pending incoming products in the location. "
         " Computed only if the location needs to check for emptiness "
-        '(has an "only empty" location storage type).',
+        '(has an "only empty" policy).',
         recursive=True,
     )
     # TODO: Maybe renaming these fields as there are already such fields
```
jbaudoux, line 70: Side note, this can be replaced by https://github.com/OCA/stock-logistics-warehouse/blob/16.0/stock_location_pending_move/models/stock_location.py

sebalix: Will put that in ROADMAP and will do it during migration to v18, if it's OK for you?

sebalix: => https://github.com/OCA/stock-logistics-warehouse/pull/2321

jbaudoux (approved):
Thanks, looks good :)

I tested the put-away computation & migration and it works

I let you squash commits

sebalix:
Commit squashed and commit message rewritten as well.

jbaudoux:
For review @rousseldenis

stock_storage_type/tests/test_storage_type_move.py
```diff
@@ -38,28 +38,33 @@ def _test_confirmed_move(self, product=None):
         return move_to_assign
 
     def test_not_only_empty_confirmed_move(self):
-        self.pallets_location_storage_type.write({"allow_new_product": "mixed"})
+        category = self.pallets_location_storage_type.storage_category_id
+        category.allow_new_product_ids.allow_new_product = "mixed"
```
sebalix, line 42: Was a bit brutal with this change, this doesn't really comply with what we are testing, need to fix that:

- set the "Allow new product" policy through a rule, not globally on the storage category (especially true for tests on cardboxes with different lots below in this file)

stock_storage_type/tests/test_storage_type_move.py
```diff
@@ -38,28 +38,33 @@ def _test_confirmed_move(self, product=None):
...
+        category = self.pallets_location_storage_type.storage_category_id
+        category.allow_new_product_ids.allow_new_product = "same"
         move = self._test_confirmed_move()
         self.assertEqual(
             move.move_line_ids.location_dest_id, self.pallets_bin_1_location
         )
 
     def test_do_not_mix_products_confirmed_move_nok(self):
-        self.pallets_location_storage_type.write({"allow_new_product": "same"})
+        self.pallets_location_storage_type.storage_category_id.write(
```
sebalix, line 65: <s>Was a bit brutal with this change, this doesn't really comply with what we are testing, need to fix that:

- set the "Allow new product" policy through a rule, not globally on the storage category (especially true for tests on cardboxes with different lots below in this file)
</s>

EDIT: that's OK in fact, this doesn't impact the test with the current data.

PR #1009, opened 2025-04-23
[16.0][FW PORT] stock_available_to_promise_release: Allow unrelease processed qties

mmequignon:
forward port of:
* https://github.com/OCA/wms/pull/971

I missed the "draft" button

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -5,11 +5,12 @@
 import itertools
 import logging
 import operator as py_operator
+from collections import defaultdict
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 from odoo.osv import expression
-from odoo.tools import date_utils, float_compare, float_round, groupby
+from odoo.tools import date_utils, float_compare, float_is_zero, float_round, groupby
 
 from odoo.addons.stock.models.stock_move import StockMove as StockMoveBase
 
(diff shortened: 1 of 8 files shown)


stock_available_to_promise_release/models/stock_move.py
```diff
@@ -87,34 +88,36 @@ def _is_unreleaseable(self):
...
-          * a picking is already printed. (The work on the picking is planned and
-            we don't want to change it)
-          * a quantity done is recorded
-          * the processed origin moves is not consumed by the dest moves.
+        the processed origin moves is not consumed by the dest moves.
         """
         self.ensure_one()
+        if self.rule_id.allow_unrelease_return_done_move:
+            return True
         origin_done_moves = origin_moves.filtered(lambda m: m.state == "done")
```
jbaudoux: Because there is also the return_picking_type_id condition that must be valid
```suggestion
        origin_done_moves = origin_moves.filtered(lambda m: m.state == "done")
        if self.rule_id.allow_unrelease_return_done_move:
            origin_done_moves = origin_done_moves.filtered(lambda m: not m.picking_type_id.return_picking_type_id)
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -685,29 +763,74 @@ def unrelease(self, safe_unrelease=False):
...
+                    moves_to_cancel.write({"propagate_cancel": False})
+                    moves_to_cancel_for_move |= moves_to_cancel
+                    qty_canceled = sum(moves_to_cancel.mapped("product_qty"))
+                # checking that for the current step (pick/pack/ship)
+                # move.product_uom_qty == step.cancelled_qty + move.returned_quanty
+                # If not the case, we have to move back goods in stock.
+                qty_to_return = qty_to_cancel - qty_canceled
+                if float_is_zero(qty_to_return, precision_rounding=rounding):
+                    continue
+                done_moves = origin_moves.filtered(lambda m: m.state == "done")
```
jbaudoux, line 809: Reduce the qty to return to what was not consumed
```suggestion
                done_moves = origin_moves.filtered(lambda m: m.state == "done")
                # in case of canceled origin_moves, the quantity to return must
                # be limited to the quantity not consumed
                done_dest_moves = done_moves.move_dest_ids.filtered(lambda m: m.state == "done")
                returnable_qty = sum(done_moves.mapped("product_qty")) - sum(done_dest_moves.mapped("product_qty"))
                qty_to_return = min(qty_to_return, returnable_qty)
                if float_compare(qty_to_return, 0, precision_rounding=rounding) <= 0:
                    continue
```

jbaudoux:
Failing test_picking_priority fixed in https://github.com/camptocamp/wms/pull/123

stock_available_to_promise_release/models/stock_route.py
```diff
@@ -7,6 +7,16 @@
 class StockRoute(models.Model):
     _inherit = "stock.route"
 
+    allow_unrelease_return_done_move = fields.Boolean(
```
rousseldenis, line 10: IMHO this introduce too big functions for return management here. A new module with only the good entry points here could be better

jbaudoux: But it's a base behavior of the unrelease mechanism of this module
I didn't want it in a separate module

jbaudoux (approved):

PR #1011, opened 2025-04-23
[IMP] stock_release_channel: Allow to sleep and reassign pickings

rousseldenis:
In some cases, we don't want to use an open channel to deliver pickings anymore. But the internal transfers may have been done.

In that case, we should not unrelease done pickings but just removing the channel and triggering an assignation (to put those transfers in another open channel).

### DIFF
--- a/stock_release_channel/models/stock_release_channel.py
+++ b/stock_release_channel/models/stock_release_channel.py
@@ -225,6 +225,10 @@ class StockReleaseChannel(models.Model):
         compute="_compute_is_action_wake_up_allowed",
         help="Technical field to check if the " "action 'Wake Up' is allowed.",
     )
+    is_action_safe_sleep_allowed = fields.Boolean(
+        compute="_compute_is_action_safe_sleep_allowed",
+        help="Technical field to check if the action 'Safe Sleep' is allowed.",
+    )
     is_release_allowed = fields.Boolean(
         compute="_compute_is_release_allowed",
         search="_search_is_release_allowed",
@@ -269,6 +273,13 @@ class StockReleaseChannel(models.Model):
     def _compute_is_release_allowed(self):
         for rec in self:
             rec.is_release_allowed = rec.state == "open" and not rec.release_forbidden
+
+    @api.depends("state")
+    def _compute_is_action_safe_sleep_allowed(self):
+        for rec in self:
+            rec.is_action_safe_sleep_allowed = bool(
+                rec.state in ["locked", "open"] and rec.open_picking_ids
+            )
 
     def _compute_show_last_picking_done(self):
         for rec in self:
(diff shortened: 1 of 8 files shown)


stock_release_channel/models/stock_release_channel.py
```diff
@@ -270,6 +274,13 @@ def _compute_is_release_allowed(self):
         for rec in self:
             rec.is_release_allowed = rec.state == "open" and not rec.release_forbidden
 
+    @api.depends("state")
+    def _compute_is_action_safe_sleep_allowed(self):
+        for rec in self:
+            rec.is_action_safe_sleep_allowed = bool(
+                rec.state in ["locked", "open"] and rec.open_picking_ids
+            )
+
```
mt-software-de, line 283: Isn't it possible to use `is_action_sleep_allowed`?
If not could we move `state in ["locked", "open"]` into an own method to be reused in both compute methods?

```suggestion
    @api.depends("state", "is_action_sleep_allowed")
    def _compute_is_action_safe_sleep_allowed(self):
        for rec in self:
            rec.is_action_safe_sleep_allowed = bool(
                rec.is_action_sleep_allowed and rec.open_picking_ids
            )

```

stock_release_channel/models/stock_release_channel.py
```diff
@@ -873,15 +884,22 @@ def action_unlock(self):
         self.write({"state": "open"})
 
     def action_sleep(self):
+        # TODO: Add safe argument in function arguments in further version to
+        # avoid context call
+        safe_sleep = self.env.context.get("release_channel_safe_sleep", False)
```
mt-software-de: Why didn't you do it now. 
Couldn't you move the code of `action_sleep` into a new method for example `def _put_to_sleep(safe=False)` and call this then from `action_sleep` and `action_safe_sleep`?

mt-software-de (changes requested):
Thx. This sounds great.

jbaudoux (commented):
What about only keeping the sleep action and show a confirmation wizard if there are started or done pickings ? Something similar to https://github.com/OCA/wms/blob/16.0/stock_release_channel_shipment_advice_deliver/models/stock_release_channel.py#L187
You can highlight `discard` button in the wizard and have a `sleep & reassign` button calling that safe sleep

stock_release_channel/models/stock_release_channel.py
```diff
@@ -873,15 +884,22 @@ def action_unlock(self):
...
         self._check_is_action_sleep_allowed()
         pickings_to_unassign = self.env["stock.picking"].search(
             self._get_picking_to_unassign_domain()
         )
         pickings_to_unassign.write({"release_channel_id": False})
-        pickings_to_unassign.unrelease()
+        pickings_to_unassign.unrelease(safe_unrelease=safe_sleep)
```
jbaudoux: We need to fix conflict with https://github.com/OCA/wms/pull/1009

If you enable to return done pickings on the route, then it will not reassign as generating a return is now considered as a safe unrelease.
Should we add another argument to prevent any return ?

cc @mt-software-de

stock_release_channel/models/stock_release_channel.py
rousseldenis: > If not could we move `state in ["locked", "open"]` into an own method to be reused in both compute methods?

@mt-software-de In fact, this is too broad as if channel is delivered but not asleep, button is still visible but has no interest as it should be when outgoing transfers are not done/cancel

stock_release_channel/models/stock_release_channel.py
rousseldenis: Ok, yes, good point

stock_release_channel/models/stock_release_channel.py
```diff
@@ -270,6 +274,13 @@ def _compute_is_release_allowed(self):
         for rec in self:
             rec.is_release_allowed = rec.state == "open" and not rec.release_forbidden
 
+    @api.depends("state")
+    def _compute_is_action_safe_sleep_allowed(self):
+        for rec in self:
+            rec.is_action_safe_sleep_allowed = bool(
+                rec.state in ["locked", "open"] and rec.open_picking_ids
```
jbaudoux, line 281: I don't think checking open_picking_ids is useful

stock_release_channel/models/stock_release_channel.py
```diff
@@ -872,16 +883,33 @@ def action_unlock(self):
...
+        pickings_to_unassign.unrelease(safe_unrelease=safe)
         self.write({"state": "asleep"})
         pickings_to_unassign._delay_assign_release_channel()
 
+    def action_sleep(self):
+        self._sleep()
+
+    def action_safe_sleep(self):
+        self._sleep(safe=True)
+        return {"type": "ir.actions.act_window_close"}
```
jbaudoux, line 911: I would not add a second button but manage it in existing `action_sleep`.
Test if there are done internal transfer.
If True, raise a wizard popup showing them and asking to confirm the reassignment of those to another channel. on confirm, call `self._sleep(safe=True)`.
Else call `self._sleep()`

jbaudoux (changes requested):

PR #1013, opened 2025-04-24
[IMP] shopfloor_reception: Allows to configure today pickings

rousseldenis:
Allows to use also additional domain and sort method.

Note: This is intended to get flexibility on first screen as pickings scheduled dates are often wrong.


shopfloor_reception/models/shopfloor_menu.py
```diff
@@ -0,0 +1,46 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import _, api, exceptions, fields, models
+
+FILTER_TODAY_SCHEDULED_PICKINGS_HELP = """
+By default, at first step, filter the available
+pickings with the ones that are scheduled for today.
+"""
+
+
+class ShopfloorMenu(models.Model):
```
mmequignon, line 11: I feel like this option is generic enough to go in shopfloor.
Could be used in other scenarios.

mmequignon (approved):
LG! Minor suggestion

shopfloor_reception/services/reception.py
```diff
@@ -76,18 +76,25 @@ def _scheduled_date_today_domain(self):
         domain.append(("scheduled_date", "<=", today_end))
         return domain
 
-    def _get_today_start_end_datetime(self):
+    def _get_today_start_end_datetime(self, naive=True):
         company = self.env.company
         tz = company.partner_id.tz or "UTC"
```
jbaudoux: should be the tz of the warehouse

shopfloor_reception/services/reception.py
```diff
@@ -76,18 +76,25 @@ def _scheduled_date_today_domain(self):
...
-        today_end = fields.Datetime.end_of(today, "day")
-        today_start_localized = (
-            pytz.timezone(tz).localize(today_start).astimezone(pytz.utc)
-        )
-        today_end_localized = pytz.timezone(tz).localize(today_end).astimezone(pytz.utc)
+        today_start = today_start_localized = fields.Datetime.start_of(today, "day")
+        today_end = today_end_localized = fields.Datetime.end_of(today, "day")
+        if not naive:
+            today_start_localized = (
+                pytz.timezone(tz).localize(today_start).astimezone(pytz.utc)
```
jbaudoux, line 93: I think it's the other way around
```suggestion
                today_start.astimezone(pytz.utc).astimezone(tz)
```

jbaudoux (changes requested):

shopfloor_reception/models/shopfloor_menu.py
rousseldenis: @mmequignon I've put it here because today's pickings are only used in reception scenario for now.

shopfloor_reception/services/reception.py
rousseldenis: I don't think so as we should transform to utc as it will be transmitted to domain.
I can add a test for that

rousseldenis:
@jbaudoux

shopfloor_reception/services/reception.py
```diff
@@ -76,18 +76,31 @@ def _scheduled_date_today_domain(self):
...
-        tz = company.partner_id.tz or "UTC"
-        today = fields.Datetime.today()
-        today_start = fields.Datetime.start_of(today, "day")
-        today_end = fields.Datetime.end_of(today, "day")
-        today_start_localized = (
-            pytz.timezone(tz).localize(today_start).astimezone(pytz.utc)
+        warehouse = self.picking_types.warehouse_id
+        tz = (
+            warehouse.partner_id.tz
+            if (len(warehouse) == 1 and warehouse.partner_id.tz)
```
jbaudoux, line 85:
```suggestion
            if (len(warehouse.partner_id.mapped("tz")) == 1 and warehouse.partner_id.tz)
```

shopfloor_reception/services/reception.py
```diff
@@ -76,18 +76,31 @@ def _scheduled_date_today_domain(self):
         domain.append(("scheduled_date", "<=", today_end))
```
jbaudoux, line 76: but this one also needs tz conversion, isn't it ?

shopfloor_reception/services/reception.py
```diff
@@ -76,18 +76,31 @@ def _scheduled_date_today_domain(self):
...
         return domain
 
-    def _get_today_start_end_datetime(self):
+    def _get_today_start_end_datetime(self, naive=True):
```
jbaudoux, line 79: what about always returning a naive but localized start/end?
and remove `p.scheduled_date*.astimezone(pytz.utc)*` from the caller

shopfloor_reception/services/reception.py
rousseldenis: Not sure about that. TZ can be False

jbaudoux: that's tested after the and

shopfloor_reception/services/reception.py
rousseldenis: This is done in `_get_today_start_end_datetime()` :smile:

shopfloor_reception/services/reception.py
rousseldenis: > what about always returning a naive but localized start/end?

Because when using it in a domain, should be naive. But using records fields, should be localized.

jbaudoux (approved):

PR #1014, opened 2025-04-24
[14.0][FIX] stock_warehouse_flow: fix empty picking

jbaudoux:
If the initial picking gets emptied, mark as cancel

This was lacking on action_confirm as flow is called with no assign

cc @mmequignon @mt-software-de

### DIFF
--- a/stock_warehouse_flow/models/stock_move.py
+++ b/stock_warehouse_flow/models/stock_move.py
@@ -26,6 +26,7 @@ class StockMove(models.Model):
         # its chained moves (if any)
         FLOW = self.env["stock.warehouse.flow"]
         move_ids_to_confirm = []
+        old_pickings = self.picking_id
         for move in self:
             if not move._apply_flow_on_action_confirm():
                 move_ids_to_confirm.append(move.id)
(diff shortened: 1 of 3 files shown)


mt-software-de (approved):
LGTM

stock_warehouse_flow/models/stock_move.py
```diff
@@ -36,6 +37,14 @@ def _action_confirm(self, merge=True, merge_into=False):
...
+        res = super(StockMove, moves_to_confirm)._action_confirm(
             merge=merge, merge_into=merge_into
         )
+        # In case the move was already assigned to a picking and action_confirm
+        # was called again, with the flow it may have changed the conditions
+        # and assigned to a different picking. If the initial picking is empty,
+        # marked it as canceled.
+        for old_picking in old_pickings:
+            if not old_picking.move_lines:
+                old_picking.state = "cancel"
```
sebalix: Should we not call `action_cancel` instead?

Odoo is also setting the `is_locked` field:
```python
    def action_cancel(self):
        self.mapped('move_lines')._action_cancel()
        self.write({'is_locked': True})
        self.filtered(lambda x: not x.move_lines).state = 'cancel'
        return True
```

jbaudoux: oh thanks, I wasn't aware they fixed it in odoo
https://github.com/odoo/odoo/commit/4a4ae31882db94080809c600209453119c30f2b4

sebalix (approved):

PR #1016, opened 2025-04-25
[FIX] shopfloor_reception_mobile: Make product info works on picking detail

rousseldenis:
In reception, barcode could not have been put yet. So, fallback on default code for scan info.

### DIFF
--- a/shopfloor_reception_mobile/static/src/scenario/reception.js
+++ b/shopfloor_reception_mobile/static/src/scenario/reception.js
@@ -346,14 +346,18 @@ const Reception = {
                 ],
             };
         },
-        picking_detail_options_for_select_move: function (move) {
+        picking_detail_options_for_select_move: function () {
             return {
                 show_title: true,
                 showActions: false,
                 list_item_options: {
                     loud_title: true,
                     title_action_field: {
-                        action_val_path: "name",
+                        action_val_path: function (record, field) {
+                            return record.product.barcode
+                                ? "product.barcode"
+                                : "product.default_code";
+                        },
                     },
                     list_item_klass_maker: this.move_card_color,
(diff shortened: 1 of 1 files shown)


rousseldenis:
@jbaudoux @lmignon @phschmidt @sbejaoui A quick win :-)

jbaudoux:
cc @mt-software-de

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -346,14 +346,18 @@ const Reception = {
...
+        picking_detail_options_for_select_move: function () {
             return {
                 show_title: true,
                 showActions: false,
                 list_item_options: {
                     loud_title: true,
                     title_action_field: {
-                        action_val_path: "name",
+                        action_val_path: function (record, field) {
+                            return record.barcode
```
mt-software-de: The record is move this doesn't contain a barcode barcode. It the product of the move should be checked instead. See https://github.com/OCA/wms/blob/867381a1a487a6e87a32179d81f233188a50d10a/shopfloor/actions/data.py#L243 https://github.com/OCA/wms/blob/867381a1a487a6e87a32179d81f233188a50d10a/shopfloor/actions/data.py#L304
```suggestion
                            return record.product.barcode
```

mt-software-de (changes requested):
Thx.

shopfloor_reception_mobile/static/src/scenario/reception.js
rousseldenis: Of course

mt-software-de (approved):
LG

jbaudoux (approved):

PR #1022, opened 2025-05-09
[16.0][ADD] sale_stock_release_channel_delivery_date: add delivery date computation

jbaudoux:
* Add a register property to allow modules to add generator to compute the earliest delivery date
* Add datetime utils to localize and naive a datetime

Generators defined in:
* https://github.com/OCA/wms/pull/1026
* https://github.com/OCA/wms/pull/1027
* https://github.com/OCA/wms/pull/1028
* https://github.com/OCA/wms/pull/1029
* https://github.com/OCA/wms/pull/1030
* https://github.com/OCA/wms/pull/1032

Depends on:
* for partner domain to support geoengine : https://github.com/OCA/wms/pull/1035
* https://github.com/OCA/wms/pull/1036

@mmequignon @grindtildeath


stock_release_channel/decorators.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+delivery_date_generators = {}
+
+
+def delivery_date_generator(step=None):
+    def decorator(func):
+        delivery_date_generators.setdefault(step, []).append(func)
```
jbaudoux: TODO: add a check on class name

sale_stock_release_channel_delivery_date/models/stock_release_channel.py
```diff
@@ -0,0 +1,52 @@
...
+                    if best_dt != previous_dt:
+                        best_generators = [gen]
+                    else:
+                        best_generators.append(gen)
+            for gen in generators:
+                gen.close()
+        return best_dt
+
+    def _is_valid_for_partner(self, partner):
+        # FIXME: add support for geoengine
```
jbaudoux: todo...

sale_stock_release_channel_delivery_date/models/sale_order.py
```diff
@@ -0,0 +1,83 @@
...
+            else:
+                order_dt = fields.Datetime.now()
+            order.expected_date = order._get_release_channel_expected_date(order_dt)
+        return res
+
+    def _get_release_channel_expected_date(self, order_dt):
+        self.ensure_one()
+        carrier = self.carrier_id
+        if not carrier:
+            # FIXME: collect from default applied on confirm
```
jbaudoux: todo...

stock_release_channel/decorators.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+delivery_date_generators = {}
```
lmignon: Your registry doesn't support multi tenants installation.  When you collect the decorated methods, you should keep the module name where it's used and when you request the generators you must filter the result based on the module installed into the current database.

lmignon (commented):
The approach is very interesting and the code is still understandable. I'm curious to see the performance impact. That's the biggest risk I see at this stage.

sale_stock_release_channel_delivery_date/models/stock_release_channel.py
jbaudoux: depends on https://github.com/OCA/wms/pull/1035

stock_release_channel/decorators.py
jbaudoux: Thanks for the feedback. I'll check if the decorator is called when the module in not installed.

in your opinion, is it a good usage of decorator or you find that overkill ? I could also just have a method in the base module returning the `delivery_date_generators` and extend that method in each module to add the module generator in the `delivery_date_generators`

jbaudoux: @lmignon I made some tests and it is working fine. When the module is not installed (i.e. activated in the DB), the decorated generator is not executed. When it is installed, it is valid for all companies inside the DB which is normal as module installation is not per company. So, if you have multiple DBs, only the db for which it is installed will be concerned

lmignon: > @lmignon I made some tests and it is working fine. When the module is not installed (i.e. activated in the DB), the decorated generator is not executed. When it is installed, it is valid for all companies inside the DB which is normal as module installation is not per company. So, if you have multiple DBs, only the db for which it is installed will be concerned

I've some doubt. It's the case while the same process has not been used to process 2 requests from 2 different databases with a different list of decorators for each database. Since your cache is set on the python module, it's initialized once the module where it's defined is loaded by the interpreter. The same cache will be therefore used by all the python modules loaded by the interpreter IOW all the addon loaded for the different databases processed by the interpreter. That's why the odoo's registry, the component's registry etc are all a dictionary by database....

sale_stock_release_channel_delivery_date/models/sale_order.py
jbaudoux: added fallback on partner carrier

stock_release_channel/decorators.py
jbaudoux: I replaced the decorator by 2 property methods with one meant to be extended in each module to register a generator

simahawk (approved):
Interesting approach. LGTM.
Let's see how it works out in the wild.

lmignon (approved):
LGTM.... let's see what the trial by fire will reveal

PR #1023, opened 2025-05-12
[IMP] shopfloor: Don't split operations at user assignation

rousseldenis:
In some flows, we don't want to split operations per user but only assign the picking and set it to printed

### DIFF
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -144,9 +144,11 @@ class StockAction(Component):
         if split:
             move_lines._extract_in_split_order(values_assigned)
         else:
-            move_lines.picking_id.filtered(
+            lines = move_lines.picking_id.filtered(
                 lambda picking: not picking.printed or not picking.user_id == user
-            ).write(values_assigned)
+            )
+            if lines:
+                lines.write(values_assigned)
         move_lines.picking_id.filtered(lambda p: p.user_id != user).user_id = user.id
 
     def unmark_move_line_as_picked(self, move_lines):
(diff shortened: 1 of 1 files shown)


shopfloor/actions/stock.py
```diff
@@ -121,21 +127,24 @@ def mark_move_line_as_picked(
...
-        move_lines._extract_in_split_order(
-            {
-                "user_id": user.id,
-                "printed": True,
-            }
-        )
+        if split:
+            move_lines._extract_in_split_order(values_assigned)
+        else:
+            move_lines.picking_id.write(values_assigned)
```
jbaudoux: As you don't create a split order, you will have multiple users working on the same picking.
If there is already a user and marked as printed, shouldn't we skip this write ?

rousseldenis: Yes, I thought about this case but not yet decided :-)

shopfloor/actions/stock.py
```diff
@@ -121,21 +127,26 @@ def mark_move_line_as_picked(
...
-                "user_id": user.id,
-                "printed": True,
-            }
-        )
+        if split:
+            move_lines._extract_in_split_order(values_assigned)
+        else:
+            move_lines.picking_id.filtered(
+                lambda picking: not picking.printed or not picking.user_id == user
+            ).write(values_assigned)
```
lmignon: @rousseldenis +You must avoid the call to `write` on an empty recordset. Even if your recordset is empty it will cost time and ressource.

lmignon (changes requested):

rousseldenis:
@jbaudoux @lmignon

lmignon (approved):

jbaudoux (approved):

PR #1024, opened 2025-05-13
[IMP] shopfloor_reception: Allows to set picking as started

rousseldenis:
Depends on:

- #1023

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -290,7 +290,8 @@ class Reception(Component):
     def _scan_line__assign_user(self, picking, line, qty_done):
         product = line.product_id
         self._assign_user_to_line(line)
-        line.qty_done += qty_done
+        stock = self._actions_for("stock")
+        stock.mark_move_line_as_picked(line, quantity=qty_done, split=False)
         if product.tracking not in ("lot", "serial") or (line.lot_id or line.lot_name):
             return self._before_state__set_quantity(picking, line)
         return self._response_for_set_lot(picking, line)
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -289,7 +289,6 @@ class Reception(Component):
 
     def _scan_line__assign_user(self, picking, line, qty_done):
         product = line.product_id
-        self._assign_user_to_line(line)
         stock = self._actions_for("stock")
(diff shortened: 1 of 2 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -277,7 +277,8 @@ def _scan_line__find_or_create_line(self, picking, move, qty_done=1):
     def _scan_line__assign_user(self, picking, line, qty_done):
         product = line.product_id
         self._assign_user_to_line(line)
```
jbaudoux: This line is now useless, isn't it ?

and the method `_assign_user_to_line` as well

jbaudoux (changes requested):

shopfloor_reception/services/reception.py
rousseldenis: But, do you use the line assignation ?

jbaudoux: `mark_move_line_as_picked` is already doing it

shopfloor_reception/services/reception.py
```diff
@@ -277,7 +277,8 @@ def _scan_line__find_or_create_line(self, picking, move, qty_done=1):
...
-        line.qty_done += qty_done
+        stock = self._actions_for("stock")
+        stock.mark_move_line_as_picked(line, quantity=qty_done, split=False)
```
lmignon, line 293: @rousseldenis with your change in #1023; every time a line is processed, an SQL update on the picking will be done since you always write the `printed` and `user_id` fields on the picking (even if the values are the same)

shopfloor_reception/services/reception.py
rousseldenis: @jbaudoux The think is this is setting `shopfloor_user_id` on lines and `mark_move_line_as_picked` do it with `user_id` on picking level.

Maybe something to think more in depth ?

jbaudoux: It's doing both https://github.com/OCA/wms/blob/88f051c26478f0768b8f6ce03e2c4f61a79bfede/shopfloor/actions/stock.py#L131

rousseldenis: @jbaudoux Solved !

rousseldenis:
@lmignon @jbaudoux Already tested by @phschmidt

lmignon (approved):

jbaudoux (approved):

PR #1025, opened 2025-05-13
[16.0][FIX] stock_release_channel_partner_delivery_window: Fix channel assignment computation

jbaudoux:
Consider only the weekday when checking if the picking can be delivered on the channel shipment date

Depends on:
* https://github.com/OCA/stock-logistics-workflow/pull/1940

@mmequignon @santostelmo

### DIFF
--- a/stock_release_channel_partner_delivery_window/models/stock_release_channel.py
+++ b/stock_release_channel_partner_delivery_window/models/stock_release_channel.py
@@ -1,3 +1,4 @@
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import fields, models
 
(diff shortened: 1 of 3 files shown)


stock_release_channel_partner_delivery_window/models/stock_release_channel.py
```diff
@@ -17,26 +18,23 @@ class StockReleaseChannel(models.Model):
...
-                channel.respect_partner_delivery_time_windows
-                and not picking.sale_id.commitment_date
-                and not partner.is_in_delivery_window(shipment_datetime)
-            ):
-                channels -= channel
-        return channels
+            )
+        return self.filtered(
+            lambda channel: not channel.respect_partner_delivery_time_windows
+            or (channel.shipment_date and channel.shipment_date.weekday() in open_days)
```
mmequignon: Are we sure about this? I mean… It looks like to me that shipment date is compared with partner's open days.
What if shipped on friday when delivery lead is 1 day?

jbaudoux: the shipment date is the date of arrival at customer :)
it's the preparation date + shipment lead time

mmequignon: then the naming is probably wrong, right? I would rename it reception_date in a future version, this is confusing.

stock_release_channel_partner_delivery_window/models/stock_release_channel.py
```diff
@@ -17,26 +18,23 @@ class StockReleaseChannel(models.Model):
...
+            return self
+        if partner.delivery_time_preference == "workdays":
+            open_days = set(range(5))
+        else:
+            open_days = set(
+                map(
+                    int,
+                    partner.delivery_time_window_ids.time_window_weekday_ids.mapped(
+                        "name"
+                    ),
```
mmequignon: I guess we would benefit from a method on the mixin to return open days.
```suggestion
        if partner.delivery_time_preference == "anytime":
            return self
        open_days = partner._get_open_days()
```
This would be kinda useful, as I suppose open days might vary depending on the country maybe.

mmequignon (approved):

stock_release_channel_partner_delivery_window/models/stock_release_channel.py
jbaudoux: Yes, this field needs renaming. I drop a note in the roadmap here https://github.com/OCA/wms/pull/1028/files#diff-70db0b8690d98e78f69865f274299cd913c8337b5bb3196ab9b4fd5fd49d297aR1

stock_release_channel_partner_delivery_window/models/stock_release_channel.py
jbaudoux: good idea
https://github.com/OCA/stock-logistics-workflow/pull/1940

PR #1029, opened 2025-05-13
[16.0][ADD] stock_release_channel_warehouse_calendar

jbaudoux:
Add generator for returning next valid delivery date

Depends on:
* https://github.com/OCA/wms/pull/1022

cc @mmequignon @grindtildeath


stock_release_channel_warehouse_calendar/models/stock_release_channel.py
```diff
@@ -0,0 +1,41 @@
...
+                delivery_date, delivery_date + batch_delta, tz=wh_tz
+            )[False]
+            for begin_dt_tz, end_dt_tz, _attendance in work_intervals:
+                while delivery_date <= end_dt_tz:
+                    if delivery_date < begin_dt_tz:
+                        delivery_date = begin_dt_tz
+                    delivery_date = yield delivery_date.astimezone(pytz.utc).replace(
+                        tzinfo=None
+                    )
+                    delivery_date = delivery_date.astimezone(pytz.utc)
```
mmequignon, line 48: I'm sorry, but I don't get why this method would expect input from `generator.send()`.
Also, I feel like wh.calendar_id is meant to compute expedition date, not delivery date.

jbaudoux: see https://github.com/OCA/wms/pull/1022/files#diff-dc4151c115a6c8830e650f7fba7647db1e88cfa2ed9588b736a3f67c24a9785cR19

~~i'll update the naming, you're right it's not helping in the understanding~~
**edit** there is already a docstring explaning it's the preparation date

sebalix (approved):

PR #1031, opened 2025-05-14
[14.0][IMP] stock_available_to_promise_release: Do not copy picking responsible to cancel picking

raumschmiede-joshuaL:
When the cancel picking is scanned by someone who isn't the responsible, an error occurs. The responsible needs to be removed manually to make it scanable. Remove the responsible after creating the picking so that anyone can scan it without error.

I'm not sure about the test changes and the commit message. If you know a better one, please suggest it.

@mt-software-de 
@jbaudoux

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -1,6 +1,7 @@
 # Copyright 2019 Camptocamp (https://www.camptocamp.com)
 # Copyright 2020 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# Copyright 2025 Raumschmiede GmbH
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 import itertools
 import logging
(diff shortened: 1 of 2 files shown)


stock_available_to_promise_release/models/stock_move.py
```diff
@@ -657,7 +658,12 @@ def _return_quantity_in_stock(self, qty_to_return_per_move):
             if product_return_moves:
                 wiz_values["product_return_moves"] = product_return_moves
                 return_wiz = self.env["stock.return.picking"].create(wiz_values)
-                return_wiz.create_returns()
+                action = return_wiz.create_returns()
+
+                cancel_picking = self.picking_id.browse(action["res_id"])
+                # Do not copy the responsible user from the source picking as somebody
+                # else could scan the new cancel picking
+                cancel_picking.user_id = False
```
jbaudoux, line 666:
```suggestion
                cancel_picking.write({"user_id": False, "printed": False"})
```

jbaudoux (commented):
Should not be marked as started/printed either

mt-software-de (commented):
With @jbaudoux suggestion it is fine.

stock_available_to_promise_release/models/stock_move.py
raumschmiede-joshuaL: I set printed in the test and executed it: 
image

It raised the error in _unrelease_not_allowed_error. So seems that if the pack picking is marked as printed the stock move is marked as unreleasable and cannot be canceled.  
Maybe this error is just related to the test setup, I will still add your suggestion. I just can't test it. Is this ok?

jbaudoux: Indeed, if it is printed, it cannot be canceled.
So all fine then :)

jbaudoux (approved):

mt-software-de (approved):

jbaudoux:
@raumschmiede-joshuaL can you also port to 16.0 ?

PR #1035, opened 2025-05-16
[16.0][IMP] stock_release_channel* : cleanup candidate domains

jbaudoux:
* Simplify the split of domain methods
* Clarify the usage and naming
* Move partner domain to partner model to allow calling from any object having a partner
  * issue https://github.com/OCA/wms/pull/1022#discussion_r2093133107 now fixed here https://github.com/OCA/wms/pull/1022/files#diff-08172cc6cf920de12bb3f1bd5e83e40e675c91c2bc520fd0e880552a6ba26a35R80
* Replace "not in" by "in"
* all modules now adapt the proper domain and don't do anymore post filtering
* moved sale dependency from `stock_release_channel_delivery_window` to `sale_stock_release_channel` as that condition on SO commitment date was also valid for public holidays
  * https://github.com/OCA/wms/pull/1036

cc @sebalix @lmignon @rousseldenis @santostelmo 

note: to merge as major


stock_release_channel/models/stock_picking.py
```diff
@@ -103,27 +104,43 @@ def _find_release_channel_possible_candidate(self):
...
 
         The condition is defined by the channel.
         """
-        return [
+        self.ensure_one()
+        domain_base = self._release_channel_possible_candidate_domain_base
+        domain = [
             ("is_manual_assignment", "=", False),
-            ("state", "!=", "asleep"),
+            ("state", "in", ("open", "locked")),
```
lmignon, line 118: :heart: This change will fix potential unexpected behaviour when the addon `stock_release_channel_shipment_advice_deliver` is installed since it adds additional state :thinking:

jbaudoux: yes I could drop this https://github.com/OCA/wms/pull/1035/files#diff-97ec0dc3db17a33c59fdc5a17ebc4bb499777a78fa7eaf19168e9cf2c25c2ea8L13

lmignon: Indeed,  I can see it in the following changes

lmignon (commented):
The code is cleaner. THank you @jbaudoux! Reading it, I don't see any risk of error or change in behavior. To be tested on our project...

stock_release_channel_geoengine/models/res_partner.py
```diff
@@ -31,3 +32,17 @@ def _compute_located_in_stock_release_channel_ids(self):
                     [("delivery_zone", "geo_intersect", rec.geo_point)]
                 )
             )
+
+    @property
+    def _release_channel_possible_candidate_domain(self):
+        self.ensure_one()
```
mt-software-de: `self.ensure_one()` will be called through super. 
```suggestion
```

stock_release_channel_geoengine/models/stock_picking.py
mt-software-de, line 1: Shouldn't this file be removed?

mt-software-de (changes requested):
This looks great. Thanks.
Just minor remarks. :-)

jbaudoux:
Thanks for the reviews. It's not yet entirely correct. I'm fixing tests and adding more tests that shows what's wrong with partner holidays and window. Those issues are in fact already there

stock_release_channel_geoengine/models/res_partner.py
jbaudoux: removed

stock_release_channel_geoengine/models/stock_picking.py
jbaudoux: idd

jbaudoux:
> Thanks for the reviews. It's not yet entirely correct. I'm fixing tests and adding more tests that shows what's wrong with partner holidays and window. Those issues are in fact already there

fixed

And:
* all modules now adapt the proper domain and don't do anymore post filtering
* moved sale dependency from `stock_release_channel_delivery_window` to `sale_stock_release_channel` as that condition was also valid for public holidays

jbaudoux:
@mt-software-de can you update your review? Then i'll squash the fixups

sebalix (approved):

jbaudoux:
fixup squashed

stock_release_channel/models/stock_picking.py
```diff
@@ -103,27 +104,43 @@ def _find_release_channel_possible_candidate(self):
...
         ]
+        domain_partner = (
+            self.partner_id._release_channel_possible_candidate_domain
+            if self.partner_id
+            else []
+        )
+        domain_extras = []
+        if self._release_channel_possible_candidate_domain_apply_extras:
+            domain_extras = self._release_channel_possible_candidate_domain_extras
+        domain = expression.AND([domain, domain_base, domain_partner] + domain_extras)
```
mt-software-de, line 131: Could this be simplified as well?
Is the property `_release_channel_possible_candidate_domain_apply_extras` really needed. 
If you don't want to apply the extra domain, (what isn't the case right now in any depending addon) you could just overwrite `_release_channel_possible_candidate_domain_extras` and return `[]`.

```suggestion
        domain_extras = self._release_channel_possible_candidate_domain_extras
        domain = expression.AND([domain, domain_base, domain_partner] + domain_extras)
```

mt-software-de (approved):
LGTM. But IMO `_release_channel_possible_candidate_domain_apply_extras` could be removed. But this isn't a blocker. Thx

stock_release_channel/models/stock_picking.py
jbaudoux: There are modules extending the domains like delivery window and public holidays for checking the date is a valid customer date
And there is the sale module that bypasses those domains when the commitment date is set https://github.com/OCA/wms/pull/1036/files
As those modules do not depend on each other, I can't have one setting the domain to empty as it could be before the other modules extending the domain

mt-software-de: Thx. I didnt checked your other PRs

PR #1036, opened 2025-05-18
[ADD][16.0] sale_stock_release_channel : candidate domain for sales order + commitment_date

jbaudoux:
* Exclude extra domains if SO commitment date
* Add base candidate domain on SO mimicking the same domain on stock.picking


sale_stock_release_channel/models/sale_order.py
```diff
@@ -0,0 +1,24 @@
...
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    @property
+    def _release_channel_possible_candidate_domain_base(self):
```
simahawk, line 12: Being a property, possible limitation: if tomorrow you need to pass an argument around, you'll have to change the api.

simahawk (approved):
LG overall

sale_stock_release_channel/models/sale_order.py
jbaudoux: I did it on purpose to forbid it !

PR #1045, opened 2025-05-28
[18.0] [MIG] stock_available_to_promise_release: Migration to 18.0

bizzappdev:


stock_available_to_promise_release/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2019 Camptocamp (https://www.camptocamp.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+
+{
+    "name": "Stock Available to Promise Release",
+    "version": "16.0.3.7.1",
```
mmequignon: Please fix version number

mmequignon (changes requested):
You'll need this module as a dependency for this to work I guess `stock_warehouse_out_pull`.
With that kind of setup in unit tests.
```python
        cls.warehouse = cls.env.ref("stock.warehouse0")
        cls.warehouse.delivery_pull = True
```

stock_available_to_promise_release/__manifest__.py
bizzappdev: @mmequignon 
Thank you for the review. It's a bit early, as the module is still a work in progress.
I'll ping you once the PR is ready for a proper review.

bizzappdev:
It's ready for review. The only open issue is the return picking test case, which we are working on fixing.

bizzappdev:
> It's ready for review. The only open issue is the return picking test case, which we are working on fixing.

The issue with the return location has been solved.

stock_available_to_promise_release/hooks.py
```diff
@@ -5,17 +5,17 @@
...
 
-def init_release_policy(cr):
-    if not sql.column_exists(cr, "stock_picking", "release_policy"):
+def init_release_policy(env):
+    if not column_exists(env.cr, "stock_picking", "release_policy"):
         # Use the default sql query instead relying on ORM as all records will
         # be updated.
         _logger.info("Creating 'release_policy' field on stock.picking")
-        env = api.Environment(cr, SUPERUSER_ID, {})
+        env = api.Environment(env.cr, SUPERUSER_ID, {})
```
simahawk: there's already an `env` , no need to init another one w/ the same cr

stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -134,12 +134,7 @@ def release_available_to_promise(self):
     def _release_link_backorder(self, origin_picking):
         self.backorder_id = origin_picking
         origin_picking.message_post(
-            body=_(
-                "The backorder <a href=# data-oe-model=stock.picking"
-                " data-oe-id=%(id)s>%(name)s</a> has been created.",
-                name=self.name,
-                id=self.id,
-            )
+            body=_("The backorder %s has been created.", self._get_html_link())
```
simahawk: `self.env._(` for all translations

simahawk (commented):
LG overall.  Only code review, no tests. Just some minor remarks.

stock_available_to_promise_release/models/stock_picking.py
bizzappdev: @simahawk Both the changes are done.

stock_available_to_promise_release/models/stock_location.py
```diff
@@ -10,10 +10,8 @@ class StockLocation(models.Model):
...
         for location in self:
-            location_domain = expression.OR(
-                [
-                    location_domain,
-                    [("location_id.parent_path", "=like", location.parent_path + "%")],
-                ]
+            paths_domain = expression.OR(
+                [[("parent_path", "=like", location.parent_path + "%")]]
             )
+            location_domain = [("location_id", "any", paths_domain)]
```
ivantodorovich: This change doesn't seem correct.

Before, location_domain was the combined domain (`OR`) for all locations in `self`.

Now, `expression.OR` is used to combine a single leaf (useless), and it's ignoring the locations in the previous iterations (❌ wrong)

I do like that you're using the new `any` operator 💚 

I think you want to do something like this here:

```py
        if not self:
            return expression.FALSE_DOMAIN
        return [
            (
                "location_id",
                "any",
                [
                    ("parent_path", "=like", location.parent_path + "%")
                    for location in self
                ]
            )
        ]
```

Although I don't really understand the use of `parent_path` directly here (prior code). Why not using `child_of` and simply do this?

```py
    return [("location_id", "child_of", self.ids)]
```

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -463,7 +464,7 @@ def _search_ordered_available_to_promise_uom_qty(self, operator, value):
             "!=": py_operator.ne,
         }
         if operator not in operator_mapping:
-            raise UserError(_("Unsupported operator %s") % (operator,))
+            raise UserError(self.env._("Unsupported operator %s") % (operator,))
```
ivantodorovich:
```suggestion
            raise UserError(self.env._("Unsupported operator %s", operator))
```

stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -53,7 +53,9 @@ def _compute_need_release(self):
 
     def _search_need_release(self, operator, value):
         if operator not in ("=", "!="):
-            raise exceptions.UserError(_("Unsupported operator: %s") % (operator,))
+            raise exceptions.UserError(
+                self.env._("Unsupported operator: %s") % (operator,)
```
ivantodorovich:
```suggestion
                self.env._("Unsupported operator: %s", operator)
```

stock_available_to_promise_release/models/stock_picking.py
```diff
@@ -97,7 +99,9 @@ def _compute_release_ready(self):
 
     def _search_release_ready(self, operator, value):
         if operator != "=":
-            raise exceptions.UserError(_("Unsupported operator %s") % (operator,))
+            raise exceptions.UserError(
+                self.env._("Unsupported operator %s") % (operator,)
```
ivantodorovich:
```suggestion
                self.env._("Unsupported operator %s", operator)
```

stock_available_to_promise_release/models/product_product.py
```diff
@@ -0,0 +1,32 @@
...
+    move_need_release_count = fields.Integer(
+        string="Moves Need Release", compute="_compute_move_need_release_count"
+    )
+
+    @api.depends("stock_move_ids.need_release")
+    def _compute_move_need_release_count(self):
+        for product in self:
+            product.move_need_release_count = len(
+                product.stock_move_ids.filtered("need_release")
+            )
```
ivantodorovich: This would scale badly over time, with thousands of stock.moves per product

```suggestion
        data = self.env["stock.move"]._read_group(
            domain=[('product_id', 'in', self.ids), ('need_release', '=', True)],
            fields=['product_id'],
            groupby=['product_id'],
        )
        counts = {
            item['product_id'][0]: item['product_id_count']
            for item in data
        }
        for product in self:
            product.move_need_release_count = counts.get(product.id, 0)
```

stock_available_to_promise_release/models/res_company.py
```diff
@@ -0,0 +1,22 @@
...
+
+    stock_reservation_horizon = fields.Integer(
+        default=0,
+        help="Compute promised quantities for order planned to be shipped "
+        "until this number of days from today.",
+    )
+    stock_release_max_prep_time = fields.Integer(
+        string="Transfer Releases Max Prep. Time",
+        default=180.0,
+        required=True,
```
ivantodorovich: Required doesn't really work well with integer fields (`0` is also a value, thus it's considered set)

If 0 really isn't allowed, then we need a constraint for this

ivantodorovich (changes requested):

stock_available_to_promise_release/models/stock_location.py
bizzappdev: The reason for using `parent_path` directly instead of `child_of` is performance-related. 
When dealing with a large location tree, `child_of` can be slower since it requires ORM processing. Using `parent_path` with a `=like` query leverages SQL indexing and improves performance. 
You can see a similar approach in Odoo's code [here](https://github.com/odoo/odoo/blob/18.0/addons/stock/models/product.py#L320).

However, per your suggestion, we have applied the domain change regarding list comprehension.

ivantodorovich: Uhm, those lines of code in Odoo are super old (10yo)
That optimization is no longer needed, the ORM is smart enough to do it in the same way now. 
See: https://github.com/odoo/odoo/blob/90f235fceff96f4dd060b142a1073ce37bd0f7a2/odoo/osv/expression.py#L889-L893

IMHO you can just use `child_of` for simplicity.

If you don't, note that I've just realized I made a mistake in my list comprehension suggestion. I missed the `expression.OR` for the right operand

bizzappdev: ahh, indeed. Odoo ORM is also using the same now. Perfect. We have changed as per your suggestions.

ivantodorovich (approved):
Thanks!

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -136,7 +137,7 @@ def _is_unrelease_allowed_on_origin_moves(self, origin_moves):
         dest_done_moves = origin_done_moves.move_dest_ids
         dest_qty_done = sum(
             m.product_uom._compute_quantity(
-                m.quantity_done,
+                sum(ml.quantity for ml in dest_done_moves if ml.picked),
```
mmequignon: are we sure about that? You're already iterating over dest_done_moves

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -716,9 +717,15 @@ def _return_quantity_in_stock(self, qty_to_return_per_move):
...
                 wiz_values["product_return_moves"] = product_return_moves
-                return_wiz = self.env["stock.return.picking"].create(wiz_values)
-                action = return_wiz.create_returns()
+                return_wiz = (
+                    self.env["stock.return.picking"]
+                    .with_context(
+                        unrelease_return_type_loc_dest_id=return_type.default_location_dest_id.id
+                    )
+                    .create(wiz_values)
+                )
```
mmequignon: Not blocking, just cosmetic:
I don't like those 7 lines one-liners.

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -899,7 +907,9 @@ def _search_picking_for_assignation_domain(self):
 
     def _get_new_picking_values(self):
         values = super()._get_new_picking_values()
-        values["release_policy"] = values["move_type"]
+        # Ensure move_type exists in values before accessing it
+        move_type = self.mapped("group_id").move_type or "direct"
```
mmequignon, line 910: Is it me or you aren't checking move_type in values?

stock_available_to_promise_release/tests/common.py
```diff
@@ -141,9 +142,13 @@ def _deliver(cls, picking, product_qty=None):
         if product_qty:
             lines = picking.move_ids.move_line_ids
             for product, qty in product_qty:
-                line = lines.filtered(lambda m: m.product_id == product)
-                line.qty_done = qty
+                line = lines.filtered(
+                    lambda m, product=product: m.product_id == product
+                )
```
mmequignon, line 147: I'm not sure I understand this change. product is already defined, right?

mmequignon (commented):
Looks very good, just a few questions and remarks

stock_available_to_promise_release/models/stock_move.py
bizzappdev: In v16, `move_type` was returned from _get_new_picking_values, but in v18, it's no longer present. So, to ensure the correct value is used, we retrieve it directly using self.mapped("group_id").move_type

stock_available_to_promise_release/tests/common.py
bizzappdev: It is a fix for pre-commit: Ruff B023 warning

stock_available_to_promise_release/models/stock_move.py
mmequignon: Then maybe the comment is misleading

sebalix:
Should we not move this module to `OCA/stock-logistics-workflow`? cc @jbaudoux

stock_available_to_promise_release/tests/common.py
mmequignon: Ok, I didn't know

jbaudoux:
> Should we not move this module to `OCA/stock-logistics-workflow`? cc @jbaudoux

should be moved to https://github.com/OCA/stock-logistics-reservation according to repo README

jbaudoux:
This fix has been merged in v14.0 https://github.com/OCA/wms/pull/1053

stock_available_to_promise_release/models/stock_move.py
bizzappdev: Thank you for noticing. It was indeed meant for move_line_ids.

mt-software-de:
> This fix has been merged in v14.0 #1053

also in v16 https://github.com/OCA/wms/pull/1056

bizzappdev:
Closing this as module is moved to the repo `stock-logistics-reservation`: https://github.com/OCA/stock-logistics-reservation/pull/10.

PR #1049, opened 2025-06-05
[FIX] stock_storage_type: Allow to have two rules for a storage type

rousseldenis:
If there are several rules for a storage type (e.g.: in Pallets, you have Pallets/Pallets UK), don't bypass the next rules after the first condition evaluation.


rousseldenis:
@sebalix @jbaudoux

jbaudoux (approved):
good catch

will have to fwd port this to v18 @sebalix

jbaudoux:
> good catch
> 
> will have to fwd port this to v18 @sebalix

done here https://github.com/OCA/stock-logistics-putaway/pull/7

PR #1050, opened 2025-06-06
[16][IMP] shopfloor_mobile_base: persist profile

hparfr:
Align with language managment: profile selection is persisted.

Without this, on each login user is asked to choose a profile.

### DIFF
--- a/shopfloor_mobile_base/static/wms/src/main.js
+++ b/shopfloor_mobile_base/static/wms/src/main.js
@@ -51,7 +51,13 @@ const register_app_components = function (components) {
 register_app_components(process_registry.all());
 register_app_components(page_registry.all());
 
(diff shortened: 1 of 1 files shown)


simahawk:
> Without this, on each login user is asked to choose a profile.

Is this because you want to link a device to a specific profile?

hparfr:
The devices in the reception aera are shared between receptionnists.
The devices in the picking aera are shared between pickers.

So each device is tied to one profile because the device  is always used in the same "service" (= physical space).

And also, If you have just one profile defined in the application it doesn't make sense to ask for it. This implementation is a quick fix for this use case.

giarve:
oh lol I just came with a similar solution for the single profile

https://github.com/OCA/wms/pull/1078

PR #1053, opened 2025-06-12
[14.0][FIX] stock_available_to_promise_release: Assign transfers after unrelease return

mt-software-de:
On a environment with a fixed procurement pick rule.

If two delivery having the same pick move as origin and one of the deliveries were canceled,
then also the none cancel delivery changed the state from assigned to confirmed.
This is because of https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/stock/wizard/stock_picking_return.py#L116. 

Now after returning the already done moves it calls action_assign on the destination pickings. 
This ensures the the pickings keeping the state `assigned`.

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -686,6 +686,14 @@ class StockMove(models.Model):
                 # Do not copy the responsible user from the source picking as somebody
                 # else could scan the new cancel picking
                 cancel_picking.user_id = False
+
+                returned_moves = return_wiz.product_return_moves.move_id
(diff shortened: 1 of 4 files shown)


mt-software-de:
cc @jbaudoux @mmequignon @i-vyshnevska @raumschmiede-joshuaL

jbaudoux (approved):

jbaudoux:
@mt-software-de do you fwd port to 16.0 ?

mt-software-de:
> @mt-software-de do you fwd port to 16.0 ?

https://github.com/OCA/wms/pull/1056

PR #1055, opened 2025-06-18
[16.0][FIX] shopfloor: Fetch qty_available only when needed

sbejaoui:
Several services use `_data_move_line` to prepare stock move line data for the Shopfloor frontend. However, qty_available is not displayed on all pages where stock move lines are handled.

Since qty_available is computed from quants, accessing it unnecessarily can lead to useless database queries. In high-concurrency environments, this can even trigger deadlocks when reads and writes occur simultaneously across many Shopfloor users.

This fix limits the use of qty_available to cases where it’s strictly needed.

image

@lmignon , @rousseldenis

### DIFF
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -97,7 +97,7 @@ class ClusterPicking(Component):
         self, move_line, message=None, popup=None, sublocation=None
     ):
         kw = {"sublocation": self.data.location(sublocation)} if sublocation else {}
-        data = self._data_move_line(move_line, **kw)
+        data = self._data_move_line(move_line, with_qty_available=False, **kw)
         return self._response(
             next_state="start_line",
             data=data,
@@ -107,9 +107,14 @@ class ClusterPicking(Component):
 
     def _response_for_scan_destination(self, move_line, message=None, qty_done=None):
         if qty_done is None:
-            data = self._data_move_line(move_line)
+            data = self._data_move_line(
+                move_line,
+                with_qty_available=False,
+            )
         else:
-            data = self._data_move_line(move_line, qty_done=qty_done)
+            data = self._data_move_line(
+                move_line, with_qty_available=False, qty_done=qty_done
+            )
         last_picked_line = self._last_picked_line(move_line.picking_id)
         if last_picked_line:
             # suggest pack to be used for the next line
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):
Great !

lmignon (approved):
Do you've analyzed the domain used in such a case. The qty available is computed for a specific location. I find it hard to believe that there are several people at the same time asking for the same quantity of the same product for the same location... In any case, if you don't need the field because you don't display it, it's time saved.

sbejaoui:
> Do you've analyzed the domain used in such a case. The qty available is computed for a specific location. I find it hard to believe that there are several people at the same time asking for the same quantity of the same product for the same location... In any case, if you don't need the field because you don't display it, it's time saved.

The quantity is not relevant; the issue is related to the same product from the same location, and it occurs consistently based on the data I analyzed in our use case.

simahawk (approved):
make sense

shopfloor/services/cluster_picking.py
```diff
@@ -406,9 +411,10 @@ def _data_move_line(self, line, **kw):
         data["batch"] = self.data.picking_batch(batch)
         data["picking"] = self.data.picking(picking)
         data["postponed"] = line.shopfloor_postponed
-        data["product"]["qty_available"] = product.with_context(
-            location=line.location_id.id
-        ).qty_available
+        if "no_qty_available" not in kw or not kw.get("no_qty_available"):
```
jbaudoux: why not simply this?
```suggestion
        if not kw.get("no_qty_available"):
```

rousseldenis: @sbejaoui

jbaudoux (approved):

shopfloor/services/cluster_picking.py
```diff
@@ -406,9 +411,10 @@ def _data_move_line(self, line, **kw):
...
+            data["product"]["qty_available"] = product.with_context(
+                location=line.location_id.id
+            ).qty_available
```
mmequignon: Is it possible to avoid double negations ? `not no_qty_available` is not really intuitive to read.
For instance, explicitely add a named argument `with_qty_available` which defaults to `True`.
When you don't need this information in the data, we can just set `with_qty_available=False`.
Then you can do this, which IMO is easier to read. and less bug prone.
```suggestion
        if with_qty_available:
            data["product"]["qty_available"] = product.with_context(
                location=line.location_id.id
            ).qty_available
```
What do you think ?

jbaudoux:
ping @sbejaoui

shopfloor/services/cluster_picking.py
sbejaoui: good point, I made the changes

sbejaoui:
changes made, please review

PR #718, opened 2023-08-10
[16.0][ADD] shopfloor_packing

sbejaoui:

PR #1063, opened 2025-06-24
[IMP] shopfloor: Add context key to control package total quantity computation

sbejaoui:
the total quantity per package is not relevant or displayed in all scenarios where the parser is used. currently, it's only used in the zone_picking scenario

this PR introduces a context key to conditionally disable the total quantity computation based on the use case. this avoids unnecessary access to stock.quant when the information isn't required, improving performance and preventing potential concurrency issues

example of usage: https://github.com/OCA/wms/pull/718/commits/e55352111d71ead8f00139f84218719609f844dc

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -131,7 +131,9 @@ class DataAction(Component):
             ("package_type_id:storage_type", ["id", "name"]),
             (
                 "quant_ids:total_quantity",
-                lambda rec, fname: sum(rec.quant_ids.mapped("quantity")),
+                lambda rec, fname: 0
+                if rec.env.context.get("no_quantity", False)
+                else sum(rec.quant_ids.mapped("quantity")),
             ),
         ]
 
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):

lmignon (approved):

shopfloor/actions/data.py
```diff
@@ -131,7 +131,9 @@ def _package_parser(self):
             ("package_type_id:storage_type", ["id", "name"]),
             (
                 "quant_ids:total_quantity",
-                lambda rec, fname: sum(rec.quant_ids.mapped("quantity")),
+                lambda rec, fname: 0
+                if rec.env.context.get("no_quantity", False)
+                else sum(rec.quant_ids.mapped("quantity")),
             ),
```
jbaudoux, line 137: shouldn't the tuple be removed completely instead of returning a wrong value?

jbaudoux (commented):
as it's only used when you `handle_complete_mix_pack`, can you by default not provide the data unless that option is enabled ?

cc @mt-software-de

shopfloor/actions/data.py
rousseldenis: @sbejaoui Indeed, it could be great as this could lead to wrong information. Could you check also if schema should not be adapted

jbaudoux:
@sbejaoui Can you look at the comments?

lmignon:
ping @sbejaoui

PR #1064, opened 2025-06-24
 [16.0][FWP][IMP] shopfloor: Ensure destination location on moves for allow move create

mt-software-de:
FWP of #973

cc @jbaudoux

### DIFF
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -266,11 +266,9 @@ class StockAction(Component):
             lines_without_package_level.location_dest_id = location_dest
             lines_without_package_level.move_id.location_dest_id = location_dest
 
-    def _unload_package(self, lines):
+    def unload_package(self, lines):
         lines.result_package_id = False
 
-    def set_destination_and_unload_lines(self, lines, location_dest, unload=False):
+    def set_destination_on_lines(self, lines, location_dest):
         self._lock_lines(lines)
         self._set_destination_on_lines(lines, location_dest)
-        if unload:
-            self._unload_package(lines)
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
+# Copyright 2025 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import _, fields
 from odoo.tools.float_utils import float_round
@@ -250,3 +251,26 @@ class StockAction(Component):
         # when no putaway is found, the move line destination stays the
         # default's of the picking type
         return any(line.location_dest_id in base_locations for line in move_lines)
+
+    def _lock_lines(self, lines):
+        self._actions_for("lock").for_update(lines)
+
+    def _set_destination_on_lines(self, lines, location_dest):
+        # when writing the destination on the package level, it writes
+        # on the moves and move lines
+        lines_with_package_level = lines.package_level_id.move_line_ids
+        lines_without_package_level = lines - lines_with_package_level
+        if lines_with_package_level:
+            lines_with_package_level.package_level_id.location_dest_id = location_dest
+        if lines_without_package_level:
+            lines_without_package_level.location_dest_id = location_dest
+            lines_without_package_level.move_id.location_dest_id = location_dest
+
+    def _unload_package(self, lines):
+        lines.result_package_id = False
+
+    def set_destination_and_unload_lines(self, lines, location_dest, unload=False):
+        self._lock_lines(lines)
+        self._set_destination_on_lines(lines, location_dest)
+        if unload:
+            self._unload_package(lines)
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -1147,7 +1147,7 @@ class ClusterPicking(Component):
 
     def _unload_write_destination_on_lines(self, lines, location):
         stock = self._actions_for("stock")
-        stock.set_destination_and_unload_lines(lines, location)
+        stock.set_destination_on_lines(lines, location)
         lines.write({"shopfloor_unloaded": True})
         for picking in lines.batch_id.picking_ids:
             picking_lines = lines.filtered(lambda x, p=picking: x.picking_id == p)
--- a/shopfloor/services/cluster_picking.py
+++ b/shopfloor/services/cluster_picking.py
@@ -1146,8 +1146,9 @@ class ClusterPicking(Component):
         return self._unload_end(batch, completion_info_popup=completion_info_popup)
 
     def _unload_write_destination_on_lines(self, lines, location):
-        lines.write({"shopfloor_unloaded": True, "location_dest_id": location.id})
-        lines.package_level_id.location_dest_id = location
+        stock = self._actions_for("stock")
+        stock.set_destination_and_unload_lines(lines, location)
+        lines.write({"shopfloor_unloaded": True})
         for picking in lines.batch_id.picking_ids:
             picking_lines = lines.filtered(lambda x, p=picking: x.picking_id == p)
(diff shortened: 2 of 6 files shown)


mmequignon (commented):
Hi @mt-software-de I tested this on 18.0, and there's something that I find confusing.
In 14.0 this `set_destination_and_unload_lines` is used in the `set_quantity__post_line` method, which first sets `result_package_id` on the move line if a package was scanned, then `set_destination_and_unload_lines` is called, which sets result_package_id to False.

As I understand/remember, we can move a product to a pallet, and we shouldn't force the `result_package_id` to False in such case.

Do you guys have some insight?
@jbaudoux @mt-software-de

mt-software-de:
> Hi @mt-software-de I tested this on 18.0, and there's something that I find confusing. In 14.0 this `set_destination_and_unload_lines` is used in the `set_quantity__post_line` method, which first sets `result_package_id` on the move line if a package was scanned, then `set_destination_and_unload_lines` is called, which sets result_package_id to False.
> 
> As I understand/remember, we can move a product to a pallet, and we shouldn't force the `result_package_id` to False in such case.
> 
> Do you guys have some insight? @jbaudoux @mt-software-de

I can not find the method `set_quantity__post_line` in v14

mmequignon:
> > Hi @mt-software-de I tested this on 18.0, and there's something that I find confusing. In 14.0 this `set_destination_and_unload_lines` is used in the `set_quantity__post_line` method, which first sets `result_package_id` on the move line if a package was scanned, then `set_destination_and_unload_lines` is called, which sets result_package_id to False.
> > As I understand/remember, we can move a product to a pallet, and we shouldn't force the `result_package_id` to False in such case.
> > Do you guys have some insight? @jbaudoux @mt-software-de
> 
> I can not find the method `set_quantity__post_line` in v14

My bad, this isn't post_line but post_move :pray: 
When a package is scanned, we set the result_package on the line https://github.com/OCA/wms/blob/14.0/shopfloor_single_product_transfer/services/single_product_transfer.py#L746
set_destination_and_unload_lines is then called, and reverts this.

mt-software-de:
> > > Hi @mt-software-de I tested this on 18.0, and there's something that I find confusing. In 14.0 this `set_destination_and_unload_lines` is used in the `set_quantity__post_line` method, which first sets `result_package_id` on the move line if a package was scanned, then `set_destination_and_unload_lines` is called, which sets result_package_id to False.
> > > As I understand/remember, we can move a product to a pallet, and we shouldn't force the `result_package_id` to False in such case.
> > > Do you guys have some insight? @jbaudoux @mt-software-de
> > 
> > 
> > I can not find the method `set_quantity__post_line` in v14
> 
> My bad, this isn't post_line but post_move 🙏 When a package is scanned, we set the result_package on the line https://github.com/OCA/wms/blob/14.0/shopfloor_single_product_transfer/services/single_product_transfer.py#L746 set_destination_and_unload_lines is then called, and reverts this.

In fact `_set_quantity__post_move` is calling `_write_destination_on_lines` https://github.com/OCA/wms/blob/c63f6d29e42dac0573e71ba93e5d77285e8eb054/shopfloor_single_product_transfer/services/single_product_transfer.py#L641 and then `stock.set_destination_and_unload_lines` https://github.com/OCA/wms/blob/c63f6d29e42dac0573e71ba93e5d77285e8eb054/shopfloor_single_product_transfer/services/single_product_transfer.py#L636 but it is not unloading it. Because the parameter `unload` is not passed to it. https://github.com/OCA/wms/blob/c63f6d29e42dac0573e71ba93e5d77285e8eb054/shopfloor/actions/stock.py#L250
I guess the naming of the method is a little bit misleading.

shopfloor/actions/stock.py
```diff
@@ -237,3 +238,20 @@ def no_putaway_available(self, picking_types, move_lines):
...
         # default's of the picking type
         return any(line.location_dest_id in base_locations for line in move_lines)
+
+    def _lock_lines(self, lines):
+        self._actions_for("lock").for_update(lines)
+
+    def _set_destination_on_lines(self, lines, location_dest):
+        # when writing the destination on the package level, it writes
+        # on the moves and move lines
+        lines.picking_id.location_dest_id = location_dest
```
jbaudoux: It doesn't look right to change the location of the picking and all its moves. Only the lines should be updated
```suggestion
        lines.location_dest_id = location_dest
        lines.package_level_id.location_dest_id = location_dest
```

mt-software-de: Yes you are right, i just restored as it is done in v14's PR.

shopfloor/services/cluster_picking.py
```diff
@@ -1146,8 +1146,9 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=None):
         return self._unload_end(batch, completion_info_popup=completion_info_popup)
 
     def _unload_write_destination_on_lines(self, lines, location):
-        lines.write({"shopfloor_unloaded": True, "location_dest_id": location.id})
-        lines.package_level_id.location_dest_id = location
+        stock = self._actions_for("stock")
+        stock.set_destination_and_unload_lines(lines, location)
+        lines.write({"shopfloor_unloaded": True})
```
jbaudoux, line 1151: What about moving this to _unload_package?

shopfloor/actions/stock.py
```diff
@@ -250,3 +251,26 @@ def no_putaway_available(self, picking_types, move_lines):
...
+        if lines_with_package_level:
+            lines_with_package_level.package_level_id.location_dest_id = location_dest
+        if lines_without_package_level:
+            lines_without_package_level.location_dest_id = location_dest
+            lines_without_package_level.move_id.location_dest_id = location_dest
+
+    def _unload_package(self, lines):
+        lines.result_package_id = False
+
+    def set_destination_and_unload_lines(self, lines, location_dest, unload=False):
```
jbaudoux: It's a bit strange to have the method named 'and unload' and have by default 'unloadFalse'
What about dropping this argument? If you don't want to unload, you can call 'set destination on lines'

shopfloor/services/zone_picking.py
```diff
@@ -1585,11 +1585,10 @@ def set_destination_all(self, barcode, confirmation=None):
...
 
     def _write_destination_on_lines(self, lines, location):
-        self._lock_lines(lines)
-        lines.location_dest_id = location
-        lines.package_level_id.location_dest_id = location
-        if self.work.menu.unload_package_at_destination:
-            lines.result_package_id = False
+        stock = self._actions_for("stock")
+        stock.set_destination_and_unload_lines(
+            lines, location, unload=self.work.menu.unload_package_at_destination
```
jbaudoux: Unload value could be moved to called method

shopfloor/actions/stock.py
```diff
@@ -250,3 +251,26 @@ def no_putaway_available(self, picking_types, move_lines):
...
+        self._lock_lines(lines)
+        self._set_destination_on_lines(lines, location_dest)
+        if unload:
```
jbaudoux: In case the method is in service 

If self.work.menu.unload_package_at_destination:

shopfloor/services/cluster_picking.py
mt-software-de: You mean `shopfloor_unloaded`? This is used for holding the information, that it should be unloaded. If all lines are done, the real unload is happening.

shopfloor/actions/stock.py
mt-software-de: Yes. Thats right.I created a separate commit to clean this up. No it is only locking the lines and sets the dest.

shopfloor/services/zone_picking.py
mt-software-de: Only in this scenario the unload is maybe true. So i removed it with just calling the unload method.

jbaudoux (approved):
That version is cleaner

jbaudoux:
@mmequignon Can we merge this and update the port to v18 ?

mt-software-de:
Ping @jbaudoux @mmequignon

PR #1071, opened 2025-06-30
[FIX] shopfloor_reception: don't reset expiration date after lot scan

sbejaoui:
if the user selects an expiration date and then scans a lot, the expiration date is unintentionally reset, causing data loss
this fix preserves the manually entered expiration date when scanning a lot

### DIFF
--- a/shopfloor_reception_mobile/static/src/scenario/reception_states.js
+++ b/shopfloor_reception_mobile/static/src/scenario/reception_states.js
@@ -8,6 +8,9 @@
     Define states for reception scenario.
     @param this VueJS component instance
 */
+
+import event_hub from "/shopfloor_mobile_base/static/wms/src/services/event_hub.js";
+
 export const reception_states = function () {
     return {
         init: {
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):

lmignon (changes requested):
You can't assign a contextualized name to a generic component. t
That doesn't make sense

shopfloor_mobile_base/static/wms/src/components/datepicker/date_picker.js
```diff
@@ -40,6 +40,7 @@ export var DatePicker = Vue.component("date-picker-input", {
                     v-model="date"
                     v-bind="attrs"
                     v-on="on"
+                    name="expiration_date"
```
simahawk: We shouldn't do this here and I would argue on why we have a specific label here if this is a generic widget.
If we really need this, we should use properties. However, I don't think you need this.

simahawk: Also, such changes to another module should have its own commit.

shopfloor_reception_mobile/static/src/scenario/reception_states.js
```diff
@@ -131,11 +131,14 @@ export const reception_states = function () {
             },
             on_scan: (barcode) => {
                 // Scan a lot
+                const input = document.querySelector('input[name="expiration_date"]');
```
simahawk: You should hook to the event provided by the widget instead and collect this info somewhere else (see `on_date_picker_selected`). As an alternative you can declare a `$ref` on the widget where it's used (see `searchbar` for example) .

simahawk (changes requested):

shopfloor_mobile_base/static/wms/src/components/datepicker/date_picker.js
sbejaoui: I was misled by the wording. At first glance, it seems like this widget is dedicated to the expiration date. 

I changed the name to "date_picker" to be more generic. Getting the value from input using the selector is way easier than adding hooks to the events

lmignon:
This PR is no more required. The code has been improved at service level into https://github.com/OCA/wms/pull/1072.

PR #1085, opened 2025-07-17
[16.0][ADD] shopfloor_priting_base/mobile

rousseldenis:
These modules allow to build base bricks to support printing


shopfloor_mobile_printing_base/static/src/i18n/i18n.de.js
```diff
@@ -0,0 +1,113 @@
+/**
+ * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
```
simahawk: leftover :)

shopfloor_mobile_printing_base/static/src/components/printing-widget.js
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
+ * @author Simone Orsi <simahawk@gmail.com>
+ * Copyright 2025 ACSONE SA/NV (https://acsone.eu)
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+ */
+
+export var LabelPrinterWidget = Vue.component("label-printer", {
```
simahawk, line 8: why not improving the `input-number-spinner` widget?

shopfloor_printing_base/components/base.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.addons.component.components.base import AbstractComponent
+
+
+class ShopFloorProcessAction(AbstractComponent):
+    """Base Component for actions"""
+
+    _inherit = "base.shopfloor.service"
+
+    def _printing_for(self, usage):
```
simahawk: I don't understand what is it used for

shopfloor_printing_base/components/printing.py
```diff
@@ -0,0 +1,29 @@
...
+    _name = "shopfloor.printing.action"
+    _inherit = "shopfloor.process.action"
+    _collection = "shopfloor.printing"
+
+    @property
+    def report_to_print(self):
+        """
+        Returns the report to print
+        """
+        return self.work.menu.label_print_report_id
```
simahawk: IMO instead of relying on a fixed report per menu item, we should pass this id into the menu data for the frontend and then the frontend will pas the id of the report to print. This way, the endpoint will work with any report.
Then I don't think we need this endpoint at all since we could use the standard endpoint for printing reports or a similar one.
Just ensure that you can pass the config to the menu or to the scenario.

simahawk (changes requested):

shopfloor_mobile_printing_base/static/src/components/printing-widget.js
rousseldenis: I hesitated

shopfloor_printing_base/components/base.py
rousseldenis: This is needed to get the printing component on flows ones (reception, ...). This allows to reuse a certain logic and also to get specific ones when needed

shopfloor_printing_base/components/printing.py
rousseldenis: Could you elaborate ? What do you mean by `into the menu data` ?

nicolas-delbovier-acsone:
I refactored the `allow_print_label` logic to ensure correct data nesting.

Previously, the variable was not placed correctly in the response. Because the Vue frontend maps its internal state based on a specific data nesting, the variable wasn't being committed to the Session Storage, causing it to be lost on refresh.

shopfloor_printing_base/services/service.py
```diff
@@ -0,0 +1,25 @@
...
+class BaseShopfloorService(AbstractComponent):
+    """Base class for REST services"""
+
+    _inherit = "base.shopfloor.service"
+
+    def _response(
+        self, base_response=None, data=None, next_state=None, message=None, popup=None
+    ):
+        data = data or {}
+        data["allow_print_label"] = self._menu.display_print_label_button
```
nicolas-delbovier-acsone: This implementation depends on PR #1115. That PR fixes the component initialization logic, ensuring self._menu is correctly populated. Without that fix, self._menu remains empty, preventing from accessing the necessary menu context.

nicolas-delbovier-acsone:
Improved the UI of the label-printer component. The updated layout provides a more robust and accessible experience on small warehouse scanner screens where space is limited.

image

nicolas-delbovier-acsone:
The recent refactoring I made on this PR is to fix `allow_print_label` injection to ensure persistent state handling in the frontend.

**The Issue**: Previously, the `allow_print_label` flag was injected into the root of the response dictionary rather than inside the specific state data defined in `_states().` Because the Shopfloor Vue application dynamically updates its local store based on the next_state payload, any data sent outside of that specific state key was ignored by the frontend’s persistence logic. This caused the flag to be lost upon browser refresh.

shopfloor_printing_base/components/printing.py
simahawk: I mean: add a field to the menu item and include this info (id + name of report) in the menu item info that is sent to the app.

shopfloor_base/services/service.py
```diff
@@ -23,11 +23,15 @@ class BaseShopfloorService(AbstractComponent):
     _collection = "shopfloor.app"
     _expose_model = None
 
-    def __init__(self, work_context):
-        super().__init__(work_context)
+    @property
```
simahawk, line 26: why changing this to properties?

sbejaoui (approved):

shopfloor_base/services/validator.py
```diff
@@ -162,10 +162,9 @@ class BaseShopfloorValidatorResponse(AbstractComponent):
     # Initial state of a workflow
     _start_state = "start"
 
-    def _states(self):
-        """List of possible next states
-
-        With the schema of the data send to the client to transition
+    def _states(self) -> dict:
+        """Returns a dict mapping next states with the schema
```
simahawk, line 166: is this change needed?

shopfloor_base/services/validator.py
```diff
@@ -181,6 +180,31 @@ def schemas(self):
     def schemas_detail(self):
         return self._actions_for("schema_detail")
 
+    def _get_global_fields_schemas(self) -> dict:
+        "Returns schemas of fields to be added in all next states data schemas"
```
simahawk: triple quotes needed

shopfloor_printing_base/components/printing.py
```diff
@@ -0,0 +1,44 @@
...
+from odoo.addons.component.components.base import AbstractComponent
+from odoo.addons.shopfloor_base.actions.base_action import get_actions_for
+
+
+class ShopFloorPrintingAction(AbstractComponent):
+    """Base Component for actions"""
+
+    _name = "shopfloor.printing.action"
+    _inherit = "shopfloor.process.action"
+    _collection = "shopfloor.printing"
```
simahawk, line 12: do we need another collection?

shopfloor_printing_base/tests/test_printing.py
```diff
@@ -0,0 +1,156 @@
...
+            kind="test",
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.loader.restore_registry()
+        super().tearDownClass()
+
+    @contextmanager
+    def _work_on_actions(self, **params):
```
simahawk, line 79: there's already a `work_on_actions` in the base module, do we need this?

shopfloor_base/services/service.py
nicolas-delbovier-acsone: This change is from https://github.com/OCA/wms/pull/1115/changes which I needed as explained above

shopfloor_base/services/validator.py
nicolas-delbovier-acsone: I found the previous docstring to be confusing. The return value is not a "list of possible next states", it is a dict mapping next state to its data schema

shopfloor_printing_base/components/printing.py
nicolas-delbovier-acsone: What collection would you want? "shopfloor.printing.base"? What value would a new collection bring?

simahawk: "printing" itself is an action. If you define another collection you have to lookup for another collection of actions. That's why you are defining "_printing_for" below.
This is way I'd say you avoid defining a new collection, unless you really need it.

shopfloor_mobile_printing_base/static/src/components/printing-widget.js
simahawk: still hesitating? 😄 
From what I can see you only need one slot above and one slot below. Correct?

nicolas-delbovier-acsone:
I just added a UI fix (because shopfloor menus can quickly become quite cluttered when we install a lot of OCA modules together)

| Before | After |
|--------|--------|
| image | image |

PR #1090, opened 2025-08-12
[16.0][IMP] shopfloor_reception*: allow over receiving

nicolas-delbovier-acsone:
This PR adds the possibility to receive more goods than ordered (in case of an error from the supplier for example)

image

image


shopfloor_reception/services/reception.py
```diff
@@ -1201,19 +1221,36 @@ def set_quantity__cancel_action(self, picking_id, selected_line_id):
...
-            return self._response_for_set_quantity(
-                picking, line, message=self.msg_store.unable_to_pick_qty()
+        if compare == 1 and not is_over_reception_confirmed:
+            line.write(previous_vals)
+            message = self._response_for_confirm_over_reception(
+                picking,
+                line,
+                quantity,
+                action,
+                message=self.msg_store.line_scanned_qty_done_higher_than_allowed(),
```
sbejaoui, line 1415: avoid the useless write,

```python
if compare == 1:
    return self._response_for_confirm_over_reception()
line.qty_done = quantity

```

shopfloor/actions/message.py
```diff
@@ -562,6 +562,15 @@ def create_new_pack_ask_confirmation(self, barcode):
...
 
+    # def over_reception_ask_confirm(self):
+    #     return {
+    #         "message_type": "warning",
+    #         "body": _(
+    #             "You confirmed you wanted to receive more goods "
+    #             "than expected. Click again to continue."
+    #         ),
+    #     }
+
```
sbejaoui:
```suggestion

```

shopfloor/actions/message.py
```diff
@@ -201,7 +201,7 @@ def confirm_location_changed(self, from_location, to_location):
         return {
             "message_type": "warning",
             "body": _(
-                "Confirm location change from %(location_from)s to " "%(location_to)s?"
+                "Confirm location change from %(location_from)s to %(location_to)s?"
```
sbejaoui: avoid this changes

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
```diff
@@ -220,7 +220,7 @@ export var PackagingQtyPicker = Vue.component("packaging-qty-picker", {
                 return "background-color: rgb(143, 191, 68)";
             }
             if (this.qty > this.qtyTodo) {
-                return "background-color: orangered";
+                return "background-color: rgb(229, 171, 0)";
```
sbejaoui: avoid this changes

shopfloor_reception/services/reception.py
```diff
@@ -828,6 +828,26 @@ def _response_for_set_quantity(
...
 
+    def _response_for_confirm_over_reception(
+        self,
+        picking,
+        line,
+        quantity,
+        action,
+        message,
+    ):
+        response = self._response(
```
sbejaoui, line 977: please add a comment here to explain the usage of this method and it's role in the process

shopfloor_reception/services/reception.py
```diff
@@ -1201,19 +1221,36 @@ def set_quantity__cancel_action(self, picking_id, selected_line_id):
...
+
+        This hook function is called when a user confirms an over-reception on a picking.
+        It can be extended to implement custom business logic, such as:
+            - Creating a new helpdesk ticket for the supplier.
+            - Sending a notification to a specific team.
+            - Automatically adjusting the purchase order quantity.
+        """
+
+    def _set_quantity__process__set_qty_and_split(
+        self, picking, line, quantity, action, is_over_reception_confirmed=False
```
sbejaoui:
```suggestion
        self, picking, line, quantity, action=None, is_over_reception_confirmed=False
```

sbejaoui (changes requested):

shopfloor_reception/tests/test_set_quantity.py
```diff
@@ -774,8 +772,8 @@ def test_move_states(self):
         )
         #
         expected_message = {
-            "body": "You cannot process that much units.",
-            "message_type": "error",
+            "body": "Please note that the scanned quantity is higher than the maximum allowed.",
```
sbejaoui, line 816: I don't see where you added this message

sbejaoui (changes requested):

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: This write is not useless, this is used to rollback the value before user confirms. 

I need to do this because the function `_set_quantity__check_quantity_done` is checking all the qty_done on the move so I need to update the move before this check but then I need to rollback before user confirms. I agree this is not beautiful since, in the case of a user confirming, we will have a useless double write. However, I suspect this is necessary to follow this pattern to avoid problems in the case of many workers working on the same move. 

I  copied this pattern from the function `_set_quantity__by_product`

shopfloor_reception/tests/test_set_quantity.py
nicolas-delbovier-acsone: This message already existed in the code:
https://github.com/acsone/wms/blob/462d83df8d7ebdfaba44936be2e48eacea6ffb08/shopfloor/actions/message.py#L761-L768

shopfloor_mobile/static/wms/src/components/packaging-qty-picker.js
nicolas-delbovier-acsone: The idea was to make the color a "warning" color instead of a "danger" one since now it is possible to over-receive. Thus, it is no more an error.

shopfloor_reception/services/reception.py
```diff
@@ -578,7 +598,9 @@ def _set_quantity__by_product(self, picking, selected_line, product):
             # If we have an error, return it, since this is also true for return lines
             if message_type == "error":
                 return response
-            compare = self._set_quantity__check_quantity_done(selected_line)
+            compare = self._set_quantity__check_quantity_done(
+                selected_line, selected_line.qty_done
```
sbejaoui:
```suggestion
                selected_line, sum(move.move_line_ids.mapped("qty_done"))
```

sbejaoui (changes requested):

sbejaoui (approved):
Code review + functional test

shopfloor_reception/README.rst
```diff
@@ -1,7 +1,3 @@
-.. image:: https://odoo-community.org/readme-banner-image
```
simahawk, line 1: Could you please rollback these README changes? 
TIP: run `export SKIP=oca-gen-addon-readme` to avoid updating readme files on each commit ;)

shopfloor_reception/services/reception.py
```diff
@@ -1201,19 +1270,44 @@ def set_quantity__cancel_action(self, picking_id, selected_line_id):
...
-        sum(move.move_line_ids.mapped("qty_done"))
-        savepoint = self._actions_for("savepoint").new()
-        line.qty_done = quantity
-        compare = self._set_quantity__check_quantity_done(line)
-        if compare == 1:
-            # If move's qty_done > to move's qty_todo, rollback and return an error
-            savepoint.rollback()
-            return self._response_for_set_quantity(
-                picking, line, message=self.msg_store.unable_to_pick_qty()
+    def _after_over_recepetion_confirmed_hook(self, picking, line):
```
```suggestion
    def _after_over_reception_confirmed_hook(self, picking, line):
```

shopfloor_reception/services/reception.py
```diff
@@ -1247,14 +1352,20 @@ def process_with_new_pack(self, picking_id, selected_line_id, quantity):
                 picking, selected_line, message=message
             )
         response = self._set_quantity__process__set_qty_and_split(
-            picking, selected_line, quantity
+            picking,
+            selected_line,
+            quantity,
+            "process_with_new_pack",
+            is_over_reception_confirmed,
```
```suggestion
            action="process_with_new_pack",
            is_over_reception_confirmed=is_over_reception_confirmed,
```
I'm not a fan of "hiding" kw args w/ positional args just because python allows to do so :)

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -85,6 +85,25 @@ const Reception = {
                     </v-row>
                 </div>
             </template>
+            <template v-if="state_is('confirm_over_reception')">
+                <div class="button-list button-vertical-list full">
+                    <v-row align="center">
+                        <v-col class="text-center" cols="12">
+                            <div>{{ state.display_info.message }}</div>
```
simahawk: `display_info` messages are normally visible on the top of the page automatically. Is this really needed?

simahawk (changes requested):
Can we have separated commits per module? The rest can be squashed.

sbejaoui:
can you rebase please

jbaudoux (changes requested):
LG.
There must an option to allow over receiving. Or block over-receiving as by default in the backend you can over-receive.
Initially is was blocked on purpose, that feature should not be lost.

nicolas-delbovier-acsone:
@jbaudoux I made the behaviour optional as you suggested. Can you have a look?

jbaudoux (approved):

PR #783, opened 2023-11-17
[14.0][ADD] shopfloor_checkout_putinpack_restriction

TDu:
Related PR:
https://github.com/OCA/stock-logistics-workflow/pull/1431
Has been merged.

PR #665, opened 2023-06-27
[14.0][IMP] shopfloor: package checkout buttons

raumschmiede-joshuaL:
Adds a menu config to allow/disallow the usage of packages

PR #1094, opened 2025-08-25
[16.0][IMP] shopfloor_reception/mobile: allow to not use packages

rousseldenis:


shopfloor_reception/data/shopfloor_scenario_data.xml
```diff
@@ -8,6 +8,7 @@
       <field name="key">reception</field>
       <field name="options_edit">
 {
+      "allow_dont_use_packs_in_reception": false,
```
jbaudoux, line 11: Can you do like in https://github.com/OCA/wms/pull/783 for checkout scenario and rely on the configuration of the picking type instead?

jbaudoux (changes requested):
Also drop changes to auto-generated files

shopfloor_reception/data/shopfloor_scenario_data.xml
jbaudoux: For the history, have a look at https://github.com/OCA/wms/pull/665#issuecomment-1816704422

jbaudoux:
@rousseldenis reminder to finish this PR

jbaudoux:
replaced by https://github.com/OCA/wms/pull/1163

PR #1097, opened 2025-09-01
[IMP] shopfloor: Allow to use picking type destination

rousseldenis:


jbaudoux (commented):
That shouldn't be needed if the destination of the move is properly managed. Maybe with dynamic routing push rule?

jbaudoux (changes requested):
You should not allow to do in shopfloor something that is not allowed in the backend.

PR #1098, opened 2025-09-01
 [16.0][FIX] shopfloor_mobile_base: display falsy but not undefined fields in detail-card.

nicolas-delbovier-acsone:
If the field is equal to `0` or `""` it would not get displayed in the card


rousseldenis (approved):

rousseldenis:
@jbaudoux

jbaudoux (approved):

PR #1100, opened 2025-09-01
[16.0][ADD] shopfloor_location_product_restriction

rousseldenis:
Depends on:

- #1097 
- https://github.com/OCA/stock-logistics-warehouse/pull/2342

Usage in:

- https://github.com/OCA/wms/pull/1099


shopfloor_location_product_restriction/services/service.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def is_dest_location_valid(self, moves, location, pick_type=False):
+        """
+        Check if destination location is restricted for
+        current moves
+        """
+        result = super().is_dest_location_valid(moves, location, pick_type=pick_type)
+        for move in moves:
+            if location._check_has_location_product_restriction(move.product_id):
+                return bool(False and result)
```
nicolas-delbovier-acsone, line 18: `False and ...`  will always return `False`. What was your idea in here? What do you want to return?

PR #1102, opened 2025-09-05
[FIX] stock_dynamic_routing: use _action_confirm result

sbejaoui:
_action_confirm returns a recordset; we should rely on that result for subsequent logic


sbejaoui:
cc/ @lmignon , @rousseldenis

lmignon (approved):

lmignon:
@sbejaoui It would be nice to make the same PR on the module into its new repo for odoo 18....

rousseldenis (approved):

PR #1104, opened 2025-09-19
[16.0] shopfloor_base: critical fixes

simahawk:
Backport of https://github.com/OCA/shopfloor-app/pull/19

### [shopfloor_base: fix _registered_endpoint_rule_keys ](https://github.com/OCA/wms/commit/43b9b2b2097cd5df7ecfb1244672863f2929747a)

This was breaking _unregister_controllers since we don't get tuples
but EndpointRule objects.

Prior to this change the call failed like
```
  File /odoo/external-src/shopfloor-app/shopfloor_base/models/shopfloor_app.py, line 166, in _registered_endpoint_rule_keys
    return [x[0] for x in self._registered_routes()]
            ~^^^
TypeError: 'EndpointRule' object is not subscriptable
```

**How to reproduce:** archive an app and sync registry (or directly call the method on an existing active app).

### [shopfloor_base: fix routing registration](https://github.com/OCA/wms/commit/7001f8c8ba18981895fdc5b59ac5d67fd9176e17)  (the most important fix)

This prevented the usage of the `fastapi` dispatcher leading - among other things - to broken handling of server side errors as exceptions are not wrapped.

**How to reproduce:** raise an exception server side (eg: override the 1st method of a scenario to raise an exception in any case) and see on the screen an alert w/ "undefined undefined" instead of a nice error message.

### TODO

@lmignon I have no time to test this on 16 so I simply backported the fixes. Could you have a look?
Migrations scripts should be also ported (w/o the "readonly" flag) iif the issue is confirmed as all the existing routes should be updated.

### DIFF
--- a/shopfloor_base/models/shopfloor_app.py
+++ b/shopfloor_base/models/shopfloor_app.py
@@ -163,7 +163,8 @@ class ShopfloorApp(models.Model):
 
     def _registered_endpoint_rule_keys(self):
         # `endpoint.route.sync.mixin` api
-        return [x[0] for x in self._registered_routes()]
+        # TODO: add tests
+        return [x.key for x in self._registered_routes()]
 
     def _register_hook(self):
         super()._register_hook()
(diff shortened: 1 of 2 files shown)


lmignon:
Thank you @simahawk. I take this PR on my TODO list.

PR #1110, opened 2025-10-31
[16.0][ADD] shopfloor_reception_helpdesk/mobile

rousseldenis:
Depends on:
* https://github.com/OCA/helpdesk/pull/875


nicolas-delbovier-acsone:
Here is what I added in the UI for selecting a motive: 
image

Let me now what you think.

nicolas-delbovier-acsone:
Improved UI:

image

nicolas-delbovier-acsone:
For the moment, the helpdesk button shows after cancel and back buttons which feels a little odd:
image

I spent quite some time attempting to position the button precisely, but the current method of direct string manipulation proved unreliable. Because the template contains multiple identical `<template>` tags, targeting the specific 'set_quantity' block via string searching is highly error-prone.

If anyone finds a better way to manipulate that template string, I'd be happy to hear about a solution 😀

nicolas-delbovier-acsone:
I added the helpdesk button on 'select_move' and 'set_lot' screens since it makes sense depending on the type of issue you might have in the warehouse

nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (approved):

jbaudoux:
@sbidoul modules names to reserve on pypi shopfloor_reception_helpdesk, shopfloor_reception_helpdesk_mobile. Thanks in advance

sbidoul:
names reserved

PR #1111, opened 2025-11-03
[16.0][FIX] stock_storage_type: Fixed wrong usage of create function

baguenth:
Using the wrong parameter in create function leads to, that for stock.quant.package all create-functions called with parameter "vals_list" will break.

TypeError: Base.create() got an unexpected keyword argument 'vals_list'

### DIFF
--- a/stock_storage_type/models/stock_quant_package.py
+++ b/stock_storage_type/models/stock_quant_package.py
@@ -57,8 +57,8 @@ class StockQuantPackage(models.Model):
         return res
 
     @api.model_create_multi
-    def create(self, vals):
-        records = super().create(vals)
+    def create(self, vals_list):
+        records = super().create(vals_list)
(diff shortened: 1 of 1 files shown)


jbaudoux (commented):
It's not a keywork argument but an argument. Can you share your stack trace? What is the caller that made the argument a keyword argument?

baguenth:
@jbaudoux It is just a regular create-call, nothing fancy.

In the end it is a straight forward fix. Simply, the parameter used in the module is wrong and not honouring the original parameter name. 

I saw this issue in other modules as well and also created pr for it. E.g. https://github.com/OCA/multi-company/pull/900

@pedrobaeza This is another pr regarding the wrong create function overwrite.

pedrobaeza (approved):
/ocabot merge patch

PR #1112, opened 2025-11-14
[16.0][IMP] stock_available_to_promise_release no search available to promise qty on done / canceled moves

santostelmo:
### Context 
I am trying to improve the processing time of [`product.pickings_auto_release`](https://github.com/OCA/stock-logistics-warehouse/blob/0639912c240d7693a86253abcc79b6be6a736248/stock_move_auto_assign_auto_release/models/product_product.py#L18)
With a large number of moves, this operation is slow and becomes slower as the number of moves increases. 
I wonder why we should look for available quantities to promise in moves that are already done or cancelled?
Perhaps there is a reason for this, but IMO we should exclude them from the search.

1. When stock move is done [_prepare_auto_assign](https://github.com/OCA/stock-logistics-warehouse/blob/0639912c240d7693a86253abcc79b6be6a736248/stock_move_auto_assign/models/stock_move.py#L15) is called.
2. Product moves with [is_auto_release_allowed](https://github.com/OCA/stock-logistics-warehouse/blob/0639912c240d7693a86253abcc79b6be6a736248/stock_move_auto_assign_auto_release/models/stock_picking.py#L38) are searched. `release_ready` is part of the search domain
3. `_search_release_ready` scan ALL moves with `ordered_available_to_promise_uom_qty > 0` no matter the state. 
https://github.com/OCA/wms/blob/da5e224621c6f128e84031f77f3a910673022f87/stock_available_to_promise_release/models/stock_move.py#L387 => This is very slow

**Speedscope before this change** (~1min)

image

**Speedscope after this change** (~12s)
image

### DIFF
--- a/stock_available_to_promise_release/models/stock_move.py
+++ b/stock_available_to_promise_release/models/stock_move.py
@@ -471,7 +471,17 @@ class StockMove(models.Model):
         }
         if operator not in operator_mapping:
             raise UserError(_("Unsupported operator %s") % (operator,))
-        moves = self.search([("need_release", "=", True)])
+
+        moves = self.search(
+            [
+                (
+                    "state",
+                    "in",
+                    ["waiting", "confirmed", "partially_available", "assigned"],
+                ),
+                ("need_release", "=", True),
+            ]
+        )
         operator_func = operator_mapping[operator]
         # computed field has no depends set, invalidate cache before reading
         moves.invalidate_recordset(["ordered_available_to_promise_uom_qty"])


santostelmo:
@jbaudoux can I please have your opinion on that ?

cyrilmanuel (approved):

stock_available_to_promise_release/models/stock_move.py
```diff
@@ -471,7 +471,10 @@ def _search_ordered_available_to_promise_uom_qty(self, operator, value):
         }
         if operator not in operator_mapping:
             raise UserError(_("Unsupported operator %s") % (operator,))
-        moves = self.search([("need_release", "=", True)])
+
+        moves = self.search(
+            [("state", "not in", ["done", "cancel"]), ("need_release", "=", True)]
```
gurneyalex: you will possibly get a faster SQL query by using the "in" operator and the reverse set of states (this allows PostgreSQL to use an index on the state column if it exists.

gurneyalex:
side note, from what I see in the code, when a move is cancelled, `need_release` is set to `False`. But since there is an index on `state` it is still interesting.

stock_available_to_promise_release/models/stock_move.py
santostelmo: In this case it also make sense to search in ["waiting", "confirmed", "partially_available",  "assigned"] 
[here](https://github.com/OCA/stock-logistics-warehouse/blob/0639912c240d7693a86253abcc79b6be6a736248/stock_move_auto_assign_auto_release/models/stock_move.py#L41)

jbaudoux (commented):
There is an index on `need_release`, and done or canceled moves should never be in need release. You more likely have a data issue.
Fix your data, remove `need_release` from any done/cancel moves or any moves that are not outgoing.

sebalix:
IMO the issue is located in `stock_move_auto_assign_auto_release` which triggers too much queries due to unwanted search operations on `release_ready` + `ordered_available_to_promise_qty`. I did a fix here which greatly improved the situation in our project, and reduce the number of queries executed a lot:
- https://github.com/OCA/stock-logistics-workflow/pull/2169

santostelmo:
> IMO the issue is located in `stock_move_auto_assign_auto_release` which triggers too much queries due to unwanted search operations on `release_ready` + `ordered_available_to_promise_qty`. I did a fix here which greatly improved the situation in our project, and reduce the number of queries executed a lot:
> 
> * [[18.0][IMP] stock_move_auto_assign_auto_release: improve performance stock-logistics-workflow#2169](https://github.com/OCA/stock-logistics-workflow/pull/2169)

@sebalix thanks a lot for that. Indeed your approach improves even more the processing time.
I close this PR and backport your improvement. 
Done here: https://github.com/OCA/stock-logistics-warehouse/pull/2468

santostelmo:
Closing in favor of this implementation https://github.com/OCA/stock-logistics-workflow/pull/2169

PR #1115, opened 2026-01-07
[FIX] shopfloor_base: Set _menu and _profile as properties

rousseldenis:
As soon as work context is not loaded yet at service init(), change them to properties to allow dynamic evaluation.

### DIFF
--- a/shopfloor_base/services/service.py
+++ b/shopfloor_base/services/service.py
@@ -23,11 +23,15 @@ class BaseShopfloorService(AbstractComponent):
     _collection = "shopfloor.app"
     _expose_model = None
 
-    def __init__(self, work_context):
-        super().__init__(work_context)
+    @property
+    def _menu(self):
         # User private attributes to not mess up w/ public endpoints
-        self._profile = getattr(self.work, "profile", self.env["shopfloor.profile"])
-        self._menu = getattr(self.work, "menu", self.env["shopfloor.menu"])
+        return getattr(self.work, "menu", self.env["shopfloor.menu"])
+
+    @property
+    def _profile(self):
+        # User private attributes to not mess up w/ public endpoints
+        return getattr(self.work, "profile", self.env["shopfloor.profile"])
 
     def _get_api_spec(self, **params):
         return ShopfloorRestServiceAPISpec(self, **params)


nicolas-delbovier-acsone (approved):

simahawk (commented):
I'm not sure to understand what you are trying to fix 🤔 
What's the use case?

I think I did that on purpose to freeze the two variables at component instance level.
See https://github.com/acsone/wms/commit/9dfe2abd09ac73ef7bd3c4938f2699549cce0fd4

nicolas-delbovier-acsone:
> I'm not sure to understand what you are trying to fix 🤔 What's the use case?
> 
> I think I did that on purpose to freeze the two variables at component instance level. See [acsone@9dfe2ab](https://github.com/acsone/wms/commit/9dfe2abd09ac73ef7bd3c4938f2699549cce0fd4)

Personally, I needed this while debugging #1085. I had the same problem that described here: the variables where not populated when I called them.

sbejaoui (approved):

simahawk:
> > I'm not sure to understand what you are trying to fix 🤔 What's the use case?
> > I think I did that on purpose to freeze the two variables at component instance level. See [acsone@9dfe2ab](https://github.com/acsone/wms/commit/9dfe2abd09ac73ef7bd3c4938f2699549cce0fd4)
> 
> Personally, I needed this while debugging #1085. I had the same problem that described here: the variables where not populated when I called them.

Debugging what and how? If you lookup the component using component machinery, if the work context used it's properly set before the component is initialized I see no reason to no have the value propagated.

If you had this issue only on tests and not while using the app, then you might have something wrong in the tests.
I'll check on that PR.

rousseldenis:
>  I see no reason to no have the value propagated.

That's the case in production with real behavior. The values are void

But why should not menu and profiles can be evaluated at runtime ?

rousseldenis:
> I think I did that on purpose to freeze the two variables at component instance level.
> See [acsone@9dfe2ab](https://github.com/acsone/wms/commit/9dfe2abd09ac73ef7bd3c4938f2699549cce0fd4)

@simahawk in fact, instance and not work context. Here, the change does not introduce drawback.

image

Could you review again ?

simahawk (approved):
let it go....

PR #1116, opened 2026-01-08
[IMP] shopfloor: Add location name to error message for invalid destination location

rousseldenis:


shopfloor/actions/message.py
```diff
@@ -191,8 +192,22 @@ def no_location_found(self):
...
+    def dest_location_not_allowed(
+        self, location: (Location | None) = None, extra_message: str = ""
+    ) -> dict:
+        """
+        This returns the message that destination location is incorrect
+        """
+        if location:
+            body = _(
+                "You cannot place it here (%(location_name)s)",
+                location_name=location.name if location else "",
```
nicolas-delbovier-acsone: The ternary operator here is redundant since you are already inside `if location:`

jbaudoux: @nicolas-delbovier-acsone @rousseldenis Can one of you fix this?

nicolas-delbovier-acsone: I'll do it

PR #1125, opened 2026-01-27
[16.0][FIX] shopfloor_reception: Incorrect move line selection after partial receipt (without pack)

nicolas-delbovier-acsone:
**Problem**

When performing a partial reception without packages, the system splits the stock.move.line into a completed line and a backorder line. Currently, the line selection logic incorrectly assumes that fully processed lines are the ones linked to a package. This causes the system to re-select the already-processed line (e.g., 3/3) instead of the remaining one (e.g., 0/7), blocking the completion of the transfer and causing data inconsistency.

**Solution**

Updated the move line selection logic to ensure that lines are selected based on the fact they have been completely processed or not.

---

This PR isolates and improves some change made in https://github.com/OCA/wms/pull/1092 as to ease review (see [this commit](https://github.com/OCA/wms/pull/1092/changes/3da08bf3c48713406f31bd7aea7a75350d49ff15))

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -287,7 +287,7 @@ class Reception(Component):
         """
         unassigned_lines = self.env["stock.move.line"]
         for line in move.move_line_ids:
-            if line.progress == 100.0:
+            if line.shopfloor_unloaded:
                 continue
             if line.shopfloor_user_id.id == self.env.uid:
                 return self._scan_line__recover(picking, line, qty_done)
@@ -335,7 +335,7 @@ class Reception(Component):
         return_line = fields.first(
             lines.filtered(
                 lambda l: not l.package_id.product_packaging_id
-                and not l.result_package_id
+                and not l.shopfloor_unloaded
                 and l.shopfloor_user_id.id in (False, self.env.uid)
             )
         )
@@ -352,7 +352,7 @@ class Reception(Component):
         return fields.first(
             lines.filtered(
                 lambda l: l.package_id.product_packaging_id == packaging
-                and not l.result_package_id
+                and not l.shopfloor_unloaded
                 and l.shopfloor_user_id.id in [False, self.env.uid]
             )
         )
@@ -550,8 +550,7 @@ class Reception(Component):
                     lot == l.lot_id
                     or (lot.name == l.lot_name and lot.product_id == l.product_id)
                 )
-                and not l.progress == 100.0
-                and not l.result_package_id
+                and not l.shopfloor_unloaded
             )
         )
         if not lines:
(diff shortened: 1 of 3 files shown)


nicolas-delbovier-acsone:
@jbaudoux

nicolas-delbovier-acsone:
We should use the field `shopfloor_unloaded` instead of relying on progress

shopfloor_reception/services/reception.py
```diff
@@ -1361,6 +1362,7 @@ def process_without_pack(self, picking_id, selected_line_id, quantity):
 
     def _post_line(self, selected_line):
         selected_line.reserved_uom_qty = selected_line.qty_done
+        selected_line.shopfloor_unloaded = True
```
jbaudoux, line 1394: It's in case you don't post to know that the product has been unloaded at destination. This indicates you don't need to process the line anymore. So when you write the destination location in case of process without pack, or when you write the destination location on a package that has no location yet, or when you put on a package that already have a destination location

jbaudoux: Seems right. Can you add a docstring to the method that it is called when the product is unloaded at destination

shopfloor_reception/services/reception.py
```diff
@@ -544,8 +546,11 @@ def _scan_line__by_lot(self, picking, lot):
...
         lines = picking.move_line_ids.filtered(
             lambda l: (
-                lot == l.lot_id
-                or (lot.name == l.lot_name and lot.product_id == l.product_id)
+                (
+                    lot == l.lot_id
+                    or (lot.name == l.lot_name and lot.product_id == l.product_id)
+                )
+                and not l.shopfloor_unloaded
                 and not l.result_package_id
```
jbaudoux: Below in fallback, you dropped `result_package_id`. I would do the same here.
Here and in fallback, it's like it's missing a test on the user. If 2 operators are processing the same lot, the second operator should not steal the move line of the first operator.

Look at `_select_line__filter_lines_by_packaging` and `_select_line__filter_lines_by_packaging__return`, they should also be aligned

shopfloor_reception/services/reception.py
```diff
@@ -544,9 +546,11 @@ def _scan_line__by_lot(self, picking, lot):
...
         lines = picking.move_line_ids.filtered(
             lambda l: (
-                lot == l.lot_id
-                or (lot.name == l.lot_name and lot.product_id == l.product_id)
-                and not l.result_package_id
+                (
+                    lot == l.lot_id
+                    or (lot.name == l.lot_name and lot.product_id == l.product_id)
+                )
+                and not l.shopfloor_unloaded
```
jbaudoux, line 553:
```suggestion
                and not l.shopfloor_unloaded
                and l.shopfloor_user_id.id in (False, self.env.uid)


```

shopfloor_reception/services/reception.py
```diff
@@ -568,7 +572,7 @@ def _scan_line__by_lot(self, picking, lot):
     def _scan_line__fallback(self, picking, barcode):
         # We might have lines with no lot, but with a lot_name.
         lines = picking.move_line_ids.filtered(
-            lambda l: l.lot_name == barcode and not l.result_package_id
+            lambda l: l.lot_name == barcode and not l.shopfloor_unloaded
```
```suggestion
            lambda l: l.lot_name == barcode and not l.shopfloor_unloaded
                            and l.shopfloor_user_id.id in (False, self.env.uid)

```

jbaudoux (approved):

PR #1126, opened 2026-01-28
[16.0][FIX] shopfloor_reception: incorrect increment of move line qty_done on manual selection.

nicolas-delbovier-acsone:
Currently, selecting a move line manually automatically increments the `qty_done` by 1. This behavior creates data inconsistencies if an operator selects a line by mistake; navigating "Back" does not revert the increment, leaving an incorrect quantity on the move line.

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1043,7 +1043,7 @@ class Reception(Component):
     def manual_select_move(self, move_id):
         move = self.env["stock.move"].browse(move_id)
         picking = move.picking_id
-        return self._scan_line__find_or_create_line(picking, move)
+        return self._scan_line__find_or_create_line(picking, move, qty_done=0)
 
     def done_action(self, picking_id, confirmation=False):
         """Mark a picking as done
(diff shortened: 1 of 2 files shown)


nicolas-delbovier-acsone:
@jbaudoux

jbaudoux:
When a product is scanned, the quantity is increased by 1 if the barcode is attached to the product or by the packaging quantity if the barcode is attached to it. However, when a line is selected, in my opinion it make sense that the quantity is not increased. Once you are on the set quantity screen, then you can set if you received 1 unit or 1 box.
@mt-software-de I think you implemented this manual selection.

jbaudoux (approved):

mt-software-de:
> When a product is scanned, the quantity is increased by 1 if the barcode is attached to the product or by the packaging quantity if the barcode is attached to it. However, when a line is selected, in my opinion it make sense that the quantity is not increased. Once you are on the set quantity screen, then you can set if you received 1 unit or 1 box. @mt-software-de I think you implemented this manual selection.

Yes i changed it. Thx.
It think it is better with your change.
But i don't know if it is a fix or feature :-)

mt-software-de (approved):
LGTM

PR #1127, opened 2026-01-28
[16.0][IMP] shopfloor_reception_mobile: display qty_done over total qty in select_move.

nicolas-delbovier-acsone:
| Before | After |
|--------|--------|
| image |image|

### DIFF
--- a/shopfloor_reception_mobile/static/src/scenario/reception.js
+++ b/shopfloor_reception_mobile/static/src/scenario/reception.js
@@ -380,6 +380,7 @@ const Reception = {
             };
         },
         picking_detail_options_for_select_move: function () {
+            const self = this;
             return {
                 show_title: true,
                 showActions: false,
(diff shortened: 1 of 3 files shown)


shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -381,6 +381,9 @@ const Reception = {
                             path: "quantity_done",
                             label: "Qty done",
                             display_no_value: true,
+                            renderer: function (rec, field) {
+                                return rec.quantity_done + " / " + rec.quantity;
+                            },
```
jbaudoux, line 386: Ideally, it should display the quantity by packaging
```suggestion
                    render_component: "packaging-qty-picker-display",
                    render_props: function (record) {
                        return self.utils.wms.move_line_qty_picker_props(record, {
                            qtyInit: record.quantity_done,
                            # qtyTodo: record.quantity,
                        });
                    },
```

And then the component could be improved to support a qtyTodo to render it the same way

Examples:
* 0 / 1000 (1 PAL, 3 TU)
* 200 (2 TU) / 1000 (1 PAL, 3 TU)

nicolas-delbovier-acsone:
@jbaudoux, Here is the new UI with your suggested implementation:

image

I am not sure if I like that... It is a bit cluttered isn't it?

nicolas-delbovier-acsone:
Also, I needed to make a [[FIX] commit](https://github.com/OCA/wms/pull/1127/changes/40a5afa14a9831259926378dd4aa4b74fd2a21b1) because of this change in another scenario:

| With FIX (same as before this PR) | Without FIX |
|--------|--------|
| image | image | 

This is because the code was passing `qtyTodo` by default to the "packaging-qty-picker-display" component which was ignored until now since it did not exist as a prop before (it only existed on the "packaging-qty-picker" component beforehand).

jbaudoux (approved):

PR #1130, opened 2026-02-12
[14.0][FIX] shopfloor_delivery_shipment: Search only docks of warehouse

mt-software-de:
If there are docks with same barcodes in different warehouses.
It depends on the ordering of docks, if dock for the right warehouse is found. 
By adding the warehouse to the domain when searching a dock it is fixed. 

@jbaudoux


mt-software-de:
ping @i-vyshnevska

jbaudoux (approved):

PR #1131, opened 2026-02-17
[14.0][IMP] shopfloor: Allow to search for locations of a company

mt-software-de:
This PR allows to search for locations of a company.
That brings us one step closer to a multi company capability. 

### DIFF
--- a/shopfloor/actions/search.py
+++ b/shopfloor/actions/search.py
@@ -91,15 +91,20 @@ class SearchAction(Component):
                 return self._make_search_result(record=record, code=barcode, type=btype)
         return self._make_search_result(type="none")
 
-    def location_from_scan(self, barcode, limit=1):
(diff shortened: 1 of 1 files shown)


jbaudoux (approved):
I guess there is the same need for the search of product

jbaudoux (commented):
You could also rely on env.company and call the search using `with_company()` to not change the signature of every methods

vuwnevska (approved):

mt-software-de:
> You could also rely on env.company and call the search using `with_company()` to not change the signature of every methods

Yes but right now there is not `with_company` on the components `shopfloor.search.action`

nicolas-delbovier-acsone (approved):

mt-software-de:
This will be closed due to changes made in https://github.com/OCA/wms/pull/1133.

PR #1133, opened 2026-02-17
[14.0][IMP] shopfloor_base: Multi company capability

mt-software-de:
Search locations for company of picking type. Create lots with company of picking type. 

Depends on https://github.com/OCA/wms/pull/1131

### DIFF
--- a/shopfloor_base/services/service.py
+++ b/shopfloor_base/services/service.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # Copyright 2020 Akretion (http://www.akretion.com)
+# Copyright 2026 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 from werkzeug.exceptions import BadRequest
@@ -298,3 +299,23 @@ class BaseShopfloorProcess(AbstractComponent):
 
     _requires_header_menu = True
     _requires_header_profile = True
+
+    @property
+    def env(self):
+        env = super().env()
+        if (
+            not hasattr(self, "work")
+            or not hasattr(self.work, "menu")
+            or not hasattr(self.work.menu, "picking_type_ids")
+        ):
+            return env
+        company = self.work.menu.picking_type_ids.company_id
+        if not company:
+            return env
+        context = dict(self.collection.env.context)
+        context.update(
+            {
+                "allowed_company_ids": company.ids,
+            }
+        )
+        return env(context=context)
--- a/shopfloor/tests/test_single_pack_transfer.py
+++ b/shopfloor/tests/test_single_pack_transfer.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # Copyright 2020 Akretion (http://www.akretion.com)
+# Copyright 2026 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo.tests.common import Form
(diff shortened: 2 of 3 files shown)


jbaudoux (approved):

shopfloor_reception/services/reception.py
```diff
@@ -1349,7 +1349,7 @@ def set_destination(
             )
         search = self._actions_for("search")
 
-        location = search.location_from_scan(location_name)
+        location = search.location_from_scan(location_name, company=picking.company_id)
```
jbaudoux: What about 
```suggestion
        search = self._actions_for("search").with_company(picking.company_id)

        location = search.location_from_scan(location_name)
```

jbaudoux:
cc @nicolas-delbovier-acsone

shopfloor_reception/services/reception.py
mt-software-de: i will change it i like your suggestion

jbaudoux: `with_company` cannot be called on a work component.

@simahawk How would you solve to provide the right company to the env of the component as it affects several search methods in shopfloor

mt-software-de: > i will change it i like your suggestion

I will give it a try tomorrow by extending `shopfloor.search.action`

nicolas-delbovier-acsone (approved):

shopfloor_reception/services/reception.py
jbaudoux: @mt-software-de You could modify shopfloor base service with something like this to define `with_company`

```
    @property
    def company(self):
        if self._company is None:
            return self.env.company
        return self._company

    @contextmanager
    def with_company(self, company):
        try:
            self._company = company
            yield
        finally:
            self._company = None


    def _actions_for(self, usage, **kw):
        return get_actions_for(self, usage, company=self.company, **kw)
```

You would need to initialize in an init method `_company = None` in that base service

In the action, you could then retrieve the company with `self.work.company` instead of using `self.env.company`

mt-software-de: Since there was a new issue in the checkout scenario, i changed the approach. 
Now the base of all scenarios are changed. `base.shopfloor.process`.
When `_actions_for` is called, it tries to set the company from `self.work.menu.picking_type_ids.company`,
this ensures that the env has the right company set. 
Maybe it is possible to set already in `__init__`, but i didn't checked. The first tests on our side worked. 

IMO. Its better to have the right company in the env already instead of passing it. Or using a new property. 

cc @jbaudoux

jbaudoux (approved):
Much better :)

nicolas-delbovier-acsone (approved):

mt-software-de:
> Much better :)

yes i think so too.

mt-software-de:
> Much better :)

I will close other PRs.

rousseldenis (approved):

shopfloor_base/services/service.py
```diff
@@ -298,3 +299,23 @@ class BaseShopfloorProcess(AbstractComponent):
 
     _requires_header_menu = True
     _requires_header_profile = True
+
+    @property
+    def env(self):
+        env = super().env()
+        if (
+            not hasattr(self, "work")
```
simahawk, line 307: Sorry, I'm late to the party but TBH I don't like this implementation 😞 

There's no need for these checks. Also, the work context will always be initialized properly before you use the env and on such "process" services the menu will always be set. In the context of `shopfloor` the picking types as well will be always set as it's mandatory on the menu items -> this change should stay in `shopfloor` if really needed.

A better approach would have been to allow the component to explicitly provide the allowed companies where needed and then override this in `shopfloor`.

Another place, probably even better, is this https://github.com/OCA/rest-framework/blob/18.0/base_rest/controllers/main.py#L170 where the env is initialized for the whole request and not just for this component.

In the entire application the `menu` comes from the header `HTTP_SERVICE_CTX_MENU_ID` which can be used there to inject the right companies.

mt-software-de: Thx for your input. No problem i have no problem to change this again. :-)
The checks were only implemented for unittests.

I will check your suggestions.

PR #1134, opened 2026-02-17
[14.0][IMP] shopfloor: Single pack transfer create package level for company of picking type

mt-software-de:
Right now the package level for new create picking is assigned to the environments company.
Setting the company to the picking type's company, the single pack transfer scenario, can be used in multi company environment. 


shopfloor/services/single_pack_transfer.py
```diff
@@ -193,7 +193,7 @@ def _create_package_level(self, package):
                 "picking_id": picking.id,
                 "package_id": package.id,
                 "location_dest_id": picking.location_dest_id.id,
-                "company_id": self.env.company.id,
+                "company_id": self.picking_types.company_id.id,
```
jbaudoux: What about ?
```suggestion
                "company_id": picking.company_id.id,
```

jbaudoux (changes requested):

shopfloor/services/single_pack_transfer.py
mt-software-de: Yes.

jbaudoux (approved):

mt-software-de:
ping @jbaudoux could retry to merge

mt-software-de:
> ping @jbaudoux could retry to merge

This wouldn't be needed anymore, if shopfloor_base is changed like it is described here https://github.com/OCA/wms/pull/1133#discussion_r2911377489

mt-software-de:
@jbaudoux Should we still merge this? After https://github.com/OCA/wms/pull/1133 is changing the env.company?
I still think it is better to use the company picking_type directly instead of the company from env

jbaudoux:
> @jbaudoux Should we still merge this? After #1133 is changing the env.company? I still think it is better to use the company picking_type directly instead of the company from env

yes it's better

/ocabot merge patch

Have you updated the copier? I'm not sure anything can be merged here before that

mt-software-de:
> > @jbaudoux Should we still merge this? After #1133 is changing the env.company? I still think it is better to use the company picking_type directly instead of the company from env
> 
> yes it's better
> 
> /ocabot merge patch
> 
> Have you updated the copier? I'm not sure anything can be merged here before that

no i didn't

mt-software-de:
> > @jbaudoux Should we still merge this? After #1133 is changing the env.company? I still think it is better to use the company picking_type directly instead of the company from env
> 
> yes it's better
> 
> /ocabot merge patch
> 
> Have you updated the copier? I'm not sure anything can be merged here before that

I updated copier.

PR #1135, opened 2026-02-17
[14.0][IMP] shopfloor: Single pack transfer owner

mt-software-de:
Allow to move packages of owner and keep owner

### DIFF
--- a/shopfloor/services/single_pack_transfer.py
+++ b/shopfloor/services/single_pack_transfer.py
@@ -180,14 +180,25 @@ class SinglePackTransfer(Component):
 
         return self._response_for_scan_location(package_level)
 
+    def _prepare_picking_vals_for_package(self, package):
+        vals = {}
+        if package.owner_id:
+            vals["owner_id"] = package.owner_id.id
+        return vals
+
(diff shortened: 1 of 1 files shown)


shopfloor/services/single_pack_transfer.py
```diff
@@ -180,14 +180,25 @@ def start(self, barcode, confirmation=None):
 
         return self._response_for_scan_location(package_level)
 
+    def _prepare_picking_vals_for_package(self, package):
+        vals = {}
+        if package.owner_id:
+            vals["owner_id"] = package.owner_id.id
```
jbaudoux, line 186: You should do that only when you want to change ownership from what I remember.

The package will have an owner when all quants inside it have that owner.
The picking should keep the quant ownership during the transfer (unless you force another one on the picking). The created moves will have the owner of the quant inside the package. Same for the move lines.

mt-software-de: Yes in the normal case it is keeping the owner, but somehow in this scenario its not the case. I need to investigate why.

PR #1140, opened 2026-02-24
[16.0][IMP] shopfloor_mobile: display absolute unit counts in packaging picker

nicolas-delbovier-acsone:
Updated the 'packaging-qty-picker' component to display the total number of base units contained within each packaging level. Previously, only relative quantities (e.g., x6 Level 2) were shown. This change provides the absolute unit count (e.g., x48 Units) to simplify encoding for the operator.

---

| Before | After |
| :---: | :---: |
| before | after |


simahawk (commented):
IMO this must be optional.
It could be a global setting via app info or flag on the menu (this gives more freedom depending on menu and profile).

@jbaudoux WDYT?

jbaudoux (changes requested):
Euh, no. That's not helping the operator. You can only count what is on the previous level 
If you take a pallet, you want to check you have 12 boxes on it. You never want to open the box to check that inside there are smaller boxes with multiple units or even smaller packagings

nicolas-delbovier-acsone:
Ok, you convienced me

PR #1144, opened 2026-03-16
[16.0][IMP]stock_warehouse_flow

AdrianaSaiz:
Summary
Add a computed field wh_total_products on stock.move that sums product_uom_qty of sibling moves sharing the same sale order and warehouse, excluding cancelled moves.

This allows flow domains to evaluate per-warehouse quantities instead of the global sale.order.total_products, which is critical when a single SO is split across multiple warehouses.

Problem
When SO lines are routed to different warehouses, flow domains referencing sale_line_id.order_id.total_products still see the global SO total. For example, a warehouse handling 10 units gets routed as B2B because the SO total is 25.

Solution
New field wh_total_products (Float, compute, non-stored) on stock.move
Helper method _get_wh_sibling_moves() returns moves from the same SO + warehouse
Compute batches all involved orders in a single query grouped by (order_id, warehouse_id) to avoid N+1
Falls back to the move's own product_uom_qty when there is no sale_line_id
Flow domain usage

No new dependencies added
Compatible with single-warehouse scenarios (same value as total_products)
Field is computed on-the-fly during _action_confirm(), when moves already have warehouse_id and sale_line_id set by procurement

Tests
All moves in one warehouse (total = full SO sum)
Moves split across 2 warehouses (each one sees only its qty in the new field)
Move without sale_line_id (fallback to own qty)
Cancelled moves excluded from total
_get_wh_sibling_moves() returns correct subset
Integration: SO confirm generates moves with correct field value

### DIFF
--- a/stock_warehouse_flow/models/stock_move.py
+++ b/stock_warehouse_flow/models/stock_move.py
@@ -1,8 +1,9 @@
 # Copyright 2022 Camptocamp SA
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# © 2026 FactorLibre - Adriana Saiz <adriana.saiz@factorlibre.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class StockMove(models.Model):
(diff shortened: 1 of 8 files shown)


stock_warehouse_flow/models/stock_move.py
```diff
@@ -15,6 +16,61 @@ class StockMove(models.Model):
...
+        moves_with_sale = self.filtered("sale_line_id")
+        for move in self - moves_with_sale:
+            move.wh_total_products = move.product_uom_qty
+        if not moves_with_sale:
+            return
+        order_ids = moves_with_sale.sale_line_id.order_id.ids
+        all_siblings = self.search(
+            [
+                ("sale_line_id.order_id", "in", order_ids),
+                ("state", "!=", "cancel"),
```
sergiobstoj, line 57: should `state` be present on `api.depends`?

stock_warehouse_flow/tests/test_wh_total_products.py
```diff
@@ -0,0 +1,210 @@
...
+                (self.product_a, 5),
+                (self.product_b, 10),
+                (self.product_c, 3),
+            ]
+        )
+        moves = self.env["stock.move"]
+        for line in order.order_line:
+            moves |= self._create_move(line, self.wh1)
+
+        for move in moves:
```
sergiobstoj: missing determinism. If moves is empty tests will silently pass, consider using mapped() + set().

```python
    self.assertEqual(len(moves), 2)
    self.assertEqual(set(moves.mapped("wh_total_products")), {18.0})
```

stock_warehouse_flow/models/stock_move.py
AdrianaSaiz: Hello Sergio, the state we filter on (!= cancel) belongs to sibling moves fetched via search(), not to the record itself.

stock_warehouse_flow/tests/test_wh_total_products.py
AdrianaSaiz: Changes applied, thanks

stock_warehouse_flow/models/stock_move.py
sergiobstoj: understood, thanks!

sergiobstoj (approved):

jbaudoux (changes requested):
The new field doesn't bring value to the engine itself.
It should be in a separate module.

If you have one delivery per warehouse, why you need to go back to the SO to count the total quantity per warehouse. Why don't you just sum the quantity of all the moves in the delivery?

PR #1146, opened 2026-03-26
[16.0][IMP] shopfloor/shopfloor_reception: prevent unexpected backorders when unmarking moves

nicolas-delbovier-acsone:
Previously, unmarking a move line as picked would trigger backorder logic (see https://github.com/OCA/wms/pull/543),
moving assigned moves into a new picking. This was functionally confusing
for UI users in the reception scenario, as canceling an action on a move line would cause items to
"disappear" from the current picking.

There is now an option on the unmarking function to choose to not trigger this split.

### DIFF
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -152,32 +152,33 @@ class StockAction(Component):
                 lines.write(values_assigned)
         move_lines.picking_id.filtered(lambda p: p.user_id != user).user_id = user.id
 
-    def unmark_move_line_as_picked(self, move_lines):
+    def unmark_move_line_as_picked(self, move_lines, split=True):
         """Reverse the change from `mark_move_line_as_picked`."""
         move_lines.write(
             {
                 "shopfloor_user_id": False,
                 "qty_done": 0,
                 "result_package_id": False,
+                "lot_id": False,
             }
         )
         pickings = move_lines.picking_id
         for picking in pickings:
-            lines_still_assigned = picking.move_line_ids.filtered(
-                lambda x: x.shopfloor_user_id
-            )
-            if lines_still_assigned:
-                # Because there is other lines in the picking still assigned
-                # The picking has to be split
-                unmark_lines = picking.move_line_ids & move_lines
-                unmark_lines._extract_in_split_order(default={"user_id": False})
-            else:
-                pickings.write(
+            still_assigned_users = picking.move_line_ids.shopfloor_user_id
+            if not still_assigned_users:
+                picking.write(
                     {
                         "user_id": False,
                         "printed": False,
                     }
                 )
+            else:
+                # Decide what to do if there are other lines in the picking still assigned
+                if split:
+                    unmarked_lines = picking.move_line_ids & move_lines
+                    unmarked_lines._extract_in_split_order(default={"user_id": False})
+                else:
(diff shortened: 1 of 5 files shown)


lmignon (approved):
As I understand, a call to `cancel` at the `set_quantity`  step no more cancel the operation and move it into a backorder but reset it to its original state.... it makes sense .

LGTM (Code review)

shopfloor_reception/services/reception.py
```diff
@@ -1305,8 +1305,13 @@ def set_quantity__cancel_action(self, picking_id, selected_line_id):
...
             if selected_line.reserved_uom_qty:
-                stock = self._actions_for("stock")
-                stock.unmark_move_line_as_picked(selected_line)
+                selected_line.write(
+                    {
+                        "shopfloor_user_id": False,
+                        "qty_done": 0,
+                        "lot_id": False,
+                    }
+                )
```
jbaudoux: Keep `unmark_move_line_as_picked` as it's managing the picking as well. Just set the `lot_id` to False afterwards

jbaudoux (changes requested):

shopfloor_reception/services/reception.py
rousseldenis: I'd prefer that too as the function could be overridden somewhere else

shopfloor/actions/stock.py
```diff
@@ -159,25 +159,24 @@ def unmark_move_line_as_picked(self, move_lines):
...
         pickings = move_lines.picking_id
         for picking in pickings:
-            lines_still_assigned = picking.move_line_ids.filtered(
-                lambda x: x.shopfloor_user_id
-            )
-            if lines_still_assigned:
-                # Because there is other lines in the picking still assigned
-                # The picking has to be split
-                unmark_lines = picking.move_line_ids & move_lines
-                unmark_lines._extract_in_split_order(default={"user_id": False})
```
jbaudoux, line 173: Can you keep this as default behavior and add an argument `split=True` to the method like it is in `mark_move_line_as_picked`. So `if split` then you still do the extract in split order, `else` you do your new solution.

jbaudoux (changes requested):
Differentiate with and without split.
Keep the existing tests for the case `split=True` (default value).
Add your new tests for the case `split=False`

The split was there for the other shopfloor scenario. For the reception, it indeed make sense to not split. `shopfloor_reception` 'is the scenario that introduced `split=False` for `mark_move_line_as_picked`

shopfloor/actions/stock.py
```diff
@@ -152,32 +152,33 @@ def mark_move_line_as_picked(
...
                     {
                         "user_id": False,
                         "printed": False,
                     }
                 )
+            else:
+                # Decide what to do if there are other lines in the picking still assigned
+                if split:
+                    unmarked_lines = picking.move_line_ids & move_lines
+                    unmarked_lines._extract_in_split_order(default={"user_id": False})
```
jbaudoux, line 179: printed should also have been removed

```suggestion
                    unmarked_lines._extract_in_split_order(default={"user_id": False, "printed", False})
```

jbaudoux (approved):
Looks good :)

Just a small remaining bug worth fixing while on it

PR #1150, opened 2026-03-31
[16.0][FIX] shopfloor_reception: fix UTC boundaries for "Today" filters

nicolas-delbovier-acsone:
The `_get_today_start_end_datetime` function returned naive datetimes without accounting for the user's local timezone. Since Odoo stores datetimes in UTC, this caused incorrect filtering on the "Pickings to Process Today" screen for users in non-UTC timezones.

Now, the function correctly calculates the start and end of the local day and converts those boundaries to UTC .

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -4,6 +4,8 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 
+from datetime import datetime, time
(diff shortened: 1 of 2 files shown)


lmignon (approved):
LGTM (Code review)

jbaudoux (approved):

jbaudoux:
Tested in shopfloor reception. Looks good

PR #1156, opened 2026-04-16
[16.0][FIX] shopfloor_reception: correct _set_quantity__by_lot signature

nicolas-delbovier-acsone:
### The problem
When a user is in the `set_quantity` state of a Reception and scans a Lot/Serial barcode, the system would raise a 500 error due to an invalid method signature in the reception component.

### Technical Reason
The dispatcher method `_set_quantity__by_barcode` uses the Shopfloor search service to identify the scanned barcode. Once identified as a lot, it calls the corresponding handler:
`handler(picking, selected_line, search_result.record)`

The `search_result.record` provides an Odoo recordset (`stock.lot`). However, the implementation of `_set_quantity__by_lot` was incorrectly expecting a raw string (`barcode` as a string) instead of a recordset. 

This mismatch caused:
1. Failure when accessing attributes like `lot.name`.
2. Inconsistency with other `_set_quantity__by_*` handlers (like product or packaging) which all receive recordsets.

### Changes
Updated `_set_quantity__by_lot(self, picking, selected_line, lot)` to correctly handle the `lot` argument as a `stock.lot` recordset.

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -754,8 +754,8 @@ class Reception(Component):
         selected_line.location_dest_id = location
         return self._response_for_select_move(picking)
 
-    def _set_quantity__by_lot(self, picking, selected_line, barcode):
-        if selected_line.lot_id.name == barcode or selected_line.lot_name == barcode:
+    def _set_quantity__by_lot(self, picking, selected_line, lot):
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
Of course! Thanks

rousseldenis:
@jbaudoux

jbaudoux (approved):

PR #1128, opened 2026-02-03
[16.0][REF] shopfloor_reception/mobile: better set_lot

nicolas-delbovier-acsone:
Previously, the `set_lot` function/endpoint carried too many responsibilities and could be called in different contexts (only a name, only a date, both). Moreover, the "set_lot" state in the shopfloor "reception" scenario called this backend endpoint immediately upon scanning a lot name or picking a date. This "eager" synchronization often resulted in:

- Partial Data: The backend receiving a lot name without an expiry date (or vice versa), which bypassed or complicated validation logic.
- Ghost Records: Useless `stock.lot` records were created in the database even if the user realized they scanned the wrong barcode or made a typo before finishing the step.

### Main Changes

This PR refactors the synchronization logic between the VueJS frontend and the Odoo backend to move from an incremental update pattern to an atomic confirmation pattern. The main changes includes:

- Local Buffering: User inputs (Lot Name and Expiry Date) are now stored locally in the component state before being sent to the backend **on button click**.
- Atomic Call: The backend function responsible for creating the lot is now triggered **only once** when the user clicks the "Continue" button.
- Real-time UI: The item-detail-card is updated reactively to reflect the buffered values, providing immediate visual feedback without touching the database.

### Also includes
- Use ISO format for communicating lot expiration date
- Prevent user from changing an existing expiration date on an existing lot
- Prevent user from creating a lot if the picking type does not allow it
- Skip set_lot screen if lot info has been found from parsing barcode
- Fix button cancel not doing anything in "set_quantity" 
---

This PR is a rework of: https://github.com/OCA/wms/pull/1072

PR #1157, opened 2026-04-17
[16.0][FIX] shopfloor_reception: make sure "set_lot" is skipped on existing lot scan

nicolas-delbovier-acsone:
### Rationale
When a user scans a barcode, Shopfloor is often able to detect the specific move line it corresponds to. Currently, even if the system identifies both the move line and the lot, the user might still be forced to navigate through the "set_lot" screen. 

This manual step feels redundant to the end-user if the data has already been uniquely identified via the scan.

---

Based on: https://github.com/OCA/wms/pull/1128


nicolas-delbovier-acsone:
I put these changes in the original PR

PR #1159, opened 2026-04-21
[16.0][FIX] shopfloor_reception/mobile: set_destination validation flow

nicolas-delbovier-acsone:
- Use `is_dest_location_valid` from shopfloor core instead of duplicating the logic inside custom reception function (same idea as in #1099)
- Fix destination validation not working in shopfloor UI

image


rousseldenis (approved):
Great! Thanks

jbaudoux (approved):

PR #1165, opened 2026-05-12
[16.0][IMP] shopfloor/base: Improve message management

rousseldenis:
This PR will enhance message management in shopfloor stack.

In some situations (like the call to `_is_dest_location_valid()` function, the reason of error is not known exactly depending on extension mechanisms that add logic to existing flows.

So, having a stored message queue in the call stack will allow to add more context to the message for operator.

This will also improve message for location content transfer when location destination is invalid (it shows the source and destination (computed) locations. It also shows the concerned product.

### DIFF
--- a/shopfloor_base/actions/message.py
+++ b/shopfloor_base/actions/message.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
 from odoo import _
 
 from odoo.addons.component.core import Component
@@ -18,6 +19,17 @@ class MessageAction(Component):
     _name = "shopfloor.message.action"
     _inherit = "shopfloor.process.action"
     _usage = "message"
+
+    _message_queue = []
+
+    @property
+    def message_queue(self):
+        return self._message_queue
+
+    def add_message(self, value):
+        if not isinstance(value, str):
+            raise TypeError("You should set a string to message queue!")
+        self._message_queue.append(value)
 
     def generic_record_not_found(self):
         return {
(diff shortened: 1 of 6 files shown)


shopfloor_base/tests/common.py
```diff
@@ -119,10 +119,15 @@ def setUpClass(cls):
         with cls.work_on_actions(cls) as work:
             cls.data = work.component(usage="data")
             cls.data_detail = work.component(usage="data_detail")
-            cls.msg_store = work.component(usage="message")
+            # cls.msg_store = work.component(usage="message")
```
jbaudoux, line 122: Can you look at this?

shopfloor_base/actions/message.py
```diff
@@ -19,6 +27,24 @@ class MessageAction(Component):
...
+            self._message_queue = []
+        return self._message_queue
+
+    def add_message(self, body, message_type="info"):
+        if not isinstance(body, str):
+            raise TypeError("You should set a string to message queue!")
+        if message_type not in MESSAGE_TYPES.keys():
+            raise TypeError("You should use a correct Shopfloor message type!")
+        if not hasattr(self, "_message_queue"):
+            self._message_queue = []
```
nicolas-delbovier-acsone, line 42: This is redundant with what is already inside `def message_queue`.

shopfloor_base/actions/message.py
```diff
@@ -1,9 +1,17 @@
...
 
 from odoo.addons.component.core import Component
 
+MESSAGE_TYPES = {
+    "info": 0,
+    "success": 1,
+    "warning": 2,
+    "error": 3,
+}
+
```
nicolas-delbovier-acsone, line 14: Why not use an `IntEnum` ?
```
from enum import IntEnum
class MessageType(IntEnum):
    INFO = 1
    SUCCESS = 2
    WARNING = 3
    ERROR = 4
```

nicolas-delbovier-acsone (changes requested):

PR #1166, opened 2026-05-15
[IMP] stock_storage_type: Don't retrieve too much values

rousseldenis:
Don't retrieve unneeded values for locations that should not compute 'location_will_contain_product_ids'.

### DIFF
--- a/stock_storage_type/models/stock_location.py
+++ b/stock_storage_type/models/stock_location.py
@@ -171,7 +171,6 @@ class StockLocation(models.Model):
         for location in self:
             if (
                 location.fill_state not in ("empty", "being_emptied")
-                and location._should_compute_will_contain_lot_ids()
                 and len(location.location_will_contain_lot_ids) > 1
             ):
                 locations_with_exception |= location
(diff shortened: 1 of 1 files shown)


stock_storage_type/models/stock_location.py
```diff
@@ -312,7 +310,14 @@ def _should_compute_will_contain_lot_ids(self):
         "do_not_mix_products",
     )
     def _compute_location_will_contain_product_ids(self):
-        for rec in self:
+        # We should not retrieve fields values for records that should not be computed
+        # as when the field is accessed the first time for a record that should be
+        # computed (in 'else'), field value for records of 'self' will be retrieved.
+        # Like 'quant_ids' for locations like 'Customers'
+        prefetch_ids = [
+            rec.id for rec in self if rec._should_compute_will_contain_product_ids()
```
jbaudoux: Do not call it twice

lmignon: > Do not call it twice

Fixed

jbaudoux (approved):

PR #1168, opened 2026-05-29
[16.0][REF] shopfloor_reception: block lot creation with a different expiration date

nicolas-delbovier-acsone:
Change the notification semantic when trying to scan or assign an existing lot with a different expiration date. Previously, this scenario raised a "warning" message type, which was misleading since the system blocks the operation as an error.

Update the message type from "warning" to "error" and simplify the message body.

| Before | After |
|--------|--------|
| image | image |


nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (approved):

PR #1169, opened 2026-05-29
[16.0][REF] shopfloor/shopfloor_reception: global boolean "allow_quantity_exceeding_demand" on all scenarios

nicolas-delbovier-acsone:
Make the `allow_over_reception` field from reception scenario available to all scenarios and rename it to `allow_quantity_exceeding_demand`

Also includes : fix accidentally deleted code (from this PR: https://github.com/OCA/wms/pull/1090)

### DIFF
--- a/shopfloor/models/shopfloor_menu.py
+++ b/shopfloor/models/shopfloor_menu.py
@@ -103,6 +103,15 @@ class ShopfloorMenu(models.Model):
             "triggering a backorder creation with the remaining lines."
         ),
     )
+    allow_quantity_exceeding_demand = fields.Boolean(
+        help="If True, prevents shopfloor error in case you process more qty than "
+        "planned."
+    )
+
+    allow_quantity_exceeding_demand_is_possible = fields.Boolean(
+        compute="_compute_allow_quantity_exceeding_demand_is_possible"
+    )
+
     # TODO: refactor handling of these options.
     # Possible solution:
     # * field should stay on the scenario and get stored in options
@@ -157,8 +166,7 @@ class ShopfloorMenu(models.Model):
         string="Show Get Work on start",
         default=False,
         help=(
-            "When enabled the user will have the option to ask "
-            "for a task to work on."
+            "When enabled the user will have the option to ask for a task to work on."
         ),
     )
     allow_get_work_is_possible = fields.Boolean(
@@ -428,7 +436,7 @@ class ShopfloorMenu(models.Model):
             ):
                 raise exceptions.ValidationError(
                     _(
-                        "Processing reserved quantities is" " not allowed for menu {}."
+                        "Processing reserved quantities is not allowed for menu {}."
                     ).format(menu.name)
                 )
 
@@ -536,6 +544,13 @@ class ShopfloorMenu(models.Model):
         for menu in self:
             menu.require_destination_package_is_possible = menu.scenario_id.has_option(
                 "require_destination_package"
+            )
+
+    @api.depends("scenario_id")
+    def _compute_allow_quantity_exceeding_demand_is_possible(self):
+        for menu in self:
+            menu.allow_quantity_exceeding_demand_is_possible = (
+                menu.scenario_id.has_option("allow_quantity_exceeding_demand")
             )
 
     @api.constrains(
--- a/shopfloor/views/shopfloor_menu.xml
+++ b/shopfloor/views/shopfloor_menu.xml
@@ -13,6 +13,16 @@
             </field>
             <group name="options" position="inside">
                 <group
+                    name="quantity_exceeding_demand"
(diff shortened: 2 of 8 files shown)


nicolas-delbovier-acsone:
Because of some error when rebasing, some code has been lost in here; https://github.com/OCA/wms/commit/813c6da718655550f2796224bec425d725429908#diff-768e3710963536fd21aa34f807f18c34629399f77a1c5a463fa66b332aef5052

image

nicolas-delbovier-acsone:
@jbaudoux

shopfloor_reception/models/shopfloor_menu.py
```diff
@@ -30,6 +33,15 @@ def _compute_filter_today_scheduled_pickings_is_possible(self):
...
             )
 
+    @api.depends("scenario_id")
+    def _compute_allow_over_reception_is_possible(self):
+        for menu in self:
+            menu.allow_over_reception_is_possible = bool(
+                menu.scenario_id.has_option(
+                    "filter_today_scheduled_pickings_is_possible"
+                )
+            )
```
mmequignon: method name and option name doesn't match, this doesn't look right.

mmequignon (changes requested):

shopfloor_reception/models/shopfloor_menu.py
nicolas-delbovier-acsone: Indeed, I'll fix this. Thanks

jbaudoux (commented):
As you are on it, would it be possible to make the option agnostic of reception. Entering a quantity done bigger than the quantity to do is a standard feature valid for any operation type.
What about `allow_quantity_exceeding_demand` ?

nicolas-delbovier-acsone:
> As you are on it, would it be possible to make the option agnostic of reception. Entering a quantity done bigger than the quantity to do is a standard feature valid for any operation type. What about `allow_quantity_exceeding_demand` ?

Good idea, I'll do that

shopfloor/views/shopfloor_menu.xml
```diff
@@ -12,6 +12,9 @@
                 />
             </field>
             <group name="options" position="inside">
+                <group name="quantity_exceeding_demand">
+                    <field name="allow_quantity_exceeding_demand" />
```
jbaudoux, line 23: Don't forget the invisible condition

nicolas-delbovier-acsone: But don't you want this option to be available on all scenarios? Thus I can show it in all menus without a condition, no?

jbaudoux: Available in the sense that it could be enabled in the scenario options :). For now, it's only for reception as that's the only scenario that was tested with this option.

nicolas-delbovier-acsone: Oh, Ok I see. I'll fix this then

shopfloor_reception/data/shopfloor_scenario_data.xml
```diff
@@ -11,7 +11,8 @@
       "auto_post_line": true,
       "allow_return": true,
       "scan_location_or_pack_first": true,
-      "allow_filter_today_scheduled_pickings": true
+      "allow_filter_today_scheduled_pickings": true,
+      "allow_quantity_exceeding_demand_is_possible": true
```
jbaudoux:
```suggestion
      "allow_quantity_exceeding_demand": true
```

jbaudoux (commented):
Can you add the migration script like this one https://github.com/OCA/wms/blob/16.0/shopfloor_reception/migrations/16.0.1.3.0/post-migrate.py ?

lmignon (approved):
LGTM. Thank you for this improvement @nicolas-delbovier-acsone ! (Code review)

jbaudoux (approved):

shopfloor/models/shopfloor_menu.py
```diff
@@ -538,6 +546,13 @@ def _compute_require_destination_package_is_possible(self):
                 "require_destination_package"
             )
 
+    @api.depends("scenario_id")
+    def _compute_allow_quantity_exceeding_demand_is_possible(self):
+        for menu in self:
+            menu.require_destination_package_is_possible = menu.scenario_id.has_option(
```
lmignon: @nicolas-delbovier-acsone Your compute method is wrong.... you assign the value to the wrong field

lmignon (changes requested):

shopfloor/models/shopfloor_menu.py
nicolas-delbovier-acsone: Bad copy-paste my bad. I'll fix it now

nicolas-delbovier-acsone:
@jbaudoux can we merge this ?

PR #1170, opened 2026-06-02
[IMP] shopfloor: Be able to manage several source locations

rousseldenis:
When there are several move lines for the same product but with several source locations, location_content_transfer was confused about which move line to return to the operator as there was a fallback on the picking.

Use directly the move line concerned by the scanned source location.


jbaudoux (commented):
@rousseldenis I'm confused by the TODOs. Is it a PR still draft?
Can you clarify?

PR #1171, opened 2026-06-02
[16.0][ADD] shopfloor_reception_dock/mobile: show docks info in picking details cards

nicolas-delbovier-acsone:
### Rationale

During reception, carrier trucks deposit goods across various physical unloading docks. Because these goods are technically in transit or sitting on an external dock, they have not yet been processed into the system's internal `stock.location`s. Consequently, they are invisible to standard inventory tracking.

---

Depends on https://github.com/OCA/stock-logistics-warehouse/pull/2584

| Select Picking | Select Move |
|--------|--------|
| image | image |


nicolas-delbovier-acsone:
@jbaudoux

rousseldenis:
@nicolas-delbovier-acsone @jbaudoux My only question: Why not avoiding to put another information on that screen that can surcharge it ?

- The current reception list displays only ready ones (with grn in our case)
- Choose a reception
  - Display the Dock and scan it
  - Do the rest of flow
- Scan a Dock
  - Choose a reception (or scan it)

What do you think of ?

nicolas-delbovier-acsone:
> @nicolas-delbovier-acsone @jbaudoux My only question: Why not avoiding to put another information on that screen that can surcharge it ?
> 
>     * The current reception list displays only ready ones (with grn in our case)
> 
>     * Choose a reception
>       
>       * Display the Dock and scan it
>       * Do the rest of flow
> 
>     * Scan a Dock
>       
>       * Choose a reception (or scan it)
> 
> 
> What do you think of ?

Are you suggesting another screen ?

Or are you suggesting that we should only display the dock in the picking card of the "select move" screen and not in the "select picking" screen?

In the later case, I updated the screenshots in the PR description -> you can see that the JS I modified to add the docks info is linked to both screen's picking card so I can not update one without updating the other. I'll need to dive deeper to find a way to decouple those 2 cards info but this seems like more work.

jbaudoux:
> @nicolas-delbovier-acsone @jbaudoux My only question: Why not avoiding to put another information on that screen that can surcharge it ?
> 
> * The current reception list displays only ready ones (with grn in our case)
> * Choose a reception
>   
>   * Display the Dock and scan it
>   * Do the rest of flow
> * Scan a Dock
>   
>   * Choose a reception (or scan it)
> 
> What do you think of ?

IMO. You choose a reception by viewing the dock and going there. Then you scan the SA/GRN, PO or product. No need to scan the dock. You could also scan a dock and filter the list if there are multiple pickings but it should remain an option and not a requirement. That would also fit nicely with the existing flow

jbaudoux (commented):
LG. What about also adding a handler for the dock in the scan document?

nicolas-delbovier-acsone:
@rousseldenis I also do not see in your idea how you deal with multiple docks for a single reception? (in case the supplier delivers in multiple steps) Should we scan all the docks then?

nicolas-delbovier-acsone:
> LG. What about also adding a handler for the dock in the scan document?

Good idea, I'll look for that

jbaudoux (changes requested):
Use singular in module name. It's an OCA convention. It's about feature, no matter the link is 2many.

shopfloor_reception_dock/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Shopfloor Reception Docks",
+    "summary": """Add docks info to shopfloor""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
```
jbaudoux, line 9: Can you add maintainer

jbaudoux:
@nicolas-delbovier-acsone dependency is merged

jbaudoux (approved):

nicolas-delbovier-acsone:
Wait for the merge, there are still issues with the names being plural. I'm fixing it

nicolas-delbovier-acsone:
> Wait for the merge, there are still issues with the names being plural. I'm fixing it

Done, you can merge now

nicolas-delbovier-acsone:
@jbaudoux can you re-try to merge ?

simahawk (approved):

PR #1172, opened 2026-06-02
[16.0][REF] shopfloor_reception: use generic "find" inside "scan_document"

nicolas-delbovier-acsone:
The `scan_document` processing step has been refactored to utilize the `find()` method from the shopfloor search action component. Previously, the logic manually iterated over the handlers and called the internal `_find_record_by_type` method directly, bypassing the broader search pipeline.

By shifting this responsibility to the base `find` method, any standardized barcode preprocessing, custom resolution logic, or GS1-128 parsing injected into the search action component is now fully honored during reception. Furthermore, this provides downstream extending modules with a unified, singular hook point to alter or enrich barcode matching behaviors globally across the app.

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1036,18 +1036,30 @@ class Reception(Component):
         return self._response_for_select_document()
 
     def _scan_document__get_handlers_by_type(self):
-        return {
-            "picking": self._scan_document__by_picking,
-            # only add the handler if scan_location_or_pack_first is disabled
-            "product": (
-                self._scan_document__by_product
-                if not self.work.menu.scan_location_or_pack_first
-                else None
-            ),
-            "packaging": self._scan_document__by_packaging,
-            "lot": self._scan_document__by_lot,
-            "origin_move": self._scan_document__by_origin_move,
-        }
+        # CRITICAL ORDERING:
+        # 1. "origin_move" BEFORE "picking".
+        # -> If a partial delivery occurred, a completed move and an active
+        # backorder picking will share the same Source Document name (e.g., SO001).
+        # Because `search.find()` loops through this dictionary in order, placing
+        # "origin_move" first ensures we trigger a return for the completed delivery
+        # instead of accidentally opening the pending backorder picking.
+        #
+        # 2. "product" BEFORE "lot"
+        # -> as it restricts the lots of the found products.
+        handlers_by_type = {}
+
+        if self.work.menu.allow_return:
+            handlers_by_type["origin_move"] = self._scan_document__by_origin_move
+
+        handlers_by_type["picking"] = self._scan_document__by_picking
+
+        if not self.work.menu.scan_location_or_pack_first:
+            handlers_by_type["product"] = self._scan_document__by_product
+
+        handlers_by_type["packaging"] = self._scan_document__by_packaging
+        handlers_by_type["lot"] = self._scan_document__by_lot
+
+        return handlers_by_type
 
     def _scan_document__get_find_kw(self):
         return {
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1074,17 +1074,22 @@ class Reception(Component):
                         single correspondance. Not tracked product
         """
         handlers_by_type = self._scan_document__get_handlers_by_type()
+
+        if not handlers_by_type:
+            return self._scan_document__fallback()
+
         search = self._actions_for("search")
         find_kw = self._scan_document__get_find_kw()
-        for handler_type, handler in handlers_by_type.items():
-            record = search._find_record_by_type(
-                barcode, handler_type, handler_kw=find_kw
-            )
-            if not record:
-                continue
-            res = handler(record, barcode)
-            if res:
-                return res
+
+        result = search.find(barcode, types=handlers_by_type.keys(), handler_kw=find_kw)
+
+        if result and result.type != "none" and (recordset := result.record):
+            handler = handlers_by_type.get(result.type)
+            if handler:
+                res = handler(recordset, barcode)
+                if res:
+                    return res
+
         return self._scan_document__fallback()
 
     def list_stock_pickings(self):


nicolas-delbovier-acsone:
@jbaudoux

shopfloor_reception/services/reception.py
```diff
@@ -1074,17 +1078,24 @@ def scan_document(self, barcode):
                         single correspondance. Not tracked product
         """
         handlers_by_type = self._scan_document__get_handlers_by_type()
+        # Filter out keys with None handlers (e.g., product if scan_location_or_pack_first)
+        supported_types = [t for t, handler in handlers_by_type.items() if handler]
```
jbaudoux: Shouldn't this be inside `_scan_document__get_handlers_by_type` ?

shopfloor_reception/services/reception.py
```diff
@@ -1074,17 +1078,24 @@ def scan_document(self, barcode):
...
-            res = handler(record, barcode)
-            if res:
-                return res
+
+        result = search.find(barcode, types=supported_types, handler_kw=find_kw)
+
+        if result and result.type != "none" and result.record:
+            handler = handlers_by_type.get(result.type)
+            if handler:
+                res = handler(result.record, barcode)
```
jbaudoux: Maybe this highlights better that the result is a recordset and not a single record

```suggestion
        if result and result.type != "none" and recordset := result.record:
            handler = handlers_by_type.get(result.type)
            if handler:
                res = handler(recordset, barcode)
```

shopfloor_reception/services/reception.py
```diff
@@ -1037,6 +1037,11 @@ def start(self):
 
     def _scan_document__get_handlers_by_type(self):
         return {
+            # ↓ MUST be placed before "picking" so that when `search.find()`
+            # evaluates the supported types sequentially, an exact source document
+            # match (origin) on a done move takes precedence over an "assigned"
+            # backorder picking that shares the same source name.
```
jbaudoux: I don't understand the comment :/

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: Indeed, that is a better place

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: I agree

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: I changed it, is it clearer now?

shopfloor_reception/services/reception.py
```diff
@@ -1036,18 +1036,25 @@ def start(self):
...
-                if not self.work.menu.scan_location_or_pack_first
-                else None
-            ),
             "packaging": self._scan_document__by_packaging,
             "lot": self._scan_document__by_lot,
-            "origin_move": self._scan_document__by_origin_move,
         }
+        # only add the handler if scan_location_or_pack_first is disabled
+        if not self.work.menu.scan_location_or_pack_first:
+            handlers_by_type.update({"product": self._scan_document__by_product})
```
jbaudoux: Product must be before lot as it restricts the lots of the found products.

jbaudoux (changes requested):

shopfloor_reception/services/reception.py
```diff
@@ -1036,18 +1036,25 @@ def start(self):
...
+        # CRITICAL ORDERING FOR THE `search.find()` PIPELINE:
+        # "origin_move" must come BEFORE "picking".
+        #
+        # Why? If a partial delivery occurred, a completed move and an active
+        # backorder picking will share the same Source Document name (e.g., SO001).
+        # Because `search.find()` loops through this dictionary in order, placing
+        # "origin_move" first ensures we trigger a return for the completed delivery
+        # instead of accidentally opening the pending backorder picking.
+        handlers_by_type = {
+            "origin_move": self._scan_document__by_origin_move,
```
jbaudoux: This element should be there only if self.work.menu.allow_return

shopfloor_reception/services/reception.py
jbaudoux: Thanks, got it

shopfloor_reception/services/reception.py
```diff
@@ -1036,18 +1036,17 @@ def start(self):
...
-                self._scan_document__by_product
-                if not self.work.menu.scan_location_or_pack_first
-                else None
-            ),
             "packaging": self._scan_document__by_packaging,
             "lot": self._scan_document__by_lot,
             "origin_move": self._scan_document__by_origin_move,
         }
+        # only add the handler if scan_location_or_pack_first is disabled
+        if not self.work.menu.scan_location_or_pack_first:
```
mmequignon, line 1056: I understand this commit, but I believe you're fixing a non-issue.
And introducing a bug, as @jbaudoux stated in a comment.
We need to keep that order as it is in order to not break current's behavior.
Having if not handler: continue in scan_document is fine to me.
Or return `{key: value for key, values in handlers_by_type.items() if value}` ?

mmequignon (commented):
I agree on the the generic find, not on current's implementation.
This needs a little rework, I believe.

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: I changed the implementation in order to keep both the filtering logic and keep the order of the elements in the dict. 

Do you agree with this new implementation?

nicolas-delbovier-acsone:
@jbaudoux are the changes here OK now? I need this somewhere else

shopfloor_reception/services/reception.py
```diff
@@ -1036,18 +1036,30 @@ def start(self):
...
+        # CRITICAL ORDERING:
+        # 1. "origin_move" BEFORE "picking".
+        # -> If a partial delivery occurred, a completed move and an active
+        # backorder picking will share the same Source Document name (e.g., SO001).
+        # Because `search.find()` loops through this dictionary in order, placing
+        # "origin_move" first ensures we trigger a return for the completed delivery
+        # instead of accidentally opening the pending backorder picking.
+        #
+        # 2. "product" BEFORE "lot"
+        # -> as it restricts the lots of the found products.
```
jbaudoux, line 1048: It now stops at the first match. I would remove that comment line.

jbaudoux:
@nicolas-delbovier-acsone needs rebase

nicolas-delbovier-acsone:
1. It seems that the refactoring in https://github.com/OCA/wms/pull/1174, already made so that the `scan_document` function uses the base `find` method.

2. For the change in `_scan_document__get_handlers_by_type`, as this was not really appreciated:
> I believe you're fixing a non-issue.

I propose that we close this PR

PR #1174, opened 2026-06-03
[16.0][REF] shopfloor: improve `find` action

nicolas-delbovier-acsone:
### Problems
The `generic_find` method in `shopfloor` suffered from different problems:
1. **Redundant Handler Scans:** Even when the parser successfully identified a highly specific barcode type (e.g., a GS1 `'lot'`), the loop structure would blindly execute database queries across *all* other active handlers (e.g., `'package'`, `'product'`, `'location'`).
2. **Suboptimal Order of Execution:** Because the parser sequence injects an `"unknown"` result type (representing the raw string fallback) at the beginning of the results collection, the method consistently evaluated this generic fallback first (and thus tested all possible types instead of only one).
3.  **Bad readibility/Dev experience:** The previous implementation of the `find` method of the `SearchAction` component relied on some `handler_kw` arg to pass down arguments to downstream functions. Now use "fluent builder pattern" to remove the need for this arg.
```python
# Example: Searching for a lot
search_action = self.component(usage="shopfloor.search.action")

# BEFORE
result = search_action.find(
    barcode,
    types=["lot"],
    handler_kw={"lot": {"products": my_products, "limit": limit}},
)

# AFTER
result = (
    search_action
    .for_products(my_products)
    .with_limit(5)
    .find(barcode, types=["lot"])
    )
``` 
4. **Parsing bypass**: The parsing would only happen when calling the base `find` function. All the other "_from_scan" functions were not fed with the parsing results bypassing parsing addons logic such as `shopfloor_gs1`.
For example, when searching a lot, scanning a GS1 with "product" info should narrow down the search of the lot to only those that match the given product. Without passing the parsing info to the `lot_from_scan` function, we have no way to do that.

---

**Also includes**
- Found a file not named properly in `shopfloor_gs1` -> renamed to not be misleading in the future / improve code navigation


nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (commented):
Definitely better but not enough.
I think I explained it in the issue. You need to loop over each result and their type to refine the result.
First result is ean, search for products. Next result is lot, search for lots of found products. Etc...

shopfloor_gs1/actions/__init__.py
```diff
@@ -1 +1 @@
-from . import search
+from . import barcode_parser
```
simahawk, line 1: is this change really needed?

nicolas-delbovier-acsone: I am not against reverting this change if really you prefer so.

I was just applying the "boy scout rule" to bring this file name in line with naming conventions while I was working in the area.

nicolas-delbovier-acsone:
> Definitely better but not enough. I think I explained it in the issue. You need to loop over each result and their type to refine the result. First result is ean, search for products. Next result is lot, search for lots of found products. Etc...

I do not understand how what you say could work? In`generic_find`, once a record is found, the method returns (thus the importance of the ordering of the types). Therefore, there is no way to "refine" results for later searches...

Are you suggesting that we should not return on first result found and always test all possible types first and then decide which record I want to link to the returned `SearchResult` in case of collisions (e.g. lot and package sharing the same name) ? Wouldn't this make the search slower while in most cases such a collision should be rare? How do we decide which record is the best to return in case of multiple records found ?

jbaudoux:
> > Definitely better but not enough. I think I explained it in the issue. You need to loop over each result and their type to refine the result. First result is ean, search for products. Next result is lot, search for lots of found products. Etc...
> 
> I do not understand how what you say could work? In`generic_find`, once a record is found, the method returns (thus the importance of the ordering of the types). Therefore, there is no way to "refine" results for later searches...
> 
> Are you suggesting that we should not return on first result found and always test all possible types first and then decide which record I want to link to the returned `SearchResult` in case of collisions (e.g. lot and package sharing the same name) ? Wouldn't this make the search slower while in most cases such a collision should be rare? How do we decide which record is the best to return in case of multiple records found ?

Yes you need to process all the results and parse according their type. The one to return is the last that made a match. If you match a lot, you return the lot. If you match only a product, you return a product.

shopfloor/actions/search.py
```diff
@@ -76,58 +98,101 @@ def _make_search_result(self, **kwargs):
...
+    def picking_from_scan(self, barcode):
+        res = self.find(barcode, types=["picking"])
+        return res.record if res else self.env["stock.picking"].browse()
+
+    def product_from_scan(self, barcode):
+        res = self.find(barcode, types=["product"])
+        return res.record if res else self.env["product.product"].browse()
+
+    def lot_from_scan(self, barcode):
+        res = self.find(barcode, types=["lot", "serial"])
```
jbaudoux: Keep the existing signature
```suggestion
    def lot_from_scan(self, barcode, products=None):
         if products:
             for_products(products)
        res = self.find(barcode, types=["lot"])
```
and no need for serial, it's calling the same method

shopfloor/services/cluster_picking.py
```diff
@@ -492,7 +492,7 @@ def scan_line(self, picking_batch_id, move_line_id, barcode, sublocation_id=None
                 picking, move_line, packaging, sublocation
             )
 
-        lot = search.lot_from_scan(barcode, products=move_line.product_id)
+        lot = search.for_products(move_line.product_id).lot_from_scan(barcode)
```
jbaudoux: revert this, make the method support the existing signature

shopfloor/services/cluster_picking.py
```diff
@@ -1076,7 +1076,7 @@ def change_pack_lot(self, picking_batch_id, move_line_id, barcode, quantity=None
         response_ok_func = self._response_for_scan_destination
         response_error_func = self._response_for_change_pack_lot
         change_package_lot = self._actions_for("change.package.lot")
-        lot = search.lot_from_scan(barcode, products=move_line.product_id)
+        lot = search.for_products(move_line.product_id).lot_from_scan(barcode)
```
jbaudoux: revert this, make the method support the existing signature

shopfloor/services/location_content_transfer.py
```diff
@@ -560,7 +560,7 @@ def scan_package(self, location_id, package_level_id, barcode):
             else:
                 return self._response_for_scan_destination(location, package_level)
 
-        lot = search.lot_from_scan(barcode, products=package_move_lines.product_id)
+        lot = search.for_products(package_move_lines.product_id).lot_from_scan(barcode)
```
jbaudoux: revert this, make the method support the existing signature

shopfloor/services/zone_picking.py
```diff
@@ -779,7 +779,7 @@ def _scan_source_lot(
         search = self._actions_for("search")
         products = self.env["product.product"].browse(product_id)
         # Could get several lots from different products, check each of them
-        lots = search.lot_from_scan(barcode, products=products, limit=None)
+        lots = search.for_products(products).with_limit(None).lot_from_scan(barcode)
```
jbaudoux: revert this, make the method support the existing signature

shopfloor/services/zone_picking.py
```diff
@@ -1452,7 +1452,7 @@ def change_pack_lot(self, move_line_id, barcode):
         response = None
         change_package_lot = self._actions_for("change.package.lot")
         # handle lot
-        lot = search.lot_from_scan(barcode, products=move_line.product_id)
+        lot = search.for_products(move_line.product_id).lot_from_scan(barcode)
```
jbaudoux: revert this, make the method support the existing signature

jbaudoux (changes requested):

shopfloor/actions/search.py
```diff
@@ -51,21 +50,44 @@ def parser(self):
...
-            "lot": self.lot_from_scan,
-            "serial": self.lot_from_scan,
-            "packaging": self.packaging_from_scan,
-            "delivery_packaging": self.delivery_packaging_from_scan,
-            "origin_move": self.origin_move_from_scan,
+            "product": self._find_product,
+            "package": self._find_package,
+            "picking": self._find_picking,
+            "location": self._find_location,
+            "location_dest": self._find_location,
```
jbaudoux: should get rid of this location_dest, we try to find a location, doesn't matter if source or destination

shopfloor/actions/search.py
```diff
@@ -51,21 +50,44 @@ def parser(self):
...
+            "lot": self._find_lot,
+            "serial": self._find_lot,
```
jbaudoux: lot is enough, no need for serial

shopfloor/actions/search.py
```diff
@@ -51,21 +50,44 @@ def parser(self):
...
+            "packaging": self._find_packaging,
+            "delivery_packaging": self._find_delivery_packaging,
+            "origin_move": self._find_origin_move,
             # Extra data can be contained in barcodes
-            "expiration_date": self.expiration_date_from_scan,
+            "expiration_date": self._find_expiration_date,
```
jbaudoux: that's not returning a record, that should not be available here

shopfloor/actions/search.py
```diff
@@ -76,58 +98,101 @@ def _make_search_result(self, **kwargs):
...
+        res = self.find(barcode, types=["delivery_packaging"])
+        return res.record if res else self.env["stock.package.type"].browse()
+
+    def origin_move_from_scan(self, barcode):
+        res = self.find(barcode, types=["origin_move"])
+        return res.record if res else self.env["stock.move"].browse()
+
+    def expiration_date_from_scan(self, barcode):
+        res = self.find(barcode, types=["expiration_date"])
+        return res.record if res else None
```
jbaudoux: this doesn't make sense to me

jbaudoux (commented):
I would cleanup this as well

shopfloor/actions/search.py
```diff
@@ -148,68 +213,58 @@ def picking_from_scan(self, barcode, use_origin=False):
...
         if not barcode:
             return model.browse()
         domain = [
             ("company_id", "=", self.env.company.id),
             ("name", "=", barcode),
         ]
-        if products:
-            domain.append(("product_id", "in", products.ids))
-        return model.search(domain, limit=limit)
+        if self._products:
```
jbaudoux: If there is a "product" in parse_results, you need to filter on that one. It could be the ean/gtin of a product or packaging

jbaudoux (changes requested):
When you search a lot, you provide the related product (for_product).
Either you scan a lot barcode and you filter on the provided products.
Or you scan a lot gs1 and the product is embedded in the barcode. You then need to filter on that product.
Then returning a lot that doesn't respect the products asked for is questionnable. In reception, you added a test on the returned lot to raise an error saying it's a wrong scanned product. But it's not intuitive that a lot for another product may be returned and other calls may not test that. Maybe that error should be raised here and catched by the caller.

Edit: see https://github.com/OCA/wms/pull/1174/changes#r3535534445. There are several cases where you ask a record for a product and you can receive a record for another product. The caller is then checking the returned record for its product.

shopfloor_reception/services/reception.py
```diff
@@ -1202,34 +1192,35 @@ def scan_lot(self, picking_id, selected_line_id, barcode):
         existing_lot = self.env["stock.lot"]
         lot_expiration_date = None
 
-        search = self._actions_for("search")
+        search = self._actions_for("search").for_products(selected_line.product_id)
         search_result = search.find(
             barcode=barcode,
-            types=["lot", "expiration_date", "product"],
-            handler_kw={"lot": {"products": selected_line.product_id}},
+            types=["lot", "expiration_date"],
```
jbaudoux: expiration_date should be returned in the parse_result if parsed. It's not a record you ask for
```suggestion
            types=["lot"],
```

shopfloor_reception/services/reception.py
```diff
@@ -1202,34 +1192,35 @@ def scan_lot(self, picking_id, selected_line_id, barcode):
...
-                and result.raw != selected_line.product_id.barcode
-            ):
-                return self._response_for_set_lot(
-                    picking,
-                    selected_line,
-                    message=self.msg_store.lot_product_mismatch(),
-                )
+        if (
+            prod_result := search_result.parse_result.get("product")
+        ) and prod_result.value != selected_line.product_id.barcode:
```
jbaudoux: You asked for a lot record, use that. You should be able to do something like
```suggestion
        if search_result.record.product_id != selected_line.product_id:
```

shopfloor_reception/services/reception.py
```diff
@@ -1202,34 +1192,35 @@ def scan_lot(self, picking_id, selected_line_id, barcode):
...
+        if exp_result := search_result.parse_result.get("expiration_date"):
+            lot_expiration_date = self._locale_date_to_datetime_utc(exp_result.value)
 
         if search_result.type == "lot":
             existing_lot = search_result.record
         if not existing_lot:
-            existing_lot = search.lot_from_scan(lot_name, selected_line.product_id)
+            existing_lot = search.for_products(selected_line.product_id).lot_from_scan(
+                lot_name
+            )
```
jbaudoux: There should be no need for a second call

shopfloor_reception/services/reception.py
```diff
@@ -1285,7 +1276,9 @@ def set_lot_confirm_action(
         lot = (
             search_result_record
             if search_result_record and search_result_record._name == "stock.lot"
-            else self._actions_for("search").lot_from_scan(lot_name, product)
+            else self._actions_for("search")
+            .for_products(product)
+            .lot_from_scan(lot_name)
```
```suggestion
            .lot_from_scan(lot_name, products=product)
```

shopfloor_reception/services/reception.py
```diff
@@ -1379,11 +1372,10 @@ def _set_quantity__by_barcode(
         self, picking, selected_line, barcode, confirmation=None
     ):
         handlers_by_type = self._set_quantity__get_handlers_by_type()
-        search = self._actions_for("search")
+        search = self._actions_for("search").for_products(selected_line.product_id)
```
jbaudoux, line 1352: Here there is an example where we ask a product or packaging for a product. The returned record may not correspond to that product and the returned record will be checked for its product in _set_quantity__by_product and _set_quantity__by_packaging.

I think it should be the same then for the lot

See for example
```
        if packaging.product_id.id != selected_line.product_id.id:
            return self._response_for_set_quantity(
                picking,
                selected_line,
                message=self.msg_store.wrong_record(packaging),
            )
```

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: The condition `search_result.record.product_id != selected_line.product_id` will never be True since we added the product on the search above like so `search = self._actions_for("search").for_products(selected_line.product_id)`. Thus, the found lot, if any, will always match the product on the line.

The search above is trying to find an already existing lot in odoo matching the product and the barcode. But the condition here is trying to assert that the barcode (GS1) we scanned is indeed for the currently selected product.

nicolas-delbovier-acsone: See the test `test_scan_lot_wrong_product`  (inside `shopfloor_reception/tests/test_scan_lot.py`) this test would fail with your suggested code.

jbaudoux: That's what is wrong. Look on the search for a product or packaging, even when you say `for_products(selected_line.product_id)`, the result can be a different product or a packaging of a different product. We should do the same for the lot. It could be a lot with a product corresponding to the product in the gs1 barcode.

nicolas-delbovier-acsone: I am not sure to understand what you mean? Would you suggest this:

In case there is both a `products` argument sent to the `lot_from_scan` function as well as a `BarcodeResult` of type "product" found in parsing, you want to refine the search **only** by the parsed result barcode.
Therefore, the parsing is, in a sense, "stronger" than the arguments passed to the function.

Am I understanding your point correctly?

I have this for now:
```python
def _find_lot(self, parse_results, btype="lot"):
    model = self.env["stock.lot"]
    barcode = self._get_parse_results_value(parse_results, btype)
    if not barcode:
        return model.browse()
    domain = [
        ("company_id", "=", self.env.company.id),
        ("name", "=", barcode),
    ]
    if product_result := parse_results.get("product"):
        domain.extend(
            [
                "|",
                # also check for False in case we do not have the barcode
                # for the product yet in odoo
                ("product_id.barcode", "in", [False, product_result.value]),
                ("product_id.default_code", "in", [False, product_result.value]),
            ]
        )

    if self._products:
        domain.append(("product_id", "in", self._products.ids))
    return model.search(domain, limit=self._limit)
```

But if I understood correctly, you would rather do this:

```diff
def _find_lot(self, parse_results, btype="lot"):
    model = self.env["stock.lot"]
    barcode = self._get_parse_results_value(parse_results, btype)
    if not barcode:
        return model.browse()
    domain = [
        ("company_id", "=", self.env.company.id),
        ("name", "=", barcode),
    ]
    if product_result := parse_results.get("product"):
        domain.extend(
            [
                "|",
                # also check for False in case we do not have the barcode
                # for the product yet in odoo
                ("product_id.barcode", "in", [False, product_result.value]),
                ("product_id.default_code", "in", [False, product_result.value]),
            ]
        )

-    if self._products:
+    elif self._products:
        domain.append(("product_id", "in", self._products.ids))
    return model.search(domain, limit=self._limit)
```

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: Indeed

shopfloor_reception/services/reception.py
jbaudoux: yes exactly

shopfloor/actions/search.py
```diff
@@ -148,68 +215,65 @@ def picking_from_scan(self, barcode, use_origin=False):
...
+        if product_result := parse_results.get("product"):
+            domain.extend(
+                [
+                    "|",
+                    # also check for False in case we do not have the barcode
+                    # for the product yet in odoo
+                    ("product_id.barcode", "in", [False, product_result.value]),
+                    ("product_id.default_code", "in", [False, product_result.value]),
+                ]
+            )
```
jbaudoux: Use instead `find(barcode, ["product", "packaging"])` as the ean/gtin can be on the product or on the packaging.

jbaudoux: Ah no need to parse again. Then call `_find_product` and `_find_packaging`

jbaudoux: without limit

shopfloor_reception/services/reception.py
```diff
@@ -1199,37 +1177,37 @@ def scan_lot(self, picking_id, selected_line_id, barcode):
...
         # Look for more info in the barcode
+        existing_lot = search_result.record or self.env["stock.lot"]
+        if (
+            prod_result := search_result.parse_result.get("product")
+        ) and prod_result.value != selected_line.product_id.barcode:
+            return self._response_for_set_lot(
+                picking,
+                selected_line,
+                message=self.msg_store.lot_product_mismatch(),
+            )
```
jbaudoux: This is now useless has InvalidProduct will be raised already in this case, isn't it?

jbaudoux (approved):

shopfloor/services/checkout.py
```diff
@@ -482,29 +483,33 @@ def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=Non
             "delivery_packaging": self._select_lines_from_delivery_packaging,
             "none": self._select_lines_from_none,
         }
-        search_result = self._scan_line_find(picking, barcode, handlers.keys())
+        try:
+            search_result = self._scan_line_find(picking, barcode, handlers.keys())
```
jbaudoux: I would drop `_scan_line_find` and do the find directly here

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: Well the test "test_scan_lot_wrong_product" inside ` /shopfloor_reception/tests/test_scan_lot.py` did not break so I guess this is still useful

shopfloor/services/checkout.py
nicolas-delbovier-acsone: Yes, I hesitaded also.

shopfloor/services/checkout.py
```diff
@@ -482,29 +483,33 @@ def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=Non
...
+            scanned_record = search_result.record
+            scanned_type = search_result.type
+        except InvalidProduct as e:
+            # Leave downstream functions deal with more precise error message
+            scanned_record = e.recordset
+            scanned_type = e.type_
```
jbaudoux: I would do like you did in the other scenario:
```
            return self._response_for_select_line(
                picking,
                message=self.msg_store.wrong_record(e.recordset),
            )
```

nicolas-delbovier-acsone: But then, I'll loose the more refined error messages in this case...

jbaudoux: which one?
I see if the product is right, you can change lot. If there multiple valid lines in packages, you need to tell which package. If you don't scan a product, there no InvalidProduct raised.
All this will remain.

nicolas-delbovier-acsone: This distinction for example:
image

shopfloor_reception/services/reception.py
jbaudoux: The test is mocking find, so you don't test what find is doing. It would now have raised InvalidProduct. The test should call the real find.

shopfloor/services/checkout.py
jbaudoux: ah yes. You could move this in the except. The idea is interresting. If there is an InvalidProduct, it searches another move line reserving the product on the location you are. It's then telling you the transfer reserving the product you scanned.

jbaudoux: The one in _select_lines_from_product (not package) :)

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: > The test is mocking find, so you don't test what find is doing.

Ah yes correct... I'll test using the UI and see what happens

nicolas-delbovier-acsone: The error is not raised because the logic we wrote yesterday is not right.

Here is the idea of the problem I encountered during testing:

image

jbaudoux: idd, need to test also `not products and not packagings`

nicolas-delbovier-acsone: Yes, I have a refactor for this that is a little bit cleaner (not much but still) and that takes this case into account. I'll push everything when tests are green locally.

jbaudoux:
FYI @sebalix @mmequignon @TDu

shopfloor/actions/search.py
```diff
@@ -76,140 +134,212 @@ def _make_search_result(self, **kwargs):
...
             return model.browse()
-        return model.search([("name", "=", barcode)], limit=1)
 
-    def picking_from_scan(self, barcode, use_origin=False):
+        packages = model.search([("name", "=", barcode)], limit=self._limit)
+
+        if packages and self._pickings:
+            invalid_pickings = self._pickings.filtered(
+                lambda p: not (p.move_line_ids.package_id & packages)
+            )
```
jbaudoux: I see no reason to make it different from _find_product or _find_lot

valid_pickings = ...
if not valid_pickings:

shopfloor/actions/search.py
```diff
@@ -76,140 +134,212 @@ def _make_search_result(self, **kwargs):
...
-        return model.search(
-            [
-                "|",
-                ("barcode", "=", barcode),
-                ("default_code", "=", barcode),
-            ],
-            limit=1,
+        products = model.search(
+            ["|", ("barcode", "=", barcode), ("default_code", "=", barcode)],
+            limit=self._limit,
```
jbaudoux: I think the limit should apply only if there is no `self._products`

```suggestion
            limit=None if self._products else self._limit,
```

Then limit the returned products
return products[:self._limit]

jbaudoux (changes requested):

shopfloor/actions/search.py
nicolas-delbovier-acsone: Indeed, nice catch

shopfloor/actions/search.py
```diff
@@ -76,140 +134,211 @@ def _make_search_result(self, **kwargs):
...
+        barcode = self._get_parse_results_value(parse_results, btype)
         if not barcode:
             return model.browse()
-        return model.search([("name", "=", barcode)], limit=1)
 
-    def picking_from_scan(self, barcode, use_origin=False):
+        packages = model.search([("name", "=", barcode)], limit=self._limit)
+
+        if packages and self._pickings:
+            valid_packages = packages & self._pickings.move_line_ids.package_id
```
jbaudoux:
```suggestion
        if packages and self._packages:
            valid_packages = packages & self._packages
```

shopfloor/actions/search.py
```diff
@@ -76,140 +134,211 @@ def _make_search_result(self, **kwargs):
...
+            if not valid_packages:
+                raise SearchInvalidPicking(recordset=packages)
```
jbaudoux:
```suggestion
                raise SearchInvalidPackage(recordset=packages)
```

shopfloor/actions/search.py
```diff
@@ -51,21 +66,64 @@ def parser(self):
...
+        work_context,
+        products=None,
+        pickings=None,
+        limit=1,
+        use_origin=False,
+        extra_domain=None,
+    ):
+        super().__init__(work_context)
+        self._products = products
+        self._pickings = pickings
```
jbaudoux:
```suggestion
        self._packages = packages
```

shopfloor/actions/search.py
```diff
@@ -51,21 +66,64 @@ def parser(self):
...
+        self._limit = limit
+        self._use_origin = use_origin
+        self._extra_domain = extra_domain
+
+    def _get_properties(self):
+        return {
+            "products": self._products,
+            "pickings": self._pickings,
```
jbaudoux:
```suggestion
            "packages": self._packages,
```

shopfloor/services/checkout.py
```diff
@@ -482,7 +483,51 @@ def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=Non
             "delivery_packaging": self._select_lines_from_delivery_packaging,
             "none": self._select_lines_from_none,
         }
-        search_result = self._scan_line_find(picking, barcode, handlers.keys())
+        try:
+            search = self._actions_for("search")
+            search_result = (
+                search.for_products(picking.move_ids.product_id)
+                .for_pickings(picking)
```
jbaudoux:
```suggestion
                .for_packages(picking.move_ids.package_id)
```

shopfloor/services/checkout.py
```diff
@@ -482,7 +483,51 @@ def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=Non
...
+                else:
+                    message = self.msg_store.product_not_found_in_current_picking(
+                        product
+                    )
+                return self._response_for_select_line(picking, message=message)
+
+            else:
+                raise e
+
+        except SearchInvalidPicking as e:
```
jbaudoux:
```suggestion
        except SearchInvalidPackage as e:
```

shopfloor/services/checkout.py
```diff
@@ -482,7 +483,51 @@ def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=Non
...
+            if e.recordset._name == "stock.quant.package":
```
jbaudoux, line 512: Should always be the case

shopfloor/services/checkout.py
```diff
@@ -482,7 +483,51 @@ def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=Non
...
+                        return_picking
+                    )
+                else:
+                    message = self.msg_store.product_not_found_in_current_picking(
+                        product
+                    )
+                return self._response_for_select_line(picking, message=message)
+
+            else:
+                raise e
```
jbaudoux, line 503:
```suggestion
            if e.recordset._name == "product.product":
                product = e.recordset[:1]
            elif e.recordset._name == "product.packaging":
                product = e.recordset[:1].product_id
            else:
                # should not happen
                raise e
            return_picking = self._get_pickings_for_product(product, limit=1)
            if return_picking:
                message = self.msg_store.reserved_for_other_picking_type(
                    return_picking
                )
            else:
                message = self.msg_store.product_not_found_in_current_picking(
                    product
                )
            return self._response_for_select_line(picking, message=message)
```

shopfloor/services/checkout.py
```diff
@@ -482,7 +483,51 @@ def scan_line(self, picking_id, barcode, confirm_pack_all=False, confirm_lot=Non
...
+                search.for_products(picking.move_ids.product_id)
+                .for_pickings(picking)
+                .find(
+                    barcode,
+                    types=handlers.keys(),
+                )
+            )
+        except SearchInvalidProduct as e:
+            if e.recordset._name == "product.product":
+                product = e.recordset
```
jbaudoux: Pay attention in case of error, it's a recordset that is returned.
```suggestion
                product = e.recordset[:1]
```

jbaudoux (commented):
That `for_pickings` is not right. It should be `for_packages`.

jbaudoux (approved):
Great !

lmignon (approved):

PR #1176, opened 2026-06-04
[16.0] [REF] shopfloor_reception: use shopfloor.search.move.line component in scan_document

nicolas-delbovier-acsone:
### Rationale
The "additional_domain" on the menu was respected when displaying the list of pickings but when scanning a picking by name, you could bypass the filter

---

- Reduces code duplication
- Fix scan_document bypassing the "additional_domain" set on the shopfloor menu

### DIFF
--- a/shopfloor/actions/move_line_search.py
+++ b/shopfloor/actions/move_line_search.py
@@ -61,10 +61,11 @@ class MoveLineSearch(Component):
     def _search_move_lines_domain(
         self,
         locations=None,
-        picking_type=None,
-        package=None,
-        product=None,
-        lot=None,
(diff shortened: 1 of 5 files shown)


nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (commented):
You need to do this for all the scan document handlers.
I would rely on search_move_lines action and adapt it to support recordset instead of record as arguments

nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (approved):
Nice cleaning :)

PR #1177, opened 2026-06-09
[16.0] [IMP] stock_storage_type: Speed up potential mix exception columns initialization

lmignon:
Before this commit, the initialization of the potential mix exception columns
on stock_location was done using the ORM, which was very slow on large databases.
This commit add an SQL query to initialize those columns in a more efficient way.

* [ ] includes #1166


lmignon:
ping @rousseldenis @jbaudoux @nicolas-delbovier-acsone

PR #1178, opened 2026-06-10
[16.0][IMP] stock_available_to_promise_release_block: add backorder auto-blocking modes

nicolas-delbovier-acsone:
Replace the boolean "Auto-block Release on Backorders" option by a selection field to support multiple blocking strategies.

Available modes are:

Never, Always, On single OUT move for customer

The new customer-based mode automatically blocks a backorder only when it would become the sole outgoing move for the customer, helping avoid shipping a delivery containing a single late product.


nicolas-delbovier-acsone:
Closed because we had no more need for this change

PR #1180, opened 2026-06-11
[16.0][BK] [18.0] shopfloor: add qty_picked helper methods

lmignon:
Description:
Port from 18.0 to 16.0:
- https://github.com/OCA/stock-logistics-shopfloor/pull/28

These helpers will be used by the back-port of the `shopfloor_single_product_transfer` addon

### DIFF
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -2,7 +2,7 @@
 # Copyright 2025 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from odoo import _, fields
-from odoo.tools.float_utils import float_round
+from odoo.tools.float_utils import float_compare, float_round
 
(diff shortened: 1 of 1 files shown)


shopfloor/actions/stock.py
```diff
@@ -275,3 +275,16 @@ def unload_package(self, lines):
...
+        qty = packaging and packaging.qty or 1
+        move_line.qty_done += qty
+
+    def move_line_check_qty_picked(self, move_line):
+        rounding = move_line.product_id.uom_id.rounding
+        qty_done = move_line.qty_done
+        qty_todo = move_line.reserved_uom_qty
+        # If qty picked is >= qty todo, then there's nothing more to pick
+        if float_compare(qty_done, qty_todo, precision_rounding=rounding) > 0:
+            return False
```
jbaudoux, line 289: I haven't checked where it's used but @nicolas-delbovier-acsone added the option in shopfloor to allow to pick more than planned. Should this option be checked here?

PR #1182, opened 2026-06-15
[16.0][ADD] shopfloor_search_filter: allows to filter which records can be scanned

nicolas-delbovier-acsone:
This module helps prevent warehouse picking mistakes by letting you control exactly what operators are allowed to scan.

When a barcode type is disabled on a menu, the system ignores it during scanning to prevent operators from selecting the wrong resource.

image


nicolas-delbovier-acsone:
Needs to be adapted after refactor: https://github.com/OCA/wms/pull/1174

jbaudoux (commented):
What's the purpose?
The scenario dictates what is supposed to be scanned according to where you are in the steps.

nicolas-delbovier-acsone:
> What's the purpose? The scenario dictates what is supposed to be scanned according to where you are in the steps.

I think I was looking for what is already implemented in the `scan_location_or_pack_first` feature. I'll close this.

PR #1186, opened 2026-06-17
[16.0][IMP] shopfloor_reception: validate product match on multi-attribute barcodes

nicolas-delbovier-acsone:
Prevent users from associating a scanned lot with an incorrect product when processing multi-attribute barcodes (such as GS1-128).

Previously, if a user scanned a GS1 barcode containing both product and lot information, the system would parse the lot string without verifying if the embedded product barcode matched the currently selected line's product. This allowed mismatched lots to be assigned to the wrong products.

### DIFF
--- a/shopfloor_reception/actions/message.py
+++ b/shopfloor_reception/actions/message.py
@@ -42,3 +42,11 @@ class MessageAction(Component):
                 qty=qty,
             ),
         }
+
+    def lot_product_mismatch(self):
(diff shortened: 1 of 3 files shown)


nicolas-delbovier-acsone:
@jbaudoux

lmignon (approved):

jbaudoux (approved):

PR #1187, opened 2026-06-17
[16.0][FIX] shopfloor_reception: ensure search result is a lot in `set_lot_confirm_action`

nicolas-delbovier-acsone:
Prior to this change, if `self.search_result.record` returned a `product.product` instead of a `stock.lot`, the method would fail further down the line when expecting lot-specific attributes

### Reproduce error in UI

You can reproduce the error in shopfloor UI by first installing `shopfloor_gs1`, create a reception picking for some product with this barcode
```
3182550780926
```
And scan this GS1 inside the "select move screen": (GTIN (01)03182550780926(17)281228(10)TestPST04)
```
01031825507809261728122810TestPST04
```
image

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1289,8 +1289,11 @@ class Reception(Component):
             )
 
         product = selected_line.product_id
-        lot = self.search_result.record or self._actions_for("search").lot_from_scan(
-            lot_name, product
+        search_result_record = self.search_result.record
+        lot = (
+            search_result_record
+            if search_result_record and search_result_record._name == "stock.lot"
+            else self._actions_for("search").lot_from_scan(lot_name, product)
         )
 
         if product.use_expiration_date and (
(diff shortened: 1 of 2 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -1276,8 +1276,11 @@ def set_lot_confirm_action(
             )
 
         product = selected_line.product_id
-        lot = self.search_result.record or self._actions_for("search").lot_from_scan(
-            lot_name, product
+        search_result_record = self.search_result.record
```
lmignon, line 1292: @nicolas-delbovier-acsone Can you provide a unittest?

lmignon (changes requested):

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: Here you go

rousseldenis (approved):

nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (approved):

lmignon (approved):

jbaudoux:
Can you rebase?

PR #1190, opened 2026-06-19
[16.0][IMP] stock_release_channel_partner_by_date: do not archive future partner dates on action_sleep

santostelmo:
When putting a release channel to sleep, restrict the deactivation of channel partner dates to only those in the past leaving today's and future configurations active.

- Update action_sleep() in the release channel model to construct an AND domain filtering by date.
- Add test_release_channel_sleep_archive_specific_date_future to ensure future and present dates are untouched.

### DIFF
--- a/stock_release_channel_partner_by_date/models/stock_release_channel.py
+++ b/stock_release_channel_partner_by_date/models/stock_release_channel.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from odoo import fields, models
+from odoo.osv import expression
 
 
 class StockReleaseChannel(models.Model):
@@ -15,12 +16,20 @@ class StockReleaseChannel(models.Model):
     )
 
     def action_sleep(self):
+        shipment_date = self.shipment_date
         res = super().action_sleep()
-        channel_dates = self.env["stock.release.channel.partner.date"].search(
-            self._get_release_channel_partner_date_domain()
-        )
+        domain = self._get_release_channel_partner_date_domain(shipment_date)
+        channel_dates = self.env["stock.release.channel.partner.date"].search(domain)
         channel_dates.write({"active": False})
         return res
 
-    def _get_release_channel_partner_date_domain(self):
-        return [("release_channel_id", "in", self.ids)]
+    def _get_release_channel_partner_date_domain(self, shipment_date=None):
+        domain = [("release_channel_id", "in", self.ids)]
+        if shipment_date:
+            domain = expression.AND(
+                [
+                    domain,
+                    [("date", "<", shipment_date)],
+                ]
+            )
+        return domain
--- a/stock_release_channel_partner_by_date/__manifest__.py
+++ b/stock_release_channel_partner_by_date/__manifest__.py
@@ -13,6 +13,7 @@
     "website": "https://github.com/OCA/wms",
     "depends": [
         "stock_release_channel",
+        "stock_release_channel_shipment_lead_time",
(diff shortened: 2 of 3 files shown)


stock_release_channel_partner_by_date/models/stock_release_channel.py
```diff
@@ -16,9 +17,13 @@ class StockReleaseChannel(models.Model):
 
     def action_sleep(self):
         res = super().action_sleep()
-        channel_dates = self.env["stock.release.channel.partner.date"].search(
-            self._get_release_channel_partner_date_domain()
+        domain = expression.AND(
+            [
+                self._get_release_channel_partner_date_domain(),
+                [("date", "<=", fields.Date.today())],
```
Ricardoalso: Couldn't this be moved into `_get_release_channel_partner_date_domain`?

I'm not sure to what extent this should be configurable at the stock.release.channel level in order to avoid breaking the module's behavior.

santostelmo: I considered that, but `_get_release_channel_partner_date_domain()` is a generic selector for partner-date records related to the channel.
Adding `("date", "<=", today)` there would make the date cutoff part of the base domain everywhere , which would change module behavior outside action_sleep().
In this PR I only want to limit the archival behavior during action_sleep(), so I kept the extra condition local to that method.

jbaudoux: The date is the delivery date. So you should add this only if there is a shipment_date and use that date instead of today. You can add a dependency on `stock_release_channel_shipment_lead_time` for that as there is no base delivery date concept on the stock release channel base module.

santostelmo: @jbaudoux I made the changes as you suggested

santostelmo:
Quick update to keep today's specific date active to allow multiple wake/sleep cycles in the same day.

stock_release_channel_partner_by_date/models/stock_release_channel.py
```diff
@@ -15,10 +16,17 @@ class StockReleaseChannel(models.Model):
     )
 
     def action_sleep(self):
+        shipment_date = self.shipment_date
         res = super().action_sleep()
-        channel_dates = self.env["stock.release.channel.partner.date"].search(
-            self._get_release_channel_partner_date_domain()
-        )
+        domain = self._get_release_channel_partner_date_domain()
+        if shipment_date:
```
jbaudoux, line 28: if there is no shipment date, shouldn't you skip the archival?

jbaudoux: It's fine like that. Would be unusual use case

stock_release_channel_partner_by_date/models/stock_release_channel.py
```diff
@@ -15,10 +16,17 @@ class StockReleaseChannel(models.Model):
...
+            domain = expression.AND(
+                [
+                    domain,
+                    [("date", "<", shipment_date)],
+                ]
+            )
```
jbaudoux, line 34: Can you move it inside `_get_release_channel_partner_date_domain` ?

jbaudoux (approved):

stock_release_channel_partner_by_date/models/stock_release_channel.py
jbaudoux: The method name is maybe a bit misleading as it's similar to the one on stock.picking. This one is used only by action_sleep

PR #1191, opened 2026-06-19
[16.0][FIX] shopfloor_reception: prevent "lot already exists" error when scanning existing lot with GS1

nicolas-delbovier-acsone:
When comparing the `date` from GS1 parsing with the expiration date from the `stock.lot`, python detects a difference because e.g. `datetime(2026, 10, 1) ` and `date(2026, 10, 1) ` are not considered equal in python.

### Bug I am trying to fix

- Install `shopfloor_reception` and `shopfloor_gs1`.
- During the "set_lot", scan a GS1 of a lot that already exists.
=> You get the error message "This lot already exists with expiration date..." (which should not happen since you did not change the expiration date, you scanned the exact date on the lot)

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1223,7 +1223,11 @@ class Reception(Component):
             if result.type == "lot":
                 lot_name = result.value
             elif result.type == "expiration_date":
-                lot_expiration_date = result.value
+                # We need to ensure we have a `datetime` object (and not a
+                # `date` one) for valid comparison with stock.lot.expiration_date
+                lot_expiration_date = datetime.combine(
+                    result.value, datetime.min.time()
+                )
             elif (
                 result.type == "product"
                 and result.raw != selected_line.product_id.barcode
--- a/shopfloor_reception/tests/test_scan_lot.py
+++ b/shopfloor_reception/tests/test_scan_lot.py
@@ -1,10 +1,8 @@
 # Copyright 2026 ACSONE SA/NV
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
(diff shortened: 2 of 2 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -1223,7 +1223,8 @@ def scan_lot(self, picking_id, selected_line_id, barcode):
             if result.type == "lot":
                 lot_name = result.value
             elif result.type == "expiration_date":
-                lot_expiration_date = result.value
+                _date = result.value
+                lot_expiration_date = datetime(_date.year, _date.month, _date.day)
```
mmequignon: I'm sorry, I don't know much about GS1, but can you please explain how dropping minutes/seconds from a datetime fixes a lot already exists error ?

nicolas-delbovier-acsone: This is simply a python problem, `datetime(2026, 10, 1)` and `date(2026, 10, 1)` are not equal

image

I need to convert the `date` back into a `datetime`

nicolas-delbovier-acsone: I will add a comment in the code for that, that'll probably help in the futur

mmequignon: I believe we are usually casting date into datetimes like so
```python
lot_expiration_date = datetime.combine(result.value, datetime.min.time())
```

nicolas-delbovier-acsone:
@jbaudoux

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: Indeed, I'll follow that convention

jbaudoux (approved):

PR #1193, opened 2026-06-22
[16.0][IMP] stock_full_location_reservation: Introduce a strict mode

lmignon:
As in some flows we need to limit the full reservation to the move line characteristics (lot, package, owner), add a strict mode that will use the Odoo core mechanism to gather the needed quant(ities).

Recover work from https://github.com/mt-software-de/wms/pull/6


lmignon:
ping @jbaudoux @rousseldenis

jbaudoux:
merged with https://github.com/OCA/wms/pull/1194

PR #1194, opened 2026-06-22
[16.0][FIX] stock_full_location_reservation: Do not unlink full location reservation moves on cancel

lmignon:
The module stock_full_location_reservation has a feature that creates a
new move for full location reservation when a move is partially reserved.
However, when the original move is canceled, the new move is also unlinked,
which can cause issues if the new move is still needed for other operations.

This commit modifies the behavior of the module to prevent the unlinking of
full location reservation moves when the original move is canceled. Instead,
the deletion of the move is deferred until the end of the transaction,
allowing other operations to complete without losing the move.

* [ ] includes 1193

This is a partial rewrite of https://github.com/mt-software-de/wms/pull/6

### DIFF
--- a/stock_full_location_reservation/models/stock_move.py
+++ b/stock_full_location_reservation/models/stock_move.py
@@ -72,7 +72,10 @@ class StockMove(models.Model):
                 product, qty, location, package
             )
         )
-        if self.picking_type_id.merge_move_for_full_location_reservation:
+        if (
+            self.product_id == product
+            and self.picking_type_id.merge_move_for_full_location_reservation
+        ):
             # To be able to be merged, the new move should use the same source location as
             # the original one.
             new_move.location_id = self.location_id
--- a/stock_full_location_reservation/models/stock_move.py
+++ b/stock_full_location_reservation/models/stock_move.py
@@ -1,6 +1,7 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 import warnings
+from typing import Literal
 
 from odoo import fields, models
 
@@ -91,7 +92,11 @@ class StockMove(models.Model):
             )
         return new_move
 
-    def _full_location_reservation(self, reservation_mode=None, **kwargs):
+    def _full_location_reservation(
+        self,
+        reservation_mode: Literal["strict", "package", "product"] | None = None,
+        **kwargs
+    ):
         if "package_only" in kwargs:
             warnings.warn(
                 "The 'package_only' parameter is deprecated. "
--- a/stock_full_location_reservation/models/stock_move.py
+++ b/stock_full_location_reservation/models/stock_move.py
@@ -1,5 +1,7 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import warnings
+
 from odoo import fields, models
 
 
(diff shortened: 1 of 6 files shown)


lmignon:
@rousseldenis @jbaudoux

stock_full_location_reservation/models/stock_move.py
```diff
@@ -30,7 +43,38 @@ def _undo_full_location_reservation(self):
...
+                    "Failed to unlink full location reservation moves %s",
+                    ids_to_unlink,
+                )
+
+        self.env.cr.postcommit.add(_deferred_unlink)
+
+    @api.autovacuum
+    def _gc_full_location_reservation_moves(self):
+        """This method is meant to be called by a cron and will delete full location
+        eservation moves that are still in cancelled state.
```
jbaudoux:
```suggestion
        reservation moves that are still in canceled state.
```

stock_full_location_reservation/models/stock_move.py
```diff
@@ -30,7 +43,38 @@ def _undo_full_location_reservation(self):
...
+
+        This should normally not be necessary as the moves should be unlinked right
+        after being cancelled into the post commit hook of the transaction,
```
jbaudoux:
```suggestion
        after being canceled into the post commit hook of the transaction,
```

stock_full_location_reservation/models/stock_move_line.py
```diff
@@ -42,23 +42,48 @@ def _get_full_location_reservable_qties(self, package_only=None):
                 ] += qty_available
         return res
 
-    def _full_location_reservation(self, package_only=None):
-        reservable_qties = self._get_full_location_reservable_qties(package_only)
+    def _full_location_reservation(self, strict=False, package_only=None):
```
jbaudoux: Will you add product_only here as well?

jbaudoux (approved):

stock_full_location_reservation/models/stock_move_line.py
lmignon: it's not planned and not part of our needs....

stock_full_location_reservation/models/stock_move.py
```diff
@@ -72,7 +72,10 @@ def _full_location_reservation_create_move(
                 product, qty, location, package
             )
         )
-        if self.picking_type_id.merge_move_for_full_location_reservation:
+        if (
+            self.product_id == product
```
mt-software-de, line 122: Could this change perhaps go into a separate commit?

stock_full_location_reservation/models/stock_move_line.py
```diff
@@ -42,23 +42,48 @@ def _get_full_location_reservable_qties(self, package_only=None):
...
-            reservable_qties[location].pop(package)
+                if not quants:
+                    continue
+
+                total_quantity = 0.0
+                for quant in quants:
+                    total_quantity += quant.available_quantity
+                # We let the core mechanism occur that will reserve
+                # the needed quants
+                line.reserved_uom_qty += total_quantity
```
mt-software-de: Can we maybe add a assert? Because strict=True and package_only=True together will not work. 
Or could we change it to one argument. Like `reservation_source` or `reservation_rule`.
Which could be package, strict and product (@jbaudoux requested implementation)

mt-software-de (changes requested):

lmignon:
@mt-software-de I implemented your requested changes can you update your review?
@jbaudoux I added a reservation_mode for product only

stock_full_location_reservation/models/stock_move_line.py
lmignon: implemented

stock_full_location_reservation/models/stock_move.py
lmignon: done

PR #1196, opened 2026-06-26
[16.0][FIX] shopfloor_reception: fix expiration dates time zone handling

nicolas-delbovier-acsone:
Following a first fix from: https://github.com/OCA/wms/pull/1191

### Reproduce the errors

- Create a db with shopfloor_reception and shopfloor_gs1 installed
- Make the tz of the warehouse's partner to something else than UTC

#### Error 1

Creating a lot from a GS1 scan does shift the time (e.g. you see "2026-12-31 01:00:00" in odoo instead of "2026-12-31 00:00:00")

#### Error 2

- Create a lot by hand for 2026-12-31 00:00:00" through odoo UI
- Scan the corresponding GS1 in the screen 'set_lot'
- You get the "Lot already exists, different expiration date" error tough you should not since the date do match

### DIFF
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -887,8 +887,9 @@ class Reception(Component):
                     result.type == "expiration_date"
                     and line.product_id.use_expiration_date
                 ):
-                    date = result.value
-                    expiration_date = datetime(date.year, date.month, date.day)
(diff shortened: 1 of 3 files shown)


nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (approved):

PR #1197, opened 2026-06-26
[16.0][IMP] shopfloor_reception: clearer error messages in case of lot product mismatch

nicolas-delbovier-acsone:
When scanning a GS1, we parse the GS1 data to ensure the product on the GS1 matches the barcode we have in odoo.

However, when we do not have a barcode for the given product in odoo, the error message was confusing.

| Before | After (barcode not found) | After (barcode mismatch) |
|--------|--------|--------|
| image | image | image |

cc @jbaudoux

### DIFF
--- a/shopfloor_reception/actions/message.py
+++ b/shopfloor_reception/actions/message.py
@@ -43,10 +43,24 @@ class MessageAction(Component):
             ),
         }
 
-    def lot_product_mismatch(self):
+    def lot_product_mismatch(self, line, lot_product):
         return {
             "message_type": "error",
             "body": _(
-                "The scanned lot does not match the selected product",
+                "The scanned lot does not match the selected product. "
+                "Current product is '%(current_product)s' but the scanned lot "
+                "is for '%(lot_product)s'.",
+                current_product=line.product_id.display_name,
+                lot_product=lot_product.display_name,
+            ),
+        }
+
+    def lot_product_not_found(self, product_barcode):
+        return {
+            "message_type": "error",
+            "body": _(
+                "The scanned lot contains the product barcode '%(product_barcode)s' "
+                "but you have no product with this barcode in Odoo.",
+                product_barcode=product_barcode,
             ),
         }
--- a/shopfloor_reception/services/reception.py
+++ b/shopfloor_reception/services/reception.py
@@ -1184,15 +1184,27 @@ class Reception(Component):
                 barcode=barcode,
                 types=["lot"],
             )
-        except SearchInvalidProduct:
+        except SearchInvalidProduct as e:
+            lot_product = e.recordset
             return self._response_for_set_lot(
                 picking,
                 selected_line,
-                message=self.msg_store.lot_product_mismatch(),
+                message=self.msg_store.lot_product_mismatch(selected_line, lot_product),
             )
 
-        # Look for more info in the barcode
         existing_lot = search_result.record or self.env["stock.lot"]
+        if not existing_lot and (
+            product_barcode := search_result.parse_result.get("product")
+        ):
+            product = search.product_from_scan(product_barcode.value)
+            if not product:
+                return self._response_for_set_lot(
+                    picking,
+                    selected_line,
+                    message=self.msg_store.lot_product_not_found(product_barcode.value),
+                )
+
+        # Look for more info in the barcode
         lot_name = barcode
         if lot_result := search_result.parse_result.get("lot"):
             lot_name = lot_result.value
--- a/shopfloor_reception/tests/test_scan_lot.py
+++ b/shopfloor_reception/tests/test_scan_lot.py
@@ -182,7 +182,39 @@ class TestScanLotName(CommonCase):
         self.assert_response(
             res,
             "set_lot",
-            self.msg_store.lot_product_mismatch(),
+            self.msg_store.lot_product_mismatch(
+                self.selected_move_line, lot_other_product.product_id
+            ),
+            data={
+                "picking": self.data.picking(self.picking),
+                "selected_move_line": self._data_for_move_lines(
+                    self.selected_move_line
+                ),
+            },
+        )
+
+    def test_scan_lot_product_not_found(self):
+        lot = self._create_lot(
+            product_id=self.product_b.id, expiration_date=datetime(2022, 7, 2)
+        )
+        unkown_barcode = lot.product_id.barcode
+        with mock.patch.object(BarcodeParser, "parse") as mock_parse:
+            mock_parse.return_value = self._get_gs1_parsing_results(lot)
+            # ↓ Delete the barcode on the product to simulate that we do not have it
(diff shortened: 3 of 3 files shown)


shopfloor_reception/services/reception.py
```diff
@@ -1232,10 +1232,16 @@ def scan_lot(self, picking_id, selected_line_id, barcode):
                 result.type == "product"
                 and result.raw != selected_line.product_id.barcode
             ):
+                message = (
+                    self.msg_store.lot_product_mismatch()
+                    if selected_line.product_id.barcode
+                    else self.msg_store.lot_has_barcode_but_product_missing()
```
jbaudoux: If I have a result of type "product", it means I scanned a barcode that matched an existing product.

If I cannot match a product, I should not enter here, isn't it?

shopfloor_reception/actions/message.py
```diff
@@ -50,3 +50,12 @@ def lot_product_mismatch(self):
                 "The scanned lot does not match the selected product",
             ),
         }
+
+    def lot_has_barcode_but_product_missing(self):
+        return {
+            "message_type": "error",
+            "body": _(
+                "The scanned lot contains a barcode but you have no "
+                "barcode for this product in Odoo.",
```
jbaudoux: Maybe also display the barcode in the message?

```suggestion
                "The scanned lot contains a product barcode but you have no "
                "product with this barcode in Odoo.",
```

shopfloor_reception/tests/test_scan_lot.py
```diff
@@ -198,6 +198,39 @@ def test_scan_lot_wrong_product(self):
             },
         )
 
+    def test_scan_lot_barcode_on_lot_but_not_on_product(self):
+        self.selected_move_line.product_id.sudo().barcode = False
+
+        expiration_date = date(2022, 7, 2)
+        other_product_barcode = "01223334444555"
```
jbaudoux: Can you split in 2 test cases:
* assert there is no product with this barcode => lot_has_barcode_but_product_missing
* assert there is a another product with this barcode => lot_product_mismatch

In both cases, self.lot exists

shopfloor_reception/services/reception.py
jbaudoux: When an expiration date is matched, what is the search_result?
To be sure it's not rerunning a seach while there is already the lot in the results.
The above `lot_name = result.value` is the lot or its barcode? I would expect `result.raw` to be the barcode. Is this redoing a useless second search?

jbaudoux (commented):
My suggestion was to fill-in the screen with the scanned data even if you could not match a product with the barcode. Because the screen to set the product data like the barcode comes after the set lot screen. But we should invert those screens so that you can set the product barcode then you should not face the issue in set lot that the product has no barcode (unless you don't have the module to collect product data).
I think if you cannot confirm it's the wrong product because the scanned product barcode doesn't exist in odoo and the selected line product has no barcode, you should consider the product is right and fill the data.

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: The parsed results do not care about matching a record in DB. Only the `record` field of a SearchResult contains the potentially matched odoo record but the rest of the info is just parsing.

See for example the debug info, if you scan this gs1:
```
(01)20261619202603(17)231212(10)test-lot
```

image

You can see that no record has been found, but the parsing still contains a "product" section.

shopfloor_reception/services/reception.py
nicolas-delbovier-acsone: This is not redoing a second search since we do not handle yet the case of scanning an expiration date: 

image


Here is an example of what looks a search_result in case of expiration date:

image

As you can see `value` and `raw` are the same in this case. I would argue that using "value" is more correct in terms of semantics but would you want that we use the `raw` field instead (to be sure we have the name string and not some other object)?

shopfloor_reception/services/reception.py
jbaudoux: Ah ok, thanks for the clarification.

Shouldn't it then be
```
            elif result.type == "product":
                if not result.record:
                    message = lot_has_barcode_but_product_missing()
                elif result.record != selected_line.product_id:
                    message = self.msg_store.lot_product_mismatch()
```
to look at resolved product only, no matter if it is resolved using product barcode of whatever else

Assuming there is a result type "product" when there is a raw/value for product

nicolas-delbovier-acsone: This is the idea I am going for indeed, but this is a bit more complex than that. There is no record for each parsed result, only one record for the whole `SearchResult` object, see:
image

The `find` function will return the first record it finds depending on the order of the types we pass in arguments

image


Since we search in this order: ["lot", "expiration_date", "product"], we must also cover the case of search_result type not "product" but not "none" either (e.g. scan the gs1 of an existing lot in odoo -> search result type is "lot" and no search on "product.product" has been made then). In which case we need to make a search on the product by hand to ensure there is no matching product in odoo.

jbaudoux (approved):

jbaudoux (commented):
> When a GS1 barcode is parsed (e.g., Application Identifier 01 for GTIN), the underlying parser strips leading zeros from the interpreted `result.value` (e.g., '05703188036063' becomes '5703188036063').

But the GTIN is 5703188036063. It's because it's coded on 14 characters that there is a leading 0 in the GS1. In odoo, it's expected to be 5703188036063. So it is currently correct, isn't it?

nicolas-delbovier-acsone:
> In odoo, it's expected to be 5703188036063. 

I did not know what is expected to be in odoo. I had this example form our client's data. Maybe its a data problem then...

I'll rollback this change.

nicolas-delbovier-acsone:
Needs to be adapted after refactor: https://github.com/OCA/wms/pull/1174

shopfloor_reception/services/reception.py
```diff
@@ -1184,15 +1184,27 @@ def scan_lot(self, picking_id, selected_line_id, barcode):
...
+        if not existing_lot and (
+            product_barcode := search_result.parse_result.get("product")
+        ):
+            product = search.product_from_scan(product_barcode.value)
+            if not product:
+                return self._response_for_set_lot(
+                    picking,
+                    selected_line,
+                    message=self.msg_store.lot_product_not_found(product_barcode.value),
+                )
```
nicolas-delbovier-acsone, line 1205: Good idea indeed, I'll do that

jbaudoux:
> Good idea indeed, I'll do that

Did I posted something ? Oups...
It's good like you did as we still need to fill the field with the lot in case no lot is found.
Cannot be done like in the other scenario

nicolas-delbovier-acsone:
> > Good idea indeed, I'll do that
> 
> Did I posted something ? Oups... It's good like you did as we still need to fill the field with the lot in case no lot is found. Cannot be done like in the other scenario

OK I do not change then

PR #1199, opened 2026-06-29
[16.0][IMP] shopfloor:  show picking type details in shopfloor menu form

nicolas-delbovier-acsone:
In some places in shopfloor, we use configuration options from the picking type. It would make the UX smoother if these options could be activated directly from the shopfloor menu instead of having to navigate to a separate window.

image

#### Other example of where this would make sense

- `shopfloor_reception_helpdesk` (https://github.com/OCA/wms/pull/1110):
  - allow_helpdesk_ticket
  - default_helpdesk_team_id

_(Don't hesitate to suggest other fields)_

cc @jbaudoux


jbaudoux (commented):
LGTM

Maybe add:
* warehouse (if group is enabled)
* location source
* location destination
* shopfloor_zero_check (shopfloor option on the picking type)
* auto_batch (module shopfloor_batch_automatic_creation)

jbaudoux (approved):

PR #1200, opened 2026-06-29
[16.0][IMP] shopfloor/reception_mobile: show product pictures in product details

nicolas-delbovier-acsone:
### 1. Show product image in product details

image

### 2. Centralize product detail card options

Centralize the rendering options in all these screens (avoid code duplication)

| set lot | set quantity | set detination |
|--------|--------|--------|
| image | image | image | 

cc @jbaudoux

### DIFF
--- a/shopfloor_reception_mobile/static/src/scenario/reception.js
+++ b/shopfloor_reception_mobile/static/src/scenario/reception.js
@@ -21,14 +21,6 @@ const Reception = {
                 v-on:found="on_scan"
                 :input_placeholder="search_input_placeholder"
             />
-            <template v-if="state_is('select_move')">
-                <item-detail-card
-                    :record="state.data.picking"
-                    :options="operation_options()"
-                    :card_color="utils.colors.color_for('screen_step_done')"
-                    :key="make_state_component_key(['reception-picking-item-detail', state.data.picking.id])"
-                />
-            </template>
             <template v-if="state_is('select_document')">
                 <manual-select
                     class="with-progress-bar"
@@ -66,6 +58,12 @@ const Reception = {
                 </div>
             </template>
             <template v-if="state_is('select_move')">
+                <item-detail-card
+                    :record="state.data.picking"
+                    :options="operation_options()"
+                    :card_color="utils.colors.color_for('screen_step_done')"
+                    :key="make_state_component_key(['reception-picking-item-detail', state.data.picking.id])"
+                />
                 <manual-select
                     :card_color="utils.colors.color_for('screen_step_done')"
                     :records="ordered_moves"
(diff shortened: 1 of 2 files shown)


rousseldenis (approved):
LGTM

lmignon (approved):

shopfloor_reception_mobile/static/src/scenario/reception.js
```diff
@@ -377,31 +375,12 @@ const Reception = {
...
-                        label: "Vendor code",
-                    },
-                    {path: "lot.name", label: "Lot"},
-                    {
-                        path: "lot.expiration_date",
-                        label: "Expiry date",
-                        renderer: (rec, field) => {
-                            return this.utils.display.render_field_date(rec, field);
-                        },
+                        path: "package_dest.name",
```
jbaudoux, line 379: This looks strange to me to have the result package in the product detail (which is the source product).

In the set destination, we see the package "Bin 1" in a separate box. I'm confused. Can you clarify?

jbaudoux (commented):
Can you split in 2 PR for the image and for the reception cleanup. It's 2 different topics and on 2 different modules.

Do you see the image while picking or only with the info button?

nicolas-delbovier-acsone:
> Do you see the image while picking or only with the info button?

Only when inside the details of the product (when hitting the little 'i' icon on a product card)

shopfloor_reception_mobile/static/src/scenario/reception.js
nicolas-delbovier-acsone: It was just because it existed previously (but only in the "set_destination" product card though):

image

jbaudoux: yes but there is no need for it as it's also displayed in a separate box. By the way, that box "Bin 1" with the destination package should be in green and not in yellow

nicolas-delbovier-acsone: > yes but there is no need for it as it's also displayed in a separate box

OK, I'll remove this in the separate PR

>  By the way, that box "Bin 1" with the destination package should be in green and not in yellow

I am not sure, the green is for stuff that we already scanned I believe. Here the destination has not yet been scanned.

nicolas-delbovier-acsone:
Closed in favor of https://github.com/OCA/wms/pull/1203 et https://github.com/OCA/wms/pull/1204

shopfloor_reception_mobile/static/src/scenario/reception.js
nicolas-delbovier-acsone: > In the set destination, we see the package "Bin 1" in a separate box. I'm confused. Can you clarify?

For me this screen is simply showing you the product you've been working with so far in one box and in another box showing you the expected destination to scan (separate box because this is related to a separate record).

PR #1201, opened 2026-06-29
[16.0][IMP] shopfloor_mobile: parse note html in detail transfer

nicolas-delbovier-acsone:
image

| Before | After |
|--------|--------|
| image | image |

cc @jbaudoux


jbaudoux (approved):
It's better but doesn't work properly in case of multiple lines in the note

PR #1202, opened 2026-06-29
[16.0] shopfloor: fix unmark

lmignon:
don't reset lot on unmark_move_line_as_picked
    
We don't want to reset the lot when unmarking a move line as picked, as it
can lead to loss of information especially it will reset the reservation
if the product is tracked by lot.

### DIFF
--- a/shopfloor/actions/stock.py
+++ b/shopfloor/actions/stock.py
@@ -159,7 +159,6 @@ class StockAction(Component):
                 "shopfloor_user_id": False,
                 "qty_done": 0,
                 "result_package_id": False,
-                "lot_id": False,
             }
         )
         pickings = move_lines.picking_id
--- a/shopfloor/tests/test_actions_stock.py
+++ b/shopfloor/tests/test_actions_stock.py
@@ -14,12 +14,13 @@ class TestActionsStock(CommonCase):
         super().setUpClass()
         with cls.work_on_actions(cls) as work:
             cls.stock = work.component(usage="stock")
+        cls.product_a.tracking = "lot"
         cls.picking = cls._create_picking(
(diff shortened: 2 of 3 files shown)


nicolas-delbovier-acsone:
If we do not reset the lot, there is no way to properly cancel all data from a line in shopfloor reception (no way to get back to 'set_lot' screen for the given line).

If the operator scanned a wrong lot, there is then no way for him to change this invalid data from shopfloor app. He will have to go back to odoo...

@jbaudoux what do you think?

**EDIT**: I've just seen you also added a fix in shopfloor_reception for the case I am mentioning, my bad

jbaudoux (approved):

shopfloor/actions/stock.py
```diff
@@ -159,7 +159,6 @@ def unmark_move_line_as_picked(self, move_lines, split=True):
                 "shopfloor_user_id": False,
                 "qty_done": 0,
                 "result_package_id": False,
-                "lot_id": False,
```
jbaudoux, line 162: Another approach would be to do that only on move lines where `move_id._should_bypass_reservation()`

jbaudoux:
> If we do not reset the lot, there is no way to properly cancel all data from a line in shopfloor (no way to get back to 'set_lot' screen for the given line).
> 
> If the operator scanned a wrong lot, there is then no way for him to change this invalid data from shopfloor app. He will have to go back to odoo...
> 
> @jbaudoux what do you think?

@nicolas-delbovier-acsone You recently added it for reception because there is no reservation and the operator sets the lot. For the other scenario where is a reservation, the reserved lot should never be lost.

nicolas-delbovier-acsone (approved):

PR #1203, opened 2026-06-30
[16.0][IMP] shopfloor_mobile: show product image in product details

nicolas-delbovier-acsone:
Split part from: https://github.com/OCA/wms/pull/1200

image

cc @jbaudoux

### DIFF
--- a/shopfloor_mobile/static/wms/src/components/detail/detail_product.js
+++ b/shopfloor_mobile/static/wms/src/components/detail/detail_product.js
@@ -6,6 +6,29 @@
 
 import {ItemDetailMixin} from "/shopfloor_mobile_base/static/wms/src/components/detail/detail_mixin.js";
 
+Vue.component("detail-product-image", {
+    props: ["record"],
+    template: `
+        <div v-if="record.image" class="detail-product-image-wrapper my-3 d-flex justify-center">
+            <v-img
+                :src="record.image"
+                max-width="256"
+                contain
+                class="rounded elevation-1"
+            >
+                <template #placeholder>
+                    <div class="d-flex align-center justify-center fill-height">
+                        <v-progress-circular
+                            color="grey-lighten-4"
+                            indeterminate
+                        />
+                    </div>
+                </template>
+            </v-img>
+        </div>
+    `,
+});
+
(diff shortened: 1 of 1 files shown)


shopfloor_mobile/static/wms/src/components/detail/detail_product.js
```diff
@@ -6,6 +6,31 @@
...
 
+Vue.component("detail-product-image", {
+    props: ["record"],
+    template: `
+        <div v-if="record.image" class="detail-product-image-wrapper my-3 d-flex justify-center">
+            <v-img
+                :src="record.image"
+                max-height="128"
+                max-width="128"
+                aspect-ratio="1"
```
jbaudoux: I don't think this is doing what you expect.
aspect-ratio 1 will force to have same width and same height which is not what you want

You cannot combine max height and max with, as soon as it exceeds both, it will squeeze without respecting the scale

128 is very small

I would only set:
```suggestion
                max-width="256"
```

shopfloor_mobile/static/wms/src/components/detail/detail_product.js
```diff
@@ -6,6 +6,31 @@
...
+                contain
+                class="rounded elevation-1"
```
jbaudoux, line 17: why rounded?

jbaudoux (commented):
I'm not fan of having the picture inside the box. I find it renders better after the box like it was done in the cluster picking

Image

shopfloor_mobile/static/wms/src/components/detail/detail_product.js
nicolas-delbovier-acsone: A purely aesthetic choice

shopfloor_mobile/static/wms/src/components/detail/detail_product.js
nicolas-delbovier-acsone: OK, I'll fix this thx for the clarification

jbaudoux (approved):

nicolas-delbovier-acsone:
> I'm not fan of having the picture inside the box. I find it renders better after the box like it was done in the cluster picking

See https://github.com/OCA/wms/pull/1210 for a picture that is out side the detailed card in the 'set_lot' state

PR #1207, opened 2026-07-01
[16.0][IMP] shopfloor_reception_packaging_dimension/mobile: enable to create a new packaging

nicolas-delbovier-acsone:
### Rationale

When warehouse operators receive goods in a packaging they do not have yet in odoo, it makes sense for them to create a new packging so that they can simply scan **one** packaging instead of each product individually.

### Implementation ideas
- Extract the product packaging form in a separate Vue component so that it can be reused
- Extend the component to also accept to modify the name of the packaging

image

| set quantity | create new packaging |
|--------|--------|
| image| image | 

cc @jbaudoux


nicolas-delbovier-acsone:
We should not create the packaging directly, we should wait until we hit "done" or at least have a "cancel" button to delete the created packaging. Otherwise, we'll surely create useless ghosts packagings due to misclicks

jbaudoux:
> We should not create the packaging directly, we should wait until we hit "done" or at least have a "cancel" button to delete the created packaging. Otherwise, we'll surely create useless ghosts packagings due to misclicks

Then maybe rename the button to "DECLARE NEW PACKAGING" instead of "CREATE NEW PACKAGING".
WDYT?

nicolas-delbovier-acsone:
> Then maybe rename the button to "DECLARE NEW PACKAGING" instead of "CREATE NEW PACKAGING". WDYT?

I decided to follow the "cancel" button implementation as this was easier. Therefore, in current implementation, "CREATE NEW PACKAGING" is valid as it is indeed creating a new packaging as soon as you hit the button

shopfloor_reception_packaging_dimension/services/reception.py
```diff
@@ -149,6 +149,62 @@ def _update_packaging_dimension(self, packaging, dimensions_to_update):
...
+        line = self.env["stock.move.line"].browse(selected_line_id)
+        packaging = (
+            self.env["product.packaging"]
+            .sudo()
+            .create(
+                {
+                    "name": "New Packaging (from Shopfloor)",
+                    "product_id": line.product_id.id,
+                }
+            )
```
jbaudoux, line 191: At this stage, the quantity (it cannot be 0 anymore in newer versions) and the level (tells what fields are required in the dimension screen) should be set.

This packaging will be immediately available in the catalog, in sales, on webshop, on reports. Having later the dimensions/barcode is ok, but the basic data should be set at creation.

nicolas-delbovier-acsone: > it cannot be 0 anymore in newer versions

The default is 1 already:

image

---

> the level (...) should be set

Done

nicolas-delbovier-acsone:
I'll close this and create a new simpler module that focuses on creating a new packaging

PR #1210, opened 2026-07-02
[16.0][IMP] shopfloor_reception/mobile: show product image in 'set_lot'

nicolas-delbovier-acsone:
image

cc @jbaudoux

### DIFF
--- a/shopfloor_reception/actions/data.py
+++ b/shopfloor_reception/actions/data.py
@@ -18,7 +18,7 @@ class DataAction(Component):
         data_detail_action = self._actions_for("data_detail")
         return res + [
             "use_expiration_date",
-            ("image_256:image", data_detail_action._product_image_url),
+            ("image_128:image", data_detail_action._product_image_url),
         ]
 
     @property
--- a/shopfloor_reception/actions/data.py
+++ b/shopfloor_reception/actions/data.py
@@ -15,7 +15,11 @@ class DataAction(Component):
         parser functions. We use *args to capture them.
         """
         res = super(DataAction, self)._product_parser
-        return res + ["use_expiration_date"]
+        data_detail_action = self._actions_for("data_detail")
+        return res + [
+            "use_expiration_date",
+            ("image_256:image", data_detail_action._product_image_url),
+        ]
 
     @property
     def _lot_parser_reception(self):
--- a/shopfloor_reception/actions/schema.py
+++ b/shopfloor_reception/actions/schema.py
@@ -15,6 +15,7 @@ class ShopfloorSchemaAction(Component):
                     "nullable": True,
                     "required": False,
                 },
+                "image": {"type": "string", "nullable": True, "required": False},
             }
         )
         return res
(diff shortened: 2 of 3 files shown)


shopfloor_reception/actions/data.py
```diff
@@ -15,7 +15,11 @@ def _product_parser(self):
         parser functions. We use *args to capture them.
         """
         res = super(DataAction, self)._product_parser
-        return res + ["use_expiration_date"]
+        data_detail_action = self._actions_for("data_detail")
+        return res + [
+            "use_expiration_date",
+            ("image_128:image", data_detail_action._product_image_url),
```
jbaudoux, line 21: You should use image_256 with max-width=256

jbaudoux (changes requested):

jbaudoux (approved):

jbaudoux (commented):
I don't know if it's a good idea to always display the image here.
The initial idea was to confirm it's the right product in case you select the line to receive instead of scanning its barcode.
However, now, if the barcode is missing, the product/packaging barcode/dimension screen is shown first to fill the barcode (if you enabled the option).

Ideally, the image should be show first so that the operator can confirm it's the right product. Especially if the operator will start filling product/packaging data. We saw a case where the operator was processing the wrong product. Maybe an intermediate screen to add.

Besides that, regarding this specific PR, maybe it's not bad to always show the image on this set lot step.
WDYT @sebalix @mt-software-de @Highcooley

mt-software-de:
> I don't know if it's a good idea to always display the image here. The initial idea was to confirm it's the right product in case you select the line to receive instead of scanning its barcode. However, now, if the barcode is missing, the product/packaging barcode/dimension screen is shown first to fill the barcode (if you enabled the option).
> 
> Ideally, the image should be show first so that the operator can confirm it's the right product. Especially if the operator will start filling product/packaging data. We saw a case where the operator was processing the wrong product. Maybe an intermediate screen to add.
> 
> Besides that, regarding this specific PR, maybe it's not bad to always show the image on this set lot step. WDYT @sebalix @mt-software-de @Highcooley

@jbaudoux thx for the ping, i don't really know if it is a bad idea to show it, i would make it configurable. 

But i think it is a bad idea, to always load the image in every call of the `_product_parser` this parser is used in many scenarios. If I understand correctly `_product_image_url` it always loads the image to simply return the url. See https://github.com/OCA/wms/blob/ac6afb43c0dd72c518b6d1e006766f03fcf686f8/shopfloor/actions/data_detail.py#L212

I was facing in the past already performance issues, because we are loading to much data which isn't used at all in a scenario. I think a dedicated parser is better.

sebalix:
I agree that always loading images could introduce latency (especially when images are stored outside of Odoo like in a cloud bucket, adding more I/O).
We could instead return the image URL instead of blob, and let the client renders it asynchronously. The drawback is to generate too much 404 errors if products have no image.

simahawk:
> I agree that always loading images could introduce latency (especially when images are stored outside of Odoo like in a cloud bucket, adding more I/O). We could instead return the image URL instead of blob, and let the client renders it asynchronously. The drawback is to generate too much 404 errors if products have no image.

If products have no image, no URL should be generated/returned.

sebalix:
> If products have no image, no URL should be generated/returned.

That would be perfect, but Odoo needs to retrieve the blob (slow) to know if there is an image... and we cannot do a `IS NOT NULL`  SQL request easily as such request will heavily depend how images are stored in Odoo.

What about a dedicated endpoint proposed by Shopfloor which returns the product image or a placeholder/blank pixel if there is none? And we always return this URL.

simahawk:
can we use the `length` fuction at SQL level?

PR #1213, opened 2026-07-07
[16.0][FIX] shopfloor_reception: invalid field dependency on "filter_today_scheduled_pickings" constraint

nicolas-delbovier-acsone:


jbaudoux (commented):
But then it's a data issue, isn't it?
When you disable the option on the scenario, you need to disable it on the menu as well.

Why it doesn't depend only on scenario_id ?

nicolas-delbovier-acsone:
> But then it's a data issue, isn't it? When you disable the option on the scenario, you need to disable it on the menu as well.

Would you prefer that we change the field to `False` when the config option is deleted?

> Why it doesn't depend only on scenario_id ?

I do not know, this is not my code. It comes from this PR: https://github.com/OCA/wms/pull/1013.

I am just trying to fix an issue I am facing. To me, this check does not make any sense (because of the explained scenario in PR description).

shopfloor_reception/models/shopfloor_menu.py
```diff
@@ -31,16 +31,3 @@ def onchange_filter_today_scheduled_pickings_is_possible(self):
         self.filter_today_scheduled_pickings = (
             self.filter_today_scheduled_pickings_is_possible
         )
-
-    @api.constrains("scenario_id", "picking_type_ids", "allow_move_create")
```
jbaudoux, line 35: There was here a wrong copy/paste

It should be
```
    @api.constrains("scenario_id", "picking_type_ids", "filter_today_scheduled_pickings")
```

jbaudoux:
> > But then it's a data issue, isn't it? When you disable the option on the scenario, you need to disable it on the menu as well.
> 
> Would you prefer that we change the field to `False` when the config option is deleted?

The option cannot be deleted unless there is a migration script that sets the option to False and that script should then also set it to False on the menu using that scenario.
I would say just fix your data locally and it should be good as your described use case is not something you are supposed to do in the interface.

shopfloor_reception/models/shopfloor_menu.py
nicolas-delbovier-acsone: Fixed

nicolas-delbovier-acsone:
>I would say just fix your data locally and it should be good as your described use case is not something you are supposed to do in the interface.

I changed the PR to only fix the `api.constraint` error then.

jbaudoux (approved):

PR #1214, opened 2026-07-07
[16.0][ADD] shopfloor_fs_product_multi_image: glue module fixing incompatibility with fs_product_multi_image

nicolas-delbovier-acsone:
### Rationale

When the module `fs_product_multi_image` is installed, the field representing the image on a product is `image`. However, in base odoo (and thus in shopfloor), the base field for an image is `image_1920` (and all the other `image_x` that are related to it).

If one installs `fs_product_multi_image` and starts changing the image in odoo, the change would not be visible in shopfloor images since shopfloor is not targetting the correct field anymore. (The user would see the previous old image from before the installation of `fs_product_multi_image` module)

cc @jbaudoux 

---

Following:
- https://github.com/OCA/wms/pull/1210 (based from it -> to be rebased)
- https://github.com/OCA/wms/pull/1203


lmignon:
This one is no more required thanks to https://github.com/OCA/storage/pull/652

PR #1216, opened 2026-07-12
[14.0][IMP] shopfloor, cluster picking: add option 'Pick by product'

sebalix:
Instead of processing move lines one by one, the operator works product by product.
When confirmed, the batch shows the first product to pick. After scanning the product and a destination bin, all its lines in the batch are processed at once (qty done and destination bin are set). Continues to the next product until all are completed.
Only applies to lines without a source package and not tracked by lot/serial.

New `start_product` screen:

image

Followed by `scan_product_destination` screen:

image

Then if other move lines with a source package or tracked by lot/serial need to be processed, it'll loop on usual `start_line` -> `scan_destination` screens.

ping @jbaudoux


jbaudoux (commented):
That's a lot of complexity added for something that won't work properly if you have several source locations, source packages or lots.

I would have solved that in the backend. There are already 2 solutions you know:
* RAU: delivery pull flow with pick. When pulling the pick, you set a fixed procurement group by product. We did a module in v14 to auto generate the procurement group
* COS: delivery pull flow without pick (e.g. pre-pick kardex). You don't generate the pick but instead have an orderpoint. You can use stock_location_orderpoint to have one rule for all products.

In both cases, you have all the moves for a product in the same picking. And it works in case of multiple source locations, packages or lot as you still process the move lines the normal way.

Then depending on how you want to work, you use cluster picking (if you want to have a batch with a defined amount of products), zone picking (no get work, operator is free to choose), location content transfer (one product, multiple destinations) or single product transfer (one move line at a time).

sebalix:
In our case we are not in multi-steps flow here (and it's a constraint, too complicated for the kind of users/WH size we address), we process directly deliveries that could be OUT or internal transfers, and our customer wants to keep each OUT/delivery recipes, but regarding the pick they don't really care about the source locations (it's often /Stock), they care about the `(batch, product)` group for most of the lines to process (ex: roll of fabric, in meters).

Lines with source packages or lots are still processed the same way, even with the option enabled, no change regarding them.

The other solution I had in mind was to put this in an entire new scenario if you think it's cluttering the current one, but in case we have to ship one product tracked by lot we'll have to process them with another scenario/picking type, that's not simple.

EDIT: lines without source package and lot are now grouped by `(product, source location)`.

PR #1219, opened 2026-07-15
[14.0][IMP] shopfloor_mobile[_base]: add FR translations (mainly cluster.picking)

sebalix:

PR #1222, opened 2026-07-20
[14.0][IMP] shopfloor_mobile: add FR translations

sebalix:
Both for `batch-picking-detail` and `detail-transfer` components.


sebalix:
@jbaudoux to complete #1219 for translations located in sub-components used by the cluster picking scenario.

jbaudoux (approved):
cc @nicolas-delbovier-acsone

PR #1223, opened 2026-07-22
[14.0][IMP] shopfloor, picking batch parser: use `display_name` instead of `name`

sebalix:
This allows to inject useful informations in the list of batches for users in the cluster picking scenario for instance, without introducing new overrides and dependencies on Shopfloor.

Our use case is to display a shipping address field we added on the batch and injected in its `display_name` if there is one.

image

image

### DIFF
--- a/shopfloor/actions/data.py
+++ b/shopfloor/actions/data.py
@@ -335,7 +335,13 @@ class DataAction(Component):
 
     @property
     def _picking_batch_parser(self):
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):
LGTM

sebalix:
@jbaudoux if you agree, I would gladly merge that little one (and port it to 16.0 + 18.0 soon after)

simahawk (approved):

PR #1225, opened 2026-07-22
[IMP] shopfloor: Allow to partially reserve in location_content_transfer

rousseldenis:
When several users are working on a content transfer (e.g.: put away from input locations), scanning the source location will result in an error as some products are already reserved.

Improve also the message for the operator

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -911,7 +911,10 @@ class MessageAction(Component):
     def new_move_lines_not_assigned(self):
         return {
             "message_type": "error",
-            "body": _("New move lines cannot be assigned: canceled."),
+            "body": _(
+                "Some products in the source location are already reserved. "
+                "It's impossible to reserve them!"
+            ),
         }
 
     def package_open(self):
--- a/shopfloor/models/shopfloor_menu.py
+++ b/shopfloor/models/shopfloor_menu.py
@@ -81,6 +81,13 @@ class ShopfloorMenu(models.Model):
         help="If you tick this box, this scenario will allow operator to move"
         " goods even if a reservation is made by a different operation type.",
     )
+    reserve_only_available_is_possible = fields.Boolean(
+        compute="_compute_reserve_only_available_is_possible"
+    )
+    allow_reserve_only_available = fields.Boolean(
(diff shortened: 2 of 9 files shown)


rousseldenis:
@jbaudoux

shopfloor/actions/message.py
```diff
@@ -911,7 +911,10 @@ def lines_different_dest_location(self):
     def new_move_lines_not_assigned(self):
         return {
             "message_type": "error",
-            "body": _("New move lines cannot be assigned: canceled."),
+            "body": _(
+                "Some products in the source location are already reserved. "
+                "It's impossible to reserve them!"
```
sebalix, line 916:
```suggestion
                "It's impossible to reserve them."
```
I would avoid the exclamation mark in messages returned to the user?

shopfloor/models/shopfloor_menu.py
```diff
@@ -81,6 +81,13 @@ class ShopfloorMenu(models.Model):
         help="If you tick this box, this scenario will allow operator to move"
         " goods even if a reservation is made by a different operation type.",
     )
+    reserve_only_available_is_possible = fields.Boolean(
+        compute="_compute_reserve_only_available_is_possible"
+    )
+    allow_reserve_only_available = fields.Boolean(
+        string="Allow to reserve only available quantities on source location",
```
sebalix, line 88: As with location content transfer we want to move the goods, is "reserve" the right word here? Suggestion:
```suggestion
    allow_moving_only_available = fields.Boolean(
        string="Allow to move only available quantities on source location",
```

sebalix (commented):
Small comments, LG

shopfloor/actions/message.py
rousseldenis: Yes, less pressure  but we know they don't read them at all :smiley:

shopfloor/models/shopfloor_menu.py
rousseldenis: In fact, this is for quant quantities reservations, the 'move' is the next step :-)

lmignon (approved):

lmignon:
@jbaudoux Ready to be merged? At least it  works in production for months...

PR #1233, opened 2026-07-29
[FIX] shopfloor: Allow to use multi valued barcodes to find product AND lot

rousseldenis:


lmignon:
@nicolas-delbovier-acsone @jbaudoux A one we should take care of and question if it's still relevant...

nicolas-delbovier-acsone:
> @nicolas-delbovier-acsone @jbaudoux A one we should take care of and question if it's still relevant...

Yes it is, I added the necessary unit test to ensure the `scan_line` behaves as expected when using multi-barcodes.

nicolas-delbovier-acsone:
@rousseldenis can we open this one? I guess there is no reason to keep it draft?

PR #1237, opened 2026-08-04
[FIX] shopfloor: cluster_picking set_destination_all: skip batch sanity check

lmignon:
When set_destination_all splits the remaining lines into a new picking
and re-attaches it to the batch, the batch may already contain a done
picking (the one processed first). Assigning batch_id then triggers
stock_picking_batch's sanity check, which raises because the batch
mixes a done picking with a draft/assigned one.

Skip the sanity check for this specific assignment via a context key,
since the new picking is a legitimate split of a batch already in
progress, not an inconsistent manual assignment.

### DIFF
--- a/shopfloor/models/stock_move.py
+++ b/shopfloor/models/stock_move.py
@@ -130,10 +130,13 @@ class StockMove(models.Model):
                 if line.package_id == line.result_package_id:
                     line.result_package_id = False
 
-        # The batch cannot be set during copy as the moves have to be first
-        # extracted to the new picking in order to have a non draft state that
-        # will succeed the batch sanity check
-        new_picking.batch_id = picking.batch_id
+        # We skip the sanity check on the batch assignment to the new picking because in
+        # some cases, the batch may contain done pickings, which would make the sanity check
+        # fail. This could be the case when using the cluster_picking set_destination_all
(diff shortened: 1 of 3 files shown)


lmignon:
@jbaudoux

shopfloor/services/cluster_picking.py
```diff
@@ -1156,6 +1156,7 @@ def set_destination_all(self, picking_batch_id, barcode, confirmation=None):
         return self._unload_end(batch, completion_info_popup=completion_info_popup)
 
     def _unload_write_destination_on_lines(self, lines, location):
+        lines = lines.with_context(skip_batch_sanity_check=True)
```
jbaudoux: Can you move that skip in
stock_move.py:136: in _extract_in_split_order
    new_picking.batch_id = picking.batch_id

and adapt the comment

jbaudoux (changes requested):

shopfloor/services/cluster_picking.py
jbaudoux: `new_picking.with_context(skip_batch_sanity_check=True).batch_id = picking.batch_id`

lmignon:
@jbaudoux done

jbaudoux (approved):
/ocabot merge patch

PR #1238, opened 2026-08-05
[16.0][FIX] shopfloor: add unique parameter to product image URL

lmignon:
The change adds a unique parameter to the product image URL, which is generated using a SHA-512 hash of the product's last update timestamp. This ensures that when a product image is updated, the URL will change, preventing caching issues and ensuring that users always see the most recent image.

### DIFF
--- a/shopfloor/actions/data_detail.py
+++ b/shopfloor/actions/data_detail.py
@@ -1,6 +1,8 @@
 # Copyright 2020 Camptocamp SA (http://www.camptocamp.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import hashlib
 from collections import defaultdict
+from datetime import datetime
 
 from odoo.tools.float_utils import float_round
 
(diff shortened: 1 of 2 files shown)


lmignon:
ping @jbaudoux

shopfloor/actions/data_detail.py
```diff
@@ -212,7 +213,10 @@ def _locations_for_product(self, record):
     def _product_image_url(self, record, field_name):
         if not record[field_name]:
             return None
-        return f"/web/image/product.product/{record.id}/{field_name}"
+        sha = hashlib.sha512(
+            str(getattr(record, "__last_update")).encode("utf-8")  # noqa: B009
```
simahawk: is it necessary to use `getattr`?

lmignon: This is the kind of mistake that can happen when taking inspiration from Odoo's code without enough critical distance.  https://github.com/odoo/odoo/blob/16.0/addons/web/models/ir_qweb_fields.py#L39

It's now changed and the code should be compatible with Odoo up to 19.0. (where '__last_update' no more exists)

nicolas-delbovier-acsone (approved):

jbaudoux (approved):
/ocabot merge patch

PR #1246, opened 2026-08-23
[FIX] stock_storage_type: respect sequence

jbaudoux:
1/ Fix. Override odoo standard putaway computation based on package type when locations sequence are defined in order to apply the configuration rules of this storage type module.

2/ Perf. When reapplying standard putaway on top of the computed putaway based on storage type, do not re-evaluate putaway based on storage type.

cc @rousseldenis @lmignon


lmignon (approved):
LGTM. It would be nice properly document this behavior into the documentation. ;-)

rousseldenis (approved):

jbaudoux:
> LGTM. It would be nice properly document this behavior into the documentation. ;-)

@lmignon Good idea. I reviewed and updated the complete documentation.

rousseldenis (approved):

lmignon (approved):
Great improvement of the documentation!!!!

jbaudoux:
Note: to merge as major as computed result may differ from before

PR #1247, opened 2026-08-25
[16.0][IMP] shopfloor: cluster picking, option to force the scan of product, lot or package

nicolas-delbovier-acsone:
#### Problem / Business Need
In standard cluster picking operations, scanning a location barcode infers and validates/selects the move planned at that location. 

However, in real-world warehouse environments, stock discrepancies frequently occur (e.g., an operator physically placed Lot B at a bin while Odoo expects Lot A, or misidentified products are stored in the location). When location scanning auto-selects the move, operators unknowingly validate incorrect physical movements without realizing a mismatch exists between physical stock and Odoo's inventory state.

To enforce strict inventory control and force immediate discrepancy detection, warehouse managers need a way to require operators to explicitly scan the **product, lot, or package** barcode rather than relying on location inference.

cc @jbaudoux


jbaudoux (commented):
I don't think bypassing the confirmation of the location is a good idea. Why not just enabling the existing option scan location first? You should then have to still scan the product. Needs testing it works properly in this scenario with qty prefill, but the purpose was this one.

nicolas-delbovier-acsone:
> I don't think bypassing the confirmation of the location is a good idea. Why not just enabling the existing option scan location first? You should then have to still scan the product. Needs testing it works properly in this scenario with qty prefill, but the purpose was this one.

The problem is that, here, we want to **completely forbid** the scan of a location, not force it to be the first thing to scan.

When I tested the option you mentioned, I could end up validating lines having only scanned the location (in case there is no ambiguity for the system -> e.g. only one lot in the location). This is not something I want to be possible.

We want to strictly enforce scanning a unique identifier (such as a lot or packaging) to prevent picking errors. In practice, if scanning just the location is allowed, operators quickly default to scanning the bin and grabbing whichever box is physically closest. If that box doesn't match the specific lot Odoo assigned, the wrong item gets picked without anyone noticing, leading to inventory mismatches between Odoo and reality.

jbaudoux:
@nicolas-delbovier-acsone 

> When I tested the option you mentioned, I could end up validating lines having only scanned the location (in case there is no ambiguity for the system -> e.g. only one lot in the location). This is not something I want to be possible.

Like I said, there could be a bug when used in combination with prefill. Also that option was put in place for a customer using other picking scenario (e.g. zone picking), not this one. So there is less feedback.

Nevertheless, the purpose of the option is to force scanning the location and then the product/lot, so that there is no doubt because a same product can be at multiple locations.

> The problem is that, here, we want to completely forbid the scan of a location, not force it to be the first thing to scan.

So compared to your intention you will still have to scan first the location but I don't think you will slow down the user. Both labels are just in front of him. However everything will be confirmed.

nicolas-delbovier-acsone:
For the record, we've been discussing with @jbaudoux and we realised that it was **indeed** possible to select a move line by scanning **only** the location (even with "scan location first" activated).

The PR has been refactored in order to switch the mentality from "prevent location scan" to "force more detailed scan" (now you **can** scan a location but it will **never** allow you to validate the move line having only scanned the location).

lmignon (approved):

jbaudoux (approved):

jbaudoux:
Oups, we are already in version `16.0.2.26.6`. And this has a migration script for `16.0.2.26.0`.
@nicolas-delbovier-acsone Can you make a quick patch to move the migration script to `16.0.2.27.1` and set the version in the `__manifest__` ?

jbaudoux:
@nicolas-delbovier-acsone Ah you blocked the ongoing merging by force pushing just on time :) Then now you need to rebase and ideally set version `16.0.2.27.0`. Then it will be good :)

jbaudoux:
@nicolas-delbovier-acsone nearly :) You also need to rename the migration directory with right version number

jbaudoux:
That cmd will increase the version of the manifest to match the version of the migration

/ocabot merge patch

PR #1248, opened 2026-08-27
[16.0][FIX] shopfloor_gs1: prevent parsing when barcode is not a valid GS1

nicolas-delbovier-acsone:
### The problem

If we scan a lot named `10D9T`, shopfloor fails to find the given lot and returns a "barcode not found". This is because the gs1 module parses this barcode into `D9T` (`10` is the AI for "lot" in GS1).

See https://ref.gs1.org/ai/

@jbaudoux

### DIFF
--- a/shopfloor_gs1/utils.py
+++ b/shopfloor_gs1/utils.py
@@ -13,7 +13,7 @@ DEFAULT_AI_WHITELIST = tuple(MAPPING_AI_TO_TYPE.keys())
 class GS1Barcode:
     """GS1 barcode parser and wrapper."""
 
-    PRIMARY_AIS = ("01", "240")
+    PRIMARY_AIS = ("00", "01", "02", "240")
 
     __slots__ = ("ai", "code", "value", "raw_value")
 
--- a/shopfloor_gs1/utils.py
+++ b/shopfloor_gs1/utils.py
@@ -13,6 +13,8 @@ DEFAULT_AI_WHITELIST = tuple(MAPPING_AI_TO_TYPE.keys())
 class GS1Barcode:
     """GS1 barcode parser and wrapper."""
 
+    PRIMARY_AIS = ("01", "240")
+
     __slots__ = ("ai", "code", "value", "raw_value")
 
     def __init__(self, **kw) -> None:
@@ -51,6 +53,14 @@ class GS1Barcode:
         return value
 
     @classmethod
+    def _is_valid_gs1(cls, parsed):
+        # Standalone secondary AIs (e.g. lot '10LOT') without primary keys
+        # prefixes are ambiguous and usually false positive plain barcodes.
+        return any(
+            element.ai.ai in cls.PRIMARY_AIS for element in parsed.element_strings
+        )
+
+    @classmethod
     def parse(cls, barcode, ai_whitelist=None, safe=True):
         """Parse given barcode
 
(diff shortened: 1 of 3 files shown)


rousseldenis:
Shouldn't we fallback to normal barcode parsing as I suppose GS1 library returns a void result ?

nicolas-delbovier-acsone:
> Shouldn't we fallback to normal barcode parsing 

The logic to fallback to normal parsing is dealt with here (inside `barcode_parser.py`):

```python
def parse(self, barcode):
        """
        This method will parse the barcode and return the
        value with its type if determined.

        Override this to implement specific parsing

        """
        # Retrieve in any case the 'unknown' parsing with raw barcode
        result = super().parse(barcode)
        if not barcode:
            return result

        parsed = GS1Barcode.parse(barcode)
        if parsed:
            for barcode_type in self.search_action._barcode_type_handler.keys():
                for parsed_item in parsed:
                    if parsed_item.ai in MAPPING_TYPE_TO_AI.get(barcode_type, tuple()):
                        result[barcode_type] = BarcodeResult(
                            type=barcode_type,
                            value=parsed_item.value,
                            raw=parsed_item.raw_value,
                        )
        return result
```

> I suppose GS1 library returns a void result ?

No the GS1 library is parsing when it should not

```python
>>> GS1Message.parse("10RTGGERER")
GS1Message(value='10RTGGERER', element_strings=[GS1ElementString(ai=GS1ApplicationIdentifier(ai='10', description='Batch or lot number', data_title='BATCH/LOT', fnc1_required=True, format='N2+X..20'), value='RTGGERER', pattern_groups=['RTGGERER'], gln=None, gtin=None, sscc=None, date=None, decimal=None, money=None)])
```

shopfloor_gs1/utils.py
```diff
@@ -50,6 +52,14 @@ def _get_value(cls, ai_value: GS1ElementString):
             value = ai_value.gtin.payload + str(ai_value.gtin.check_digit)
         return value
 
+    @classmethod
+    def _is_valid_gs1(cls, parsed):
+        # Standalone secondary AIs (e.g. lot '10LOT') without primary keys
+        # prefixes are ambiguous and usually false positive plain barcodes.
+        return any(
+            element.ai.ai in cls.PRIMARY_AIS for element in parsed.element_strings
+        )
```
jbaudoux, line 61: For lot or serial, you indeed need a product primary ais 01, 02, 240
But if it's a package (SSCC), you can have only 00.
Would you then add it also to the list of primary?
00, 01, 02, 240 ?

jbaudoux (changes requested):
Can you add a test for a package "(00)PACK1" ?

Question: I see in the other tests it's a string with parenthesis. The library supports both and just drop the parenthesis?

nicolas-delbovier-acsone:
> Question: I see in the other tests it's a string with parenthesis. The library supports both and just drop the parenthesis?

For now, the code tries both HRI and normal barcode format, the library has different functions for each apparently:

```python
    @classmethod
    def parse(cls, barcode, ai_whitelist=None, safe=True):
        """Parse given barcode

        :param barcode: valid GS1 barcode
        :param ai_whitelist: ordered list of AI to look for()
        :param safe: break or not if barcode is invalid

        :return: a list of `GS1Barcode` instances.
        """
        res = []
        try:
            # TODO: we might not get an HRI...
            parsed = GS1Message.parse_hri(barcode)
        except ParseError:
            try:
                parsed = GS1Message.parse(barcode)
            except ParseError:
                if not safe:
                    raise
                parsed = None
```

shopfloor_gs1/utils.py
nicolas-delbovier-acsone: I'll add those as primary AI's as well then

jbaudoux (approved):

simahawk (approved):

PR #1250, opened 2026-09-01
[16.0][FIX] shopfloor: handle invalid product error gracefully on lot scan

nicolas-delbovier-acsone:
When scanning a barcode containing both product and lot info for a product
different from the current line, a `SearchInvalidProduct` exception was
uncaught in `cluster_picking`, leading to an unhandled 500 internal server
error in the Shopfloor application.

cc @jbaudoux


shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -195,8 +199,7 @@ def test_scan_line_lot_error_several_packages(self):
             line.lot_id.name,
             {
                 "message_type": "warning",
-                "body": "This lot is part of multiple"
-                " packages, please scan a package.",
+                "body": "This lot is part of multiple packages, please scan a package.",
             },
```
sebalix, line 203: It should probably be replaced by `self.msg_store.lot_multiple_packages_scan_package()` message (from `actions/message.py`).

shopfloor/tests/test_cluster_picking_scan_line.py
```diff
@@ -215,8 +218,7 @@ def test_scan_line_lot_error_in_one_package_and_unit(self):
...
```
sebalix, line 222: Same

shopfloor/tests/test_cluster_picking_scan_line.py
nicolas-delbovier-acsone: I agree completely but hard-coded error messages instead of using the msg_store can be found all over the place in shopfloor unit tests. I did not meant to make the changes to those lines in here, pre-commit did it.

I would say that we can avoid this problem for this PR and I'll open another one fixing all the hard-coded messages in shopfloor tests so that we keep concerns separated. What do you think?

sebalix: Yes sure :+1:

jbaudoux (approved):

shopfloor/services/cluster_picking.py
```diff
@@ -496,7 +501,11 @@ def scan_line(self, picking_batch_id, move_line_id, barcode, sublocation_id=None
                 picking, move_line, packaging, sublocation
             )
 
-        lot = search.lot_from_scan(barcode, products=move_line.product_id)
+        try:
+            lot = None
+            lot = search.lot_from_scan(barcode, products=move_line.product_id)
```
lmignon, line 506: Why `lot = None`? Could be removed

jbaudoux: No, it cannot. But would be maybe more understandable if outside try.

nicolas-delbovier-acsone: If the variable is not initialized, the rest of the code will fail in case `lot_from_scan` raises an error.

lmignon (approved):

jbaudoux:
@nicolas-delbovier-acsone needs rebase for merging

PR #1252, opened 2026-09-02
[16.0][REF] shopfloor: use msg_store instead of copy pasting messages

nicolas-delbovier-acsone:
There has been this bad practice for a while in Shopfloor unit tests to copy paste the expected messages instead of using the `msg_store` to call the message functions.

This leads to a codebase that is less maintainable since changes in the wording now needs to be performed in multiple places (does not respect DRY principle).

### DIFF
--- a/shopfloor/actions/message.py
+++ b/shopfloor/actions/message.py
@@ -201,7 +201,7 @@ class MessageAction(Component):
         return {
             "message_type": "warning",
             "body": _(
-                "Confirm location change from %(location_from)s to " "%(location_to)s?"
+                "Confirm location change from %(location_from)s to %(location_to)s?"
             )
             % dict(location_from=from_location.name, location_to=to_location.name),
         }
@@ -277,6 +277,9 @@ class MessageAction(Component):
 
     def wrong_record(self, record):
         return {"message_type": "error", "body": self._wrong_record_msg(record._name)}
+
+    def wrong_bin(self):
+        return {"message_type": "error", "body": _("Wrong bin")}
 
     def no_lot_for_barcode(self, barcode):
         return {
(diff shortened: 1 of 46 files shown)


nicolas-delbovier-acsone:
@sebalix Here is the PR we were talking about here: https://github.com/OCA/wms/pull/1250#discussion_r3904235810

sebalix (approved):
Thank you! There were a lot of them... LG

shopfloor/actions/message.py
```diff
@@ -1055,3 +1067,60 @@ def reserved_for_other_picking_type(self, picking):
             "message_type": "error",
             "body": body,
         }
+
+    def negative_quantity_not_allowed(self):
+        return {
+            "body": _("Negative quantity not allowed."),
+            "message_type": "error",
+        }
```
jbaudoux: not used?

shopfloor/services/checkout.py
```diff
@@ -1453,10 +1444,7 @@ def set_packaging(self, picking_id, package_id, package_type_id):
         package.package_type_id = packaging
         return self._response_for_summary(
             picking,
-            message={
-                "message_type": "success",
-                "body": _("Packaging changed on package {}").format(package.name),
-            },
+            message=self.msg_store.packaging_changed_on_package(package),
```
jbaudoux: Packaging was in v14, now it's package type
```suggestion
            message=self.msg_store.package_type_changed_on_package(package),
```

shopfloor/actions/message.py
```diff
@@ -1055,3 +1067,60 @@ def reserved_for_other_picking_type(self, picking):
...
+    def products_processed_as_raw_products(self):
+        return {
+            "message_type": "success",
+            "body": _("Product(s) processed as raw product(s)"),
+        }
+
+    def packaging_changed_on_package(self, pack):
+        return {
+            "message_type": "success",
+            "body": _("Packaging changed on package %s", pack.name),
```
```suggestion
            "body": _("Package type changed on package %s", pack.name),
```

jbaudoux (commented):
canceled instead of cancelled

shopfloor/actions/message.py
nicolas-delbovier-acsone: Indeed, I removed it

nicolas-delbovier-acsone:
> canceled instead of cancelled

Are you sure we need to change this? This comes in many places of the codebase. Is it really relevant here?

jbaudoux:
> > canceled instead of cancelled
> 
> Are you sure we need to change this? This comes in many places of the codebase. Is it really relevant here?

Always good to fix it when it's spotted

PR #1253, opened 2026-09-02
[16.0][IMP] shopfloor: replace raw HTML links with _get_html_link()

nicolas-delbovier-acsone:
Use Odoo standard `_get_html_link()` helper method on recordsets when posting
chatter messages in `stock.move` and `stock.picking`. This removes hardcoded
HTML string formatting and ensures consistent link generation.

@jbaudoux


jbaudoux (approved):
/ocabot merge patch

PR #1259, opened 2026-09-08
[16.0][FIX] shopfloor: cluster picking: unload partialy processed

jbaudoux:
Follow-up of the full bin improvement.

Special use case:
A partially available line gets its availability increased after being picked.
Currently, the cluster picking scenario always assumed that all move lines are fully picked. If it wasn't the case, the move line was split during the processing of the line.
When the move line reservation is modified after being picked, we need to split the move line and it will be processed in a backorder.

Warning: I also removed the copy=True from the move lines shopfloor flags

cc @lmignon @TDu @sebalix

### DIFF
--- a/shopfloor/tests/test_cluster_picking_unload.py
+++ b/shopfloor/tests/test_cluster_picking_unload.py
@@ -320,6 +320,91 @@ class ClusterPickingSetDestinationAllCase(ClusterPickingUnloadingCommonCase):
             message=self.service.msg_store.batch_transfer_complete(),
         )
 
+    def test_set_destination_all_picking_partial(self):
+        """Set destination on lines partially processed for one transfer.
+
+        Process a partially available move line.
+        Increase the stock after picking causing a line partially processed.
+        Check a backorder is created.
+        """
+        self.one_line_picking.do_unreserve()
+        location = self.one_line_picking.location_id
+        move = self.one_line_picking.move_ids
+        product = move.product_id
+        qty = move.product_uom_qty
+        # note: there are 2 move lines => qty * 2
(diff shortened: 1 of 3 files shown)


lmignon (changes requested):
@jbaudoux LGTM, even though the change in the copy behaviour could be a little concerning.

Unfortunately, the way you assert the result in the new test is not deterministic: the order of the lines may vary from one run to another.

jbaudoux:
> @jbaudoux LGTM, even though the change in the copy behaviour could be a little concerning.

@lmignon I was hoping to have failed tests with the change but it's like there is nothing that test that :(

shopfloor/tests/test_cluster_picking_unload.py
```diff
@@ -320,6 +320,91 @@ def test_set_destination_all_picking_unassigned(self):
...
+            next_state="start",
+            message=self.service.msg_store.batch_transfer_complete(),
+        )
+        # since the whole batch is complete, we expect the batch and all
+        # pickings to be 'done'
+        self.assertRecordValues(
+            move_lines.mapped("picking_id"), [{"state": "done"}, {"state": "done"}]
+        )
+        self.assertRecordValues(
+            move_lines.sorted("qty_done"),
```
lmignon, line 367: 😆

lmignon (approved):

jbaudoux:
> > @jbaudoux LGTM, even though the change in the copy behaviour could be a little concerning.
> 
> @lmignon I was hoping to have failed tests with the change but it's like there is nothing that test that :(

move lines shopfloor fields are tested :)

PR #1260, opened 2026-09-08
[16.0] wms: remove myself from the maintainers

guewen:
I do not work on these modules and have no capacity to maintain them.


sebalix:
Still, thank you for your past work on Shopfloor BTW :)

PR #1088, opened 2025-07-25
[16.0] shopfloor_mobile: align w/ v18 linting

simahawk:
Apply the same linting changes from v18 that apply as well to 16, to ease bkp/fwd porting.

PR #1268, opened 2026-09-11
[16.0][FIX] shopfloor_mobile: pass remaining quantity across packaging loop iterations

nicolas-delbovier-acsone:
Bug introduced in https://github.com/OCA/wms/pull/1088

When calculating quantities per packaging in `_product_qty_by_packaging`, the initial `qty` parameter was passed to `_qty_by_pkg` at every packaging level without updating the remaining unallocated quantity.

As a result, smaller packaging units calculated their required quantity against the full initial amount rather than the remainder left by larger packaging units. This inflated the total calculated in `compute_qty()`, triggering a reactive watcher feedback loop that compounded the total on every tick until overflowing to `NaN`.

### How to reproduce the error

image

- Enter a shopfloor screen with the "packaging-qty-picker" inside
- _(Optional)_ Ensure you only have one `product.packaging` for this product
- Input something ">=" to the qty of the given packaging

cc @jbaudoux @lmignon

### DIFF
--- a/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.esm.js
+++ b/shopfloor_mobile/static/wms/src/components/packaging-qty-picker.esm.js
@@ -89,11 +89,15 @@ export var PackagingQtyPickerMixin = {
         _product_qty_by_packaging: function (pkg_by_qty, qty) {
             const self = this;
             const res = {};
+            let _remaining_qty = qty;
             // Const min_unit = _.last(pkg_by_qty);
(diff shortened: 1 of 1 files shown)


lmignon (approved):
😱 An annoying one, and not an easy one to debunk! BTW, the code readability is improved...

nicolas-delbovier-acsone:
There is an error in current implementation

image

jbaudoux (approved):

PR #1269, opened 2026-09-11
[16.0][IMP] shopfloor: move validation

jbaudoux:
Move recent picking validation improvement in cluster picking to the stock action validating the moves.
No moves are created by shopfloor in the cluster picking but for the generic validation we need to consider this.
Adapt the "backorder never" test for location content transfer to set is_shopfloor_created.
    
Marking a move as picked should never raise the picked quantity exceeds the need.

Add a `get_qty_picked` service method.

In `mark_move_line_as_picked`, always check for concurrent work unless we don't split lines (shopfloor_reception)

cc @Tdu @lmignon @sebalix


lmignon (commented):
@jbaudoux The simplifications are really valuable: they remove duplicated code and make the way the different processes are covered more orthogonal. 😍 

I ran the test suite in my local environment successfully. Once again, the failing tests seem to be related to other addons being installed when the tests are run.

TDu (approved):

jbaudoux:
Added a fix for partially available moves to ensure a backorder is created when all available move lines have been processed. Otherwise a split order is created.
The split of partially available moves is not necessary when a backorder is created. The method is moved to extract_and_action_done.

Restrict back the method split_other_move_lines to only split moves with other move lines. Rely on split_unavailable_qty for partially available moves.
Due to this change, mark_move_line_as_picked without quantity doesn't split anymore unavailable quantities. It needs to be now called explicitly before in the location content transfer scenario.

jbaudoux:
Added a fix for move lines partially available and fully picked that sees their reserved quantity increased, so that the new move line after the split has no result package in the cluster picking

shopfloor/models/stock_move.py
```diff
@@ -83,13 +83,38 @@ def split_other_move_lines(self, move_lines, intersection=False):
         return self.browse()
 
     def split_unavailable_qty(self):
-        """Put unavailable qty of a partially available move in their own
-        move (which will be 'confirmed').
+        """Split partially available moves
+
+        Create a new 'confirmed' move with the unavailable qty and make the
+        move 'assigned'.
```
lmignon:
```suggestion
        Create a new 'confirmed' move with the unavailable qty and make the
        original one 'assigned'.
```

lmignon:
A lot of changes and an impressive cleanup of the code! After a careful review of your PR, everything looks good to me (but as we know, the evils is into the details). The only remaining step is to confirm this with functional tests on our database.

jbaudoux:
Thanks for the reviews. I improved a bit a few docstring

PR #1270, opened 2026-09-14
[16.0][FIX] shopfloor_mobile_base: resolve language code via translation registry fallback

nicolas-delbovier-acsone:
When an Odoo user is configured with a regional language variant (e.g., `fr_BE`),
`switch_language` was setting `$i18n.locale` directly to `fr_BE`. Because
translations are registered under BCP 47 tag `fr-FR`, VueI18n failed to find
matching keys and defaulted to English.

Pass `lang_id` through `translation_registry.ensure_lang()` inside
`switch_language` to resolve regional variants to their registered fallback.

---

> [!NOTE]
> For now, the translations are registered under specific locale name such as "fr-FR" (see https://github.com/OCA/wms/blob/541024427a0d260330a5aab197339d44a95ab452/shopfloor_mobile_base/static/wms/src/i18n/i18n.fr.js#L139-L140). Shouldn't we register the translations under base "fr" name and only add specific variants in case we need them?


nicolas-delbovier-acsone:
I convert this one to draft because I want https://github.com/OCA/wms/pull/1281 to be merged first in order to change all the locale (fr-FR") variants declarations for base language code ("fr") ones.

PR #1271, opened 2026-09-14
[16.0][FIX] stock_picking_completion_info: only invalidate completion_info while the feature is in use

baguenth:
## Problem

Every `stock.move` write that changes the state invalidates `completion_info`
on the whole `stock.picking` model, because the `api.depends` cannot express
all conditions that change the value. This runs once per move for every
transfer processed — also on databases where **no** operation type has
*Display Completion Info* enabled. In that case the field is `"no"` for every
picking and there is nothing to invalidate.

## Fix

Read the flag through a registry cache (`ormcache`) and skip the invalidation
while no operation type shows the information. The cache is cleared when the
flag changes, and when an operation type with the flag is created or deleted.
Archived operation types count as enabled: their pickings still compute the
field, and archiving does not touch the flag.

The move `write()` also returns the result of `super()` again instead of a
bare `True`.


baguenth:
Hey @guewen , @simahawk ,

kindly asking for a review :)

QuatschSalat (approved):

stock_picking_completion_info/models/stock_picking.py
```diff
@@ -12,6 +12,40 @@ class PickingType(models.Model):
         " completion"
     )
 
+    @api.model
+    @tools.ormcache()
+    def _completion_info_enabled(self):
```
lmignon: Maybe you could have a cached method returning a dict mapping IDs to their `display_completion_info` value, and only invalidate the cache in the `stock.move` `write()` method when the move's picking type has `display_completion_info == True`.

lmignon (changes requested):
If I understand this correctly, your changes becomes useless once a picking type is configured to display completion_info. In that case, as mentioned above, I think a better approach would be to invalidate the cache based on the picking type's display_completion_info configuration.

baguenth:
> If I understand this correctly, your changes becomes useless once a picking type is configured to display completion_info. In that case, as mentioned above, I think a better approach would be to invalidate the cache based on the picking type's display_completion_info configuration.

Thank you for the review.

Avoiding the invalidation entirely when no picking type has the feature enabled is exactly what the change was aiming for — it is a performance improvement for installations that do not use the feature at all, and there the whole question of granularity never arises.

I agree the decision can be more granular, and I will make it so. One detail on the specific approach though: keying it on the move's picking type would miss the chained case. `completion_info` depends on `move_ids.common_dest_move_ids.state`, so a move's state change affects the transfers that share a destination with it — and the flag that matters is the type of those transfers, not the type of the transfer holding the move. In a two-step delivery where the pick type has the feature off and the ship type has it on, we would stop invalidating a value that did change.

What I will do instead: keep the cached lookup you suggest, but resolve the transfers actually reached by the move via `common_dest_move_ids`, keep the ones whose picking type shows the information, and invalidate `completion_info` on that recordset rather than on the whole model. That is narrower than today in every case, and still correct for chains. The short-circuit stays as a cheap first check.

sebinthomasst (approved):

lmignon (approved):
Your code looks logic and makes sense (Code review only)

baguenth:
Hey @guewen, @simahawk,

can you maybe help getting this integrated?

baguenth:
Hey @jbaudoux, @sebalix,

Could one of you take a look at merging this? It has approvals already and should be ready to merge.

Thanks! :)

PR #1272, opened 2026-09-14
[16.0][IMP] shopfloor: add _check_move_line_qty_picked and deprecate _split_qty_to_be_done

jbaudoux:
Some cleanup:
* Added in service a generic method `_check_move_line_qty_picked()`
* `_split_qty_to_be_done()` is now deprecated and replaced by `_split_partial_quantity_to_be_done()`

@Tdu @lmignon @nicolas-delbovier-acsone @sebalix 

Requires improvements from:
* https://github.com/OCA/wms/pull/1269


shopfloor/services/checkout.py
```diff
@@ -845,28 +845,17 @@ def _change_line_qty(
...
-                    "message_type": "error",
-                }
-            else:
-                new_line = self.env["stock.move.line"]
-                if qty_done > 0:
-                    new_line, qty_check = move_line._split_qty_to_be_done(
-                        qty_done,
-                        split_partial=False,
-                        result_package_id=False,
-                    )
```
jbaudoux, line 860: I dropped this because it appears that calling this method with split_partial=False is doing nothing except checking the quantity is not negative which is now done with `_check_move_line_qty_picked`

shopfloor/actions/message.py
```diff
@@ -768,15 +768,6 @@ def selected_lines_qty_done_higher_than_allowed(self, line):
             ),
         }
 
-    def line_scanned_qty_done_higher_than_allowed(self):
```
lmignon, line 771: This removal makes at least one test fail... 😏

shopfloor/models/stock_move_line.py
```diff
@@ -178,13 +179,18 @@ def _split_pickings_from_source_location(self):
...
         """Check qty to be done for current move line. Split it if needed.
 
         :param qty_done: qty expected to be done
         :param split_partial: split if qty is less than expected
             otherwise rely on a backorder.
         """
+        _logger.warning(
+            "`_split_qty_to_be_done` is deprecated "
+            "and replaced by `_split_partial_quantity_to_be_done`"
+        )
```
lmignon, line 193: I think it would be better to use Python's warnings.warn() with DeprecationWarning rather than logging a warning.

This is the standard mechanism for marking APIs as deprecated, and it also allows callers and tests to filter or explicitly handle the warning.

Don't forget to use stacklevel=2 so that the warning points to the caller rather than to the deprecated method itself and therefore eases to detect place where the code must be adapted.


```suggestion
import warnings
....

        warnings.warn(
            "`_split_qty_to_be_done` is deprecated "
            "and replaced by `_split_partial_quantity_to_be_done`",
            DeprecationWarning,
            stacklevel=2,
        )
```

lmignon (changes requested):

shopfloor/actions/message.py
jbaudoux: I put it back :)

lmignon (approved):

PR #1278, opened 2026-09-17
[16.0][IMP] shopfloor_reception_mobile: add an input to filter the moves

nicolas-delbovier-acsone:
| Header | Header |
|--------|--------|
| image | image | 

cc @jbaudoux


jbaudoux (commented):
As you do that frontend side only, does filter as you type?

nicolas-delbovier-acsone:
> As you do that frontend side only, does filter as you type?

Yes it does !

jbaudoux (approved):

jbaudoux (approved):

PR #1281, opened 2026-09-18
[16.0][I18N] shopfloor_mobile_*: add fr translations

nicolas-delbovier-acsone:
Note that I did not follow the convention for the translation keys for new translations.

IMO, Natural Language translation keys are better because they provide a better fallback in case of error (you do not see "btn.confirm" in case the registery got a problem/your venv is not up to date etc.). This is closer to what we do in odoo with `_("Some String")`.

This is also easier to dev as you see directly inside one file what the user sees (no need to open `i18n/en.json` to see what is displayed).

We can follow the convention to always use English in the .js files and only provide translation files for *other* languages (thus getting rid of all unnecessary `en.json` files).

cc @jbaudoux


nicolas-delbovier-acsone:
@sebalix Jacques told me to ping you on this one. Any thoughts about the way I suggest to translate?

rousseldenis (approved):

nicolas-delbovier-acsone:
For the record, I've been convinced by @simahawk to keep using keys for translations instead of plain English strings. I'll change this PR to use this approach.

shopfloor_mobile/static/wms/src/components/detail/detail_lot.esm.js
```diff
@@ -80,7 +80,7 @@ Vue.component("detail-lot", {
 
 
         <div class="suppliers mb-4" v-if="record.product.suppliers.length">
-            <separator-title>Suppliers</separator-title>
+            <separator-title>{{ $("Suppliers") }}</separator-title>
```
lmignon: This one seems to not follow the naming convention ...
```suggestion
            <separator-title>{{ $("detail.lot.suppliers") }}</separator-title>
```

shopfloor_mobile/static/wms/src/i18n/en.json
```diff
@@ -40,6 +40,33 @@
   "detail.transfer.qty_reserved": "Qty reserved",
   "detail.transfer.qty_in_stock": "Qty in stock",
   "detail.transfer.edit_carrier": "Edit carrier",
+  "detail.location.reserved_products": "Reserved products",
+  "detail.location.available_products": "Available products",
+  "detail.lot.expiry_date": "Expiry date",
+  "detail.lot.removal_date": "Removal date",
+  "detail.lot.vendor_code": "Vendor Code",
+  "detail.lot.vendor_name": "Vendor Name",
+  "detail.lot.packaging": "Packaging",
```
lmignon, line 50:
```suggestion
  "detail.lot.packaging": "Packaging",
  "detail.lot.suppliers": "Suppliers",
```

shopfloor_mobile/static/wms/src/i18n/fr.json
```diff
@@ -38,6 +38,33 @@
   "detail.transfer.qty_reserved": "Qté réservée",
   "detail.transfer.qty_in_stock": "Qté en stock",
   "detail.transfer.edit_carrier": "Modifier le transporteur",
+  "detail.location.reserved_products": "Produits réservés",
+  "detail.location.available_products": "Produits disponibles",
+  "detail.lot.expiry_date": "Date d'expiration",
+  "detail.lot.removal_date": "Date de retrait",
+  "detail.lot.vendor_code": "Code Fournisseur",
+  "detail.lot.vendor_name": "Nom Fournisseur",
+  "detail.lot.packaging": "Conditionnement",
```
lmignon, line 48:
```suggestion
  "detail.lot.packaging": "Conditionnement",
  "detail.lot.suppliers": "Fournisseurs",
```

lmignon (changes requested):

lmignon (changes requested):
What a big job....  unfortunately one translation is missing..

lmignon (approved):

rousseldenis (approved):

